PR #2, opened 2022-05-11
Incorporate jobqueue from image-builder

lzap:
Motivation:

Kafka is a bad tool to use for task queues, it is a data streaming platform rather than a traditional message bus. RabbitMQ, SQS or Redis should be prefered, however, these are unmanaged services by SRE. The only managed service at the moment is Postgres, however, modern Postgres do provide a lot of tools to implement tasks queues very efficiently via LISTEN/NOTIFY and SKIP LOCKED features.

* https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/
* https://webapp.io/blog/postgres-is-the-answer/

This is a POC of taking advantage of a very nice package from image-builder:

https://github.com/osbuild/osbuild-composer/tree/main/internal/jobqueue

We have the very same problem - to create job queue for one or more workers to perform actions (AWS RES API, polling of status, sending notifications). The library is currently in the internal package but we are discussing with folks to extract it into a separate project.

It provides everything we need: queue implemented on pgx via LISTEN/NOTIFY, good performance via FOR UPDATE SKIP LOCKED, JSONB for arguments and result, multiple types and channels, job dependency, heartbeat for longer jobs (polling AWS EC2 API), rescheduling of dead jobs, cancelation and the implementation is clean and lightweight.

This PR is an attempt to use a copy of the library for POC, once we confirm it works and we want to go this way, we can use the extracted library. Couple of changes will be required in the library, namely creating adapter for logging (we do not use logrus) and creating an adapter for connection pool (it uses pgxpool and we must use database/sql pool because of sqlx or gorm). But these changes are small.

Alternatives:

* https://eng.uber.com/cherami-message-queue-system/ (by Uber, requires persistence)
* https://github.com/RichardKnop/machinery (no Postgres, supports workflows, complex)
* https://github.com/btubbs/pgq (simple, pgx, skip locked support, no listen/notify support)

Todo:

* postgresql-contrib-13.4-3.fc35.x86_64 must be installed - put this into README


lzap:
I am going to create a new PR, struggling with GoLand GitHub support. Might end up giving up and using just CLI git. :)

PR #4, opened 2022-05-12
Migration respect configuration

lzap:
Migration was previously hardcoded to localhost, it now respects configuration via ENV variables. I also improved logging and it shows migration history after each migration too.

### DIFF
--- a/internal/db/connection.go
+++ b/internal/db/connection.go
@@ -3,11 +3,13 @@ package db
 import (
 	"fmt"
 	"github.com/RHEnVision/provisioning-backend/internal/config"
-	pgx "github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/log/zerologadapter"
+	_ "github.com/jackc/pgx/v4/stdlib"
 	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
+	"net/url"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


ezr-ondrej (approved):
Looks sane to me, thanks @lzap ! :)

internal/db/connection.go
```diff
@@ -3,29 +3,43 @@ package db
...
+	} else {
+		return fmt.Sprintf("%s://%s@%s:%d/%s",
+			prefix,
+			url.QueryEscape(config.GetLoggingConfig().Database.User),
+			config.GetLoggingConfig().Database.Host,
+			config.GetLoggingConfig().Database.Port,
+			config.GetLoggingConfig().Database.Name)
+	}
+
+}
```
lzap, line 38: The prefix is needed because golang-migrations has its own format of database URL (pgx vs postgres). But we still use pgx in both cases.

internal/db/migrations.go
```diff
@@ -38,17 +41,38 @@ func Migrate() {
...
+		var version int
+		var appliedAt time.Time
+		if err := rows.Scan(&version, &appliedAt); err != nil {
+			mlog.Fatal().Err(err).Msg("Error scanning schema history")
+		}
+		mlog.Info().Msgf("Version %d was applied %v", version, appliedAt)
+	}
+	if err := rows.Err(); err != nil {
+		mlog.Fatal().Err(err).Msg("Error iterating rows")
 	}
```
lzap, line 76: Iterating rows via the low level Scan function is quite verbose ;-)

internal/db/migrations.go
```diff
@@ -38,17 +41,38 @@ func Migrate() {
 	mlog := log.Logger.With().Bool("migration", true).Logger()
```
lzap, line 41: All logs related to migrations go into zerolog with migration=true flag so we can easily filter them.

internal/db/migrations.go
```diff
@@ -38,17 +41,38 @@ func Migrate() {
...
 	d, err := iofs.New(fs, "migrations")
 	if err != nil {
-		mlog.Fatal().Err(err)
+		mlog.Fatal().Err(err).Msg("Error reading migrations")
```
lzap, line 44: FYI the Fatal() call will not exit the program unless Msg is called too.

PR #6, opened 2022-05-13
DAO implementation, accounts table, seed script

lzap:
Also drop down-migrations, these do not appear to be very helpful in golang-migrate because when migration fails, it sets the "dirty" flag and database must be fixed manually anyway. We can add down migrations later if we need to.

The patch also adds seeds support, this will be helpful for setting up a working development setup. By deafult this is turned off, use DB_SEED_SCRIPT=dev_small to use it.


README.md
```diff
@@ -80,6 +82,10 @@ Here are few points before you start contributing:
 * Database connection is at `internal/db`, cloud connections are in 'internal/clouds'.
 * Do not introduce `utils` or `tools` common packages, they tend to grow.
 * Keep the line of sight (happy code path).
+* PostgreSQL version we currently use in production is v14+ so take advantage of all modern SQL features.
+* Use `BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY` for primary keys.
+* Use `TEXT` for string columns, do not apply any limits in the DB: https://wiki.postgresql.org/wiki/Don%27t_Do_This#Text_storage
+* Keep security in mind: https://github.com/RedHatInsights/secure-coding-checklist
```
ezr-ondrej, line 88: Should these be rationalized by ADR? I'd be ok with a quickie directly to git, summarizing your thoughts, how you got to these conclusions, I guess not much discussion needed, but having rationale backing these decisions would be useful ^_^

lzap: I've taken the PK and TEXT decisions from https://github.com/RedHatInsights/vuln4shift-backend/blob/master/dbadmin/migrations/001_init_schema.up.sql and the checklist is from https://github.com/RedHatInsights/vuln4shift-backend/blob/master/pull_request_template.md as well. I can remove whatever you like from this PR, no big deal, I just thought its useful.

lzap:
While working on DAO and REST, I find SshPublicKey hard to type, any objections to name the model as PubKey (table pubkeys) and REST endpoint /public_keys ?

lzap:
I am pushing initial DAO and REST API (POC only) implementation for Accounts.

Do not merge, the implementation suffers from an interesting bug - I am getting all records pointing to the same struct for some reason :)

```
lzap@mone provisioning-backend % curl http://localhost:8000/accounts
[{"ID":4,"AccountNumber":"77","OrgID":"000077"},{"ID":4,"AccountNumber":"77","OrgID":"000077"},{"ID":4,"AccountNumber":"77","OrgID":"000077"},{"ID":4,"AccountNumber":"77","OrgID":"000077"}]
```

I will take a look on Monday, unless someone else can figure this one out.

lzap:
Fixed the problem, lint will fail tho. I will add error wrapping to DAO, however, I will need to turn off statement close check for the dao package:

```
internal/dao/sqlx/account_dao.go:25:46: Rows/Stmt was not closed (sqlclosecheck)
	daoImpl.getById, err = db.DB.PreparexContext(ctx, getById)
	                                            ^
internal/dao/sqlx/account_dao.go:29:43: Rows/Stmt was not closed (sqlclosecheck)
	daoImpl.list, err = db.DB.PreparexContext(ctx, list)
```

As it is stated in the Go documentation: https://pkg.go.dev/database/sql#Stmt

If a Stmt is prepared on a DB, it will remain usable for the lifetime of the DB. When the Stmt needs to execute on a new underlying connection, it will prepare itself on the new connection automatically.

I am taking advantage of this in the DAO package, each DAO implementation prepares statements on the connection and when executed it the connection pool automatically creates them on the fly. They are never closed, that is what this lint is all about, but in this case it makes sense. We could close the statements at the end of each request but thats actually not needed.

README.md
ezr-ondrej: No need, I'll add the ADR into git myself, don't get me wrong I'm all up for it, and I don't mind slightly bigger PR at the beggining, but we will for sure wonder in the future why it is there, so I'd like to have that written down somewhere, but I'm happy to do that! :)

lzap:
Hmm the linter on CI errors out while its passing locally: Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: goanalysis_metalinter: execinquery: package \"dao\" (isInitialPkg: true, needAnalyzeSource: true): runtime error: invalid memory address or nil pointer dereference

lzap:
Fixed, I upgraded GHA actions to v3 where appropriate and also pinned the goloangci-lint action to version 1.45 which works fine.

README.md
lzap: Let me know if you want me to remove this from the PR.

internal/services/account_service.go
```diff
@@ -0,0 +1,51 @@
+package services
+
+import (
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/payloads"
+	"github.com/go-chi/render"
+	"net/http"
+)
+
+func ListAccounts(w http.ResponseWriter, r *http.Request) {
```
ezr-ondrej, line 10: Should we even expose Accounts in API? it seems unecessary as the service should always run in a context of one account IMHO, right?

internal/db/migrations/002_utility_functions.up.sql
```diff
@@ -0,0 +1,11 @@
...
+
+CREATE OR REPLACE FUNCTION empty(t TEXT)
+    RETURNS BOOLEAN AS
+$empty$
+BEGIN
+    RETURN t ~ '^[[:space:]]*$';
+END;
+$empty$ LANGUAGE 'plpgsql';
+
+COMMIT;
```
ezr-ondrej, line 11: there's still quite a lot of end lines missing ^_^

internal/services/account_service.go
lzap: Yes I know, but I wanted to create few example API calls with DAO so we have at least something, I had to start with Accounts because of dependencies. We will either delete REST operations completely or create an administrative role. It would be good if you could figure out if administrative roles are allowed by SRE or is something that is a good practice.

internal/db/migrations/002_utility_functions.up.sql
lzap: Fixed with `git ls-files -z | while IFS= read -rd '' f; do tail -c1 < "$f" | read -r _ || echo >> "$f"; done`

Added editorconfig for go project.

api/requests/get-account.http
```diff
@@ -0,0 +1,3 @@
+// @no-log
+GET http://localhost:8000/accounts/1 HTTP/1.1
```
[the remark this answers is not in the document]
lzap: Dropped, that is a test file for HTTP Client plugin for GoLand/VSCode and others.

internal/db/migrations/003_accounts.up.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+CREATE TABLE accounts
+(
+    id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    account_number TEXT NOT NULL UNIQUE CHECK (NOT empty(account_number)),
```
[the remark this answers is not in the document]
lzap: We don't know for sure, if I have to choose between get that info and store it to only drop it later or ignore this and solve when we encounter problems, I choose one. There is a transition period planned and some services will only accept the account number for some time, although there is the translation service why not to keep those numbers in our database since it will come for free in requests anyway? Dropping a column is literally a DDL one-liner.

lzap:
Squashed into a single commit.

adiabramovitch:
Great work, really helped. 
@ezr-ondrej Do you have any comments?

internal/db/migrations/003_accounts.up.sql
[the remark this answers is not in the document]
ezr-ondrej: I am not against keeping the account_number, I want to allow a blank value for it. There are already some CUs that will not have `account_number` and such CUs would not be able to use our service :)

ezr-ondrej:
Yeah I still believe we should allow blank accounts :thinking: but we can talk about that later, but would be better not to change migrations ^_^

lzap:
WAIT

lzap:
So I am assuming that NOT NULL will not work with UNIQUE, but let me try maybe NULLs are not taken into account.

lzap:
You were right, NULL values are allowed regardless of the unique constraint, therefore let's go with that. Also I removed the ssh key table from the seed script - that was a leftover. Finally, I added go documentation to the account model type.

internal/db/migrations/003_accounts.up.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+CREATE TABLE accounts
+(
+    id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    account_number TEXT UNIQUE CHECK (NOT empty(account_number)),
```
ezr-ondrej, line 6: :orange_heart:

ezr-ondrej (approved):
I'm good, thanks @lzap and thanks @adiabramovitch for the review !

PR #7, opened 2022-05-14
Add deploy to Ephemeral configs and guide

ezr-ondrej:


Makefile
```diff
@@ -15,7 +15,7 @@ strip: build
 
 .PHONY: build-podman
 build-podman:
-	podman build .
+	podman build -t provisioning-backend -f build/Dockerfile .
```
lzap, line 18: I use podman remote build (podman connects to a remote box where podman is daemonized and performs the actual build) and this fails for some reason (perhaps a bug in the protocol).

Could you create a symlink so I can still build it manually by issusing `podman build .`? Do this:

    ln -f build/Dockerfile Containerfile

And commit the symlink into git, if there are no objections.

Makefile
```diff
@@ -43,4 +43,4 @@ migrate: build-pbmigrate
 .PHONY: generate-migration
 MIGRATION_NAME?=unnamed
 generate-migration:
-	migrate create -ext sql -dir api/migrations -seq -digits 3 $(MIGRATION_NAME)
\ No newline at end of file
+	migrate create -ext sql -dir api/migrations -seq -digits 3 $(MIGRATION_NAME)
```
lzap, line 46: Updated my IDE configuration, surprisingly this was not the default :)

build/Dockerfile
```diff
@@ -5,10 +5,8 @@ WORKDIR /build
 COPY . .
 RUN make prep build strip
 
-FROM registry.access.redhat.com/ubi8/ubi:latest
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
lzap, line 8: Nice catch, minimal is the way to go.

build/Dockerfile
```diff
@@ -5,10 +5,8 @@ WORKDIR /build
...
 COPY --from=build /build/pbapi /pbapi
 COPY --from=build /build/pbmigrate /pbmigrate
 USER 1001
-EXPOSE 8000
-EXPOSE 5000
```
lzap, line 13: Is this left intentionally?

cmd/pbapi/main.go
```diff
@@ -61,14 +61,14 @@ func main() {
...
+	log.Info().Msg("Starting new instance on 8000/9000")
 	srv := http.Server{
 		Addr:    fmt.Sprintf(":%d", 8000),
 		Handler: r,
 	}
 
 	msrv := http.Server{
-		Addr:    fmt.Sprintf(":%d", 8080),
+		Addr:    fmt.Sprintf(":%d", 9000),
 		Handler: mr,
```
lzap, line 72: Can you extract these into config variables? Let's merge your Viper first I guess.

deploy/bonfire.example.yaml
```diff
@@ -0,0 +1,17 @@
...
+  path: bonfire_configs/ephemeral_apps.yaml
+
+# (optional) define any apps locally. An app defined here with <name> will override config for app
+# <name> in above fetched config.
+apps:
+- name: provisioning
+  components:
+    - name: service
+      host: local
+      repo: <path_to_service_dir>
```
lzap, line 16: This seems like from a template... just saying. I haven't tried this yet.

deploy/clowdapp.yaml
```diff
@@ -1,32 +1,44 @@
...
             image: ${IMAGE}:${IMAGE_TAG}
             initContainers:
               - name: run-migrations
+                image: "${IMAGE}:${IMAGE_TAG}"
                 command:
-                  - bash
+                  - sh
                   - -c
-                  - pbmigrate
+                  - /pbmigrate
```
lzap, line 36: I haven't tried, but there is no reason to run our app in shell, can you simply run it directly? You will have to give it full path to the binary but that's easy to figure out I guess.

deploy/clowdapp.yaml
```diff
@@ -35,10 +47,8 @@ objects:
               timeoutSeconds: 120
             readinessProbe:
               failureThreshold: 3
-              httpGet:
-                path: /
+              tcpSocket:
                 port: 8000
-                scheme: HTTP
```
lzap, line 41: Is this change intentional? I would expect the readiness to be exactly like liveness.

deploy/clowdenv.yaml
```diff
@@ -0,0 +1,68 @@
...
+        path: "/metrics"
+
+      # Creates a kafka pod in the targetNamespace
+      kafka:
+        namespace: default
+        clusterName: crc-cluster
+        mode: local
+
+      # Clowder supports postgres 10 and 12. Specify the name
+      # and other details in the clowdapp
```
lzap, line 39: Perhaps delete? Out of date info.

deploy/clowdenv.yaml
```diff
@@ -0,0 +1,68 @@
...
+      metrics:
+        port: 9000
+        mode: operator
+        path: "/metrics"
+
+      # Creates a kafka pod in the targetNamespace
+      kafka:
+        namespace: default
+        clusterName: crc-cluster
+        mode: local
```
lzap, line 36: I suggest to remove Kafka for now, deployments will be faster I guess. But feel free to keep it.

deploy/clowdenv.yaml
```diff
@@ -0,0 +1,68 @@
...
+        mode: none
+
+      # Deploys a local minio pod for object storage
+      objectStore:
+        mode: minio
+        port: 9000
+
+      # Deploys a local redis pod
+      inMemoryDb:
+        mode: redis
```
lzap, line 53: Can you remove both? We will unlikely use S3 and I would love to stay away from Redis if we can.

docs/deploy-ephemeral.md
```diff
@@ -0,0 +1,22 @@
...
+
+1. Copy `deploy/bonfire.example.yaml` to `deploy/bonfire.yaml`
+2. set `<path_to_service_dir>` to the path to the local path of this repo
+3. run following
+
+```
+make build-podman
+podman push provisioning-backend:latest quay.io/envision/provisioning-backend
+bonfire deploy --source local --local-config-path ./deploy/bonfire.yaml provisioning --set-image-tag quay.io/envision/provisioning-backend=latest
+```
```
lzap, line 22: I guess direct pushing to envision quay.io account is temporary solution, isn't it?

internal/config/global_config.go
```diff
@@ -1,80 +1,97 @@
...
 
 import (
+	"fmt"
 	"github.com/rs/zerolog"
+	"strings"
 	"sync"
 	"time"
 
-	"github.com/caarlos0/env/v6"
-	"github.com/joho/godotenv"
```
lzap, line 9: Please run `go mod tidy` to remove those from go.mod as well.

internal/config/global_config.go
```diff
@@ -1,80 +1,97 @@
...
+	err = viper.Unmarshal(&config)
 	if err != nil {
 		panic(err)
 	}
-	configInitialized = true
 }
 
 func GetLoggingConfig() *GlobalConfig {
 	configMutex.Lock()
 	defer configMutex.Unlock()
```
lzap, line 93: Remove mutex completely, function `init` is guaranteed to be called during package importing (very early in program loading: https://www.developer.com/languages/inti-function-golang/

I like this approach much better, this looks like a great fit for config loading.

lzap:
Nice, that `go vet` error is being fixed in my PR.

lzap:
Wait I just stumbled upon: https://github.com/RHEnVision/provisioning-backend/pull/8

That one also adds Viper, I guess you accidentally left hunks in this PR. I will re-review there as well.

docs/deploy-ephemeral.md
ezr-ondrej: by default, you're supposed to push a tag calculated from current commit, but I guess we should make a decision for the best solution. Might be necessary to use own quay accounts or at least personalized tags :thinking:

deploy/clowdapp.yaml
ezr-ondrej: nah, I just figured there is no bash so I changed to sh, but we might try running directly, it is probably the best :shrug:

deploy/bonfire.example.yaml
ezr-ondrej: It is a template, that is why it says `example` in the file name, I might add a make command to customize it.

lzap:
FYI I noticed you pushed into this (origin) repo, prefer pushing to your own fork. Maybe we can disable this in the settings?

deploy/clowdapp.yaml
lzap: Just delete sh and -c and just run /pbmigrate. The same for the service itself - no need to have another process running in the container :)

cmd/pbapi/main.go
```diff
@@ -61,14 +61,14 @@ func main() {
...
-	log.Info().Msg("Starting new instance on 8000/8080")
+	log.Info().Msg("Starting new instance on 8000/9000")
 	srv := http.Server{
 		Addr:    fmt.Sprintf(":%d", 8000),
 		Handler: r,
 	}
 
 	msrv := http.Server{
-		Addr:    fmt.Sprintf(":%d", 8080),
+		Addr:    fmt.Sprintf(":%d", 9000),
```
lzap, line 71: Can you extract these into viper variables? When running without containers ports can be occupied by other processes.

deploy/clowdenv.yaml
```diff
@@ -0,0 +1,68 @@
...
+
```
lzap, line 54: Let's remove objectStore and Redis, we are not using them at the moment.

lzap (changes requested):
Couple of remarks, please either add Dockerfile/Containerfile into .gitignore or do me a symlink, there is a bug in podman which prevents me from building the image:

    ln -f build/Dockerfile Containerfile

I am really trying hard to stick with podman :)

deploy/clowdapp.yaml
ezr-ondrej: I've not been able to make the readiness probe working, but might have been another issue, let me try :)

deploy/clowdenv.yaml
ezr-ondrej: I'll extract the clowdenv into a separate PR, because it is not used for Ephemeral, where the env is predefined, so clowdenv is for local minicube only, which I did not tested.

Makefile
```diff
@@ -15,7 +15,8 @@ strip: build
 
 .PHONY: build-podman
 build-podman:
-	podman build .
+	ln -f build/Dockerfile Containerfile
```
lzap, line 18: I thought to commit the symlink to git :) Just remove this line, create such symlink and commit it to the git please, thank you!

internal/config/config.go
```diff
@@ -2,15 +2,19 @@ package config
...
+	"github.com/rs/zerolog"
+
 	clowder "github.com/redhatinsights/app-common-go/pkg/api/v1"
 	"github.com/spf13/viper"
 )
 
 var config struct {
+	Deployment struct {
+		MetricsPort int
+	}
```
lzap, line 17: Hmmm why deployment? All the variables are actually deployment options. I would prefer Metrics.Port here. We should try to keep the service clean of k8s/openshift stuff I think.

internal/routes/all_routes.go
```diff
@@ -1,15 +1,24 @@
...
 import (
+	"net/http"
+
 	s "github.com/RHEnVision/provisioning-backend/internal/services"
 	"github.com/go-chi/chi/v5"
 )
 
 func SetupRoutes(r *chi.Mux) {
+	r.Get("/ping", s.StatusService)
+	r.Mount("/api/provisioning", apiRouter())
```
lzap, line 12: Where this comes from? Is this a SRE/k8s requirement? If not, I would not include 'provisioning' prefix, 'api' is okay why not, but I see no value in this.

lzap (changes requested):
Sorry for the delay, I have some concerns about the prefix and config name.

.gitignore
```diff
@@ -7,3 +7,5 @@ TAGS
 /pbmigrate
 /vendor/
 /configs/local.yaml
+/deploy/bonfire.yaml
+Containerfile
```
lzap, line 11:
```suggestion
/Containerfile
```

Makefile
lzap: Can you add a comment why is that?

internal/routes/all_routes.go
lzap: It is required by SRE.

ezr-ondrej: Yeah, our app is mounted on this prefix and all our routes comming from k8s are under this prefix

lzap (approved):
Nice thanks, I would love to have the mountpoint configurable at some point. But this is good enough.

PR #8, opened 2022-05-14
Use viper for config managemet

ezr-ondrej:
This moves us to using Viper config management as other projects in ConsoleDot.
Viper has a lot of funcionality we do not need, but it can handle ephemeral config easily.
It allows flexibility in the future config management options.

ADR: PROV00002


internal/config/config.go
```diff
@@ -0,0 +1,116 @@
...
+		Stream  string
+	}
+	FeatureFlags struct {
+		Environment string
+	}
+	ExitOnPanic bool
+}
+
+var config GlobalConfig
+var configMutex sync.Mutex
```
lzap, line 43: Just drop the mutex it is no longer necessary. This is cleaner.

internal/config/config.go
```diff
@@ -0,0 +1,116 @@
...
+
+		viper.Set("database.host", cfg.Database.Hostname)
+		viper.Set("database.port", cfg.Database.Port)
+		viper.Set("database.user", cfg.Database.Username)
+		viper.Set("database.password", cfg.Database.Password)
+		viper.Set("database.name", cfg.Database.Name)
+	} else {
+		fmt.Println("Loading local configuration")
+
+		// read missing default environmental values from .env.default
```
lzap, line 66: Looks like you deleted `.env.default` so this is no longer valid statement.

internal/config/config.go
```diff
@@ -0,0 +1,116 @@
...
+		panic(err)
+	}
+}
+
+func Get() *GlobalConfig {
+	configMutex.Lock()
+	defer configMutex.Unlock()
+
+	return &config
+}
```
lzap, line 94: Actually would you mind simply removing the `Get()` alltogether? There is no need for it since it is no longer lazy-initialized. Also remove the Global struct and simply expose the substructs, perhaps move ExitOnPanic to Logging struct. We will need to use configuration a lot and this is much nicer to write:

    config.Database.Password

internal/config/config.go
ezr-ondrej: :+1: thanks

internal/config/config.go
ezr-ondrej: I've tried my best, do you think this is good enough implementation? :)

Makefile
```diff
@@ -43,4 +43,4 @@ migrate: build-pbmigrate
 .PHONY: generate-migration
 MIGRATION_NAME?=unnamed
 generate-migration:
-	migrate create -ext sql -dir api/migrations -seq -digits 3 $(MIGRATION_NAME)
\ No newline at end of file
+	migrate create -ext sql -dir api/migrations -seq -digits 3 $(MIGRATION_NAME)
```
lzap, line 46:
```suggestion
	migrate create -ext sql -dir internal/db/migrations -seq -digits 3 $(MIGRATION_NAME)
```

@adiabramovitch spotted this, since you are touching this can you also fix it to prevent merge conflict :)

internal/config/config.go
```diff
@@ -0,0 +1,121 @@
...
+	Logging LoggingT
+	Cloudwatch CloudwatchT
+	Features FeatureFlagsT
+}
+
+var config GlobalConfigT
+var Database *DatabaseT
+var Logging *LoggingT
+var Cloudwatch *CloudwatchT
+var Features *FeatureFlagsT
```
lzap, line 51: Uh those T suffixes are ugly, why you have them? It's completely okay to have the very same variable and type name, it's actually a very good design pattern to follow.

Also can you drop `GlobalConfigT` type and `config` variable completely? They appear to be unused, at least I don't see any use.

internal/config/config.go
```diff
@@ -0,0 +1,121 @@
...
+
+func init() {
+	var err error
+
+	viper.SetDefault("logging.level", 1)
+	viper.SetDefault("database.host", "localhost")
+	viper.SetDefault("database.port", 5432)
+	viper.SetDefault("database.loglevel", 1)
+	viper.SetDefault("cloudwatch.enabled", false)
```
lzap, line 60: These default values do not match with the YAML file, what is the source of "truth" in this design? I would prefer to have just a single place where we keep all the defaults. If this is the place and YAML is not, then please comment all the lines in the YAML file so its all just comments so we indicate to the user that these are all just examples. Maybe you can drop a comment in the YAML that the effective default values are in this file.

internal/config/config.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if clowder.IsClowderEnabled() {
+		fmt.Println("Loading Clowder configuration")
```
lzap, line 63: I would prefer to to do any standard output in the init() functions, unless you insist. For errors you do panic and that is fine, but for success path I would prefer not to write anything.

internal/config/config.go
lzap: It's great except the naming and unused `config` variable, see above.

lzap:
Ok I have few comments on defaults and global variable config. Also feel free to fix the go vet line:

    log.logger.Info().Msgf(format, v)

should be

    log.logger.Info().Msgf(format, v...)

I will rebase on top of this.

Makefile
ezr-ondrej: I've dropped the change as I've just added the endline.

internal/config/config.go
ezr-ondrej: I've commented out the yaml file

internal/config/config.go
ezr-ondrej: The only I'd like to keep is the one informing you there is no `local.yaml` I do not have better idea how to inform user it might be a good idea to do in devel.

ezr-ondrej:
I've fixed both the typos in https://github.com/RHEnVision/provisioning-backend/pull/11

configs/defaults.yaml
```diff
@@ -0,0 +1,35 @@
...
+#   group: group
+#   stream: stream
+
+# When panic error is not rescued, exit the application instead of returning 500.
+
+# Feature flags:
+# development or production
+# feature_flags:
+#   environment: development
+#   exit_on_panic: false
```
lzap, line 35: This needs to be moved to line 30.

internal/config/config.go
```diff
@@ -0,0 +1,118 @@
...
+func DumpConfig(logger zerolog.Logger) {
+	if IsProduction() {
+		logger.Warn().Msg("dumping configuration in production mode")
+	}
+	configCopy := config
+	configCopy.Database.Password = "*****"
+	// configCopy.AWS.Key = "*****"
+	// configCopy.AWS.Secret = "*****"
+	// configCopy.AWS.Session = "*****"
+	logger.Info().Msgf("%+v", configCopy)
```
lzap, line 117: Oh for AWS (CloudWatch) we still use ENV variables, its okay for now. I will refactor that later.

configs/defaults.yaml
ezr-ondrej: Why? I've put that to feature_flags in config as well

internal/config/config.go
ezr-ondrej: alright, I've figured how to do that, it's working now :tada:

internal/config/config.go
ezr-ondrej: Yeah, if we could merge and follow-up that would be great, I'd love to unblock #7

configs/defaults.yaml
lzap: Oh okay than move the comment to the appropriate place? I mean comments for flags are off.

lzap (approved):
WFM TY

PR #9, opened 2022-05-15
Add Implementation of sources table and dao

adiabramovitch:


api/migrations/00001_create_sources_table.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS sources(
+id SERIAL
+source_id  INT PRIMARY KEY,
+name VARCHAR(100) NOT NULL,
+auth_id INT, 
+org_id INT,
```
ezr-ondrej, line 6: Should we have index on `org_id`?

adiabramovitch: Yes :)
 I have a few bugs I am fixing and then I'll push the updated version

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,95 @@
...
+	"fmt"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/db"
+	m "github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/jmoiron/sqlx"
+)
+
+const (
+	create     = `INSERT INTO sources (source_id, name, org_id, auth_id) VALUES ($1, $2, $3, $4) RETURNING created_at`
```
lzap, line 14: At any time you can opt-in to sqlx.NamedStatement where you can use `:this, :style, :of, :arguments` but I am fine with this. Anything up to let's say five is still readable.

It is typically good idea to also put ID to RETURNING clause so you also get ID as well so the value is returned back to the payload with ID.

adiabramovitch:
So I think before I push the service itself I want to get your input so far. That way I can make the changes to the logic itself :). @lzap @ezr-ondrej @amirfefer

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,95 @@
...
+	getById    = `SELECT * FROM sources WHERE source_id = $1;`
+	list       = `SELECT * FROM sources ORDER BY id LIMIT 100 OFFSET $1`
```
lzap, line 16: Yeah LIMIT is hardcoded for now, we can fix that later and make this a URL parameter or something, not sure how this is done in OpenAPI so let's just hardcode 100 for now.

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,95 @@
...
+	getByOrgId = `SELECT * FROM sources WHERE org_id = $1 ORDER BY id LIMIT 100;`
```
lzap, line 17:
```suggestion
	getByOrgId = `SELECT * FROM sources WHERE org_id = $1 ORDER BY id LIMIT 1;`
```

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,95 @@
...
+	daoImpl.getByOrgId, err = db.DB.PreparexContext(ctx, getByOrgId)
+	if err != nil {
+		fmt.Println(err.Error())
+		return nil, err
+	}
+	return &daoImpl, nil
+}
+
+func init() {
+	println("Importing dao/sqlx source")
```
lzap, line 57: This was just a debugging message I accidentally pushed, please remove :)

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,95 @@
...
+	GetSourceDao = getSourceDao
+}
+
+func (s *sourceDaoSqlx) Create(source *m.Source) error {
+	_, err := db.DB.Exec("INSERT INTO sources (source_id, name, org_id, auth_id) VALUES ($1, $2, $3, $4) RETURNING created_at", source.SourceID, source.Name, source.OrgID, source.AuthID)
+	//_, err := s.create.Exec(source)
```
lzap, line 63: Is there a problem? I haven't tried insert yet, I guess it should work this way...

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (s *sourceDaoSqlx) List(offset uint64) ([]*m.Source, error) {
+	result := []*m.Source{}
+	err := s.list.Select(&result, offset)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
```
lzap, line 86: Let's perhaps prepare our DAO API for both limit and offset and for now we can hardcode it to 100/0 in the service function.

internal/db/migrations/002_create_sources_table.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS sources;
```
lzap, line 1: FYI I propose to not do down migrations in my other PR, so we can just delete this one later on. I don't think it is worth it.

internal/dao/sqlx/source_dao.go
adiabramovitch: Do you mean to have the limit as a parameter as we did in offset?

internal/payloads/ami_payload.go
```diff
@@ -0,0 +1,48 @@
+package payloads
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/RHEnVision/provisioning-backend/internal/helpers"
+)
```
lzap, line 8: No `helper` or `utils` or `tools` please. Just come up with some proper package name :)

internal/dao/sqlx/source_dao.go
adiabramovitch: > At any time you can opt-in to sqlx.NamedStatement where you can use `:this, :style, :of, :arguments` but I am fine with this. Anything up to let's say five is still readable.

Can you give me an example? I am not sure I understood

internal/db/migrations/002_create_sources_table.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE sources
+(
+    id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    auth_id TEXT NOT NULL UNIQUE CHECK (NOT empty(auth_id)),
+    name TEXT UNIQUE CHECK (NOT empty(name)),
+    source_id TEXT NOT NULL UNIQUE CHECK (NOT empty(source_id)),
+    org_id TEXT NOT NULL UNIQUE CHECK (NOT empty(org_id)),
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+
+);
```
lzap, line 10: I my other PR I create a new table accounts as well, we need a foreign key reference from this table, this will be 1:1 association. Obviously you can only do this when my PR is merged so fine for the POC for now.

internal/models/source.go
```diff
@@ -0,0 +1,10 @@
+package models
+
+type Source struct {
+	ID        uint64 `db:"id"`
+	SourceID  string `db:"source_id"`
+	Name      string `db:"name"`
+	OrgID     string `db:"org_id"`
+	AuthID    string `db:"auth_id"`
+	CreatedAt string `db:"created_at"`
+}
```
lzap, line 10: So I am already confused what source_id and auth_id is at this point, I suggest that from now on we will document all models pretty extensively. I suggest to have documentation for the model itself (what it represents) as well as for each individual columns. @adiabramovitch @ezr-ondrej @amirfefer

internal/payloads/ami_payload.go
```diff
@@ -0,0 +1,48 @@
...
+					Type: "aws",
+					Options: Options{
+						ShareWithAccounts: []string{os.Getenv("ACCOUNT_ID")},
+					},
+				},
+			},
+		},
+	}
+	payloadBytes, err := json.Marshal(data)
+	helpers.PanicOn(err)
```
lzap, line 46: Let's avoid these "generic error handler helpers", the proper way to do this is really check for error via `if` and pass error as a return value from this function. Welcome to the Go world :)

Panic should only be used if you would create an API that will also call `defer` and then translate it to an error, which is not this case: https://go.dev/doc/effective_go#panic

internal/payloads/create_source_payload.go
```diff
@@ -0,0 +1,62 @@
...
+		],
+		"authentications": [
+				{
+						"resource_type": "Application",
+						"resource_name": "provisioning",
+						"username": "%s",
+						"authtype":"provisioning-arn"
+				}
+		]
+	}`, sourceName, sourceName, ARN)
```
lzap, line 58: There is literally "%s"? Just checking, smells like some printf formatting issue :)

internal/payloads/ami_payload.go
adiabramovitch: Yes It is a leftover I forgot to remove, I'll use the logger, but saw that I have cyclic import issue which I need to solve

internal/payloads/create_source_payload.go
adiabramovitch: there isn't an issue from what I saw, this is the way it works.
When I tried %s without the quotes it didn't recognize the value of the parameter.

internal/payloads/ami_payload.go
```diff
@@ -0,0 +1,48 @@
...
+	return string(payloadBytes)
+}
```
lzap, line 48: So `payloads` package was meant for our OpenAPI payloads only, this feel like this is a request into sources API? I think we should create a new package for this. I suggest `clients`, there you can create subfolder for each service. Looks like edge-api has this:

```
[lzap@nuc pkg]$ find clients/
clients/
clients/client.go
clients/clients_suite_test.go
clients/clients_test.go
clients/fdo
clients/fdo/client.go
clients/fdo/client_test.go
clients/fdo/fdo_suite_test.go
clients/imagebuilder
clients/imagebuilder/client.go
clients/imagebuilder/client_test.go
clients/imagebuilder/mock_imagebuilder
clients/imagebuilder/mock_imagebuilder/client.go
clients/inventory
clients/inventory/client.go
clients/inventory/mock_inventory
clients/inventory/mock_inventory/inventory.go
clients/playbookdispatcher
clients/playbookdispatcher/client.go
```

They have just a single go file per client, which is not good practice. They also have the same name `client.go` which makes navigation by typing file name harder. I would propose this:

* `internal/clients/sources/sources_client.go` - CRUD REST functions
* `internal/clients/sources/sources_client_test.go` - tests (edge-api use mocking)
* `internal/clients/sources/sources_payloads.go` - structs

Something like that.

internal/payloads/create_source_payload.go
```diff
@@ -0,0 +1,62 @@
...
+
+	body := strings.NewReader(reqStr)
+	return body
```
lzap, line 61: Wow, okay okay. I have no problems with that, this is also fast :) Just make sure arguments are JSON-escaped.

internal/payloads/error_payload.go
```diff
@@ -27,30 +32,40 @@ func (e *ErrResponse) Error() error {
...
-		Message:        "Error rendering response.",
+		RequestId:      ctx.Value(ctxval.RequestIdCtxKey).(string),
+		HTTPStatusCode: 500,
+		Message:        "Cannot fetch record from DB",
 	}
 }
 
 var ErrNotFound = &ErrResponse{HTTPStatusCode: 404, Message: "Resource not found"}
-var ErrParamParsingError = &ErrResponse{HTTPStatusCode: 500, Message: "Cannot parse parameters"}
-var ErrDeleteError = &ErrResponse{HTTPStatusCode: 500, Message: "Cannot delete resource"}
```
lzap, line 56: Yeah errors package needs more love, I am on that in my other PR.

lzap:
Great work! Looks very nice.

internal/payloads/ami_payload.go
adiabramovitch: Great Idea! thanks. 
should I create one for AWS as well?

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (s *sourceDaoSqlx) GetByOrgId(orgId uint64, limit uint64) ([]*m.Source, error) {
+	result := []*m.Source{}
+	err := s.getByOrgId.Select(&result, orgId, limit)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
```
adiabramovitch, line 94: I have a problem here, maybe you would know why it happens. 
I am running: 
```
var orgID uint64 = 1
var limit uint64 = 1
sourcesArr, err := sourceDao.GetByOrgId(orgID, limit)
```
and I am getting : 
`error="cannot convert 1 to Text"
`
Why is that a problem? I saw you did the same.

internal/dao/sqlx/source_dao.go
lzap: Its okay no need for a change, just for your info: http://jmoiron.github.io/sqlx/#namedParams

internal/dao/sqlx/source_dao.go
lzap: Yes see here: https://github.com/RHEnVision/provisioning-backend/pull/6/files#diff-c01fb12b3a88fd02171a0823d52829763a9e9a0a29698c3fecf3faecbdfce3dbR14

internal/payloads/error_payload.go
lzap: https://github.com/RHEnVision/provisioning-backend/pull/6/files#diff-4b50e8af15cd6a0dc76bc19957e33530d7bc3d431b0c1c16f1ad022dbfdca2a7

go.mod
```diff
@@ -3,10 +3,13 @@ module github.com/RHEnVision/provisioning-backend
 go 1.16
 
 require (
+	github.com/aws/aws-sdk-go v1.17.7
```
lzap, line 6: What adds AWS SDK v1? We should only use v2, maybe this is a leftover? Try to do `go mod tidy`.

internal/clients/ec2/ec2_client.go
```diff
@@ -0,0 +1,38 @@
...
+		return "", err
+	}
+
+	for _, instance := range resultRegions.Regions {
+		instanceJSON, _ := json.MarshalIndent(instance, "", "  ")
+		result = result + string(instanceJSON)
+	}
+
+	return result, err
+}
```
lzap, line 38: Yeah, I renamed clouds/ to clients/ so it matches your package. And I am changing my mind about the package name - let's use aws/ instead of ec2/ because we can put both ec2 and cloudwatch APIs into the same connection (client interface).

https://github.com/RHEnVision/provisioning-backend/pull/13

internal/clients/imagebuilder/image_builder_client.go
```diff
@@ -0,0 +1,84 @@
...
+	var err error
+
+	req, err = http.NewRequest(method, endpoint, body)
+	if err != nil {
+		log.Println("Error : something terrible happen -> ", err)
+		return nil, err
+
+	}
+
+	req.SetBasicAuth(os.Getenv("RED_HAT_USER"), os.Getenv("RED_HAT_PASS"))
```
lzap, line 24: You can now use the config structure, is this how authentication will be done in production tho?

internal/clients/ec2/ec2_client.go
lzap: Correction: let's actually use ec2 and cw separately, there is a big difference between the two - cw is connected for the whole application globally while ec2 we will create on the fly. So keep ec2 here, I will rename it.

ezr-ondrej: I think we can't use the same connection, cloudwatch will have different credentials then the ec2 connection IMHO :thinking:

internal/clients/imagebuilder/image_builder_client.go
ezr-ondrej: No, this is probably not correct, I'm testing it out in ephemeral and would love to get something similar for local, but we will get this info in a header and we should pass that onto an image builder, so image builder has the same info as we do.
I'll work on stubbing this header out in local env.

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,94 @@
...
+	daoImpl.create, err = db.DB.PreparexContext(ctx, createSource)
+	if err != nil {
+		fmt.Println(err.Error())
+		return nil, err
+	}
+	daoImpl.getByOrgId, err = db.DB.PreparexContext(ctx, getSourceByOrgId)
+	if err != nil {
+		fmt.Println(err.Error())
+		return nil, err
+	}
```
lzap, line 52: I refactored error handling, see in 'main' for good examples on how to do this.

internal/dao/sqlx/source_dao.go
lzap: See the 'main' branch for working INSERT example.

internal/models/source.go
ezr-ondrej: I'm totally up for it, is there some standards for documenting Go code inline?

internal/clients/ec2/ec2_client.go
adiabramovitch: Is it something we should add to the config? like creds for cloudwatch and for others?

internal/clients/imagebuilder/image_builder_client.go
adiabramovitch: I am positive it is not the correct way in general but for the PoC I authorized like that. 
Should I change it?

lzap:
It is very important to get the data model right, please let's not rush this is. I can take a look in the morning.

internal/services/sources_service.go
```diff
@@ -0,0 +1,132 @@
...
+		renderError(w, r, payloads.NewRenderError(r.Context(), "create source in db", err))
+		return
+	}
+
+	if err := render.Render(w, r, payloads.NewSourceResponse(source)); err != nil {
+		renderError(w, r, payloads.NewRenderError(r.Context(), "source", err))
+	}
+}
+
+func DeleteSource(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
ezr-ondrej: how! that's high level :) ok I'll take it :)

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (dao *sourceDaoSqlx) NameForError() string {
+	return dao.name
+}
+
+func init() {
+	dao.GetSourceDao = getSourceDao
+}
+
+func (dao *sourceDaoSqlx) Create(ctx context.Context, source *models.Source) (*models.Source, error) {
```
ezr-ondrej, line 63: No idea why linter doesn't complain here, but please rename `dao` to something else so we do not shadow the package `dao`.

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,124 @@
...
+	query := createSource
+	stmt := dao.create
+	err := stmt.GetContext(ctx, source, source.AccountId, source.SourceID, source.Name, source.TenantID, source.AuthID)
```
ezr-ondrej, line 66: You're not using the cool feature of sqlx scanning Lukas has described ^_^

ezr-ondrej, line 66: You're not using the cool feature of sqlx scanning Lukas has described ^_^

internal/models/source_model.go
```diff
@@ -0,0 +1,23 @@
+package models
+
+type Source struct {
+	// Required auto-generated PK.
+	ID uint64 `db:"id" json:"id"`
+
+	// Source ID in sources. Unique.
+	SourceID uint64 `db:"source_id" json:"source_id"`
```
ezr-ondrej, line 8: Do we need a `SourceID`, when we have the `AuthID`? In my mind I can query sources api just with the AuthID and it tells me the id of the source as well.
Is this to distinguish what provider this auth is for? If so, I'd rather used some internal identifier (e.g. [reuse what Lukas has](https://github.com/RHEnVision/provisioning-backend/blob/139ed1a25c67ae8d5d4a5e1da8460f25c445283c/internal/models/model_constants.go) ) and map it from sources

ezr-ondrej, line 8: Do we need a `SourceID`, when we have the `AuthID`? In my mind I can query sources api just with the AuthID and it tells me the id of the source as well.
Is this to distinguish what provider this auth is for? If so, I'd rather used some internal identifier (e.g. [reuse what Lukas has](https://github.com/RHEnVision/provisioning-backend/blob/139ed1a25c67ae8d5d4a5e1da8460f25c445283c/internal/models/model_constants.go) ) and map it from sources

ezr-ondrej (commented):
The code looks good, but I do not fully understand the data structure, but I'll be happy to get educated why we need it this way :)

internal/models/source_model.go
adiabramovitch: from what I have checked the only thing needed is the auth id but Mikel from sources told me to store the source id as well. I will check it again with him.

internal/models/source_model.go
```diff
@@ -0,0 +1,23 @@
...
+
+	// User-facing name. Required.
+	Name string `db:"name" json:"name"`
+
+	// Tenant ID. Required.
+	TenantID uint64 `db:"tenant_id" json:"tenant_id"`
```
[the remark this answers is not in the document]
adiabramovitch: So I am not sure either. 
from the PoC I have made, the only thing we need in order to fetch a source is the **authentication Id**  and in order to authenticate we need to have an  x-rh-identity header which is explained here under headers: https://consoledot.pages.redhat.com/docs/dev/services/sources/integration.html#_prepare_an_endpoint_for_the_availability_checking

I am not sure the full meaning of tenant/account. I know we should have the Org Id in order to authenticate and it is one of the tenant attributes in sources.

internal/dao/sqlx/source_dao.go
adiabramovitch: Do you mean like that? 
`err := stmt.GetContext(ctx, source, source)`

ezr-ondrej: Yeah, he is not using it very often :) it is no biggie :) we can even mix it, I don't care too much.
This is just bit more writing, but I'm totatly fine leaving it as is ^_^

internal/models/source_model.go
lzap: Trying out to sort our my understanding, so this is source id: https://github.com/RedHatInsights/sources-api-go/blob/main/model/source.go#L24

internal/models/source_model.go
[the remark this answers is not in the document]
lzap: This is source's internal "account" id: https://github.com/RedHatInsights/sources-api-go/blob/main/model/source.go#L41

@ezr-ondrej it lives here https://github.com/RedHatInsights/sources-api-go/blob/main/model/tenant.go and they have PK, org id and EBS account which is named "external_tenant" for some reason.

internal/dao/sqlx/source_dao.go
adiabramovitch: I think it is more readable but how ever you prefer :)

internal/models/source_model.go
```diff
@@ -0,0 +1,23 @@
...
+
+	// Associated Account model. Required.
+	AccountId uint64 `db:"account_id" json:"account_id"`
+
+	// Authentication ID in sources lined to a spesific source in order to fetch the ARN
+	AuthID string `db:"auth_id" json:"auth_id"`
```
lzap, line 20: And this is their text-id if I am not mistaken: https://github.com/RedHatInsights/sources-api-go/blob/main/model/authentication.go#L17

I am fine with storing this as text if that's what they recommend.

internal/models/source_model.go
```diff
@@ -0,0 +1,23 @@
...
+	SourceID uint64 `db:"source_id" json:"source_id"`
+
+	// User-facing name. Required.
+	Name string `db:"name" json:"name"`
+
+	// Tenant ID. Required.
+	TenantID uint64 `db:"tenant_id" json:"tenant_id"`
+
+	// Associated Account model. Required.
+	AccountId uint64 `db:"account_id" json:"account_id"`
```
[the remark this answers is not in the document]
adiabramovitch: I think I need an explanation here
 what to the account Id here stands for, and how does it relate to sources? 
 In sources we want to generate the identity header from the EBS and org Id, why do we need the account as well? what is the difference?

internal/models/source_model.go
adiabramovitch: it is stored here as a text

internal/models/source_model.go
adiabramovitch: checked it, we need the source id in order to delete it from sources. by deleting the authentication the source isn't deleted automatically.

internal/models/source_model.go
lzap: Actually from the code it looks like authid is actually UUID. I will ask on the channel for confirmation, but if that's the case we could save ton of data by using the native UUID postgres type. Unfortunately the standard driver does only support storing UUID as text or 16 bytes, however, native pgx driver would allow us to use the native type.

I am making a note for my "Switch to Scanny/PGX driver" proposal about this, that is significant saving for data and indices. Let's keep it as string for now.

internal/dao/sqlx/source_dao.go
ezr-ondrej: Then let's leave it as is :+1:

internal/models/source_model.go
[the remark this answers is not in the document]
ezr-ondrej: oh right, they are sending account_id as tenant, so we can map that through our accounts table on `account_number = tenant`

internal/models/source_model.go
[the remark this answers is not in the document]
ezr-ondrej: that's what I'm saying as well under the tenant. if tenant = account_number, we can only keep account_id refering to our table and map it through `tenant = account_number`

internal/models/source_model.go
ezr-ondrej: ok, let's keep the source, but can we map sourceID to provider type? can we distinguish between auth for azure and auth for amazon?

internal/dao/dao_interfaces.go
```diff
@@ -33,3 +33,13 @@ type PubkeyResourceDao interface {
...
 	Delete(ctx context.Context, id uint64) error
 }
+
+var GetSourceDao func(ctx context.Context) (SourceDao, error)
+
+type SourceDao interface {
+	GetById(ctx context.Context, id uint64) (*models.Source, error)
+	List(ctx context.Context, limit uint64, offset uint64) ([]*models.Source, error)
+	Create(ctx context.Context, s *models.Source) (*models.Source, error)
+	GetByAccountId(ctx context.Context, accountId uint64, limit uint64, offset uint64) ([]*models.Source, error)
```
adiabramovitch, line 43: Maybe `ListByAccountId` is better?

lzap:
Let's just have a call with Sources API for their advice on what exactly are we supposed to keep in our DB. I would like to put as high resistance as we can to avoid storing duplicate data, maybe we can get away just with calling their API and in-memory LUR cache.

internal/models/source_model.go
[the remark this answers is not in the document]
lzap: I agree, let's get more info from sources guys on what exactly are we supposed to store in our db (if anything) and then we can decide. If we ever store anything, we will associate our ACCOUNTS table with this which will provide ACCOUNT_NUM/ORG_ID for all records automatically, there is no point in storing their tenant_id as it is the same data. But let's see what they have to say to this.

internal/dao/dao_interfaces.go
lzap: Yeah for slices we should name DAO functions ListXYZ.

internal/models/source_model.go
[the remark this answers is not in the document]
lzap: Yes, Sources Tenant model = our Account model.

lzap:
Just for the record, we will call Sources API on-demand in the initial version, later on we will decide what exactly to store in the db.

PR #12, opened 2022-05-17
Create pubkey model and a POC REST API

lzap:
Creates a pubkey model, table, migration and POC API. This is not OpenAPI yet!

Also fixed configuration camel case and added missing AWS credentials config.

Please test before merging.

**Checklist**


README.md
```diff
@@ -87,7 +87,7 @@ Here are few points before you start contributing:
 * Do not introduce `utils` or `tools` common packages.
 * Keep the line of sight (happy code path).
 * PostgreSQL version we currently use in production is v14+ so take advantage of all modern SQL features.
-* Use `BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY` for primary keys.
+* Use `BIGINT GENERATED BY DEFAULT AS IDENTITY (START 101) PRIMARY KEY` for primary keys. This allows us to use 1-100 in our seed data (see above).
```
lzap, line 90: I found an issue with our seed script, GENERATED BY allows explicit creation of PKs which we take advantage of there (so you can create meaningful test data), however sequence will start with 1 which will lead to error for the very first INSERT statement, this starts sequence from 101 so we have 100 primary key numbers to work with. it is a change of existing migration which we should not generally do but I would like to do it for this case so its not incorrectly done in the history.

lzap:
Fixed bunch of problems, further improved error logging. This introduces some migration changes, I will ask you to wipe out database and migrate over. I would like to have our initial migrations to be clean and correct if you don't mind :)

README.md
ezr-ondrej: I do not like to be a party booper, but I do not love the resolution of that issue by lets just start at random high number. It could easily lead to issues again once we have 100 records and seems like quite hidden issue.

Wouldn't it be easier to just `ALTER SEQUENCE serial RESTART WITH <last seed +1>;` after the seed data? It is easily editable while adding new seeds and clear why it is there :thinking:

internal/db/seeds/dev_small.sql
```diff
@@ -1,15 +1,20 @@
...
-    ('13', '000013'),
-    ('15', '000015'),
-    (NULL, '000042'),
-    (NULL, '000077')
+    (1, '13', '000013'),
+    (2, '15', '000015'),
+    (3, NULL, '000042'),
+    (4, NULL, '000077')
+ON CONFLICT DO NOTHING;
+
```
ezr-ondrej, line 14: Would this work?
```sql
ALTER SEQUENCE accounts_id_seq RESET 5;
```

lzap: Yes, but we will forget to do this for sure, then its a runtime error.

ezr-ondrej: In devel tho, right? and it is easy to discover, tho if we ever hit 100 seeds it would be IMHO quite hard to discover what is going on...

README.md
lzap: I considered this too, decided to do the start approach because I felt like it is safer. I implemented a function that resets everything automatically, so no need to worry about this anymore.

lzap:
Pushed a new version, there is a new function that is called from the seed script which will automatically correct sequences for all tables. This way, we will never forget as this will work out of box.

Any other comments @ezr-ondrej @adiabramovitch ?

internal/db/seeds/dev_small.sql
```diff
@@ -1,15 +1,22 @@
...
+       (4, NULL, '000077')
 ON CONFLICT DO NOTHING;
 
+INSERT INTO pubkeys(id, account_id, name, body)
+VALUES (1, 1, 'lzap-ed25519-2021',
+        'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap')
+ON CONFLICT DO NOTHING;
+
+-- Reset all primary key sequences (columns named "id") to the maximum value.
+SELECT reset_sequences();
```
ezr-ondrej, line 20: briliant ^_^

ezr-ondrej (approved):
Thanks @lzap, lets get this in! :+1:

PR #13, opened 2022-05-18
Rename clouds/ package to clients/ so we can add sources too

lzap:
This prepares for Adi's PR: https://github.com/RHEnVision/provisioning-backend/pull/9


lzap:
Do not merge yet, I want to also rename "subpackage" and implement key upload.

PR #14, opened 2022-05-18
Configurable mounting point

ezr-ondrej:
Please advice on the config option names and nesting :cry: 

**Checklist**

### DIFF
--- a/cmd/pbapi/main.go
+++ b/cmd/pbapi/main.go
@@ -63,11 +63,11 @@ func main() {
 	// Routes for metrics
 	mr := chi.NewRouter()
 	mr.Get("/", statusOk)
-	mr.Handle("/metrics", promhttp.Handler())
+	mr.Handle(config.Prometheus.Path, promhttp.Handler())
 
-	log.Info().Msgf("Starting new instance on port 8000 with prometheus on %d", config.Prometheus.Port)
+	log.Info().Msgf("Starting new instance on port %d with prometheus on %d", config.Application.Port, config.Prometheus.Port)
 	srv := http.Server{
-		Addr:    fmt.Sprintf(":%d", 8000),
+		Addr:    fmt.Sprintf(":%d", config.Application.Port),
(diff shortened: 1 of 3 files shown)


cmd/pbapi/main.go
```diff
@@ -64,11 +64,11 @@ func main() {
 	// Routes for metrics
 	mr := chi.NewRouter()
 	mr.Get("/", statusOk)
-	mr.Handle("/metrics", promhttp.Handler())
+	mr.Handle(config.Prometheus.Path, promhttp.Handler())
 
-	log.Info().Int("webPort", 8000).Int("metricsPort", config.Prometheus.Port).Msg("Starting new instance")
+	log.Info().Int("webPort", config.Application.Port).Int("metricsPort", config.Prometheus.Port).Msg("Starting new instance")
```
lzap, line 69: Could you actually drop these ints from the logging message here? Or explain what is the benefit. Structured logs are good for searching (give me all logs for a particular request, account, user, organization etc).

internal/config/config.go
```diff
@@ -58,12 +65,16 @@ var Features = &config.FeatureFlags
 func init() {
 	var err error
 
+	viper.SetDefault("app.name", "provisioning")
+	viper.SetDefault("app.pathPrefix", "api")
```
lzap, line 69: I find this a bit unreadable, would you mind refactoring this to just app.prefix being simply "/api/provisioning"? Ideally this value would have a validator that it is actually an absolute path.

lzap (changes requested):
Looks good, for better transparency I suggest to go with simply "prefix".

internal/config/config.go
ezr-ondrej: well app name is configured through Clowder, but the prefix (api) is "fixed" so I'd rather go with having static prefix and configure only the name if that would look better

lzap (approved):
Thanks.

PR #15, opened 2022-05-19
Add image-builder client

adiabramovitch:


internal/clients/imagebuilder/image_builder_client.go
```diff
@@ -0,0 +1,80 @@
...
+
+const URL = "https://console.redhat.com/api/image-builder/v1/compose"
+
+func RequestImageBuilder(client *http.Client, endpoint string, method string, orgId string, body *strings.Reader) ([]byte, error) {
+	var req *http.Request
+	var err error
+
+	req, err = http.NewRequest(method, endpoint, body)
+	if err != nil {
+
```
adiabramovitch, line 20: @lzap So I saw you used regular `fmt.ErrorOf in /clouds.`
how do you want to deal with logging errors? 
(This function will be called as part of the sources service)
 The same question goes for sources_client.go

internal/clients/imagebuilder/image_builder_client.go
```diff
@@ -0,0 +1,80 @@
...
+		return err
+	}
+
+	fmt.Println("the AMI's ID is:", string(req))
+	return nil
+}
+
+func createAMIrequest() (string, error) {
+	data := AmiResponse{
+		Distribution: "rhel-84",
```
Gundersanne, line 60: So this is a quirk on our end, but the non-GA images shouldn't really be used as they're not EUS (long term support) images and aren't updated anymore. They're essentially frozen-in-time rhel 8. For technical reasons we don't have rolling rhel-8/9 distros, but we could add the alias in the image-builder layer. 

This should still build of course, but as soon as you run dnf update you'll get rhel-86 (current GA) and there will be a lot to update.

internal/clients/imagebuilder/image_builder_client.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strings"
+)
+
+const URL = "https://console.redhat.com/api/image-builder/v1/compose"
```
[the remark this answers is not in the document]
adiabramovitch: as a struct or a constant?

internal/clients/imagebuilder/image_builder_payloads.go
```diff
@@ -0,0 +1,19 @@
...
+}
+type UploadRequest struct {
+	Type    string  `json:"type"`
+	Options Options `json:"options"`
+}
+type ImageRequests struct {
+	Architecture  string        `json:"architecture"`
+	ImageType     string        `json:"image_type"`
+	UploadRequest UploadRequest `json:"upload_request"`
+}
```
Gundersanne, line 19: You can generate these structs based on the spec file. See https://github.com/osbuild/image-builder/blob/main/internal/composer/client.go#L1 for an example of the magic `// go:generate` comment.  You still need to download the actual spec for that though.

For that I'd recommend to work against the spec currently in production (https://console.redhat.com/api/image-builder/v1/openapi.json). If you pick the production spec (vs stage or upstream), you can upgrade/release as you please as backward compatibility is guaranteed within a major version, but not forward compatibility (relying on new features/endpoints not in production yet).

internal/clients/imagebuilder/image_builder_client.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("the AMI's ID is:", string(req))
+	return nil
+}
+
+func createAMIrequest() (string, error) {
+	data := AmiResponse{
```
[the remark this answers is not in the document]
Gundersanne: Note that the image name doesn't actually have any effect in the end result. The ami itself will be named composer-api-$uuid. It's a purely consoledot presentation thingy.

internal/clients/imagebuilder/image_builder_client.go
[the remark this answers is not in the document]
lzap: As a config option I believe, I assume that URL of stage cluster is different (console.stage.redhat.com).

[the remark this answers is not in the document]
lzap: The idea is to use IB directly, so I assume we need to make the URL configurable and then OpenShift will fill the right URL in for us? Or is that a constant for all environments?

internal/clients/imagebuilder/image_builder_client.go
lzap: So in my current proposal I have a type named ec2.Client which encapsulates AWS SDK client connection, Context as well as Zerolog logger from the context. Let's definitely use the logger from context which also has context information (request id etc). For now, I am using fmt.Errorf for simplicity, but going forward I think we should declare our own error types because clients will eventually do error out at some point and we need to handle them nicely.

https://github.com/RHEnVision/provisioning-backend/pull/16/files#diff-a2ed92d8fa157beb7e8104db92ae53957c91a288f949aea765823bdb2025a651

internal/clients/imagebuilder/image_builder_client.go
```diff
@@ -0,0 +1,80 @@
...
+	var err error
+
+	req, err = http.NewRequest(method, endpoint, body)
+	if err != nil {
+
+		return nil, err
+
+	}
+
+	req.Header.Set("x-rh-identity", orgId)
```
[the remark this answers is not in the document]
lzap: We should probably not use the gateway for accessing IB, therefore no identity will be necessary.

But important thing is log correlation - is there any kind of correlation id supported across all services?

internal/clients/imagebuilder/image_builder_payloads.go
lzap: Whatever we can achieve by generation, we should use generation. @adiabramovitch please create Makefile targets for these.

lzap:
Is this breaking up this PR: https://github.com/RHEnVision/provisioning-backend/pull/9 ?

ezr-ondrej:
> Is this breaking up this PR: #9 ?

Yes, this got extracted from #9 to make it smaller :)

lzap:
Ok my PR was merged so you should now have the ec2 client with logging. See the comments above.

internal/clients/imagebuilder/image_builder_client.go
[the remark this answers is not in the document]
Gundersanne: It's an openshift feature, for openshift services it's `<service>.<pod_namespace>.svc.cluster.local`; so that would be `image-builder.image-builder-(prod|stage).svc.cluster.local` (initial url I gave isn't correct).

The url can be mapped using env variables, via openshift template parameters per namespace in app-intferface (they're predefined there).

lzap:
So this is the confit that worked for me:

```yaml
package: ib
generate:
  models: true
  client: true
```

Then create a makefile target with something like this:

```
wget https://raw.githubusercontent.com/osbuild/image-builder/main/internal/v1/api.yaml --force
oapi-codegen -config config.yaml api.yaml
```

This way we can generate all 3rd party clients, assuming OpenAPI generator generates actually working code. I bet there will be some edge cases.
```

File: configs/ib_config.yaml [hunk 7]
```diff
@@ -0,0 +1,9 @@
+--- 
+generate:
+  - client
+  - types
+  - chi-server
```
adiabramovitch, line 5: What do we want to generate? I tried models but it does not work for some reason, do you maybe know why?

lzap: Just a client and types, that is all.

adiabramovitch:
> Looks sane to me, I'd like to also have a stub for tests, but we can add that once we write service that depends on the IB. Only one nit: Could we add the new config into the `configs/defaults.yaml` so we have a full list there?

Do you mean adding a new struct for image builder configuration? 
I guess that now the only configurable thing is supposed to be the URL?

configs/defaults.yaml
```diff
@@ -27,6 +27,10 @@
 #   secret:
 #   session:
 
+# Image Builder configuration
+# ib:
+#   url: 
+
```
adiabramovitch, line 33: @ezr-ondrej like that?

configs/defaults.yaml
```diff
@@ -27,6 +27,10 @@
 #   secret:
 #   session:
 
+# Image Builder configuration
+# ib:
```
ezr-ondrej, line 31: I think this needs to be `imageBuilder` (it needs to be the same as in the nested config, for viper to pick that up

configs/ib_config.yaml
```diff
@@ -0,0 +1,8 @@
+--- 
+generate:
+  - client
+  - types
+package: ib
+output: ./internal/clients/ib/client.gen.go
```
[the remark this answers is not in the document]
ezr-ondrej: I like it

ezr-ondrej (approved):
Awesome! great job @adiabramovitch ! :+1:

internal/config/main_config.go
```diff
@@ -64,6 +67,7 @@ var Logging = &config.Logging
 var Cloudwatch = &config.Cloudwatch
 var AWS = &config.AWS
 var Features = &config.FeatureFlags
+var ImageBuilderURL = &config.ImageBuilder.URL
```
lzap, line 70: We only store sub-hashes to variables, but not a big deal. Nicely done indeed.

lzap:
The Makefile actually does not work as expected, fixed in https://github.com/RHEnVision/provisioning-backend/pull/39

PR #16, opened 2022-05-24
Pubkey and PubkeyResource models, DAO, API

lzap:
Adds pubkey and pubkey_resource models, tables, DAO and API. I also included transactions because the key upload operation must use one.

Not implemented yet: tasks. I dropped a TODO comment in a place where it should be a separate task.

Ready for review.

@adiabramovitch apologies for again such a big PR messing around with DB, DAO and AWS, I am trying to tackle all the important things. I am trying my best to explain all the changes.


lzap:
Rebased, also added transaction handling. This was more work than I expected due to the DAO package we use.

cmd/pbapi/main.go
```diff
@@ -66,7 +65,7 @@ func main() {
 	mr.Get("/", statusOk)
 	mr.Handle("/metrics", promhttp.Handler())
 
-	log.Info().Int("webPort", 8000).Int("metricsPort", config.Prometheus.Port).Msg("Starting new instance")
+	log.Info().Msgf("Starting new instance on port 8000 with prometheus on %d", config.Prometheus.Port)
```
lzap, line 68: I see no value in introducing these logging fields, why would we search logs by port number?

configs/defaults.yaml
```diff
@@ -20,6 +20,13 @@
...
+# AWS provisioning account (EC2, STS...)
+# aws:
+#   region: us-east-1
+#   key:
+#   secret:
+#   session:
+
 # AWS CloudWatch logger
 # cloudwatch:
 #   enabled: false
```
lzap, line 32: The patch introduces second AWS credentials, the one we already have are only for cloudwatch logging. This one is used for actual provisioning (the master RH account).

internal/clients/cloudwatchlogs/cwl_client.go
```diff
@@ -1,27 +1,19 @@
...
 var CWL *cloudwatchlogs.Client
 
 func Initialize() {
 	cache := aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(
 		config.Cloudwatch.Key, config.Cloudwatch.Secret, config.Cloudwatch.Session))
 
-	EC2 = ec2.New(ec2.Options{
-		Region:      config.Cloudwatch.Region,
-		Credentials: cache,
-	})
```
lzap, line 24: This was moved into a separate file.

internal/ctxval/ctx_getters.go
```diff
@@ -13,10 +14,11 @@ func GetUInt64Value(ctx context.Context, key CommonKeyId) uint64 {
...
-// GetLogger returns logger or nil when not in the context
+// GetLogger returns logger or the standard global logger when not in the context
+// or when context is nil.
 func GetLogger(ctx context.Context) *zerolog.Logger {
-	if ctx.Value(LoggerCtxKey) == nil {
-		return nil
+	if ctx == nil || ctx.Value(LoggerCtxKey) == nil {
+		return &log.Logger
 	}
 	logger := ctx.Value(LoggerCtxKey).(zerolog.Logger)
```
lzap, line 23: Its useful to either return logger with context OR the global logger.

internal/dao/dao_interfaces.go
```diff
@@ -2,10 +2,39 @@ package dao
...
+	// GetContext from sqlx library
+	GetContext(ctx context.Context, dest interface{}, query string, args ...interface{}) error
+	// SelectContext from sqlx library
+	SelectContext(ctx context.Context, dest interface{}, query string, args ...interface{}) error
+}
+
+// A TxFn is a function that will be called with an initialized `Transaction` object
+// that can be used for executing statements and queries against a database.
+type TxFn func(Transaction) error
+
```
lzap, line 36: This might look scary, but what this interface really does is it allows our DAO code to be completely independent of underlaying implementation. All DAO functions can simply use `Transaction` interface regardless if its coming either from a connection pool or from an explicit transaction. See below for how to use the TxFn function.

internal/dao/dao_interfaces.go
```diff
@@ -14,11 +43,20 @@ type AccountDao interface {
 	List(ctx context.Context, limit, offset uint64) ([]*models.Account, error)
 }
 
-var GetPubkeyDao func(ctx context.Context) (PubkeyDao, error)
+var GetPubkeyDao func(ctx context.Context, tx Transaction) (PubkeyDao, error)
```
lzap, line 46: All `GetXXXDao` functions now require a tx interface, which is typically sqlx.DB but it can be also sqlx.Tx for explicit transactions.

internal/dao/dao_interfaces.go
```diff
@@ -14,11 +43,20 @@ type AccountDao interface {
...
 	Delete(ctx context.Context, id uint64) error
 }
+
+var GetPubkeyResourceDao func(ctx context.Context, tx Transaction) (PubkeyResourceDao, error)
+
+type PubkeyResourceDao interface {
+	Create(ctx context.Context, pkr *models.PubkeyResource) error
+	Update(ctx context.Context, pkr *models.PubkeyResource) error
+	Delete(ctx context.Context, id uint64) error
+}
```
lzap, line 62: This is a new resource.

internal/dao/dao_transaction.go
```diff
@@ -0,0 +1,53 @@
...
+				logger.Error().Err(err).Msg("Cannot rollback database transaction")
+				return
+			}
+		}
+	}()
+
+	logger.Trace().Msg("Starting database transaction")
+	err = fn(tx)
+	return err
+}
```
lzap, line 53: This function simplifies use of transactions, see below on how to use it.

internal/db/migrations/002_utility_functions.up.sql
```diff
@@ -28,4 +46,14 @@ BEGIN
...
+CREATE OR REPLACE FUNCTION prevent_tag_update()
+  RETURNS trigger AS
+$prevent_tag_update$
+BEGIN
+  NEW.tag := OLD.tag;
+  RETURN NEW;
+END;
+$prevent_tag_update$ LANGUAGE 'plpgsql';
+
 COMMIT;
```
lzap, line 59: Another bunch of helper functions, database must be recreated in order to re-migrate it. Sorry, I do not want to do this in separate migrations to keep migration history clean.

internal/db/migrations/004_pubkeys.up.sql
```diff
@@ -8,4 +8,21 @@ CREATE TABLE pubkeys
...
 );
 
+CREATE TABLE pubkey_resources
+(
+  id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  tag TEXT NOT NULL DEFAULT random_string(15),
+  pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
+  provider INTEGER NOT NULL CHECK (valid_provider(provider)),
+  handle TEXT NOT NULL
+);
```
lzap, line 18: Column `tag` is a random string that is used as a handler in the future. I considered UUID but I think it is too long and users might be bothered with such a long tag. The tag does not need to be unique, its just something that is hard to guess.

internal/models/model_constants.go
```diff
@@ -0,0 +1,8 @@
+package models
+
+const (
+	ProviderTypeUnknown int = iota
+	ProviderTypeAWS
+	ProviderTypeAzure
+	ProviderTypeGCE
+)
```
lzap, line 8: Instead of introducing a table with providers, I propose to use constants (1 = AWS, 2 = Azure, 3 = GCE).

internal/models/pubkey_resource.go
```diff
@@ -0,0 +1,27 @@
...
+
+	// Resource handle (id). Format is provider-dependant. Required.
+	Handle string `db:"handle" json:"handle"`
+}
+
+// FormattedTag returns ID and Tag concatenated in a safe way for clouds. That means
+// it does not start with a number, only includes alpha-num characters and dash.
+func (p *PubkeyResource) FormattedTag() string {
+	return fmt.Sprintf("pk-%d-%s", p.ID, p.Tag)
+}
```
lzap, line 27: This is how tag looks like: `pk-14-9c9xMYHuJgoet46VpBsd` it contains our PK (which is unique) and a "secret".

internal/services/pubkey_service.go
```diff
@@ -15,25 +19,67 @@ func CreatePubkey(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		renderError(w, r, payloads.NewInitializeDAOError(r.Context(), "pubkey DAO", err))
-		return
-	}
+	// TODO: extract this into job queue
+	err := dao.WithTransaction(r.Context(), db.DB, func(tx dao.Transaction) error {
+		pkDao, err := dao.GetPubkeyDao(r.Context(), tx)
+		if err != nil {
+			return payloads.NewInitializeDAOError(r.Context(), "pubkey DAO", err)
+		}
```
lzap, line 27: This is how you do transactions with this proposal - you simply call a function and pass `tx` into Dao "getter", that's all. All the rest is the same.

internal/dao/dao_interfaces.go
ezr-ondrej: Could this have a default value? I am bit worried this creates more boilerplate then necessary, but if it is easy to create such Transaction object to use, I'm ok.

internal/db/migrations/004_pubkeys.up.sql
ezr-ondrej: Isn't the usefulness of such a tag lower if uniqueness at least per CU account is not guaranteed?
If I can't reliably verify that this key is or is not present in the cloud provider using this `tag`, I do not see much value of it.

ezr-ondrej (commented):
I'm sharing my thoughts, but I might have missed something as there is bit too much to process :)

lzap:
> Could this have a default value? I am bit worried this creates more boilerplate then necessary, but if it is easy to create such Transaction object to use, I'm ok.

What you mean by default value? Go does not support default function arguments or overriding function names. I would need to create something like `GetPubkeyDaoDefaults` extra function, associate it for all DAOs, is this worth it?

> Isn't the usefulness of such a tag lower if uniqueness at least per CU account is not guaranteed?

The idea is that a tag is not unique and it provides an extra check to prevent malicious primary key change. My patch stores tag in the form of "prefix-PK-tag". Primary key is what is guarantee uniquness for a particular resource and tag will be only used to verify the record is the one we expect.

internal/dao/dao_interfaces.go
lzap: It could only be a second function. I think it is not worth the effort, `db.DB` will always be named this way even if we change its type.

internal/db/migrations/004_pubkeys.up.sql
lzap: Answered below.

ezr-ondrej:
> is this worth it?

No, definitelly not worth it :)


> Primary key is what is guarantee uniquness for a particular resource and tag will be only used to verify the record is the one we expect.

But we need the resource id in cloud provider to look it up, is that `handle`?

ezr-ondrej (approved):
Thanks @lzap, I merge it with a promisse of no more large PRs :stuck_out_tongue: 
Please demo the basics of this, so we (others) know what is happening here, when convenient :)

lzap:
We have decided to remove the Transaction interface to simplify this a bit, instead of long-running transactions we are making decision to only allow transactions in DAO implementations, therefore service layer will need to do extra logic in order to ensure atomicity. In this case, generation of tag will need to be done in go rather than in sql. So when reading this patch, you can skip the Transaction interface, it will be removed in a followup PR.

PR #17, opened 2022-05-24
Add testing block to clowdapp spec

mshriver:
This is used by the bonfire iqe-cji smoke testing process.

Context: 
https://consoledot.pages.redhat.com/docs/dev/getting-started/ephemeral/smoke.html#_clowdapp_settings

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -18,6 +18,8 @@ objects:
         service: provisioning
     spec:
       envName: ${ENV_NAME}
+      testing: 
+        iqePlugin: hms-provisioning
       deployments:
         - name: service
           minReplicas: ${{MIN_REPLICAS}}


ezr-ondrej (commented):
Thanks @mshriver !
Can you pls expand on what this section does?
I've not found similar section in files I've based this on [ib](https://github.com/osbuild/image-builder/blob/8d4fe74b42920849dc4f0e004d9c7fa2e69c45e6/templates/image-builder.yml) and [sources](https://github.com/RedHatInsights/sources-api-go/blob/0e9730edde70dcbdc9b58d4f64da295fbdcf49ca/deploy/clowdapp.yaml)

ezr-ondrej (approved):
Whatever, the name can be changed, but I'd love to add automated image building to get that tag working. :)

mshriver:
The reference here is to the iqe-tests image tag in quay, which does exist as hms-provisioning:
https://quay.io/repository/cloudservices/iqe-tests

ezr-ondrej:
Oh, that make sense, thanks for the explanation! ^_^

PR #18, opened 2022-05-26
Register pgx logging properly

lzap:
Logging from the pgx driver was not working properly because the connection was not made from the registered driver.

### DIFF
--- a/internal/db/connection.go
+++ b/internal/db/connection.go
@@ -5,6 +5,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/log/zerologadapter"
+	"github.com/jackc/pgx/v4/stdlib"
 	_ "github.com/jackc/pgx/v4/stdlib"
 	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


internal/db/connection.go
```diff
@@ -39,15 +40,19 @@ func GetConnectionString(prefix string) string {
...
 
+	// register and setup logging configuration
 	connStr := GetConnectionString("postgres")
 	connConfig, err := pgx.ParseConfig(connStr)
 	if err != nil {
 		return errors.Wrap(err, "unable to parse database configuration")
 	}
-	connConfig.Logger = zerologadapter.NewLogger(log.Logger)
-	connConfig.LogLevel = pgx.LogLevel(config.Database.LogLevel)
+	if config.Database.LogLevel > 0 {
```
ezr-ondrej, line 49: We say `-1` = trace, so this is probably not correct?

lzap: Actually no, log level of pgx driver is documented as follows:

```
# PostgreSQL
# database:
#   host: localhost
#   port: 5432
#   name:
#   user:
#   password:
#   maxOpenConn: 25
#   maxIdleConn: 20
#   maxLifetime: 5m
#   maxIdletime: 10m
#   # 1-none, 2-error, 3-warn, 4-info, 5-debug, 6-trace
#   loglevel: 1

# logging:
#   # Overall log level (-1 trace, 0 debug, 1 info ... 5 panic)
#   level: 1
#   # Write logs to standard output in human-readable form
#   stdout: false
```

There are two log levels at the moment, granted, we could improve this and parse strings into numbers at some point. I would say out of scope for now :)

ezr-ondrej (approved):

PR #20, opened 2022-05-26
Add unit test for model

ezr-ondrej:
Needs to add stubbing for Dao, which is bit clunky.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -113,7 +113,7 @@ var validateGroupStreamError = errors.New("config error: Cloudwatch enabled but 
 var validateInvalidEnvironmentError = errors.New("config error: Environment must be production or development")
 
 func Validate() error {
-	if envMatch, _ := regexp.MatchString(`^(production|development)$`, Features.Environment); !envMatch {
+	if envMatch, _ := regexp.MatchString(`^(production|development|test)$`, Features.Environment); !envMatch {
 		return fmt.Errorf("%w: %s", validateInvalidEnvironmentError, Features.Environment)
 	}
 
--- a/internal/models/pubkey_resource_test.go
+++ b/internal/models/pubkey_resource_test.go
@@ -1,0 +1,15 @@
+package models
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTagFormat(t *testing.T) {
+
+	pubkey := PubkeyResource{12, "randomtag", 25, 1, "aws-handle"}
+
+	// assert equality
+	assert.Equal(t, "pk-12-randomtag", pubkey.FormattedTag(), "the tag format is incorrect")
+}
--- a/go.mod
+++ b/go.mod
@@ -21,4 +21,5 @@ require (
 	github.com/rs/xid v1.4.0
 	github.com/rs/zerolog v1.26.1
 	github.com/spf13/viper v1.7.0
+	github.com/stretchr/testify v1.7.1
 )
--- a/go.sum
+++ b/go.sum
@@ -1112,8 +1112,9 @@ github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UV
 github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81PSLYec5m4=
 github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5cxcmMvtA=
 github.com/stretchr/testify v1.6.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
-github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
 github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
+github.com/stretchr/testify v1.7.1 h1:5TQK59W5E3v0r2duFAb7P95B6hEeOyEnHRa8MjYSMTY=
+github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/subosito/gotenv v1.2.0 h1:Slr1R9HxAlEKefgq5jn9U+DnETlIUa6HfgEzj0g5d7s=
 github.com/subosito/gotenv v1.2.0/go.mod h1:N0PQaV/YGNqwC0u51sEeR/aUtSLEXKX9iv69rRypqCw=
 github.com/syndtr/gocapability v0.0.0-20170704070218-db04d3cc01c8/go.mod h1:hkRG7XYTFWNJGYcbNJQlaLq0fg1yr4J4t/NcTQtrfww=


internal/services/pubkey_service.go
```diff
@@ -9,9 +12,18 @@ import (
...
 
+var (
+	GetPubkeyDao = (&PubkeyStore{}).GetPubkeyDao
+)
+
+type PubkeyStore struct{}
+
+func (*PubkeyStore) GetPubkeyDao(ctx context.Context) (dao.PubkeyDao, error) {
+	return (dao.GetPubkeyDao(ctx, db.DB))
+}
```
ezr-ondrej, line 25: I wanted to stub out the database completely and this was the only way I figured.
I'm reimplementing this method `GetPubkeyDao` to return `PubkeyStoreMock`

The db.DB makes it harder to stub out db, that is why I cant use
`var GetPubkeyDao = dao.GetPubkeyDao` but I need to actually add this "ugly" wrapper

internal/config/config.go
```diff
@@ -2,12 +2,14 @@ package config
...
-	"github.com/rs/zerolog"
 	"regexp"
 	"strings"
 	"time"
 
+	"github.com/rs/zerolog"
+
 	clowder "github.com/redhatinsights/app-common-go/pkg/api/v1"
 	"github.com/spf13/viper"
 )
```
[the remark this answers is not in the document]
lzap: This should prevent it, if you rebase on top should solve this: https://github.com/RHEnVision/provisioning-backend/pull/21

internal/config/config.go
```diff
@@ -80,6 +82,8 @@ func init() {
 		viper.Set("database.password", cfg.Database.Password)
 		viper.Set("database.name", cfg.Database.Name)
 		viper.Set("prometheus.port", cfg.MetricsPort)
+	} else if IsTestRun() {
+		viper.Set("featureFlags.environment", "test")
```
lzap, line 86: Isn't this missing a new field in the config type? Then add it to the example YAML config. Or I am not following.

internal/models/pubkey_resource_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTagFormat(t *testing.T) {
+
+	pubkey := PubkeyResource{12, "randomtag", 25, 1, "aws-handle"}
+
+	// assert equality
+	assert.Equal(t, "pk-12-randomtag", pubkey.FormattedTag(), "the tag format is incorrect")
+}
```
lzap, line 15: This is a good unit test candidate indeed 👍🏻

internal/services/pubkey_service.go
```diff
@@ -79,7 +91,7 @@ func CreatePubkey(w http.ResponseWriter, r *http.Request) {
 }
 
 func ListPubkeys(w http.ResponseWriter, r *http.Request) {
-	pubkeyDao, err := dao.GetPubkeyDao(r.Context(), db.DB)
+	pubkeyDao, err := GetPubkeyDao(r.Context())
```
lzap, line 94: It depends on which layer you want to stub things - if you want to replace DAO with stubbed functions, then no change is required on this line. You can simply pass `db.DB` which will be probably `nil` and it's a non-issue as you will not use it at all.

Or you can choose to stub at the driver level, then implement a new `db.Transaction` and pass it into sqlx. But this is much more challenging option, I think you want to replace DAO instead.

internal/services/pubkey_service_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	var pubkeys []models.Pubkey
+
+	if err := json.NewDecoder(rr.Body).Decode(&pubkeys); err != nil {
+		t.Errorf("Error decoding response body: %v", err)
+	}
+
+	assert.Equal(t, 2, len(pubkeys), "expected two pubkeys in response json")
+}
```
lzap, line 71: This almost feels like an integration test, I think you want to stub DAO instead and then test this service unchanged.

lzap (changes requested):
Getting a little bit tricky but hopefully the way we implemented DAO pays off now.

lzap:
First of all, I believe that SQL should never be subject of unit tests. These should be only written for functions or "modules" that do some non-database logic. The moment SQL is involved, I believe it is not worth spending time stubbing SQL and using that time to simply write integration test which actually requires database with some fixtures. There are several approaches how to achieve data stability for tests:

* Create a transaction for each test, then rollback in the end, PostgreSQL supports nested transactions so this should work for any code including with transactions. I sort of prefer this one.
* Create new database for each test, in postgres this can be achieved by creating a new schema for each test. This is what Sources API project does: https://github.com/RedHatInsights/sources-api-go/blob/main/dao/main_test.go and then in tests you call `swich_schema`: https://github.com/RedHatInsights/sources-api-go/blob/main/dao/meta_data_dao_test.go but running migrations for each test can be slower and slower as you go, so I would probably push for cloning existing schema for each test instead: https://wiki.postgresql.org/wiki/Clone_schema

Anyway, if you really want to try stubbing, then he dao package is designed in a way that it should be easy to add more implementations. Simply create a new implementation in a separate package `dao.stub` or something. You simply pass it `dao.Transaction` interface implementation that does nothing as you will not use it in our DAO. Or you can choose to pass simply `nil` but I am not sure if that would not be compile error. The only problem is that in Go you cannot do conditional import. So I suggest something like this:

```go
	// Configuration must be loaded as first
	// Performs initialization of DAO implementation, must be initialized before any database packages.
	// Stub DAO overwrites factory functions only when environment is set test
	_ "github.com/RHEnVision/provisioning-backend/internal/dao/stub"
```

Then in your new DAO implementation make sure to overwrite the "factory" function with your own one only when test environment is present:

```go
func init() {
	// If you make sure that config is imported before this DAO package, then config fill be
	// already initialized at this point, so you can write something like:

	if config.Features.Environment == "test" {
		dao.GetAccountDao = getAccountStubDao
	}
}
```

Other option to do this is to use Go build tags, that is probably more clean: https://pkg.go.dev/go/build when you can create a special package that imports implementation package depending on a build flag. If its "test" then import the stub package, if its not "test" then import "sqlx" implementation. No configuration would be needed in the `init()` function in this case.

_Github was down and it deleted my biggest comment from the review._

internal/config/config.go
ezr-ondrej: This has already a key in the yaml file, we just only had developement and production, no test.

internal/config/config.go
```diff
@@ -128,6 +132,10 @@ func Validate() error {
 	return nil
 }
 
+func IsTestRun() bool {
+	return flag.Lookup("test.v") == nil
+}
```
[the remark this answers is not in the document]
ezr-ondrej: Stack overflow, This is top recommended method to check wheter you're in a `go test` run

internal/services/pubkey_service_test.go
ezr-ondrej: I believe this is minimum required to test that the method does what it is supposed to :) Thus Unit test.
Integration test should not IMHO skip even the server.

lzap:
My conclusion: If you remove all the database-related stuff from this PR and add some info about unit testing to the README, let's merge this. No issues at all, we use testify, let's try few sprints with this and see if we can improve later but it looks good enough to me.

Then you can work on integration tests where, as we discussed, we will need to stub API calls for sure. I am not against stubbing database if it makes sense for some tests, for others we can implement schema switching/copying/rollbacks.

lzap:
We have decided to actually only implement DAO stubs for unit test database stubbing, we can probably talk about tests with real database later on for integration tests. But the gist of our approach is to avoid real database (and real REST API calls) completely in unit tests (of course) and integration tests too if possible.

internal/config/config.go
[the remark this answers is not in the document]
lzap: Does this work? According to some threads, it looks like this flag is only set when you do "go test -v". Can you actually add a unit test for this? :) Just check if IsTestRun() is actually true. Also don't you need to run `flag.Parse()` first? It appears so, but maybe viper calls that.

internal/models/pubkey_resource_test.go
lzap: Note we don't need to test edge cases because database ensures that both PK and Tag are never nil or blank. That is going to change today, but I will add an extra test for this.

lzap (changes requested):
Add one extra test.

lzap (approved):
Thanks :)

PR #21, opened 2022-05-27
Enforce goimports to prevent import ordering conflicts

lzap:
We are already hitting problems with import ordering, this enforces it via GHA.

I am also mergine jobs into three items: formatting, linting, testing. There is no point in running each in its own container.


ezr-ondrej (approved):
I do not love the job naming, but happy to have discussion on it in a PR I'll propose :)
No reason to block this for naming :+1:

PR #22, opened 2022-05-27
Simplify transactions

lzap:
Let's make a decision that will simplify transactions - they can only be done in dao functions. Although it is convenient to use SQL transactions in the service package for example for a whole REST API call, long-lived transactions is something that should not be abused. This patch refactors the helper WithTransaction function in a way that it can be only used within a dao implementation (e.g. sqlx). Error compensation is used to delete resource when EC2 AWS upload operation fails.


internal/dao/sqlx/account_dao.go
```diff
@@ -23,83 +24,83 @@ type accountDaoSqlx struct {
...
 		return nil, NewPrepareStatementError(ctx, &daoImpl, listAccounts, err)
 	}
 
 	return &daoImpl, nil
 }
 
-func (dao *accountDaoSqlx) NameForError() string {
-	return dao.name
+func (di *accountDaoSqlx) NameForError() string {
+	return di.name
```
lzap, line 53: FYI I renamed all reciever variables to `di` because this was shadowing imported `dao` package.

I think what's ugly in Go is that the language allows variable shadowing. Superb talk on this topic: https://www.youtube.com/watch?v=zPd0Cxzsslk

internal/dao/sqlx/pubkey_resource_dao.go
```diff
@@ -4,12 +4,13 @@ import (
...
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/db"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/jmoiron/sqlx"
 )
 
 const (
-	createPubkeyResource     = `INSERT INTO pubkey_resources (pubkey_id, provider, handle) VALUES ($1, $2, $3) RETURNING id, tag`
+	createPubkeyResource     = `INSERT INTO pubkey_resources (pubkey_id, provider, handle, tag) VALUES ($1, $2, $3, $4) RETURNING id, tag`
```
lzap, line 13: This was actually a bug I missed.

internal/db/migrations/002_utility_functions.up.sql
```diff
@@ -18,15 +18,6 @@ BEGIN
...
 
--- Random alpha-num tag string, not guaranteed to be unique
-CREATE OR REPLACE FUNCTION random_string(i INTEGER)
-  RETURNS TEXT AS
-$random_string$
-BEGIN
-  RETURN translate(encode(gen_random_bytes(i), 'base64'), '+/', 'xX');
-END;
-$random_string$ LANGUAGE 'plpgsql';
-
```
lzap, line 29: Moved to Go.

internal/db/migrations/002_utility_functions.up.sql
```diff
@@ -51,7 +42,9 @@ CREATE OR REPLACE FUNCTION prevent_tag_update()
   RETURNS trigger AS
 $prevent_tag_update$
 BEGIN
-  NEW.tag := OLD.tag;
+  IF OLD.tag != '' AND OLD.tag != NEW.TAG THEN
+    RAISE EXCEPTION 'tag is read-only';
+  END IF;
```
lzap, line 47: This now raises an error on overwrite attempt. Also guarded on the Go level.

internal/db/migrations/004_pubkeys.up.sql
```diff
@@ -11,7 +11,7 @@ CREATE TABLE pubkeys
 CREATE TABLE pubkey_resources
 (
   id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
-  tag TEXT NOT NULL DEFAULT random_string(15),
+  tag TEXT NOT NULL DEFAULT '',
   pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
   provider INTEGER NOT NULL CHECK (valid_provider(provider)),
   handle TEXT NOT NULL
```
lzap, line 17: Full database reset needed, sorry, again!

internal/models/tag_generate.go
```diff
@@ -0,0 +1,25 @@
...
+	return n.Text(62)
+}
+
+// GenerateTag sets Tag field to a random base62 (base64 without / or +) with a length
+// of 20 characters. Values are cryptographically pseudo-random and pseudo-unique with
+// probability of getting same values the same as for random UUID.
+func GenerateTag() string {
+	// generate from more bytes because fixed-size base62 cannot be guaranteed at this size
+	return randomBase62(20)[0:20]
+}
```
lzap, line 25: UUID has 128bits and the same entropy as UUID is 15-16 bytes which is 19-21 bytes encoded. I did not do the correct math, but this is what I get:

https://go.dev/play/p/hqM6hq7KGgr

This is definitely "almost always" unique.

internal/services/pubkey_service.go
```diff
@@ -20,57 +19,43 @@ func CreatePubkey(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		var e *payloads.ResponseError
-		if errors.As(err, &e) {
-			renderError(w, r, e)
-		} else {
-			renderError(w, r, payloads.NewUnknownError(r.Context(), err))
-		}
+		renderError(w, r, payloads.NewDAOError(r.Context(), "create pubkey resource", err))
 		return
 	}
```
lzap, line 60: This is now very much simplified (no transactions):

* Create pubkey
* Generate tag
* Upload to AWS
* Create pubkey resource (tag with handle)

ezr-ondrej (approved):
cool I like this solution quite a lot :+1:
I'll test in ephemeral

internal/dao/sqlx/account_dao.go
ezr-ondrej: just watched it, should we start using `go vet -shadow` ? :)

lzap: Totally a good idea, feel free to submit a patch, we probably have many already.

PR #23, opened 2022-05-29
Environment aware config initialization

ezr-ondrej:
Refs: HMSPROV-77

### DIFF
--- a/.github/workflows/gotest.yml
+++ b/.github/workflows/gotest.yml
@@ -43,4 +43,4 @@ jobs:
         with:
           go-version: "1.16"
       - run: |
-          go test ./...
+          make test
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,12 @@ lint:
 	golangci-lint run
 
 .PHONY: migrate
-migrate: build-pbmigrate
-	pbmigrate
+migrate:
+	go run ./cmd/pbmigrate
+
+.PHONY: test
+test:
+	go test --tags=test ./...
 
 .PHONY: generate-migration
 MIGRATION_NAME?=unnamed
--- a/cmd/pbapi/main.go
+++ b/cmd/pbapi/main.go
@@ -29,6 +29,8 @@ func statusOk(w http.ResponseWriter, _ *http.Request) {
 }
 
 func main() {
+	config.Initialize()
+
 	// initialize stdout logging and AWS clients first
 	log.Logger = logging.InitializeStdout()
 	cloudwatchlogs.Initialize()
--- a/cmd/pbmigrate/main.go
+++ b/cmd/pbmigrate/main.go
@@ -9,6 +9,8 @@ import (
 )
 
 func main() {
+	config.Initialize()
+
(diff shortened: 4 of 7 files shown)


.github/workflows/gotest.yml
```diff
@@ -43,4 +43,4 @@ jobs:
         with:
           go-version: "1.16"
       - run: |
-          go test ./...
+          make test
```
lzap, line 46: Nice make should be "the entrypoint" for all our CI.

internal/config/config.go
```diff
@@ -63,15 +63,18 @@ var AWS = &config.AWS
...
-
 	viper.SetDefault("logging.level", 1)
 	viper.SetDefault("database.host", "localhost")
 	viper.SetDefault("database.port", 5432)
 	viper.SetDefault("database.loglevel", 1)
 	viper.SetDefault("cloudwatch.enabled", false)
 	viper.SetDefault("prometheus.port", 9000)
 
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
```
lzap, line 74: One thing that looks nasty is ordering of init() calls, I am not able to tell if this line is called after or before `ENV_config.go`. This could lead to some nasty bugs. It looks like that Go compiler calls them in lexicographic order: https://go.dev/ref/spec#Package_initialization

So you were simply lucky this time but I suggest to extract all the three init functions into separate files and naming it the following way:

* 0_config_common.go
* 5_config_test.go
* 5_config_development.go

This way it is crystal clear even after 5 years what is going on here.

internal/config/devel_config.go
```diff
@@ -0,0 +1,26 @@
+//go:build !test && development
+// +build !test,development
```
lzap, line 2: Out of curiosity, is this some new/old syntax?

lzap (changes requested):
One comment about possible bugs on init() ordering that could be solved by more explicit naming of files.

internal/config/devel_config.go
ezr-ondrej: yeah, new/old syntax ... my formatter wants to include both :shrug:

internal/config/config.go
ezr-ondrej: I've added numbered prefixes, but I decided to prefix with environment, because config_test clashes with test file for config. :(

lzap:
A conflict.

lzap:
Wait it does not work!

ezr-ondrej:
> Looks good, when you run app unconfigured you get: `panic: config error: Environment must be production or development` which is a bit cryptic, but not scope of this PR. I will fix that.

There was an error there, this should not happen in developement, by default we should run developement environment

ezr-ondrej:
> Wait it does not work!

it does now I believe

lzap:
Now it works, thanks.

ezr-ondrej:
For the reference, we've removed `development` flag and now developement is the default

PR #24, opened 2022-05-30
Drop production compile tag and shadow linting

lzap:
It does not make sense to me at this point, only makes things more complicated. We should use compilation flags only when we really need some files/packages to be not compiled - that's definitely a good case for testing (stubbing of DAO). For production, I think it is a good idea to be as close as for development environment.

This updates the Makefile and makes test flags configurable, some tests could be actually skipped with compilation flags in the future.

This now works:

* go build & run
* make build & run
* make test

The second commit: Variable shadowing is a lint error now

It is one of the most dangerous things in Go, everything can be shadowed with nasty consequences. It comes at a cost, this is no longer valid:

```
err := func()
err := func()
```

Go compiler allows this for errors, go vet does not allow to align with this exception.


internal/config/main_config.go
```diff
@@ -68,21 +67,6 @@ var Cloudwatch = &config.Cloudwatch
...
-	viper.SetDefault("database.port", 5432)
-	viper.SetDefault("database.loglevel", 1)
-	viper.SetDefault("cloudwatch.enabled", false)
-	viper.SetDefault("prometheus.port", 9000)
-	viper.SetDefault("prometheus.path", "/metrics")
-
-	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
-	viper.AutomaticEnv()
-}
-
```
lzap, line 85: I moved this out of the initialization file, so the purpose is clear enough.

Makefile
```diff
@@ -49,7 +51,7 @@ migrate:
 
 .PHONY: test
 test:
-	go test --tags=test ./...
+	go test -tags=$(TEST_TAGS) ./...
```
lzap, line 54: Do `make test TEST_TAGS=test,sql,unit,whatever` in the future.

internal/config/5_real_config.go
```diff
@@ -1,5 +1,5 @@
-//go:build !test && !production
-// +build !test,!production
+//go:build !test
+// +build !test
```
ezr-ondrej, line 2: We need to avoid loading this file in prod where we do not have the configuration files so this fails :thinking:

lzap: Let's just check this with Exists. If we want to have defaults in YAML, let's use Go embedding feature.

.github/workflows/gotest.yml
```diff
@@ -17,16 +17,13 @@ jobs:
...
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
           go-version: "1.16"
-      - name: vet
-        run: |
-          go vet ./...
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
```
lzap, line 29: The golangci-lint tools itself contains `go vet` so we were linting some rules twice.

.golangci.yml
```diff
@@ -11,9 +11,14 @@ linters:
     - sql
     - module
     - unused
+  enable:
+    - predeclared
```
lzap, line 15: This also disallows redefining predeclared variables like `true` etc. This is a valid Go code:

```go
true = false
```

Yeah, nasty.

ezr-ondrej:
Thanks :+1:

PR #26, opened 2022-05-30
Add unit test for service

ezr-ondrej:
Needs to add stubbing for Dao

Fixes: HMSPROV-78

**Checklist**


lzap (changes requested):
It's okay although the test could be less dependent on what's pre-configured by default.

But here is an idea - when a test modifies data in any way, since the mock is stored globally, it will fail other tests. I think for this reason each mock need to have a Reset() function which will set all the values back to initial state. There should be perhaps some kind of "registry" of all mocks with a global ResetAll() function for all. Then in each test we need to have `defer stubs.ResetAll()` that will perform the reset. Alternatively we could use https://github.com/stretchr/testify#suite-package Suite but that one I don't like too much.

ezr-ondrej:
We are storing the Dao in the testing context now and it should be viable option, but it has caviat of having to store it as `any` and infer type on retrieval, which is unfortunate, maybe in the future we might just store a global map of the stubbed DAOs, but that would require a cleaning mechanism.

internal/dao/stubs/pubkey_dao.go
```diff
@@ -0,0 +1,50 @@
+package stubs
```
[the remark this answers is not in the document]
ezr-ondrej: I think it should not be necessary, we just should never require the stub DAO in production code, but I'd not be against it, tho if I add the flag, it trips the linter: https://github.com/RHEnVision/provisioning-backend/runs/6658737817?check_suite_focus=true

internal/dao/stubs/pubkey_dao.go
```diff
@@ -0,0 +1,54 @@
...
+func WithPubkeyDao(parent context.Context, init_store []*models.Pubkey) context.Context {
+	ctx := context.WithValue(parent, pubkeyCtxKey, PubkeyDaoStub{init_store})
+	return ctx
+}
+
+func getPubkeyDao(ctx context.Context) (dao.PubkeyDao, error) {
+	var err error
+	pkdao, ok := ctx.Value(pubkeyCtxKey).(dao.PubkeyDao)
+	if !ok {
+		err = &ContextReadError{}
```
lzap, line 37: I think you could simply use `fmt.Errorf` here, not big deal for mocks. Please refactor this later, the ContextReadError looks like a common error that should not live in pubkey_dao.go.

ezr-ondrej: Linter wanted this, but I will try to improve so it is easier to read.

PR #27, opened 2022-05-31
Add STS client

adiabramovitch:

**Checklist**

linked to https://github.com/RHEnVision/provisioning-backend/pull/9


internal/clients/sources/sources_payloads.go
```diff
@@ -0,0 +1,60 @@
...
+	SourceName          string `json:"source_name"`
+	ApplicationTypeName string `json:"application_type_name"`
+}
+
+type Authentication struct {
+	ResourceType string `json:"resource_type"`
+	ResourceName string `json:"resource_name"`
+	Username     string `json:"username"`
+	Authtype     string `json:"authtype"`
+}
```
lzap, line 60: Is this generated? If so, can you provide information on how you generated this?

internal/clients/sts/sts_client.go
```diff
@@ -0,0 +1,36 @@
...
+	output, err := stsClient.AssumeRole(context.TODO(), &sts.AssumeRoleInput{
+		RoleArn:         aws.String(arn),
+		RoleSessionName: aws.String("name"),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return output.Credentials, nil
+}
```
lzap, line 29: Please create a client type (struct) with sts.Client and context.Context fields and make it a reciever of AssumeRole. This way, a single client can be used multiple times within a request. Basically the template can be ec2 which is already in main:

```go
type Client struct {
	sts     *sts.Client
	ctx     context.Context
	log     zerolog.Logger
}

func NewSTSClient(ctx context.Context) *Client {
    // see internal/clients/ec2 for an example
}

func (c *Client) AssumeRole(arn string) (*types.Credentials, error) {
    // ... now you have c.sts, c.ctx and c.log to work with
}
```

internal/clients/sts/sts_client.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/aws/aws-sdk-go-v2/service/sts/types"
+)
+
+func AssumeRole(arn string) (*types.Credentials, error) {
+	stsClient, err := CreateSTSClient()
+	if err != nil {
+		return nil, err
+	}
```
lzap, line 18: The function should have client as a reciever, therefore it should not be creating any client.

internal/clients/sts/sts_client.go
```diff
@@ -0,0 +1,36 @@
...
+
+	output, err := stsClient.AssumeRole(context.TODO(), &sts.AssumeRoleInput{
```
lzap, line 20: Then you can use context from the reciever.

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,124 @@
...
+	getSourceByOrgId = `SELECT * FROM sources WHERE org_id = $1 ORDER BY id LIMIT 1;`
+)
+
+type sourceDaoSqlx struct {
+	name       string
+	getById    *sqlx.Stmt
+	list       *sqlx.Stmt
+	create     *sqlx.Stmt
+	getByOrgId *sqlx.Stmt
+	deleteById *sqlx.Stmt
```
lzap, line 28: This statement does not appear to be prepared.

internal/dao/sqlx/source_dao.go
```diff
@@ -0,0 +1,124 @@
...
+	err := stmt.GetContext(ctx, result, orgId)
+	if err != nil {
+		return nil, NewGetError(ctx, dao, query, err)
+	}
+	return result, nil
+}
+
+func (dao *sourceDaoSqlx) Delete(ctx context.Context, id uint64) (*models.Source, error) {
+	query := deleteSourceById
+	stmt := dao.deleteById
```
lzap, line 115: This will fail.

internal/db/migrations/005_create_sources_table.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE sources
+(
+    id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    auth_id TEXT NOT NULL UNIQUE CHECK (NOT empty(auth_id)),
+    name TEXT UNIQUE CHECK (NOT empty(name)),
+    source_id TEXT NOT NULL UNIQUE CHECK (NOT empty(source_id)),
+    org_id TEXT NOT NULL UNIQUE CHECK (NOT empty(org_id)),
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+
+);
```
lzap, line 10: Empty line, can you get rid of it?

internal/models/source_model.go
```diff
@@ -0,0 +1,20 @@
+package models
+
+type Source struct {
+	// Required auto-generated PK.
+	ID uint64 `db:"id" json:"id"`
+
+	// Source ID in sources
```
lzap, line 7:
```suggestion
	// Source ID in sources. Unique.
```

internal/models/source_model.go
```diff
@@ -0,0 +1,20 @@
...
+	SourceID string `db:"source_id" json:"source_id"`
```
lzap, line 8: Is this it: https://github.com/RedHatInsights/sources-api-go/blob/main/model/source.go#L24

If so, that is a uint64 why this is a string in our database?

internal/models/source_model.go
```diff
@@ -0,0 +1,20 @@
...
+
+	// User-facing name. Required.
+	Name string `db:"name" json:"name"`
```
lzap, line 11: Is this it: https://github.com/RedHatInsights/sources-api-go/blob/main/model/source.go#L30

If so, why this has UNIQUE constraint? We need to be careful with uniqueness because that can usually by only implemented per account, not globally.

internal/models/source_model.go
```diff
@@ -0,0 +1,20 @@
...
+
+	// Organization ID. Required.
+	OrgID string `db:"org_id" json:"org_id"`
```
lzap, line 14: What is this? TenantID? https://github.com/RedHatInsights/sources-api-go/blob/main/model/tenant.go

If so, this needs to be uint64 primary key and we should use the correct term "tenant_id" as well. We need to be very careful here, because tenant, account/external_tenant and org_id are three different numbers of a same entity.

Also, we absolutely need to associate every single record in our database with our own account_id which is our own primary table for accounts. That one also contains org_id and account_number. Please create a column and a field for that: `account_id BIGINT NOT NULL REFERENCES accounts(id),`

internal/models/source_model.go
```diff
@@ -0,0 +1,20 @@
...
+
+	// Authentication ID in sources lined to a spesific source in order to fetch the ARN
+	AuthID string `db:"auth_id" json:"auth_id"`
```
lzap, line 17: Is this it: https://github.com/RedHatInsights/sources-api-go/blob/main/model/authentication.go#L17

If API allows, I would prefer to use `DbID` which is very much more efficient uint64 (vs string).

internal/routes/all_routes.go
```diff
@@ -35,6 +35,16 @@ func apiRouter() http.Handler {
...
 		})
 	})
+	r.Route("/sources", func(r chi.Router) {
+		r.Get("/", s.ListSources)
+		r.Post("/", s.CreateSource)
+		r.Route("/{ID}", func(r chi.Router) {
+			r.Get("/", s.GetSource)
+			r.Delete("/", s.DeleteSource)
+
+		})
```
lzap, line 45: Empty line.

internal/routes/all_routes.go
```diff
@@ -35,6 +35,16 @@ func apiRouter() http.Handler {
...
+	})
 
 	return r
+
```
lzap, line 49: Ditto.

internal/services/sources_service.go
```diff
@@ -0,0 +1,132 @@
...
+
+func CreateSource(w http.ResponseWriter, r *http.Request) {
+	payload := &payloads.SourceRequest{}
+	if err := render.Bind(r, payload); err != nil {
+		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), err))
+		return
+	}
+
+	// An ARN linked to the policy and role that are needed for our application
+	arn := os.Getenv("ARN")
```
lzap, line 65: This needs to be a configuration value.

internal/services/sources_service.go
```diff
@@ -0,0 +1,132 @@
...
+		return
+	}
+
+	// Delete source from sources
+	_, err = sources.DeleteSource(w, r, source.SourceID, payload.OrgID)
+	if err != nil {
+		renderError(w, r, payloads.NewDAOError(r.Context(), "delete source from sources", err))
+		return
+	}
+
```
lzap, line 131: I would probably recommend first sources then db, so in case the request fails user can try again. Also empty line.

adiabramovitch:
> Ok this is HUGE, it not only adds STS client but also sources DAO, sources client, I would appreciate if you could break this up into smaller pieces. Lots of comments, my biggest concerns are around sources model.

please address only the second commit. the first commit is related to a different PR. I left it there because I want them to be merged in a specific order.

internal/clients/sources/sources_client.go
```diff
@@ -0,0 +1,118 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"strings"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+)
+
+const URL = "http://localhost:8000/api/sources/v3.1"
```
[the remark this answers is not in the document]
adiabramovitch: to add it to the defaults.yaml like this: 
`cloudwatch:`
  enabled: false
  group: group
  stream: stream

sources:
  URL: http://localhost:8000/api/sources/v3.1`
@lzap

internal/services/sources_service.go
adiabramovitch: I agree but I have a problem, I need to get the source_Id first in order to delete it in sources. 
I didn't want to query the DB twice - one to get the source_id and the other time to delete the source itself with the provided id. 
Do you have a better suggestion?

internal/dao/sqlx/source_dao.go
adiabramovitch: It does fail, not sure why. can you explain? how is it different from the other function?

lzap:
Looks good, just create a "contructor" function in ec2 package that will take Client from STS function so users can easily create ec2 clients via sts client.

lzap:
Let's just have a call with Sources API for their advice on what exactly are we supposed to keep in our DB. I would like to put as high resistance as we can to avoid storing duplicate data, maybe we can get away just with calling their API and in-memory LUR cache.

lzap (approved):
Perfect thank you!

PR #29, opened 2022-06-01
Dont upload pubkey

lzap:
Moves pubkey upload away from the CRUD operations so QA can actually test the POC API now.

Also pins golangci-lint to v1.45 in the makefile too to prevent a bug that @adiabramovitch ran into.


internal/models/pubkey_resource.go
```diff
@@ -9,23 +9,27 @@ type PubkeyResource struct {
...
 }
 
-// FormattedTag returns ID and Tag concatenated in a safe way for clouds. That means
+// FormattedTag returns Tag concatenated in a safe way for clouds. That means
 // it does not start with a number, only includes alpha-num characters and dash.
+// Tag is
 func (p *PubkeyResource) FormattedTag() string {
-	return fmt.Sprintf("pk-%d-%s", p.ID, p.Tag)
+	return fmt.Sprintf("pk-%s", p.Tag)
 }
```
lzap, line 33: I am actually changing the tag format from pk-ID-TAG to just pk-TAG. I think the probability of getting same TAG twice is low. I am increasing length to 22 characters which is in base62 sligtly bigger entropy than UUID. Chances are low.

Because I have a chicken-and-egg problem - do I create resource in the DB first (thus getting ID and Tag), upload and then update handle or the other way? Its complicated, I think its much easier to simply generate good enough TAG and live with it.

ezr-ondrej (approved):
I love how it simplifies CreatePubkey ^_^
I'd argue we do not need the CreatePubkeyResource endpoints, but we will see :)

internal/models/pubkey_resource.go
ezr-ondrej: makes sense, lets hope we won't be unlucky ^_^ it also avoid exposing internal ids, which might have caused some traction with security team :)

ezr-ondrej: You've not changed the test for this method tho ;)

ezr-ondrej (changes requested):
The tag formatting change causes test failure :)

lzap:
Damn yeah, fixed.

ezr-ondrej (approved):
Needs rebase, as pin had a separate PR, but then it is ready to go in :+1:

ezr-ondrej:
Thanks @lzap ! :+1:

PR #33, opened 2022-06-03
Replace golang-migrate with tern

lzap:
This is a proposal and working patch to replace golang-migrate with tern. I found golang migrate problematic for couple of reasons which are all solved in tern: https://github.com/jackc/tern

* Biggest concern is that not every migration is done in a transaction. Even when each SQL file is enclosed with BEGIN/COMMIT this does not work as expected - when something fails, golang-migrate leaves the database in a "dirty" state asking the user to manually fix this. Tern, on the other hand, is designed by default to work with transactions, migration either succeeds or fails, there is no dirty state. Content guys made a code that allows that, but I don't see a reason to use a tool that is not designed well.
* The second big concern are errors - error reporting in golang-migrate is generic and not very helpful. Tern supports native PostgreSQL error parsing with very detailed error messages including line and column extraction.
* Golang-migrate has no validation of file naming, specifically for sequenced migrations. We currently have some code that performs some validations, with Tern this all can be dropped.
* We agreed not to write down migrations for now. In Tern there are no up and down migrations as separate files, its a single file and down part is optional via a separator. 
* Tern is more simple, less code to worry about, yet powerful and its from the respected PGX PostgreSQL driver author.

Example error message from the new code:

```
10:43AM INF Seeding 'dev_small' hostname=mone.home.lan
10:43AM INF Dialing PostgreSQL server host=nuc hostname=mone.home.lan module=pgx
10:43AM ERR Exec args=[] err="ERROR: syntax error at or near \"XXX\" (SQLSTATE 42601)" hostname=mone.home.lan module=pgx pid=2241330 sql="--\n-- Testing seed data, to execute this file during migration set DB_SEED variable to \"dev_small\".\n-- Keep this file idempotent. Always specify primary keys in the range of 1-100.\n--\nBEGIN;\n\nINSERT INTO accounts(id, account_number, org_id)\nVALUES (1, '13', '000013'),\n       (2, '15', '000015'),\n       (3, NULL, '000042'),\n       (4, NULL, '000077'XXX)\nON CONFLICT DO NOTHING;\n\nINSERT INTO pubkeys(id, account_id, name, body)\nVALUES (1, 1, 'lzap-ed25519-2021',\n        'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap')\nON CONFLICT DO NOTHING;\n\n-- Reset all primary key sequences (columns named \"id\") to the maximum value.\nSELECT reset_sequences();\n\nCOMMIT;\n"
10:43AM INF closed connection hostname=mone.home.lan module=pgx pid=2241330
10:43AM FTL ERROR: syntax error at or near "XXX" (SQLSTATE 42601)
LINE 11:        (4, NULL, '000077'XXX)
                                   ^
 error="ERROR: syntax error at or near \"XXX\" (SQLSTATE 42601)" hostname=mone.home.lan seed=true
```


ezr-ondrej (approved):
If the migrations work, I'm totally fine with this, sounds reasonable, ADR: https://gitlab.cee.redhat.com/satellite-services/arch/-/merge_requests/5

lzap:
Rebased, fixed lint issues.

ezr-ondrej (approved):
ADR approved and we this works.
Note: We need an easier way of execution of drop_all, as you do not want to run migrations before droping the db

PR #37, opened 2022-06-13
Add support for X-RH identity

ezr-ondrej:
Add support for passing down the identity into the application.
Add support for stubbing the identity with constant identity for 
testing.

**Checklist**

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -125,6 +125,7 @@ github.com/armon/go-radix v0.0.0-20180808171621-7fddfc383310/go.mod h1:ufUuZ+zHj
 github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a/go.mod h1:lB+ZfQJz7igIIfQNfa7Ml4HSf2uFQQRzpGGRXenZAgY=
 github.com/aws/aws-sdk-go v1.15.11/go.mod h1:mFuSZ37Z9YOHbQEwBWztmVzqXrEkub65tZoCYDt7FT0=
 github.com/aws/aws-sdk-go v1.17.7/go.mod h1:KmX6BPdI08NWTb3/sm4ZGu5ShLoqVDhKgpiN924inxo=
+github.com/aws/aws-sdk-go v1.38.51/go.mod h1:hcU610XS61/+aQV88ixoOzUoG7v3b31pl2zKMmprdro=
 github.com/aws/aws-sdk-go-v2 v1.8.0/go.mod h1:xEFuWz+3TYdlPRuo+CqATbeDWIWyaT5uAPwPaWtgse0=
 github.com/aws/aws-sdk-go-v2 v1.9.2/go.mod h1:cK/D0BBs0b/oWPIcX/Z/obahJK1TT7IPVjy53i/mX/4=
 github.com/aws/aws-sdk-go-v2 v1.16.2/go.mod h1:ytwTPBG6fXTZLxxeeCCWj2/EMYp/xDUgX+OET6TLNNU=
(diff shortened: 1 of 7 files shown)


go.sum
```diff
@@ -125,6 +125,7 @@ github.com/armon/go-radix v0.0.0-20180808171621-7fddfc383310/go.mod h1:ufUuZ+zHj
 github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a/go.mod h1:lB+ZfQJz7igIIfQNfa7Ml4HSf2uFQQRzpGGRXenZAgY=
 github.com/aws/aws-sdk-go v1.15.11/go.mod h1:mFuSZ37Z9YOHbQEwBWztmVzqXrEkub65tZoCYDt7FT0=
 github.com/aws/aws-sdk-go v1.17.7/go.mod h1:KmX6BPdI08NWTb3/sm4ZGu5ShLoqVDhKgpiN924inxo=
+github.com/aws/aws-sdk-go v1.38.51/go.mod h1:hcU610XS61/+aQV88ixoOzUoG7v3b31pl2zKMmprdro=
```
lzap, line 128: For the record - the "platform-go-middlewares" library depends on AWS SDK v1, however, we will not utilize that feature (logrus cloudwatch SDK logging).

internal/routes/all_routes.go
```diff
@@ -20,6 +21,7 @@ func pathPrefix() string {
 
 func apiRouter() http.Handler {
 	r := chi.NewRouter()
+	r.Use(identity.EnforceIdentity)
```
lzap, line 24: So this middleware stores the identity in the context with key of different type. That is safe and the correct way to work with context, however, since fetching can be quite confusing, can you please make a helper in the ctxval package, we already have the function that returns logger. That could be quite useful already.

internal/testutils/dummy_identity.go
```diff
@@ -0,0 +1,45 @@
+package testutils
```
lzap, line 1: I am trying to hate all utils/tools packages, can we try to figure out something better? I think that this could actually go into the ctxval package where I ask you to put the Get func..

internal/testutils/dummy_identity.go
```diff
@@ -0,0 +1,45 @@
...
+	"testing"
+
+	"github.com/redhatinsights/platform-go-middlewares/identity"
+)
+
+// accountNumber to be used in the tests.
+const accountNumber = "12345"
+
+// orgId to be used in the tests.
+const orgId = "abc-org-id"
```
lzap, line 17: To be consistent with reality, orgId is also a decimal number.

If something is weak in mocking, than it's the test data - very often out of reality.

lzap (changes requested):
Couple of nitpicks.

go.sum
ezr-ondrej: Is there something we can do about that? (don't use the dependency?)

internal/routes/all_routes.go
ezr-ondrej: There is a function for it in the identity package, but we can wrap that, if you feel it's useful :)

internal/testutils/dummy_identity.go
ezr-ondrej: are you sure? I've copied it from other testing data, but it is entirely possible those data were wrong, so just asking if we have a correct understanding of what orgID should look like ^_^

internal/testutils/dummy_identity.go
ezr-ondrej: I do not like mixing production and testing helper code, but if we make it compile only with test flag and make sure it's clear from the name, I'm happy to do so, WDYT?

internal/testutils/dummy_identity.go
lzap: I am not asking to mix code, I am asking for a better name for the package. it can live in internal/test/xxx.

the WithIdentity function I think might be promoted to the production code, I find it useful, we have the same in Foreman as "[User.as](http://user.as/)_admin"

internal/ctxval/ctx_getters.go
```diff
@@ -32,3 +33,7 @@ func GetRequestId(ctx context.Context) string {
 	}
 	return ctx.Value(RequestIdCtxKey).(string)
 }
+
+func GetIdentity(ctx context.Context) identity.XRHID {
+	return identity.Get(ctx)
+}
```
lzap, line 39: Nice thanks I think this is consistent.

lzap (changes requested):
Everything is great, only rename the "testutils" package to something more appropriate like "tests/identity" or something like that.

ezr-ondrej:
I believe I've addressed that, I'll take a look at if we can move that, and use the same code in production in some follow-up, where I'll try to use the header :)

lzap (approved):
Looks so much better! Thanks.

internal/services/pubkey_service_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	dao_stubs "github.com/RHEnVision/provisioning-backend/internal/dao/stubs"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
-	"github.com/RHEnVision/provisioning-backend/internal/testutils"
+	"github.com/RHEnVision/provisioning-backend/internal/test/identity"
```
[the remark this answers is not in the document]
ezr-ondrej: That works for me: #42 :+1:

PR #38, opened 2022-06-13
List sources and Get source using Sources Client

adiabramovitch:
Depends on https://github.com/RHEnVision/provisioning-backend/pull/31

### DIFF
--- a/internal/db/seeds/dev_small.sql
+++ b/internal/db/seeds/dev_small.sql
@@ -8,7 +8,8 @@ INSERT INTO accounts(id, account_number, org_id)
 VALUES (1, '13', '000013'),
        (2, '15', '000015'),
        (3, NULL, '000042'),
-       (4, NULL, '000077')
+       (4, NULL, '000077'),
+       (5, '6089719', '000016')
 ON CONFLICT DO NOTHING;
 
 INSERT INTO pubkeys(id, account_id, name, body)
--- a/internal/payloads/sources_payload.go
+++ b/internal/payloads/sources_payload.go
@@ -1,0 +1,33 @@
+package payloads
+
+import (
+	"net/http"
+
+	sources "github.com/RHEnVision/provisioning-backend/internal/clients/sources"
+	"github.com/go-chi/render"
+)
+
+type SourceResponse struct {
+	*sources.Source
+}
+
+func (s *SourceResponse) Bind(_ *http.Request) error {
+	return nil
+}
+
+func (s *SourceResponse) Render(_ http.ResponseWriter, _ *http.Request) error {
+	return nil
+}
+
+func NewListSourcesResponse(sl *[]sources.Source) []render.Renderer {
+	sList := *sl
+	list := make([]render.Renderer, 0, len(sList))
+	for i := range sList {
+		list = append(list, &SourceResponse{Source: &sList[i]})
+	}
+	return list
+}
+
+func NewShowSourcesResponse(s *sources.Source) render.Renderer {
+	return &SourceResponse{Source: s}
+}
--- a/internal/routes/all_routes.go
+++ b/internal/routes/all_routes.go
@@ -22,6 +22,13 @@ func pathPrefix() string {
 func apiRouter() http.Handler {
 	r := chi.NewRouter()
 	r.Use(identity.EnforceIdentity)
+	r.Route("/sources", func(r chi.Router) {
+		r.Get("/", s.ListSources)
+		r.Route("/{ID}", func(r chi.Router) {
+			r.Get("/", s.GetSource)
+		})
+	})
+
 	r.Route("/accounts", func(r chi.Router) {
 		r.Get("/", s.ListAccounts)
 		r.Route("/{ID}", func(r chi.Router) {
--- a/internal/services/sources_service.go
+++ b/internal/services/sources_service.go
@@ -1,0 +1,49 @@
+package services
+
+import (
+	"net/http"
+
+	sources "github.com/RHEnVision/provisioning-backend/internal/clients/sources"
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/payloads"
+	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/render"
+)
+
+func ListSources(w http.ResponseWriter, r *http.Request) {
+	client, err := sources.NewClientWithResponses(*config.SourcesURL)
+	if err != nil {
+		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "sources client", err))
+		return
+	}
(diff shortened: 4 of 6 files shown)


internal/routes/all_routes.go
```diff
@@ -39,5 +39,12 @@ func apiRouter() http.Handler {
 		})
 	})
 
+	r.Route("/sources", func(r chi.Router) {
+		r.Get("/", s.ListSources)
+		r.Route("/{ID}", func(r chi.Router) {
+			r.Get("/", s.GetSource)
+		})
+	})
+
```
adiabramovitch, line 48: Need to decide if sources/aws or sources?provider=aws

lzap:
I've created a PR which slightly modifies the Makefile so tell me what you think: https://github.com/RHEnVision/provisioning-backend/pull/39

I see a pattern for configuration: for each REST API we will need a configurable option, so perhaps it is worth creating a common configuration group:

```yaml
restEndpoints:
  imageBuilder:
    url: xxxx
  sources:
    url: xxxx
    otherOption: xxxx
```

We can only export RestEndpoints struct, no need to export all values.

adiabramovitch:
> 

Sounds good to me, we can always change it if ou would need more information about the client.

internal/routes/all_routes.go
adiabramovitch: yes I agree

lzap:
There are conflicts FYI.

internal/payloads/sources_payload.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"net/http"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients/sources"
+	"github.com/go-chi/render"
+)
+
+type ShowSourceResponse struct {
+	sources.Source
+}
```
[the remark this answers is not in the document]
lzap: Also I would use pointer in here, generally speaking not using pointers is only good for small structs as a very tight optimalization technique.

internal/db/seeds/dev_small.sql
```diff
@@ -5,7 +5,7 @@
 BEGIN;
 
 INSERT INTO accounts(id, account_number, org_id)
-VALUES (1, '13', '000013'),
+VALUES (1, '6089719', '000013'),
```
adiabramovitch, line 8: I have changed it in order to authenticate both sources and our service :)

ezr-ondrej:
Are we adding tests, or shall that be a follow-up (which I'm totally fine with)? :)

internal/clients/sources/sources_client.gen.go
```diff
@@ -1,7 +1,7 @@
 // Package SourcesApi provides primitives to interact with the openapi HTTP API.
 //
 // Code generated by unknown module path version unknown version DO NOT EDIT.
-package SourcesApi
+package sources
 
 import (
 	"bytes"
```
lzap, line 7: Please do not make manual edits to generate files. Remove it from this patch, I will do a PR to fix this. It looks like generation does not work correctly even.

lzap (changes requested):
Lookin' good!

internal/services/sources_service.go
```diff
@@ -0,0 +1,61 @@
...
+	resp, err := client.ListSourcesWithResponse(r.Context(), &sources.ListSourcesParams{}, AddIdentityHeader)
+	if err != nil {
+		// TODO: use the errors supported by the JSON 400 OR 404
+		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), err))
+		return
+	}
+
+	if err := render.RenderList(w, r, payloads.NewListSourcesResponse(resp.JSON200.Data)); err != nil {
+		// TODO: use the errors supported by the JSON 400 OR 404
+		renderError(w, r, payloads.NewRenderError(r.Context(), "list sources", err))
```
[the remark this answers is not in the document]
adiabramovitch: No, In the resp struct there are also errors documented which I want to show as well on a different PR.

internal/services/sources_service.go
```diff
@@ -0,0 +1,61 @@
...
+		return
+	}
+
+}
+
+func GetSource(w http.ResponseWriter, r *http.Request) {
+	id := chi.URLParam(r, "ID")
```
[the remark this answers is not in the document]
adiabramovitch: I want it as a string not an int

internal/clients/sources/sources_client.gen.go
lzap: FYI the PR is here, WIP: https://github.com/RHEnVision/provisioning-backend/pull/46

internal/services/sources_service.go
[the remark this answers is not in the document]
lzap: Sorry disregard.

adiabramovitch:
@lzap , @ezr-ondrej I will open today PRs for adding more error handling and the UI dropdown. 
Regarding tests, I'll add those as well in a different PR I want the PRs to be as small as possible. 

@lzap  I addressed your comments.

ezr-ondrej (commented):
I do not know how to test a happy path (how do I seed sources data?), but when sources return 404, we panic...

internal/db/seeds/dev_small.sql
```diff
@@ -8,7 +8,8 @@ INSERT INTO accounts(id, account_number, org_id)
 VALUES (1, '13', '000013'),
        (2, '15', '000015'),
        (3, NULL, '000042'),
-       (4, NULL, '000077')
+       (4, NULL, '000077'),
+       (5, '6089719', '000016')
```
adiabramovitch, line 12: You need to make sure you have this org_id linked to this account number in sources. 
(It is there on default) @ezr-ondrej

lzap: Sure makes sense, but not all our endpoints will interact with sources. Our own accounts still make sense.

internal/services/sources_service.go
```diff
@@ -0,0 +1,53 @@
...
+
+func ListSources(w http.ResponseWriter, r *http.Request) {
+	client, err := sources.NewClientWithResponses(*config.SourcesURL)
+	if err != nil {
+		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "sources client", err))
+		return
+	}
+	resp, err := client.ListSourcesWithResponse(r.Context(), &sources.ListSourcesParams{}, AddIdentityHeader)
+	if err != nil {
+		// TODO: use the errors supported by the JSON 400 OR 404
```
lzap, line 21: What does this mean? Obviously we want to use 404 when a resource is not found, which is probably not the case for listing operation.

The code you have is good, you return 3rd party error when unknown error occurs which is HTTP 500 which is correct I think. That is the code for "unknown backend errors".

lzap (dismissed):
It's perfect, one nitpick about the "New" in the error message.

ezr-ondrej (approved):
Let's get it in, lets get the filtering by provider and tests in follow ups :+1:

PR #39, opened 2022-06-13
Split updating and generating of clients

lzap:
This PR fixes Makefile which was broken and it splits the generation into two steps.

The new target *update-clients* performs `wget` and updates the OpenAPI specs. We will run this only when we want to update API.

The existing target only generates Go code from OpenAPI, it is no longer a phony target and it works as expected - the tool will only be executed when there was no change in YAML files or OpenAPI.

I wanted to exclude all `gen.go` files from the linter, but it looks like both fmt and lint tools do already either ignore the generated files or the generated code is good enough.


internal/clients/image_builder/client.gen.go
```diff
@@ -15,7 +15,6 @@ import (
 	"strings"
 
 	"github.com/deepmap/oapi-codegen/pkg/runtime"
-	"github.com/go-chi/chi/v5"
 )
```
lzap, line 18: Also the generated code by @adiabramovitch was generated with "server" code. We don't actually need that.

Makefile
```diff
@@ -29,16 +29,15 @@ prep:
...
-generate-ib-client:
-	cd internal/clients/image_builder
-	wget https://raw.githubusercontent.com/osbuild/image-builder/main/internal/v1/api.yaml
+.PHONY: update-clients
+update-clients:
+	wget -O ./configs/ib_api.yaml -qN https://raw.githubusercontent.com/osbuild/image-builder/main/internal/v1/api.yaml
+
+generate-clients: internal/clients/image_builder/client.gen.go
+
+internal/clients/image_builder/client.gen.go: configs/ib_config.yaml configs/ib_api.yaml
```
adiabramovitch, line 38: Why is that written twice?

ezr-ondrej: it is a specification of make target, so this file is a target, that should get created here and depends on those two files (if those change, makefile will know to re-make the target. It's quite cool, but bit messy xD

ezr-ondrej (commented):
You've removed the generated file, I thought the strategy is to keep it in the code?
Was that intentional, or accident?

ezr-ondrej (approved):
Thanks @adiabramovitch  for the explanation, in that case I'm good here, I'll let @adiabramovitch to merge if she's ok as well ^_^

Makefile
adiabramovitch: but how is this file created in the first place?

ezr-ondrej: right, you're correct, this will probably not create the file, as we do not redirect the output.

lzap: It does create the file, I do not file PRs which do not work (unless stated in the description) :-) The configuration YAML actually specifies the output directory.

@adiabramovitch yeah this is how the ancient GNU Make (or any other makefile software) treats dependencies and there is no other way to do it. It typically looks nicer, granted, our path is long and we can make it nicer with variables later on.

lzap:
Yep thanks all is good.

PR #40, opened 2022-06-13
Fix bonfire config path

amirfefer:
the bonfire config yaml file  has renamed due to depreciation

### DIFF
--- a/deploy/bonfire.example.yaml
+++ b/deploy/bonfire.example.yaml
@@ -4,7 +4,7 @@
 appsFile:
   host: gitlab
   repo: insights-platform/cicd-common
-  path: bonfire_configs/ephemeral_apps.yaml
+  path: bonfire_configs/ephemeral_apps_deprecated.yaml
 
 # (optional) define any apps locally. An app defined here with <name> will override config for app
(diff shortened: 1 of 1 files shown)


ezr-ondrej (approved):
We'll need to find a proper solution, but great workaround! :+1:
@gburges would you know what is the non-deprecated solution here? ^_^

gburges:
> We'll need to find a proper solution, but great workaround! +1 @gburges would you know what is the non-deprecated solution here? ^_^

Bonfire uses the ephemeral namespace conf now in app-interface by default now - I think this is the 'unsupported' fallback option, but i'll find out for sure how non-app0interfaced apps can use bonfire dependency

ezr-ondrej:
Awesome! :) the [documentation](https://consoledot.pages.redhat.com/docs/dev/getting-started/ephemeral/deploying.html#_local_config_source) is still pointing to [default config](https://github.com/RedHatInsights/bonfire/blob/master/bonfire/resources/default_config.yaml), which now does not work, happy to contribute the correct way to the docs if we figure out what is the correct way :)

PR #41, opened 2022-06-13
Update deploy ephemeral doc for mac users

amirfefer:


ezr-ondrej (approved):
Thanks! It's great to be mac friendly service ^_^

PR #45, opened 2022-06-14
Add identities to .http test clients

lzap:
SSIA

This is only used for testing, GoLand and VSCode do support these.

### DIFF
--- a/scripts/rest_examples/http-client.env.json
+++ b/scripts/rest_examples/http-client.env.json
@@ -1,0 +1,8 @@
+{
+  "dev": {
+    "hostname": "localhost",
+    "port": "8000",
+    "prefix": "api/provisioning",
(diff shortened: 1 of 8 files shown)


ezr-ondrej (approved):
cool! thanks @lzap !

scripts/rest_examples/http-client.env.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dev": {
+    "hostname": "localhost",
+    "port": "8000",
+    "prefix": "api/provisioning",
+    "identity": "eyJpZGVudGl0eSI6IHsidHlwZSI6ICJVc2VyIiwgImFjY291bnRfbnVtYmVyIjoiMTMiLCAiaW50ZXJuYWwiOnsib3JnX2lkIjoiMDAwMDEzIn19fQo="
+  }
```
lzap, line 7: We can probably create multiple of these and name them "user account 13 org_id 000013" or something like that so ppl can switch easily.

PR #46, opened 2022-06-15
Fix and regenerate clients

lzap:
They apparently changed the configuration format, on their README they have the old syntax:

https://github.com/deepmap/oapi-codegen#using-oapi-codegen

But if you click on the example they link from the README it changed:

https://github.com/deepmap/oapi-codegen/blob/master/internal/test/externalref/externalref.cfg.yaml

This was causing the generator to fail with YAML parsing error. This fixes it, also I renamed the generated file to `client.gen.go` for both cases for consistency. Also excluded all generated files from some linting problems.

I think we should add a GHA task to also check if generated files aren't any different from what we have I guess.


lzap:
Perhaps this explicit check would do it:

```
lzap@mone provisioning-backend % [[ internal/clients/image_builder/client.gen.go -nt configs/ib_api.yaml ]] && echo OK
OK
```

lzap:
Looks like the generated output does not compile anymore, weird:

```
# github.com/RHEnVision/provisioning-backend/internal/clients/sources
Error: internal/clients/sources/client.gen.go:108:2: EndpointAvailabilityStatus is not a type
Error: internal/clients/sources/client.gen.go:109:2: EndpointAvailabilityStatus is not a type
Error: internal/clients/sources/client.gen.go:115:2: EndpointCreateAvailabilityStatus is not a type
Error: internal/clients/sources/client.gen.go:116:2: EndpointCreateAvailabilityStatus is not a type
Error: internal/clients/sources/client.gen.go:122:2: EndpointEditAvailabilityStatus is not a type
Error: internal/clients/sources/client.gen.go:655:6: EndpointAvailabilityStatus redeclared in this block
	previous declaration at internal/clients/sources/client.gen.go:107:69
Error: internal/clients/sources/client.gen.go:694:6: EndpointCreateAvailabilityStatus redeclared in this block
	previous declaration at internal/clients/sources/client.gen.go:114:[81](https://github.com/RHEnVision/provisioning-backend/runs/6895367816?check_suite_focus=true#step:4:82)
Error: internal/clients/sources/client.gen.go:739:6: EndpointEditAvailabilityStatus redeclared in this block
	previous declaration at internal/clients/sources/client.gen.go:121:77
Error: internal/clients/sources/client.gen.go:799:6: EndpointReadAvailabilityStatus redeclared in this block
	previous declaration at internal/clients/sources/client.gen.go:128:77
Error: internal/clients/sources/client.gen.go:6707:6: BulkCreateResponse redeclared in this block
	previous declaration at internal/clients/sources/client.gen.go:556:6
Error: internal/clients/sources/client.gen.go:122:2: too many errors
```

lzap:
Well that was fast, I think the sources spec suffers from this bug in the generator: https://github.com/deepmap/oapi-codegen/issues/549

The recent version 0.11 made some changes and introduced bug. Older versions do work fine. There appears to be switch to use the old behavior, I am trying that.

adiabramovitch:
I have said that they have a problem with their generated file. I had to d changes their because two different structs were generated and some code was duplicated, you can see that I changes it to  BulkCreatePayloadResponse to solve it.

lzap:
> I have said that they have a problem with their generated file. I had to d changes their because two different structs were generated and some code was duplicated, you can see that I changes it to BulkCreatePayloadResponse to solve it.

I missed that, well, making changes to generated files is not a great plan for the long-term. If we cannot get away with excluding, then we will have to write our own client by hand.

configs/ib_config.yaml
```diff
@@ -1,8 +1,6 @@
---- 
-generate:
-  - client
-  - types
+---
 package: image_builder
+generate:
+  models: true
```
adiabramovitch, line 4: this did not work for me adding the models here, it did not recognize this option. and it is not written as an option in oapi.
Can you explain?

ezr-ondrej: I think they've changed the options in the new version, but I can't find documentation for this new format...

lzap: You used an old version, the current (1.11) only supports this syntax.

Go ahead and try my patch, checkout and then:

```
make install-tools
rm internal/clients/*/*.gen.go
make generate-clients
```

lzap: Just to explain a little bit more, the key command you need to run is:

    go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@latest

This will get you 1.11 today.

adiabramovitch:
So I have success to generate the clients, is there something more to check or we can merge it?

ezr-ondrej (approved):
If that works, lets get it in! :)

lzap:
Thanks, if you are missing any endpoints just add them via the "tag" functionality. You will see list of tags in the schema itself.

PR #47, opened 2022-06-15
Add make for purging DB

ezr-ondrej:
We should purge DB before migrations as migrations can fail and then 
there is no easy way to purge DB.


lzap (approved):
TY

PR #48, opened 2022-06-15
Update client config

ezr-ondrej:


lzap:
I am almost done with this in https://github.com/RHEnVision/provisioning-backend/pull/46 :)

We need more changes tho. The latest version of the generator does not work with sources spec.

ezr-ondrej:
> I am almost done with this in #46 :)

I was just testing it out ^_^

PR #49, opened 2022-06-15
Replace gomigrate by tern in tools

ezr-ondrej:
@lzap this is not needed now, right?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ run:
 .PHONY: update-clients
 update-clients:
 	wget -O ./configs/ib_api.yaml -qN https://raw.githubusercontent.com/osbuild/image-builder/main/internal/v1/api.yaml
-	wget -O ./configs/sources_api.json -qN https://raw.githubusercontent.com/RedHatInsights/sources-api-go/main/public/openapi-3-v3.1.json 
+	wget -O ./configs/sources_api.json -qN https://raw.githubusercontent.com/RedHatInsights/sources-api-go/main/public/openapi-3-v3.1.json
 
 generate-clients: internal/clients/image_builder/client.gen.go internal/clients/sources/sources_client.gen.go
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -47,7 +47,6 @@ install-tools:
 	go install golang.org/x/tools/cmd/goimports@latest
 	# pin for a bug: https://github.com/golangci/golangci-lint/issues/2851
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.45.2
-	go install -tags 'postgres' github.com/golang-migrate/migrate/v4/cmd/migrate@latest
```
lzap, line 50: Can you replace this with

    go install github.com/jackc/tern@latest

instead? Actually the tern CLI is quite useful and we can use it, just add `tern.conf` into .gitignore and people can configure their database and then issue the tern command to operate on it (migrate, get status, create new migration file).

Makefile
```diff
@@ -32,7 +32,7 @@ run:
 .PHONY: update-clients
 update-clients:
 	wget -O ./configs/ib_api.yaml -qN https://raw.githubusercontent.com/osbuild/image-builder/main/internal/v1/api.yaml
-	wget -O ./configs/sources_api.json -qN https://raw.githubusercontent.com/RedHatInsights/sources-api-go/main/public/openapi-3-v3.1.json 
+	wget -O ./configs/sources_api.json -qN https://raw.githubusercontent.com/RedHatInsights/sources-api-go/main/public/openapi-3-v3.1.json
```
lzap, line 35: I have that in my other PR just drop it if you can.

Makefile
ezr-ondrej: sure :)

lzap (approved):
Yeah, that's useful. Thanks.

PR #50, opened 2022-06-15
Add filtering by source type when listing sources

adiabramovitch:
This PR is one draft mode until M1 or until sources will fix their filtering bug :)

### DIFF
--- a/internal/services/sources_service.go
+++ b/internal/services/sources_service.go
@@ -16,7 +16,12 @@ func ListSources(w http.ResponseWriter, r *http.Request) {
 		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "sources client", err))
 		return
 	}
-	resp, err := client.ListSourcesWithResponse(r.Context(), &sources.ListSourcesParams{}, AddIdentityHeader)
+
+	name := r.URL.Query().Get("provider")
+	filter := sources.QueryFilter("[source_type][name]=" + name)
+
+	resp, err := client.ListSourcesWithResponse(r.Context(), &sources.ListSourcesParams{Filter: &filter}, AddIdentityHeader)
+
 	if err != nil {
 		renderError(w, r, payloads.New3rdPartyClientError(r.Context(), "list sources", err))
 		return


internal/services/sources_service.go
```diff
@@ -10,21 +10,26 @@ import (
 	"github.com/go-chi/render"
 )
 
+// var sourceTypeName = map[string]int{"google": 7, "amazon": 5, "azure": 12}
+
```
lzap, line 14: You will find find our cloud constants at model.ProviderType (that's the type, there are three constants there).

Are these numbers 7, 5, 12 from sources database? What are these? Database IDs?

lzap: I am afraid we will have to perform an extra query to fetch SourceType and from there we get the proper database ID.

lzap: Oh it looks like sources support some kind of advanced filtering, that could actually work (filter by source type name): https://github.com/RedHatInsights/sources-api-go/blob/2d5f69436137b7e8e3f96fccb35c50355f5fe140/util/filtering.go#L19

Need to dig deeper.

adiabramovitch: yes, these are the ids from sources regarding the source_type_id when I looked at /source_types

lzap: This looks, better, now, we must not expose other's constants/names through our API. We need our own ids/names for target environments. Can you write code which will actually map the three providers we have defined in `model_constants.go` to SourceType names?

I think we should stick with our own naming convention, whatever we end up using we should use for all our API endpoints. I named the constants as "providers", but let's pick better term if folks don't like it.

internal/services/sources_service.go
```diff
@@ -16,7 +16,11 @@ func ListSources(w http.ResponseWriter, r *http.Request) {
 		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "sources client", err))
 		return
 	}
-	resp, err := client.ListSourcesWithResponse(r.Context(), &sources.ListSourcesParams{}, AddIdentityHeader)
+
+	// name := chi.URLParamFromCtx(r.Context(), "source_type")
+	filter := sources.QueryFilter("[source_type][name]=amazon")
+	resp, err := client.ListSourcesWithResponse(r.Context(), &sources.ListSourcesParams{Filter: &filter}, AddIdentityHeader)
+
```
adiabramovitch, line 23: From what I understood this is the correct way in order to add filters to their request but it does not work for me. 
When I am requesting sources directly, it works: 
curl --location -g  --request GET 'http://localhost:8000/api/sources/v3.1/sources?filter[source_type][name]=amazon' --header 'x-rh-identity: IDENTITY'
but when I am trying it through our service this is the query that is being formed using: https://github.com/RHEnVision/provisioning-backend/blob/eeb276f93a8629d114b2a89e72ea8c978e6bc3bc/internal/clients/sources/client.gen.go#L1229
```golang
http://localhost:8000/api/sources/v3.1/sources?filter=%3F%5Bsource_type%5D%5Bname%5D%3Damazon
``` 
and this is the error I am getting on sources side :
```golang
runtime error: index out of range [1] with length 1 goroutine 106 
```
@lzap, @ezr-ondrej Can you take a look as well?

internal/services/sources_service.go
```diff
@@ -16,7 +16,12 @@ func ListSources(w http.ResponseWriter, r *http.Request) {
 		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "sources client", err))
 		return
 	}
-	resp, err := client.ListSourcesWithResponse(r.Context(), &sources.ListSourcesParams{}, AddIdentityHeader)
+
+	name := r.URL.Query().Get("provider")
+	filter := sources.QueryFilter("[source_type][name]=" + name)
+
+	resp, err := client.ListSourcesWithResponse(r.Context(), &sources.ListSourcesParams{Filter: &filter}, AddIdentityHeader)
```
adiabramovitch, line 23: from what I understood this is the correct way to filter their requests but unfortunately it does not work dor me.
When I am requesting sources directly It works: 
```golang
curl --location -g  --request GET 'http://localhost:8000/api/sources/v3.1/sources?filter[source_type][name]=amazon' --header 'x-rh-identity: IDENTITY' 
```

But when I am trying to use our service this is the query that is being formed using: https://github.com/RHEnVision/provisioning-backend/blob/eeb276f93a8629d114b2a89e72ea8c978e6bc3bc/internal/clients/sources/client.gen.go#L1229  
```golang
http://localhost:8000/api/sources/v3.1/sources?filter=%5Bsource_type%5D%5Bname%5D%3Damazon
```
This is not working and I am getting this error from sources side: 
```golang
"message":"[PANIC RECOVER] runtime error: index out of range [1] with length 1 goroutine 121 
```
@lzap , @ezr-ondrej  Can you take a look as well?

adiabramovitch: I found the bug and asked sources on their channel.

lzap: Okay thanks for info, ping us once this is ready for testing. I have my local setup already.

PR #52, opened 2022-06-16
Add sources integration an interface and implementation

adiabramovitch:

**Checklist**

### DIFF
--- a/internal/clients/sources/sources_impl.go
+++ b/internal/clients/sources/sources_impl.go
@@ -1,0 +1,15 @@
+package sources
+
+import (
+	"context"
+
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+)
+
+func getSourcesClient(ctx context.Context) (SourcesIntegration, error) {
+	return NewClientWithResponses(*config.SourcesURL)
+}
+
+func init() {
+	GetSourcesClient = getSourcesClient
+}
--- a/internal/clients/sources/sources_interface.go
+++ b/internal/clients/sources/sources_interface.go
@@ -1,0 +1,12 @@
+package sources
+
+import (
+	"context"
+)
+
+var GetSourcesClient func(ctx context.Context) (SourcesIntegration, error)
+
+type SourcesIntegration interface {
+	ShowSourceWithResponse(ctx context.Context, id ID, reqEditors ...RequestEditorFn) (*ShowSourceResponse, error)
+	ListSourcesWithResponse(ctx context.Context, params *ListSourcesParams, reqEditors ...RequestEditorFn) (*ListSourcesResponse, error)
+}
--- a/internal/payloads/sources_payload.go
+++ b/internal/payloads/sources_payload.go
@@ -3,7 +3,7 @@ package payloads
 import (
 	"net/http"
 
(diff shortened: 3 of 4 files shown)


internal/clients/sources/sources_interface.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+)
+
+var GetSourcesImpl func(ctx context.Context) (SourcesIntegration, error)
+
+type SourcesIntegration interface {
+	GetById(ctx context.Context, id string) (*Source, error)
+	List(ctx context.Context) (*[]Source, error)
```
ezr-ondrej, line 11: I'd have `ListSourcesWithResponse` and `ShowSourceWithResponse` methods here, that would copy the method interface from the real thing. Wrapping can be useful, but I'd love to stick with the real thing until we hit the need for wrapping..

lzap:
Nice, so we can mock on this level too.

internal/clients/sources/sources_real.go
```diff
@@ -0,0 +1,12 @@
+package sources
+
+import (
+	"context"
+)
+
+var GetSourcesImpl func(ctx context.Context) (SourcesIntegration, error)
```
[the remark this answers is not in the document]
ezr-ondrej: Yeah, probably a bit better name :+1:

internal/clients/sources/sources_impl.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+)
+
+func getSourcesClient(ctx context.Context) (SourcesIntegration, error) {
+	return NewClientWithResponses(*config.SourcesURL)
+}
+
+func init() {
+	GetSourcesClient = getSourcesClient
+}
```
lzap, line 15: Ah can you also rename the file to `sources_interface.go`? I have missed that.

internal/clients/sources/sources_real.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"context"
+)
+
+var GetSourcesClient func(ctx context.Context) (SourcesIntegration, error)
+
+type SourcesIntegration interface {
+	ShowSourceWithResponse(ctx context.Context, id ID, reqEditors ...RequestEditorFn) (*ShowSourceResponse, error)
+	ListSourcesWithResponse(ctx context.Context, params *ListSourcesParams, reqEditors ...RequestEditorFn) (*ListSourcesResponse, error)
+}
```
lzap, line 12: Sorry this should have been `sources_interface.go`.

internal/clients/sources/sources_impl.go
lzap: Correction, this one is actually correct, the other one.

internal/clients/sources/sources_real.go
lzap: One detail: We use RequestEditorFn interface from the generated code, ideally we should abstract that.

lzap (approved):
Abstracting RequestEditorFn and *Response structs is probably something we can do later or when needed.

PR #53, opened 2022-06-16
Use int64 for PKs in OpenAPI spec

lzap:
We use BIGINT for PKs in the database, these are signed int64. In models we have uint64 for some reason, this is incorrect and it cannot be mapped to OpenAPI (which does not support unsigned integer types).

This PR fixes the model and also improves generation of OpenAPI (renamed output artifacts, reordered paths/components, added yaml `---` header, improved documentation a bit). Then I regenerated the specs.

I also found that the phony name for the dbpurge was incorrect (dropdb) so I synced it on "purgedb" and also created make clean target :)


ezr-ondrej (approved):
Awesome, nice improvements, that's what I've aimed at by merging the previous PR, but did not tought it through.

lzap:
Ok we need the validator now, @amirfefer :)

PR #54, opened 2022-06-17
Scripts for quick sources backend setup

lzap:
Based on @adiabramovitch 's work, I created those scripts which should help new users to get Sources backend up and running in minutes. I automated everything, so even someone with a fresh Fedora should get it running just by executing a script or two.

See the README.sources file for additional information. I would love someone to test this, I did my best but my Fedora was upgraded from let's say Fedora 19 so it's a tad different configuration than others :)

Cheers


scripts/sources.conf
```diff
@@ -0,0 +1,24 @@
...
+export DATABASE_HOST=localhost
+export DATABASE_PORT=5432
+export DATABASE_NAME=sources_devel
+export DATABASE_USER=sources_user
+# Local users are considered trusted via pg_hba.conf but
+# password must be present otherwise sources won'b boot.
+export DATABASE_PASSWORD=trusted
+export CACHE_HOST=localhost
+export CACHE_PORT=6379
+export CACHE_PASSWORD=no_authentication
```
ezr-ondrej, line 17: I do not think these are used :thinking: I was looking for them in their code and it is not there...

scripts/sources.conf
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+#
+# Do not edit this file, create sources.local.conf for overriding.
+#
+
+# Some random port because 8000 is typically already used.
+export PORT=8131
```
ezr-ondrej, line 7: can we also set `METRICS_PORT`? They have 9000 as well as we do and it could collide, right?

ezr-ondrej (commented):
nice, few nitpicks, but so minor I'm okay merging as is :shrug:

scripts/sources.conf
lzap: @adiabramovitch can you comment? You had it there so perhaps there is a reason.

scripts/sources.conf
lzap: Good point, added.

scripts/sources.conf
adiabramovitch: I don't recall, If you didn't need those, maybe they are redundant.

adiabramovitch:
Maybe we should add the org id to the account number that was seeded to the DB. (In order to authenticate both sources and provisioning)

ezr-ondrej:
> Maybe we should add the org id to the account number that was seeded to the DB. (In order to authenticate both sources and provisioning)

That's already in our `dev_small`, right?

ezr-ondrej (approved):
Duplication is fine I guess, I'd get it in, it works and breaks nothing :)
@adiabramovitch anything to add?

scripts/sources.conf
lzap: Ok let's drop it then.

lzap:
Dropped the tree REDIS unused variables.

I am not sure how you want me to add account number, the poppullator does not support it:

```
$ ag Getenv
config/parser.go
83:	logLevel := os.Getenv("LOG_LEVEL")
91:	sourcesHost := os.Getenv("SOURCES_API_HOST")
97:	sourcesPort := os.Getenv("SOURCES_API_PORT")
107:	concurrentRequests := os.Getenv("CONCURRENT_REQUESTS")
125:	numberTenants := os.Getenv("NUMBER_OF_TENANTS")
167:	sourcesPerTenant := os.Getenv("SOURCES_PER_TENANT")
180:	rhcConnectionsPerTenant := os.Getenv("RHC_CONNECTIONS_PER_TENANT")
193:	applicationsPerSource := os.Getenv("APPLICATIONS_PER_SOURCE")
206:	endpointsPerSource := os.Getenv("ENDPOINTS_PER_SOURCE")
218:	authenticationsPerResource := os.Getenv("AUTHENTICATIONS_PER_RESOURCE")
```

lzap:
Let me take a look on the seeder application to see how they actually create tenants. We can probably create our own one via curl too.

adiabramovitch:
> > because they are saving only the account number and not the org id, and without adding this we can authenticate both of the services.
> 
> Ah! got it!
> 
> > Let me take a look on the seeder application to see how they actually create tenants. We can probably create our own one via curl too.
> 
> Could we just have a DB query adding static org_id into the db? :) I'd not worry about it too much to make it pretty, org_id should go away in a year anyway :)

I might be mistaken. 
I asked @lzap  to try a simple curl request through our service to check if it is authenticating both services, I think the seeding of the x-rh-identity occurs automatically.

lzap:
Yep I can confirm, a simple curl that lists something automatically adds account number into the tenants table. It ignores org id tho, I think this is a bug (org id has only been added recently). Example:

```
curl --location --request GET 'http://localhost:8131/api/sources/v3.1/authentications' --header 'x-rh-identity: eyJpZGVudGl0eSI6IHsidHlwZSI6ICJVc2VyIiwgImFjY291bnRfbnVtYmVyIjoiNjA4OTcxOSIsICJpbnRlcm5hbCI6eyJvcmdfaWQiOiIwMDAwMTMifX19Cg==' | json_pp
```

lzap:
Just for the record, I think we can take the same approach then. When we encounter a request with account/org ids, we would find or create the record in our accounts database. Do we want to create a ticket for that @ezr-ondrej ?

lzap:
Created: https://issues.redhat.com/browse/HMSPROV-106

ezr-ondrej (approved):
While testing I found it bit challanging to having to swich directories to `/scripts` for this to work properly, but that's the only caveat, otherwise it works well, I don't want to block this on that, it is already quite useful as-is :)

ezr-ondrej:
Thanks @lzap, great job! :+1:

PR #55, opened 2022-06-19
Add tests for ListSources and GetSource

adiabramovitch:
@lzap, @ezr-ondrej  Not sure if those tests are sufficient but please share your thoughts, this PR is based on Ondrej's previous testing.

### DIFF
--- a/internal/clients/sources/stubs/sources_stub.go
+++ b/internal/clients/sources/stubs/sources_stub.go
@@ -1,0 +1,52 @@
+package stubs
+
+import (
+	"context"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients/sources"
+)
+
+type sourcesCtxKeyType string
+
+var sourcesCtxKey sourcesCtxKeyType = "sources-interface"
+
+type SourcesIntegrationStub struct {
+	store *[]sources.Source
+}
+
+func init() {
+	sources.GetSourcesClient = getSourcesClientStub
+}
+
+type contextReadError struct{}
+
+func (m *contextReadError) Error() string {
+	return "failed to find or convert dao stored in testing context"
+}
+
+func WithSourcesIntegration(parent context.Context, init_store *[]sources.Source) context.Context {
+	ctx := context.WithValue(parent, sourcesCtxKey, &SourcesIntegrationStub{init_store})
+	return ctx
+}
+
+func getSourcesClientStub(ctx context.Context) (si sources.SourcesIntegration, err error) {
+	var ok bool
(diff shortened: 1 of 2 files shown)


internal/clients/sources/stubs/sources_stub.go
```diff
@@ -0,0 +1,56 @@
...
+type sourcesCtxKeyType string
+
+var sourcesCtxKey sourcesCtxKeyType = "sources-interface"
+
+type SourcesIntegrationStub struct {
+	store *[]sources.Source
+}
+
+func init() {
+	sources.GetSourcesClient = getSourcesClient
```
ezr-ondrej, line 18: This could have been simplified :) I use one more method, because I'm calling the method from two different places with differet params, but we have just one caller with fixed params here :)
```suggestion
	sources.GetSourcesClient = getSourcesClientStub
```

ezr-ondrej (approved):
Two general not important comments:

- I'd not necessarily expect this to be as dynamic as our  database stubs, so I'd be ok with making the data static, but this is also fine, it might be useful
- I'd like to verify the account header being sent as well, but not necessarily in this PR and not necessarily as part of these tests (I don't know what would be the best)

So I'm happy with what we have here :)

internal/clients/sources/stubs/sources_stub.go
adiabramovitch: Do you mean like that? :) 
```golang
func init() {
	sources.GetSourcesClient = getSourcesClientStub
}

func getSourcesClientStub(ctx context.Context) (si sources.SourcesIntegration, err error) {
	var ok bool
	if si, ok = ctx.Value(sourcesCtxKey).(*SourcesIntegrationStub); !ok {
		err = &contextReadError{}
	}
	return si, err
}
```

adiabramovitch: > Two general not important comments:
> 
> * I'd not necessarily expect this to be as dynamic as our  database stubs, so I'd be ok with making the data static, but this is also fine, it might be useful
> * I'd like to verify the account header being sent as well, but not necessarily in this PR and not necessarily as part of these tests (I don't know what would be the best)
> 
> So I'm happy with what we have here :)

I thought of using the fixtures idea like we used to but I wasn't sure what is the right approach, therefore I am asking now :)

ezr-ondrej: I've not been sure how to fixture neighter, so I'd just not for now and keep the approach you used.
What I meant was having static data in thu stub directly, but this is bit more flexible, so lets just keep what you have :)

adiabramovitch:
@lzap  any thoughts?

lzap (approved):
No comments.

PR #57, opened 2022-06-21
Validate spec and clients via Makefile

lzap:
I wanted to test https://github.com/RHEnVision/provisioning-backend/pull/51 locally but I have realized we don't do it in the Makefile. So I moved it there.

On top of that, I implemented the same validation for OpenAPI clients so they are never out of date from the spec we have in the repo.

### DIFF
--- a/.github/workflows/gotest.yml
+++ b/.github/workflows/gotest.yml
@@ -44,7 +44,7 @@ jobs:
           make test
 
   openapi:
-    name: validate and compare spec
+    name: validate and compare spec and client code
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Might deserve a rename of the GH job now, as it does more, but apart of it awesome stuff! :orange_heart:

lzap:
Renamed job description, the name seems okay: "openapi".

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #58, opened 2022-06-21
Add application type endpoints for sources

ezr-ondrej:
We generated only sources endpoints, but we also need the application 
types.
This is adding that into the config and regenerates the client.

### DIFF
--- a/internal/clients/sources/client.gen.go
+++ b/internal/clients/sources/client.gen.go
@@ -152,6 +152,34 @@ const (
 	SourceEditAvailabilityStatusUnavailable        SourceEditAvailabilityStatus = "unavailable"
 )
 
+// AppMetaData defines model for AppMetaData.
+type AppMetaData struct {
+	// The ID of the application the metadata belongs to
+	ApplicationTypeId *string `json:"application_type_id,omitempty"`
+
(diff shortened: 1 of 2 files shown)


internal/clients/sources/client.gen.go
```diff
@@ -2658,6 +3267,159 @@ func (c *ClientWithResponses) UnpauseSourceWithResponse(ctx context.Context, id
...
+		response.JSON404 = &dest
+
+	}
+
+	return response, nil
+}
+
 // ParseBulkCreateResponse parses an HTTP response from a BulkCreateWithResponse call
 func ParseBulkCreateResponse(rsp *http.Response) (*BulkCreateResponse, error) {
 	bodyBytes, err := ioutil.ReadAll(rsp.Body)
```
lzap, line 3425: Was this change intentional? Why you added bunch of methods?

ezr-ondrej: It was, we have figured out we need more endpoints and we now only filter by enpoint tags, so we've added tag that has the endpoint required. I'd be happy to filter out what we do not use, but now we have just tags... so that is the control we use. We might end up dropping the `sources` tag in the end, but we are using it for now.

PR #59, opened 2022-06-21
Add route for openapi spec

amirfefer:
This PR adds two routes: 
1. `<basePath>/openapi.json` for serving the static spec file
2. `<basePath>/docs` - auto docs by [redoc](https://github.com/Redocly/redoc) 

Screen Shot 2022-06-21 at 23 35 17


internal/routes/all_routes.go
```diff
@@ -36,6 +45,10 @@ func apiRouter() http.Handler {
 		})
 	})
 
+	r.Get("/openapi.json", func(w http.ResponseWriter, r *http.Request) {
```
ezr-ondrej, line 48: Is that necessary to gate behind Identity enforcement and should it live under the path prefix?

cmd/spec/path.yaml
```diff
@@ -4,7 +4,7 @@ info:
   license:
     name: GPL-3.0
   title: provisioning-api
-  version: 0.0.2
+  version: 1.0.0
```
ezr-ondrej, line 7: Are we sure we want to bump directly to 1.0? I'd wait with that for the time we go live (to production), no?

lzap: I agree that 1.0.0 is the best start, let's just not touch the version until our official release and only then let's start bumping the number. No guarantees for now.

internal/routes/all_routes.go
lzap: This should not be gated indeed, just put this to the root namespace.

lzap:
Nice, I wonder if we could also run https://github.com/Redocly/redocly-cli utility on GHA to validate the OpenAPI against the same codebase that is used for the docs to ensure it's always correct. Not mandatory part of this PR of course.

internal/config/main_config.go
```diff
@@ -74,6 +77,7 @@ var AWS = &config.AWS
 var Features = &config.FeatureFlags
 var ImageBuilderURL = &config.RestEndpoints.ImageBuilder.URL
 var SourcesURL = &config.RestEndpoints.Sources.URL
+var OpenAPIPath = &config.OpenAPI.Path
```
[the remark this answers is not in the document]
ezr-ondrej: Yeah I agree that exporting the final config options might lead to a lot of variables for no reason.
With this suggestion you write `OpenAPI.Path` instead of `OpenAPIPath`, which is maybe even bit more readable :)

cmd/spec/path.yaml
amirfefer: The main idea is to start already with 1.0.0, and start bumping it only after official deployment
BTW, the provisioning API is going to be consumed by other services in stage env, so it makes sense to me to start versioning when our API is been used.

internal/routes/all_routes.go
amirfefer: the docs are metadata mainly for devs, that's why I was thinking to gate it behind the identity

internal/config/main_config.go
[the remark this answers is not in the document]
amirfefer: I agree, we will end with many vars.
I'll fix the other vars in a separate PR for easy testing :)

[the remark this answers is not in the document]
lzap: Sure, I just wanted to comment on this one :)

internal/routes/all_routes.go
```diff
@@ -21,7 +30,7 @@ func pathPrefix() string {
 
 func apiRouter() http.Handler {
 	r := chi.NewRouter()
-	r.Use(identity.EnforceIdentity)
+	r.Use(identity.EnforceIdentity, redocMiddleware)
```
lzap, line 33: This code inserts a documentation middleware to all API requests. I do not understand why we would need a middleware for data endpoints, this slows down the request processing and adds another function call to the stack which can become crowded very soon.

I suggest to create a path that will be dedicated for OpenAPI, something like:

```go
r.Route("/openapi", func(r chi.Router) {
 r.Use(redocMiddleware)
 r.Get("/spec.json", func(w http.ResponseWriter, r *http.Request) {
  http.ServeFile(w, r, *config.OpenAPIPath)
  })
}
```

This way the middleware will only be used for this particular path. Also identity is not really needed for OpenAPI docs, we want to advertise our API, make it as easy to learn as possible. That is the purpose of this PR and your change is a huge contribution. Let's not "hide" it behind some obscure custom HTTP header for no reason.

lzap (changes requested):
I don't like adding middleware for all API requests, let's only add it for the documentation requests. Perhaps by creating a custom path is a way, or whatever you find better.

cmd/spec/path.yaml
lzap: We never get to the v1 state, I am a big advocate of releasing 1.0.0 as soon as possible.

What is important is the contract which is currently: "no guarantees, this is a beta". And I think it is fair to say that. Maybe if OpenSPEC supports a "beta" flag let's set it.

amirfefer: I haven't found any beta flag in the spec, so the description workaround seems fine to me

internal/config/main_config.go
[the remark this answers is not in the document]
ezr-ondrej: we are addressing the rest in #60

internal/config/0_common_config.go
```diff
@@ -10,13 +10,15 @@ import (
...
 	viper.SetDefault("app.port", 8000)
+	viper.SetDefault("app.version", "v1")
 	viper.SetDefault("logging.level", 1)
 	viper.SetDefault("database.host", "localhost")
 	viper.SetDefault("database.port", 5432)
 	viper.SetDefault("database.loglevel", 1)
 	viper.SetDefault("cloudwatch.enabled", false)
 	viper.SetDefault("prometheus.port", 9000)
 	viper.SetDefault("prometheus.path", "/metrics")
+	viper.SetDefault("openapi.path", "./api/openapi.gen.json")
```
ezr-ondrej, line 21: We need to add this file to the container (copy to the container in the [Dockerfile](https://github.com/RHEnVision/provisioning-backend/blob/main/build/Dockerfile)), otherwise it's gona be a failure in deployments.

internal/routes/all_routes.go
```diff
@@ -7,16 +7,28 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	s "github.com/RHEnVision/provisioning-backend/internal/services"
 	"github.com/go-chi/chi/v5"
+	redoc "github.com/go-openapi/runtime/middleware"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
+func redocMiddleware(handler http.Handler) http.Handler {
```
ezr-ondrej, line 14: If lukas is ok with the implementation now, I guess it is fine, but I remember us talking about splitting this into another PR, or am I mistaken? :)

internal/routes/all_routes.go
```diff
@@ -36,6 +48,8 @@ func apiRouter() http.Handler {
 		})
 	})
 
+	r.Get("/openapi.json", s.ServeOpenAPISpec)
```
ezr-ondrej, line 51: this is still being gated by the identity enforcement, but if you do not see a good way of not gating it, I guess we can keep it for now.
But I do not like the position of this route. I think it should be first route, but for sure it looks weird if its in the middle, looks kinda random.

ezr-ondrej (commented):
Good progres, I do not have any comments on the docs anymore, so I hope it's not going to block us from getting this in asap, but I'd like adding the docs in different PR, but let see if everyone is ok with that as-is now.

internal/routes/all_routes.go
amirfefer: we have two routes for the spec file, one is gated by the identity (`<pathPrefx>/openapi.json`) for consoleDot usage, and the "upstream" use (`<root>/spec.json`) which is not gated by it.
The docs (redoc) endpoint is not gated by the identity either (`<root>/docs`)

amirfefer:
@lzap, could you please have another look?

build/Dockerfile
```diff
@@ -8,5 +8,6 @@ RUN make prep build strip
 FROM registry.access.redhat.com/ubi8/ubi-minimal
 COPY --from=build /build/pbapi /pbapi
 COPY --from=build /build/pbmigrate /pbmigrate
+COPY --from=build /build/api /api
```
lzap, line 11: Can you use embed package instead? We do the same for migrations already to avoid having to copy additional files with the binary, it's really easy to use specifically for exposing a web directory:

https://pkg.go.dev/embed

lzap (changes requested):
Other than that, its looking good.

build/Dockerfile
lzap: From the documentation:

_The //go:embed directive accepts multiple space-separated patterns for brevity, but it can also be repeated, to avoid very long lines when there are many patterns. The patterns are **interpreted relative to the package** directory containing the source file. The path separator is a forward slash, even on Windows systems. Patterns **may not contain ‘.’ or ‘..’** or empty path elements, nor may they begin or end with a slash. To match everything in the current directory, use ‘*’ instead of ‘.’. To allow for naming files with spaces in their names, patterns can be written as Go double-quoted or back-quoted string literals._

You need to have the Go source file in the same directory as the files you want to embed, that is the limitation. But that's doable, we can drop a file in the /api directory that's not big deal. Simply change this to:

```go
// api/embed.go

package api

import (
	"embed"
	"net/http"
)

//go:embed *
var embeddedSpecs embed.FS

func ServeOpenAPISpec(w http.ResponseWriter, r *http.Request) {
	http.FileServer(http.FS(embeddedSpecs))
}
```

Then use this in the main or all_routes.go file by simply importing the api via:

```go

import (
 "github.com/RHEnVision/provisioning-backend/api"
)

// api.ServeOpenAPISpec()
```

internal/routes/all_routes.go
```diff
@@ -4,47 +4,60 @@ import (
...
 func apiRouter() http.Handler {
 	r := chi.NewRouter()
-	r.Use(identity.EnforceIdentity)
-	r.Route("/sources", func(r chi.Router) {
-		r.Get("/", s.ListSources)
-		r.Route("/{ID}", func(r chi.Router) {
-			r.Get("/", s.GetSource)
+
+	r.Get("/openapi.json", api.ServeOpenAPISpec)
+	r.Group(func(r chi.Router) {
```
amirfefer, line 35: I created this group for excluding the `openapi` endpoint  from the identity middleware

cmd/spec/main.go
```diff
@@ -86,9 +86,9 @@ func main() {
 
 	jsonB, err := json.MarshalIndent(doc, "", "  ")
 	checkErr(err)
-	err = ioutil.WriteFile("./api/openapi.gen.json", jsonB, 0644) // #nosec G306
+	err = ioutil.WriteFile("./api/openapi.json", jsonB, 0644) // #nosec G306
```
amirfefer, line 89: the `.gen` suffix chamges the embed mechanism, the consumers seek for `openapi.json`, not `opemapi.gen.json`

lzap: I was hoping for some alias, maybe symlink would work. It's very practical to have all generated files marked this way, my IDE even warns me about the fact that the file was likely generated.

However, I am fine with this. But the problem is this change was not reflected in `Makefile` so several targets do not work anymore. Can you either fix the `Makefile` or possibly try symlink or some kind of HTTP alias to make it work too?

lzap (changes requested):
Looks good to me, the `Makefile` needs an update tho.

amirfefer:
`embed` doesn't support symlink files, but according to the documentation:
> For embedding a single file, a variable of type string or []byte is often best. The FS type enables embedding a tree of files, such as a directory of static web server content

Therefore embedding FS for just the `openapi.json` is redundant, we can just embed the json spec as `[]byte` and serve it as is  :)

lzap:
Found it, it's at http://localhost:8000/docs/ actually. Merging, thanks!

PR #60, opened 2022-06-22
Change List Sources to consider application type

adiabramovitch:

**Checklist**

### DIFF
--- a/internal/config/5_real_config.go
+++ b/internal/config/5_real_config.go
@@ -13,7 +13,7 @@ import (
 func init() {
 	if clowder.IsClowderEnabled() {
 		cfg := clowder.LoadedConfig
-
+		viper.Set("restEndpoints.sources.appId", 11)
 		viper.Set("featureFlags.environment", "production")
 		viper.Set("database.host", cfg.Database.Hostname)
 		viper.Set("database.port", cfg.Database.Port)
@@ -24,6 +24,7 @@ func init() {
 		viper.Set("prometheus.path", cfg.MetricsPath)
 	} else {
 		viper.SetDefault("featureFlags.environment", "developement")
+		viper.SetDefault("restEndpoints.sources.appId", 11)
 
 		viper.AddConfigPath("./configs")
 		viper.SetConfigName("defaults")
--- a/internal/config/5_test_config.go
+++ b/internal/config/5_test_config.go
@@ -7,4 +7,5 @@ import "github.com/spf13/viper"
 
 func init() {
 	viper.Set("featureFlags.environment", "test")
+	viper.Set("restEndpoints.sources.appId", 11)
 }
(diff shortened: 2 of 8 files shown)


internal/config/main_config.go
```diff
@@ -60,7 +60,8 @@ var config struct {
 			URL string
 		}
 		Sources struct {
-			URL string
+			URL   string
+			AppId string
```
ezr-ondrej, line 64: Should we define a static value for this in devel and test atleast?

ezr-ondrej (commented):
One nit about the config default, that would make it easier in devel and test, but overall :+1:

internal/config/main_config.go
adiabramovitch: I think it needs to be configurable for devel mode, because we are using sources locally

internal/config/5_real_config.go
```diff
@@ -13,7 +13,7 @@ import (
 func init() {
 	if clowder.IsClowderEnabled() {
 		cfg := clowder.LoadedConfig
-
+		viper.Set("sources.appId", 11)
```
ezr-ondrej, line 16: this doesn't work (my bad) sorry
```suggestion
		viper.Set("restEndpoints.sources.appId", 11)
```

internal/config/5_real_config.go
```diff
@@ -24,6 +24,7 @@ func init() {
 		viper.Set("prometheus.path", cfg.MetricsPath)
 	} else {
 		viper.SetDefault("featureFlags.environment", "developement")
+		viper.SetDefault("sources.appId", 11)
```
ezr-ondrej, line 27:
```suggestion
		viper.SetDefault("restEndpoints.sources.appId", 11)
```

internal/config/5_real_config.go
adiabramovitch: Thank you!

ezr-ondrej (approved):
Thanks @adiabramovitch ! :+1:

internal/config/5_test_config.go
```diff
@@ -7,4 +7,5 @@ import "github.com/spf13/viper"
 
 func init() {
 	viper.Set("featureFlags.environment", "test")
+	viper.Set("restEndpoints.sources.appId", 11)
```
lzap, line 10: Are we sure that the database ID is stable and same across environments? This does not look right.

adiabramovitch: It is the id of our application at sources service (on stage). It can be changed only when using devel env

ezr-ondrej: I'll ask Sources, but I'd love to avoid a request just for the appId.
But if it will need to be, so be it ^_^ We need the config anyway I believe

PR #61, opened 2022-06-23
Add fetch arn functionality 

adiabramovitch:
Depends on https://github.com/RHEnVision/provisioning-backend/pull/60


internal/services/sources_service.go
```diff
@@ -47,3 +49,12 @@ func GetSource(w http.ResponseWriter, r *http.Request) {
 		renderError(w, r, payloads.NewRenderError(r.Context(), "show source", err))
 	}
 }
+
+func fetchARN(ctx context.Context, client sources.SourcesIntegration, sourceId string) (string, error) {
+	resp, err := client.ListSourceAuthenticationsWithResponse(ctx, sourceId, &sources.ListSourceAuthenticationsParams{}, AddIdentityHeader)
```
ezr-ondrej, line 54: we will need to filter by our `application_type` here right? otherwise it would return all auths for this source (we are sure this source has our auth as well, but if the source is superkey source, it may have auths for multiple apps, right?

ezr-ondrej (commented):
Apart of the one issue in the superkey case, I'm very happy here ^_^

internal/services/sources_service.go
adiabramovitch: the resouce_id is not the appid of our application (provisioning). it is the application created for a source.

adiabramovitch: the resource_id is not equal to the appId (provisioning).

lzap: And @adiabramovitch updated me on the existence of : https://github.com/RedHatInsights/sources-api-go/blob/ce5b90a2c2cbf51612ff96016cb288c02a24bca4/dao/seeds/source_types.yml#L69

lzap:
So we had a long conversation with Adi about this, the wrap up is that we are fine with simply fetching all authentications for a single source (via `/sources/123456/authentications`) and to find out the ARN via simple iteration. A single source can only have few items so it's not a problem. We can probably create a TODO issue to figure out filtering, looks like this must be implemented in the Sources API.

adiabramovitch:
@lzap Should I test the function fetchARN here or test it in the last PR when listing instance_types?

internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,16 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
 
 	return newClient, nil
 }
+
+func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
+	log.Trace().Msg("Listing AWS EC2 instance types")
+	input := &ec2.DescribeInstanceTypesInput{
+		MaxResults: aws.Int32(100),
```
adiabramovitch, line 106: For reference: 
https://github.com/RHEnVision/provisioning-backend/pull/66/commits/83a5454dd691c07f26c66180c76637dd86aced2c#r907188833

internal/payloads/sources_payload.go
```diff
@@ -7,6 +7,9 @@ import (
 	"github.com/go-chi/render"
 )
 
+type SourceID struct {
+	SourceId string `json:"source_id"`
+}
```
lzap, line 12: This is not needed, the source_id should be part of the URL.

internal/services/instance_types_service.go
```diff
@@ -0,0 +1,77 @@
...
+	sourceId := &payloads.SourceID{}
+	b, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return "", fmt.Errorf("unable to read response body %w", err)
+	}
+	if err := json.Unmarshal(b, &sourceId); err != nil {
+		return "", fmt.Errorf("unable to unmarshel response body to source id %w", err)
+	}
+	return sourceId.SourceId, nil
+}
```
lzap, line 77: This is not needed, the source id should be part of the URL.

Also, the chi library provides automatic JSON parsing, see Pubkey.Create function as an example.

lzap (changes requested):
Ok first round of review, size of this PR is just about right, there was no need to split it up into three. It's just holding you back when switching branches and everything is even harder to review.

internal/routes/all_routes.go
```diff
@@ -48,5 +48,11 @@ func apiRouter() http.Handler {
 		})
 	})
 
+	r.Route("/instance_types", func(r chi.Router) {
+		// TODO: source_id is supposed to be saved in a global session
+		r.Post("/", s.ListInstanceTypes)
+
+	})
+
```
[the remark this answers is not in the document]
adiabramovitch: So this endpoint was built as a mock using the UI, we checked that passing a source from the UI works and in which way. 
this was a "check-up" which should b changed according to the correct implementation. 
I'll update the PR to match the generic requirements

internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,16 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
 
 	return newClient, nil
 }
```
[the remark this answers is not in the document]
adiabramovitch: I am not sure I understand. 
the config that we have in our app is the secret for our application (our AWS account). 
the secrets we are fetching from sources belong to the customers. 
why should I override those if the secret's purpose/value is different?

internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,16 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
...
+
+func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
+	log.Trace().Msg("Listing AWS EC2 instance types")
+	input := &ec2.DescribeInstanceTypesInput{
+		MaxResults: aws.Int32(100),
+	}
+	resp, err := c.ec2.DescribeInstanceTypes(c.context, input)
+	if err != nil {
+		return nil, fmt.Errorf("cannot list instance types: %w", err)
+	}
```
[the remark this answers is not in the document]
adiabramovitch: Is that enough? :) 

```golang	
if len(resp.InstanceTypes) == 100 {
     return nil, fmt.Errorf("cannot list more then 100 instance types")
}
```

internal/payloads/instance_types_payload.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"net/http"
+
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/go-chi/render"
+)
+
+type InstanceTypeResponse struct {
+	*types.InstanceTypeInfo
```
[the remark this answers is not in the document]
lzap: I don't know which one is it, but on the EC2 launch page it's the instance identifier. I think it's the InstanceType.

For the memory, cpu etc I think it's okay to ignore for now, or if you want just pick whatever you like. Ideally matching what's in the EC2 WebUI today to match customers expectations.

Btw I just checked and there's over 500 types in the current version of AWS SDK :) Iteration might be necessary pretty soon.

internal/clients/ec2/ec2_client.go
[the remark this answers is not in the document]
lzap: Yup, the linter will complain about this form but you will figure it out :)

internal/clients/ec2/ec2_client.go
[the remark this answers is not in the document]
lzap: Oh disregard, I keep forgetting we use our account with role change.

internal/routes/all_routes.go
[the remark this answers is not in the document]
lzap: No problem, let me know when this is ready for re-review.

internal/clients/sources/sources_interface.go
```diff
@@ -9,4 +9,5 @@ var GetSourcesClient func(ctx context.Context) (SourcesIntegration, error)
 type SourcesIntegration interface {
 	ShowSourceWithResponse(ctx context.Context, id ID, reqEditors ...RequestEditorFn) (*ShowSourceResponse, error)
 	ListApplicationTypeSourcesWithResponse(ctx context.Context, appId ID, params *ListApplicationTypeSourcesParams, reqEditors ...RequestEditorFn) (*ListApplicationTypeSourcesResponse, error)
+	ListSourceAuthenticationsWithResponse(ctx context.Context, id ID, params *ListSourceAuthenticationsParams, reqEditors ...RequestEditorFn) (*ListSourceAuthenticationsResponse, error)
 }
```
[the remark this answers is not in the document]
adiabramovitch: Should I remove it from the function signature?

internal/services/sources_service.go
adiabramovitch: I added the correct filtering after asking on sources channel

internal/payloads/instance_types_payload.go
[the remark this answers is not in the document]
adiabramovitch: For now, I kept only the identifier- the instance type.

internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,22 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
...
+	input := &ec2.DescribeInstanceTypesInput{
+		MaxResults: aws.Int32(100),
+	}
+
+	resp, err := c.ec2.DescribeInstanceTypes(c.context, input)
+	if err != nil {
+		return nil, fmt.Errorf("cannot list instance types: %w", err)
+	}
+
+	log.Info().Msgf("Number AWS EC2 instance types: %d", len(resp.InstanceTypes))
```
lzap, line 114: Should be probably debug, that is pretty detailed information.

internal/clients/ec2/ec2_errors.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"fmt"
+)
+
+var OperationNotPermit = fmt.Errorf("operation not permitted")
+
+func OperationNotAllowedError(op string) error {
+	return fmt.Errorf("OperationNotPermit %w : %s", OperationNotPermit, op)
+}
```
lzap, line 11: The name for this error imply that this was an operation that was not permitted. Looks like this is a feature that hasn't been implemented to me. But this is okay.

What I would like us to use is Err suffix for all error values, which is the standard practice in Go library. This would become `OperationNotPermittedErr`. On top of that, you must not use `fmt.Errorf` to create the error because then you will have the linter complaining about this. All it wants you to do is this:

```go
OperationNotPermittedErr := errors.New("operation not permitted")
```

Then simply return this value: `return nil, OperationnotPermittedErr` it is as easy as that. The caller simply can compare the error: `if err == OperationNotPermittedErr` too. Errors are values (Go proverb), I love that :-)

internal/clients/sources/sources_interface.go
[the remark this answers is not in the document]
lzap: It is up to you, from what I understand it is an implementation detail of the generated HTTP client library which can be automatically added to all requests so the API does not have to expose it at all.

configs/defaults.yaml
```diff
@@ -42,7 +42,7 @@
 #     url:
 #   sources:
 #     url:
-#     appId:
+#     appTypeId:
 
 # featureFlags:
 #   # development or production
```
lzap, line 48: Please add some documentation in this example file.

internal/services/instance_types_service.go
```diff
@@ -0,0 +1,58 @@
...
+
+	crd, err := stsClient.AssumeRole(arn)
+	if err != nil {
+		renderError(w, r, payloads.New3rdPartyClientError(r.Context(), "assume role sts", err))
+		return
+	}
+
+	newEC2Client, err := ec2Client.CreateEC2ClientFromConfig(crd)
+	if err != nil {
+		renderError(w, r, payloads.NewAWSError(r.Context(), "Cant create new ec2 client", err))
```
lzap, line 44: Error messages should not be caps. Also once below.

internal/services/sources_service.go
```diff
@@ -47,3 +49,39 @@ func GetSource(w http.ResponseWriter, r *http.Request) {
...
+
+func fetchARN(ctx context.Context, client sources.SourcesIntegration, sourceId string) (string, error) {
+	// Get all the authentications linked to a specific source
+	resp, err := client.ListSourceAuthenticationsWithResponse(ctx, sourceId, &sources.ListSourceAuthenticationsParams{}, AddIdentityHeader)
+	if err != nil {
+		return "", fmt.Errorf("cannot list source authentication: %w", err)
+	}
+
+	// Filter authentications to include only auth where resource_type == "Application"
+	// TODO: use sources client filtering after they solve their bug
```
lzap, line 61: Can we remove TODO from here and create an issue in our project that will reference their issue? From my experience, this TODO will likely stays forever if the solution is good enough :-) My IDE warns about TODO comments on every commit, that is the reason.

internal/services/sources_service.go
```diff
@@ -47,3 +49,39 @@ func GetSource(w http.ResponseWriter, r *http.Request) {
...
+	auths := *authentications
+	list := make([]sources.AuthenticationRead, 0, len(auths))
+	for _, auth := range auths {
+		if *auth.ResourceType == "Application" {
+			list = append(list, auth)
+		}
+	}
+	//TODO: assumption: each source has one authentication linked to
+	return list[0]
+}
```
lzap, line 87: Please write an unit test for this function.

internal/services/sources_service.go
```diff
@@ -47,3 +49,39 @@ func GetSource(w http.ResponseWriter, r *http.Request) {
...
+
+func filterSourceAuthentications(authentications *[]sources.AuthenticationRead) sources.AuthenticationRead {
+	auths := *authentications
+	list := make([]sources.AuthenticationRead, 0, len(auths))
+	for _, auth := range auths {
+		if *auth.ResourceType == "Application" {
+			list = append(list, auth)
+		}
+	}
+	//TODO: assumption: each source has one authentication linked to
```
lzap, line 85: A left over? This needs to be implemented.

This function must return error when it would return invalid data.

lzap (changes requested):
Looks great, there are some left over TODOs still.

internal/clients/ec2/ec2_errors.go
adiabramovitch: Do you mean that there is no need for that function? 
I created it because of the linter complaints, that way there are no warnings from the linter. 

do you mean to change it to: 
```golang
OperationNotPermittedErr := errors.New("operation not permitted")
	if len(resp.InstanceTypes) == 100 {
		return nil, OperationNotPermittedErr
	}
```

because that way I still get a complaint from the linter: 
```golang
do not define dynamic errors, use wrapped static errors instead: "errors.New(\"operation not permitted\")"
```

internal/services/sources_service.go
adiabramovitch: Is that enough? :)
```golang
// Assumption: each source has one authentication linked to it
	if len(list) > 1 {
		return sources.AuthenticationRead{}, fmt.Errorf("there is more then one authentication linked to this source %s", sourceId)
	}
	return list[0], nil
```

internal/clients/ec2/ec2_errors.go
adiabramovitch: So I am not sure when is it correct to use fmt.Errorf and when should I used errors.New?

lzap: So errors.New is when you want to create new variable (value) that is the "source" error. In Go, error is an interface that implements Error function and that is all. It can be any value or type, but what is very typical in Go is that you simply use plain strings as error types. And this is what errors.New returns, implementation of error that is just a string. They are typically all lowercase (no sentence) in the Go standard library. An example:

https://cs.opensource.google/go/go/+/master:src/internal/oserror/errors.go

In Go 1.13 a new Wrapper interface was introduced which allows to "wrap" errors into errors which is very useful. Typically, an API should always return its own errors wrapping underlying errors like "file not found". The concept is really easy, all you need to do is to implement Unwrap function for your own error. Similarly like errors.New which returns a simple string, there is a slightly more complex wrapped string which can be created with several ways, one of it is fmt.Errorf which returns such implementation.

To know more how to actually work with wrapped errors I recommend: https://go.dev/blog/go1.13-errors

adiabramovitch: Thanks!

internal/services/sources_service_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/clients/sources/stubs"
 	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
 	"github.com/stretchr/testify/assert"
+	"k8s.io/utils/pointer"
```
lzap, line 15: This is literally one-liner and we would pull the largest Go project in the world just because of this line:

https://github.com/kubernetes/utils/blob/3a6ce19ff2f9/pointer/pointer.go#L159

Can you simply copy and paste the function into our own package instead? Please avoid utils package, name it simply "pointer".

lzap (changes requested):
Looks good, the only thing is to avoid pulling in such a huge dependency because of a single line (kubernetes).

internal/services/sources_service_test.go
adiabramovitch: Thank you for noticing! :)

lzap (approved):
Thank you, merging.

internal/services/sources_service.go
```diff
@@ -47,3 +49,44 @@ func GetSource(w http.ResponseWriter, r *http.Request) {
...
+	auths := *authentications
+	list := make([]sources.AuthenticationRead, 0, len(auths))
+	for _, auth := range auths {
+		if *auth.ResourceType == "Application" {
+			list = append(list, auth)
+		}
+	}
+	// Assumption: each source has one authentication linked to it
+	if len(list) > 1 {
+		return sources.AuthenticationRead{}, sources.MoreThenOneAuthenticationForSourceErr
```
ezr-ondrej, line 89: Is this correct in case of superkey Source? The superkey source can have multiple authentications (for multiple apps) IMHO.

adiabramovitch: Good point, I think It will be better to get all the auths and search for the one with the correct resource Id. 
I'll fix it tomorrow.

PR #62, opened 2022-06-23
Add list instance types to EC2 client

adiabramovitch:


internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,17 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
 
 	return newClient, nil
 }
+
+func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
+	log.Trace().Msgf("Listing AWS EC2 instance types")
```
lzap, line 104: Use `Msg` when not using formatting:

```suggestion
	log.Trace().Msg("Listing AWS EC2 instance types")
```

internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,17 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
...
+	input := &ec2.DescribeInstanceTypesInput{
+		//TODO: Which filters do we want to support?
```
lzap, line 106: No filters for this one I assume.

internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,17 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
...
+		// What is The maximum number of results to return for the request in a single page?
```
lzap, line 107: Now, AWS API limits responses to 100 items. This is not meant for pagination but for data throttling. Since we will always need all types, we want to implement iteration. I would like this to be implemented even if there are not more than 100 types at the moment because they can add more later and then we have a bug.

We can test this feature by setting the MaxResults to let's say 20 to test it all correctly correct the results.

I think it would be good idea to cache this list at some point, not at this stage tho.

adiabramovitch: Not sure I totally understand, do you want me to implement getting all the instance types (using the token) and after receiving all the info listing it?

PR #63, opened 2022-06-23
add error types to spec 

amirfefer:
This PR refactors the errors payload and transforms it into an interface for a simpler consuming and extending,
this also allows adding different error types into openapi spec.

We can leverage this interface instead of creating a new error payload for every new service, and keep our code dry

### DIFF
--- a/api/openapi.gen.yaml
+++ b/api/openapi.gen.yaml
@@ -31,13 +31,13 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/v1.ResponseError"
+                $ref: "#/components/schemas/v1.NotFoundError"
         "500":
           description: 'Returned when an internal error occurs.'
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/v1.ResponseError'
+                $ref: '#/components/schemas/v1.InternalRequestError'
   /pubkeys:
     get:
       operationId: getPubkeyList
@@ -51,6 +51,12 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/v1.Pubkey'
+        "500":
+          description: 'Returned when an internal error occurs.'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/v1.InternalRequestError'
 components:
   schemas:
     v1.Account:
@@ -61,6 +67,27 @@ components:
           format: int64
           type: integer
         org_id:
+          type: string
+      type: object
+    v1.BadRequestError:
+      properties:
+        msg:
+          type: string
+        request_id:
+          type: string
+      type: object
+    v1.InternalRequestError:
+      properties:
+        msg:
+          type: string
+        request_id:
+          type: string
+      type: object
+    v1.NotFoundError:
+      properties:
+        msg:
+          type: string
+        request_id:
           type: string
       type: object
     v1.Pubkey:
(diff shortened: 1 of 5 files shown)


internal/payloads/error_payload.go
```diff
@@ -10,6 +10,23 @@ import (
 	"github.com/go-chi/render"
 )
 
+type APIError interface {
+	Error() string
+	GetStatus() int
+	SetMessage(string)
+	Unwrap() error
+	Render(_ http.ResponseWriter, r *http.Request) error
+}
```
lzap, line 19: Does this not expose RequestId intentionally?

api/openapi.gen.yaml
```diff
@@ -63,6 +69,27 @@ components:
...
+        request_id:
+          type: string
+      type: object
+    v1.NotFoundError:
+      properties:
+        msg:
+          type: string
+        request_id:
+          type: string
+      type: object
```
lzap, line 92: I don't see the use of the interface in the schema, I would expect that there will be some kind of common component that matches the interface, but here it uses structs directly. Can you explain in more detail what was the motivation?

Because RequestId is not even available through the interface, yet it is mapped here directly from the type.

lzap (changes requested):
Looks good, got some comments for the interface itself.

But what I am missing is where the interface is getting used. I don't see any use of it neither in the codebase nor in the schema. Can you explain? Maybe this is something you want to leverage in a followup PR?

amirfefer:
> But what I am missing is where the interface is getting used. I don't see any use of it neither in the codebase nor in the schema. Can you explain? Maybe this is something you want to leverage in a followup PR?

@lzap, Thanks for the review!
yes. I'd like to keep PRs clear and small,  the motivation here is to consume/extend errors as oop paradigm, instead of creating more and more similar payloads (which differ only by the message and status), we have just started and we already have more than 200 hundred repetitive lines of code here

lzap:
From all I can see, the interface is not used by the schema generator, it does appear to use the types directly. I don't understand, can you clarify?

> the motivation here is to consume/extend errors as oop paradigm, instead of creating more and more similar payloads (which differ only by the message and status), we have just started and we already have more than 200 hundred repetitive lines of code here

I am all for improving, but I don't understand how this is supposed to help. Your patch adds 120 lines and removes 81 lines, effectively adding 39 lines in the `internal/payloads/error_payload.go` file and I would like just to see what is the purpose of the interface. I just like you to defend why this is necessary. Which OOP paradigm is on your mind?

amirfefer:
I was thinking to seprate between the error type and the error payload, which an interface can make it simpler. i.e:
```go
result, err := service.someAction()
	if err != nil {
		var apiError errors.APIError
		switch err.(type) {
		case *service.XNotFoundError:
			apiError = payload.errors.NewNotFound("Could not find X", err)
		case *service.YNotFoundError:
			apiError =  payload.errors.NewNotFound("Could not find Y", err)
                case *service.AWSInternalError:
                        apiError = payload.errors.NewInternalError("Failed to coonect AWS", err)
		default:
			apiError = errors.NewInternalServerError()
		}
    renderError(w, r, apiError)
```

Anyway, since the PR's scope is basically to add types to our openapi, and following to our offline discussion, let's remove the interface from this PR, and we can discuss later on about improving error payloads rendering.

amirfefer:
Let's drop the interface, for now, it's not related to the spec file :)
I create #71  for adding error responses for the spec file for easy manual editing

PR #64, opened 2022-06-26
Add sources to spec file

amirfefer:
This PR adds sources type and paths to the spec file
blocked by #63

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -77,6 +77,58 @@
           }
         },
         "type": "object"
+      },
+      "v1.Source": {
+        "properties": {
+          "app_creation_workflow": {
+            "type": "string"
+          },
+          "availability_status": {
+            "type": "string"
+          },
+          "created_at": {
+            "format": "date-time",
+            "type": "string"
+          },
+          "id": {
+            "type": "string"
+          },
+          "imported": {
+            "type": "string"
+          },
+          "last_available_at": {
+            "format": "date-time",
+            "type": "string"
+          },
+          "last_checked_at": {
+            "format": "date-time",
+            "type": "string"
+          },
+          "name": {
+            "type": "string"
+          },
+          "paused_at": {
+            "format": "date-time",
+            "type": "string"
+          },
+          "source_ref": {
+            "type": "string"
+          },
+          "source_type_id": {
+            "type": "string"
+          },
+          "uid": {
+            "type": "string"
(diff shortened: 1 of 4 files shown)


internal/payloads/error_payload.go
```diff
@@ -10,6 +10,23 @@ import (
 	"github.com/go-chi/render"
 )
 
+type APIError interface {
+	Error() string
+	GetStatus() int
```
lzap, line 15: The recommended way for getter is just `Status()` but it's okay. We already have few in the codebase.

internal/payloads/error_payload.go
```diff
@@ -28,155 +45,177 @@ type ResponseError struct {
 	Context context.Context `json:"-"`
 }
 
-func (e *ResponseError) Render(_ http.ResponseWriter, r *http.Request) error {
-	render.Status(r, e.HTTPStatusCode)
-	return nil
+type InternalServerError struct {
+	ResponseError
 }
```
lzap, line 50: Can you use simply type alias here? Will the generator accept that? That feels much more simple than embedding:

```go
type InternalServerError ResponseError
```

All the return statements will be shorter than.

internal/payloads/error_payload.go
```diff
@@ -28,155 +45,177 @@ type ResponseError struct {
...
+			Err:            err,
+			Context:        ctx,
+		},
 	}
 }
 
-func NewURLParsingError(ctx context.Context, paramName string, err error) *ResponseError {
+type BadRequestError struct {
+	ResponseError
+}
```
lzap, line 173: Ditto.

internal/payloads/error_payload.go
```diff
@@ -28,155 +45,177 @@ type ResponseError struct {
...
+			Err:            err,
+			Context:        ctx,
+		},
 	}
 }
 
-func NewNotFoundError(ctx context.Context, err error) *ResponseError {
+type NotFoundError struct {
+	ResponseError
+}
```
lzap, line 105: Ditto.

lzap (changes requested):
Almost there, I think we can use type aliases for the custom errors instead of embedding for cleaner code.

lzap (approved):
So as I stated above, it's not okay to expose internal data structures from 3rd parties, however, this is not purpose of this patch. We can fix this later.

PR #66, opened 2022-06-27
Add list instance types to EC2 client and add region support to clients

adiabramovitch:
depends on #61

### DIFF
--- a/internal/clients/ec2/ec2_client.go
+++ b/internal/clients/ec2/ec2_client.go
@@ -83,7 +83,7 @@ func (c *Client) DeleteSSHKey(handle string) error {
 }
 
 func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client, error) {
-	newCfg, err := cfg.LoadDefaultConfig(c.context, cfg.WithCredentialsProvider(
+	newCfg, err := cfg.LoadDefaultConfig(c.context, cfg.WithRegion(config.AWS.Region), cfg.WithCredentialsProvider(
 		credentials.NewStaticCredentialsProvider(*crd.AccessKeyId, *crd.SecretAccessKey, *crd.SessionToken),
 	))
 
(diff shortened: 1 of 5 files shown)


internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,16 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
 
 	return newClient, nil
 }
+
+func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
+	log.Trace().Msg("Listing AWS EC2 instance types")
+	input := &ec2.DescribeInstanceTypesInput{
+		// What is The maximum number of results to return for the request in a single page?
+	}
```
lzap, line 107: The AWS API is designed to return between 0 - 100 items, the API returns items in random order and sorting or limit/offset is not possible. At some point, perhaps not in the initial version, we will hit the 100 limit in the API and we would need to:

* fetch results in multiple calls
* cache the result on the backend
* implement pagination in our API/UI

For the record, here is my community question and answers: https://repost.aws/questions/QUD4vDlkcRTBWVEg2gyC5_7w/sort-results-from-ec-2-api

We can probably skip this for types and hardcode an error if this returns more than 90 items (to be able to fix this later). Or if you want to go deep into this, let's implement multicalling now (unsure how to name this, its not pagination).

lzap: Iteration is perhaps a good name for this, pagination for dealing with pages.

internal/clients/ec2/ec2_client.go
```diff
@@ -99,3 +99,16 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
...
+
+func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
+	log.Trace().Msg("Listing AWS EC2 instance types")
+	input := &ec2.DescribeInstanceTypesInput{
+		MaxResults: aws.Int32(100),
+	}
+	resp, err := c.ec2.DescribeInstanceTypes(c.context, input)
+	if err != nil {
+		return nil, fmt.Errorf("cannot list instance types: %w", err)
+	}
```
adiabramovitch, line 111: So I am not sure if I am going to even receive an error if there are more than 100 records.

lzap: My point was to actually return (own) error when the list is `len(n) == 100`

lzap:
For the record, this PR is going to be squashed into a single PR.

PR #67, opened 2022-06-27
Add instance types service and payload

adiabramovitch:
depends on #66, #61


lzap:
For the record, this PR is going to be squashed into a single PR.

PR #71, opened 2022-06-28
Add error responses to the spec file

amirfefer:
This adds errors `responses` to the spec file under `components` for easy referencing from a route.
For motivation see [Reusing responses](https://swagger.io/docs/specification/describing-responses/)


lzap (approved):
Perfect! Thank you.

PR #72, opened 2022-06-29
Add sources under dependencies in clowdapp.yaml

akhil-jha:
Adding sources as dependency so it gets deployed when provisioning-backend gets deployed.


mshriver (approved):
Deployment against this includes all the sources-api components and deploys them successfully.

lzap (approved):
Thanks!

PR #74, opened 2022-06-30
Add list instance types to spec

adiabramovitch:
Depends on #61

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -48,6 +48,14 @@
         },
         "type": "object"
       },
+      "v1.InstanceType": {
+        "properties": {
+          "id": {
+            "type": "string"
+          }
+        },
+        "type": "object"
+      },
       "v1.Pubkey": {
         "properties": {
           "account_id": {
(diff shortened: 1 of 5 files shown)


api/openapi.gen.json
```diff
@@ -48,6 +48,7 @@
         },
         "type": "object"
       },
+      "v1.InstanceType": {},
```
lzap, line 51: The generated type (component) is empty, this is because you need to add Go tags to the type in order to have the field exported:

https://github.com/adiabramovitch/provisioning-backend/blob/cf19486b4b50f3e19234dc8d0c0b53167a62d566/internal/payloads/instance_types_payload.go#L11

api/openapi.gen.yaml
```diff
@@ -83,6 +83,29 @@ paths:
...
           $ref: "#/components/responses/InternalError"
+  /instance_types/{source_id}:
+      get:
+        description: 'Return a list of instance types'
+        operationId: getListInstanceTypes
+        parameters:
+          - name: source_id
+            in: path
+            required: true
+            description: 'Database ID of a source.'
```
lzap, line 94: This is the source database ID, is it? Can you make the description more clear that this is actually ID from the sources database, not ours?

lzap (changes requested):
Looks good, other than the type is empty.

lzap (approved):
RTM TY

amirfefer (approved):
Thanks @adiabramovitch ! 
spec looks 👍

lzap (approved):
Thanks merging.

PR #76, opened 2022-07-08
HMSPROV-125: Add pubkeys to the OpenAPI

lzap:
SSIA

https://issues.redhat.com/browse/HMSPROV-125


docs/openapi.md
```diff
@@ -41,6 +41,13 @@ paths:
                   $ref: '#/components/schemas/v1.<YOUR_PAYLOAD>' # a reference to the registered type 
 ```
 
+Operation naming convention:
+
+* GET /resource - getResourceList
+* POST /resource - addResource
```
ezr-ondrej, line 47: Just a NIT, I do not care much, just curious: I see others using `create` instead of `add` is there a reason to use add ?

lzap: No other reason then that I saw that in an OpenAPI example here: https://swagger.io/docs/specification/paths-and-operations/ and we already used add too. I can rename if you want, this is basically an internal identifier which is used by client generators at most.

lzap:
Rebased.

I added one extra small commit, a helper function was created in a separate package for no reason. I would like us to avoid "utils" or "helpers" package, this just moves it into the routes package where it belongs.

docs/openapi.md
lzap: I renamed this to create*. I have no strong feelings about this.

adiabramovitch (approved):

PR #77, opened 2022-07-11
Add status code for sources client error

adiabramovitch:


adiabramovitch:
Would like your opinion, we have two other functions we use from sources client, but we are handling the errors differently from the patch I added. 
What do you think is the best approach? 
should I change the fmt.Errorf to a struct that holds the status code like `ResponseError`?
@lzap , @ezr-ondrej

internal/payloads/error_payload.go
```diff
@@ -55,6 +55,20 @@ func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
 	}
 }
 
+func SourcesClientError(ctx context.Context, message string, err error, status int) *ResponseError {
+	msg := fmt.Sprintf("3rd Party Client error: %s: %v", message, err)
```
lzap, line 59:
```suggestion
	msg := fmt.Sprintf("Sources error: %s: %v", message, err)
```

lzap:
> What do you think is the best approach?

This is ok.

> should I change the fmt.Errorf to a struct that holds the status code like ResponseError?

Not sure what you mean exactly, but your patch looks clean. The user typically wants to see: HTTP error, message, request id. That's all.

adiabramovitch:
we have these two functions as well: 
https://github.com/RHEnVision/provisioning-backend/blob/9e64452d3dd2485e4838ae7ff9eb38ed9e890273/internal/services/sources_service.go#L55
https://github.com/RHEnVision/provisioning-backend/blob/9e64452d3dd2485e4838ae7ff9eb38ed9e890273/internal/services/sources_service.go#L67
ow should I refer to the HTTP status code using a different error format? we are not rendering here any payload.

lzap:
In Go, you should solve this by declaring an error value in the services package and then using that:

```
var SourceNotFoundErr = errors.new("source not found in sources app")

...
if err != nil {
  return nil, SourceNotFoundErr
}
```

Then in the caller, you need to actually compare the error and create the proper return JSON structure depending on if the error is the one that represents 404 or in case of all other errors just return the generic one.

adiabramovitch:
```golang
	resp, err := client.ListSourceAuthenticationsWithResponse(ctx, sourceId, &sources.ListSourceAuthenticationsParams{}, AddIdentityHeader)
	if err != nil {
		if resp.StatusCode() == 404 {
			return "", sources.AuthenticationForSourcesNotFoundErr
		}
	}
```

Is that what you meant? which status codes do we want to refer to?

lzap:
> Is that what you meant? which status codes do we want to refer to?

Yes, but do not enclose it into the if err statement, error is only returned when no value is returned I think. I would assume that for 404 or even 500 the err is actually `nil`. But test this, I might be wrong.

Then in the caller (`ListInstanceTypes`) simply compare the err against SourceNotFoundErr and render appropriate content.

lzap:
Just a more nicely way of what I am trying to say: https://pliutau.com/handle-http-request-errors-in-go/ :)

adiabramovitch:
> > Is that what you meant? which status codes do we want to refer to?
> 
> Yes, but do not enclose it into the if err statement, error is only returned when no value is returned I think. I would assume that for 404 or even 500 the err is actually `nil`. But test this, I might be wrong.
> 
> Then in the caller (`ListInstanceTypes`) simply compare the err against SourceNotFoundErr and render appropriate content.

Yes you are right I saw that for 500 the error is nil

lzap:
Yeah Go standard library actually describes that behavior, I was just lazy reading it all:

_An error is returned if the Client’s CheckRedirect function fails or if there was an HTTP protocol error. A non-2xx response doesn’t cause an error._

lzap:
This actually brings an interesting point, for all client requests, we should:

* check for protocol errors (error is not nit)
* check for 404 (to return a custom error)
* check for non-200 (should be a generic client error too)

adiabramovitch:
> 



> This actually brings an interesting point, for all client requests, we should:
> 
> * check for protocol errors (error is not nil)
> * check for 404 (to return a custom error)
> * check for non-200 (should be a generic client error too)

Yes, for image builder we will need to implement it when we will start to use it. 
I'll add those errors to sources service

internal/services/sources_service.go
```diff
@@ -42,7 +49,15 @@ func GetSource(w http.ResponseWriter, r *http.Request) {
...
-		renderError(w, r, payloads.New3rdPartyClientError(r.Context(), "show source", err))
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "show source", err, resp.StatusCode()))
+		return
+	}
+	if resp.StatusCode() == 404 {
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "source not found", err, resp.StatusCode()))
+		return
+	}
+	if resp.StatusCode() < 200 || resp.StatusCode() > 299 {
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "sources client error", err, resp.StatusCode()))
```
adiabramovitch, line 60: I don't like it, I think it makes the code less readable, do you have a different suggestion?

lzap: You can ether check the first character of Status or compare the StatusCode just like that. I would probably prefer >= 300 which is a bit cleaner than 299. Just go ahead and create a helper function in services/status_check.go something like `IsStatus2xx(code int)` and use that :)

Just do not introduce any `utils` or `helpers` package please, we already had one sneaked in in routes/helpers and I hate it already :)

internal/services/sources_service.go
```diff
@@ -68,6 +84,7 @@ func fetchARN(ctx context.Context, client sources.SourcesIntegration, sourceId s
 	if err != nil {
 		return "", fmt.Errorf("cannot list source authentication: %w", err)
 	}
+	_ = sources.HandleSourcesErrors(resp.StatusCode())
```
lzap, line 87: Oh this does not work, the `return` statement in `HandleSourcesErrors` returns from that function, but not from the call site (here). That's why "error handling functions" are generally a bad idea because here you actually need to write:

```
err := sources.HandleSourcesErrors(resp.StatusCode())
if err != nil {
  return err
}
```

So basically the function is 7 lines long and to call it you need to use 4 lines. When you write it directly (inline) that is 6 lines, so you save 2 lines. I do not think this is worth it.

internal/services/sources_service.go
lzap: I see the same pattern you use both in services and client packages, therefore I suggest to create a brand new package "parsing". I think we could also move the function ParseInt64 there as it is similar problem kind of.

internal/services/instance_types_service.go
```diff
@@ -23,6 +24,10 @@ func ListInstanceTypes(w http.ResponseWriter, r *http.Request) {
 
 	arn, err := fetchARN(r.Context(), sourcesClient, sourceId)
 	if err != nil {
+		if errors.Is(err, sources.ApplicationNotFoundErr) || errors.Is(err, sources.AuthenticationForSourcesNotFoundErr) {
+			renderError(w, r, payloads.SourcesClientError(r.Context(), "can't fetch arn from sources", err, 404))
+			return
+		}
 		renderError(w, r, payloads.New3rdPartyClientError(r.Context(), "can't fetch arn from sources", err))
```
lzap, line 31: I actually suggest to get rid of New3rdPartyClientError alltogether and only keep SourcesClientError with 404 or 500.

internal/services/sources_service.go
```diff
@@ -20,10 +20,17 @@ func ListSources(w http.ResponseWriter, r *http.Request) {
 	}
 	resp, err := client.ListApplicationTypeSourcesWithResponse(r.Context(), config.Sources.AppTypeId, &sources.ListApplicationTypeSourcesParams{}, AddIdentityHeader)
 	if err != nil {
-		renderError(w, r, payloads.New3rdPartyClientError(r.Context(), "list sources", err))
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "list sources", err, resp.StatusCode()))
+		return
+	}
+	if resp.StatusCode() == 404 {
```
lzap, line 26: This could be other helper function like `IsHTTPNotFound` or something like that.

internal/services/instance_types_service.go
adiabramovitch: What about other clients like ec2/sts?

internal/clients/sources/sources_errors.go
```diff
@@ -5,3 +5,16 @@ import (
...
+
+func HandleSourcesErrors(statusCode int) error {
+	if statusCode == 404 {
+		return AuthenticationForSourcesNotFoundErr
+	}
+	if statusCode < 200 || statusCode > 299 {
+		return SourcesClientErr
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
adiabramovitch: I will remove the function but I still need the errors I created in sources_errors.go.  right?

internal/services/instance_types_service.go
lzap: Yeah I thought to have separate errors per each service.

internal/services/sources_service.go
```diff
@@ -19,11 +20,19 @@ func ListSources(w http.ResponseWriter, r *http.Request) {
...
-		renderError(w, r, payloads.New3rdPartyClientError(r.Context(), "list sources", err))
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "list sources", err, statusCode))
+		return
+	}
+	if parsing.IsHTTPNotFound(statusCode) {
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "sources not found", err, statusCode))
+		return
+	}
+	if !parsing.IsHTTPStatus2xx(statusCode) {
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "sources client error", err, statusCode))
```
adiabramovitch, line 33: Wodering if this is a problem, when running the test the status received is 0 which is indeed not a 2xx status code, so the operation resulted in that error.
```golang
 	handler.ServeHTTP(rr, req) // the error is received here which causes the test to fail
	if status := rr.Code; status != http.StatusOK {
		t.Errorf("Handler returned wrong status code. Expected: %d. Got: %d.", http.StatusOK, status)
	}
```

adiabramovitch: here @ezr-ondrej

ezr-ondrej: can we change the test to return 200 status?

internal/services/instance_types_service.go
```diff
@@ -23,7 +24,11 @@ func ListInstanceTypes(w http.ResponseWriter, r *http.Request) {
 
 	arn, err := fetchARN(r.Context(), sourcesClient, sourceId)
 	if err != nil {
-		renderError(w, r, payloads.New3rdPartyClientError(r.Context(), "can't fetch arn from sources", err))
+		if errors.Is(err, sources.ApplicationNotFoundErr) || errors.Is(err, sources.AuthenticationForSourcesNotFoundErr) {
+			renderError(w, r, payloads.SourcesClientError(r.Context(), "can't fetch arn from sources", err, 404))
```
adiabramovitch, line 28: would split is in to two if's to have more accurate msg

internal/services/sources_service.go
ezr-ondrej: I didn't test, but this IMHO fixes the issue: https://gist.github.com/ezr-ondrej/a5d6d8e5da3525766bc9e6bc22c51c59

ezr-ondrej (approved):
Yup, we can later on refactor the stubs to be bit more DRY, as those responses are mostly the same, but that is fine for now :+1:
I think this is quite an improvement :+1:

ezr-ondrej:
Looks great! :+1: Thanks @adiabramovitch and @lzap for the review! :)

lzap:
Just for the record, I did not review this PR. It was in Draft mode so I created various "draft comments" because it does not make much sense to do full review for drafts. Then it was turned to Ready for Review and only you reviewed it and merged it moments after.

I'd propose to keep PRs in Ready for Review state at least 24 hours so we can all see the final code, unless it's something you really need to get in. We've already spent some time together chatting about this, I was literally looking forward seeing the final solution today ;-)

Don't get me wrong, the code is perfect, I just want to comment on the workflow.

ezr-ondrej:
Sorry, I've seen all your comments addressed so I've merged, but yeah agreed, leaving it in ready to merge for some time is a good practice...

PR #78, opened 2022-07-11
Get sources service URL

ezr-ondrej:
In Clowder environments, we need to figure out where are our 
dependencies located and fetch the URL from Clowder config

### DIFF
--- a/internal/config/5_real_config.go
+++ b/internal/config/5_real_config.go
@@ -13,7 +13,6 @@ import (
 func init() {
 	if clowder.IsClowderEnabled() {
 		cfg := clowder.LoadedConfig
-		viper.Set("restEndpoints.sources.appTypeId", 11)
 		viper.Set("featureFlags.environment", "production")
 		viper.Set("database.host", cfg.Database.Hostname)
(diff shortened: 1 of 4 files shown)


internal/config/5_real_config.go
```diff
@@ -13,7 +13,6 @@ import (
 func init() {
 	if clowder.IsClowderEnabled() {
 		cfg := clowder.LoadedConfig
-		viper.Set("restEndpoints.sources.appTypeId", 11)
```
lzap, line 16: We have the same below, and also in the `5_test_config.go`. Just saying, I am not entirely sure about this, so for non-clowder environment we say the type id is 11? That is still weird.

lzap:
The patch makes sense. Leaving the rest on @adiabramovitch

adiabramovitch:
LGTM 😊

adiabramovitch (approved):

lzap:
Just for the record, I merged, my concern about the number 11 for test/non-clowder environment is a not big deal. We need some default value anyway, might be perhaps an empty value with an error requiring you to set something, that would be more clean I think but it's not purpose of this PR to change this.

PR #79, opened 2022-07-11
Split reservation into detail and noop type

lzap:
Implements new provider type: no operation (noop). This is a special type that is used for reservations which actually do nothing and immediately finish background job. This can be really helpful for testing to verify our job backend operates normally.

For the record, this is the current list of providers:

* 0 - reserved
* 1 - noop
* 2 - AWS
* 3 - Azure (not implemented)
* 4 - GCE (not implemented)

To use no operation provider, simply create a new empty reservation, it requires no arguments:

```http
POST http://{{hostname}}:{{port}}/{{prefix}}/reservations/noop HTTP/1.1
X-Rh-Identity: {{identity}}
```

Store the reservation ID from the reply and note the current status "Created" and that both finished_at and success are both NULL values (Valid = false).

```json
{
  "id": 2,
  "provider": 1,
  "created_at": "2022-07-19T14:16:26.064033Z",
  "status": "Created",
  "finished_at": {
    "Time": "0001-01-01T00:00:00Z",
    "Valid": false
  },
  "success": {
    "Bool": false,
    "Valid": false
  }
}
```

Here is a room for improvement, provider is returned as integer not as string, also finished_at/success are nullable fields and this is how sql.Null* types are marshaled. We can improve this later.

The job backend will start processing this reservation shortly, wait few seconds and then make GET /reservations request to list all reservations and filter out the reservation with the ID (2 in the example above). Repeat until success or finished_at has Valid set to true. The status should be set to "Finished", this concludes a success test case. When the job backend is not operating correctly, the reservation gets stuck forever in the "Created" state. Inspect why the workers are not picking up the job.

The patch also splits the AWS details (pubkey id for now, will be expanded later) into a separate one-to-one associated table and model. This allows us to create different payloads for different cloud providers easily. The current AWS implementation creates a pubkey upload (AWS key-pair import) job. To call it simply:

```http
POST http://{{hostname}}:{{port}}/{{prefix}}/reservations/aws HTTP/1.1
X-Rh-Identity: {{identity}}

{
  "pubkey": {
    "existing_id": 1
  }
}
```

The reservation endpoint is currently not in the OpenAPI as it is changing (nullable fields, new fields for AWS).

Next steps:

* Add more AWS fields to the payload, model and database.
* Create second job that will launch the instance (called after the pubkey upload job). You can utilize job dependencies from dejq library to create a dependant job that will be only executed if the pubkey was successfully uploaded which can be checked by lookup up for PubkeyResrouce (must have valid handle).
* Utilize account id from the context for all database queries (specifically pubkey handling).
* Improve the code and find bugs.


internal/db/migrations/005_reservations.sql
```diff
@@ -1,8 +1,9 @@
 CREATE TABLE reservations
 (
   id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  provider INTEGER NOT NULL CHECK (valid_provider(provider)),
```
ezr-ondrej, line 4: can you also add source_id? I believe we will need it, so once we adjust the migration already...

internal/services/reservations_service.go
```diff
@@ -44,57 +45,64 @@ func CreateReservation(w http.ResponseWriter, r *http.Request) {
...
-		// TODO: IMPORTANT! Must perform GetByIdAndAccount to prevent pubkey hijack
-		logger.Debug().Msgf("Validating existing pubkey %d", *payload.Pubkey.ExistingID)
-		pk, err = pkDao.GetById(r.Context(), *payload.Pubkey.ExistingID)
-		if err != nil {
-			var e *dao.NoRowsError
-			if errors.As(err, &e) {
-				renderError(w, r, payloads.NewNotFoundError(r.Context(), err))
-			} else {
-				renderError(w, r, payloads.NewDAOError(r.Context(), "get pubkey by id", err))
+	if payload.Provider != models.ProviderTypeNoop {
```
ezr-ondrej, line 72: I get it would be bit overkill to add another table for noop, but given the service code is quite different anyway, wouldn't it be better to have special enpoint for noop?

ezr-ondrej (commented):
Looks reasonable, but do you think separating the noop endpoint from the production one would make sense?

lzap:
After discussion we concluded with:

1. Creation should be done in separate endpoints per provider type (/resevration/aws) leading to separate handler functions
2. Listing, however, will list all types
3. Reservation model should only carry minimum amount of data and details should be in polymorfic assosiated table (e.g. ReservationDetailAWS). PubkeyID should be moved to the detail table.
4. Noop type will have no detail table associated.
5. Create reservation endpoint request payload will have only the detail as all other fields are not needed (generated, from header, from URL)

lzap:
Rebased, updated description and renamed the PR to better capture the change.

ezr-ondrej (approved):
Overall I like it, conceptually it is well designed.
I'd just go ahead as is, it is main code path, so it will get tested quite thoroughly.
I'm missing tests, which makes me bit nervous, but we can work on that in follow-ups :)

adiabramovitch:
So I think few changes are needed here after #98 was merged :)

ezr-ondrej:
Thanks @lzap ! Great patch.

PR #80, opened 2022-07-11
Fetch Application Type from sources

ezr-ondrej:
Depends on #78

Fetch the sources ID on the first request (the first time the client is instantiated)


lzap:
I merged the dependency.

internal/clients/sources/sources_impl.go
```diff
@@ -2,12 +2,54 @@ package sources
...
-	return NewClientWithResponses(config.Sources.URL)
+	client, err := NewClientWithResponses(config.Sources.URL)
+	if err != nil {
+		return client, err
+	}
+	if config.Sources.AppTypeId == "" {
+		ctxval.GetLogger(ctx).Info().Msg("Fetching the Application Type ID of Provisioning for Sources")
+		resp, err := client.ListApplicationTypes(ctx, &ListApplicationTypesParams{})
+		if err != nil {
+			ctxval.GetLogger(ctx).Warn().Msgf("Failed to fetch ApplicationTypes from sources: %w", err)
```
adiabramovitch, line 32: I think you need to add it as an error in sources_error.go 
and to use errors.New to fix that error.

adiabramovitch, line 32: Maybe you need to change it to err.Error() and %v ?

adiabramovitch:
looks good to me after rebasing and fixing the lining issue :)

lzap (changes requested):
I have some concerns around concurrency and unmarshaling.

internal/clients/sources/sources_impl.go
```diff
@@ -2,12 +2,54 @@ package sources
...
+	Data []AppType `json:"data"`
+}
+
 func getSourcesClient(ctx context.Context) (SourcesIntegration, error) {
-	return NewClientWithResponses(config.Sources.URL)
+	client, err := NewClientWithResponses(config.Sources.URL)
+	if err != nil {
+		return client, err
+	}
+	if config.Sources.AppTypeId == "" {
```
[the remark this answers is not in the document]
ezr-ondrej: well... how bad it is, if we fetch the value multiple times? :thinking: setting should be mutexted, I'd agree, but I feel like it doesn't matter much if we fetch the id 30 times at the beggining ... but let me try to take a look if I can figure something not-too ugly :)

internal/clients/sources/sources_impl.go
```diff
@@ -2,12 +2,54 @@ package sources
...
+		ctxval.GetLogger(ctx).Info().Msg("Fetching the Application Type ID of Provisioning for Sources")
+		resp, err := client.ListApplicationTypes(ctx, &ListApplicationTypesParams{})
+		if err != nil {
+			ctxval.GetLogger(ctx).Warn().Msgf("Failed to fetch ApplicationTypes from sources: %w", err)
+			return client, fmt.Errorf("Failed to fetch ApplicationTypes: %w", err)
+		}
+		defer resp.Body.Close()
+		b, err := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
ezr-ondrej: It works, so there was some other issue :shrug:

internal/config/main_config.go
```diff
@@ -141,3 +141,8 @@ func DumpConfig(logger zerolog.Logger) {
 	configCopy.AWS.Session = replacement
 	logger.Info().Msgf("Configuration: %+v", configCopy)
 }
+
+func SetSourcesAppTypeId(appTypeId string) bool {
+	config.RestEndpoints.Sources.AppTypeId = appTypeId
+	return true
+}
```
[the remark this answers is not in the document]
ezr-ondrej: I feel like this is bit safer, to go through a setter, but if you do not like it, I'm ok to remove it.
I think ultimatelly we should store the AppTypeId elsewhere, it is probably a bad fit for config given that most of the cases (apart of test we need to fetch it)

internal/clients/sources/sources_impl.go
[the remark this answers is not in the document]
ezr-ondrej: what about this? I've just added mutex around fetching the appType, you should not need appTypeId prior having the client, so this should be effectively working as expected...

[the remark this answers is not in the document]
lzap: Mutex does not solve the thread-local cache issue unless also read and write is also protected. To do that, the best option is to make the config value unexported and create a "getter" method that will use the mutex. This will ensure that no other goroutine is able to read an invalid (empty/nil) value unless it's been correctly set.

I suggest sync.Once in a "getter" method. That is just like mutex, but cleaner.

internal/config/main_config.go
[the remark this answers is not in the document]
lzap: Lazy initialization will only work when you cannot set values, you can only read it (via a single function - a "getter" is how I call it).

internal/clients/sources/sources_impl.go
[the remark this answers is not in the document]
ezr-ondrej: Yeah, let's do it right, I agree. In that case I'd move that from config completely, or created a specialized structure for it.

IMHO, for this usecase this is fine, because the read doesn't happen until there is a client, but you're right that this code sets an example, so better do it right.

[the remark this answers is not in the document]
ezr-ondrej: Once is not useable, as we need to pass context, so I've used mutex, this is best efort, but happy to talk it over ^_^

internal/clients/sources/sources_impl.go
```diff
@@ -2,14 +2,75 @@ package sources
...
+}
+
 func getSourcesClient(ctx context.Context) (SourcesIntegration, error) {
-	return NewClientWithResponses(config.Sources.URL)
+	client, err := NewClientWithResponses(config.Sources.URL)
+	if err != nil {
+		return client, err
+	}
+	if err = lazyLoadAppId(ctx, client); err != nil {
+		return client, err
```
lzap, line 38: Isn't better to return `nil, err`? I mean, the caller will unlikely be able to continue.

internal/clients/sources/sources_impl.go
```diff
@@ -2,14 +2,75 @@ package sources
...
+
+type dataElement struct {
+	Data []AppType `json:"data"`
+}
+
+func getAppTypeId() string {
+	lazyLoad.Lock()
+	defer lazyLoad.Unlock()
+	return appTypeId
+}
```
lzap, line 30: Here I have the biggest concern for the future - the caller can accidentally call this function before getSourcesClient and it is not very obvious at all that this must be called after, otherwise locking will not even help and the problem persists - a goroutine/thread can cache an empty value.

So I suggest the solution - remove locking from this function and simply return error or just panic if this is called with the uninitialized value. This will prevent from accidental calls in the future.

Alternatively, you could use https://pkg.go.dev/sync/atomic to ensure the value is never cached, but you'd need to create a string wrapper (Generics hello!). I am fine with just panicking.

lzap (changes requested):
Ok I was really looking into how to utilize sync.Once to improve the code, but it looks like the code only needs a single synchronization block - the initialization method. Left few comments tho, we should panic when the value is read before initialization happens which you do not prevent from.

internal/clients/sources/sources_impl.go
ezr-ondrej: Ok, panicing sounds good here I believe ^_^

lzap (approved):
Not huge fan of this, but it works. But I really hate this.

One last idea before we commit to this approach - what if we get rid of lazy initialization alltogether? We do explicit initialization in multiple places (database, cloudwatch, logging), so if we did the same in `main.go` here, it would work:

```
sources.Initialize(context.Background())
```

The Initialize function could prepare everything and if we do the initialization before we actually start HTTP handlers, it is guaranteed to be safe so no mutexes are necessary, also the AppTypeId could be just a normal value again and not a function. This seems MUCH cleaner.

lzap:
Another option would be simply to always fetch the source ID from sources in each request but creating a global LRU cache with a sane expiration like 10 minutes. That's what Ondrej likes the most :)

internal/cache/lru.go
```diff
@@ -0,0 +1,16 @@
...
+const AppTypeKey = "app_type"
+
+var Cache *lru.Cache
+
+func init() {
+	var err error
+	if Cache, err = lru.New(LruSize); err != nil {
+		panic(err)
+	}
+}
```
lzap, line 16: The interface could be cleaner with exported function, but it's fine.

internal/clients/sources/sources_impl.go
```diff
@@ -2,14 +2,68 @@ package sources
...
 	GetSourcesClient = getSourcesClient
 }
+
+func (c *ClientWithResponses) GetProvisioningTypeId(ctx context.Context, reqEditors ...RequestEditorFn) (string, error) {
+	if appTypeId, ok := cache.Cache.Get(cache.AppTypeKey); ok {
+		return appTypeId.(string), nil
+	}
+	if err := loadAppId(ctx, c); err != nil {
+		return "", err
+	}
```
lzap, line 38: The function could have returned the key itself so additional lookups are not necessary, but it's ok.

internal/clients/sources/sources_impl.go
```diff
@@ -2,14 +2,69 @@ package sources
...
+	if appTypeId, ok := cache.Cache.Get(cache.AppTypeKey); ok {
+		return appTypeId.(string), nil
```
lzap, line 40: A function cache.GetAppTypeKey would be really beneficial because it would hide the implementation, in Go 1.18 Generics could further simplify this.

lzap (approved):
So much cleaner, thanks. I have few nitpicks but we can merge not.

internal/clients/sources/sources_impl.go
ezr-ondrej: yeah I was thinking about that, but then realized it should always be string so it is kinda fine... but I can implement that if you feel it's better :)

internal/clients/sources/sources_impl.go
ezr-ondrej: like this? :)

internal/clients/sources/sources_impl.go
ezr-ondrej: I'd rather not in this PR if it's fine with you

lzap (approved):
Great thanks!

PR #81, opened 2022-07-12
Build the sources URL from Endpoint

ezr-ondrej:
Endpoint is a struct, we need to build the URL from it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ clean:
 build-podman:
 	# remote podman build has problem with -f option, so we link the file as workaround
 	ln -f build/Dockerfile Containerfile
-	podman build -t provisioning-backend .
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -23,7 +23,7 @@ clean:
 build-podman:
 	# remote podman build has problem with -f option, so we link the file as workaround
 	ln -f build/Dockerfile Containerfile
-	podman build -t provisioning-backend .
+	podman build --build-arg quay_expiration=2d -t provisioning-backend .
```
ezr-ondrej, line 26: As we do not use build-podman on the main branch, I thought it would be helpful to add expiration to quay for containers build by build-podman as those are all testing images ATM

lzap (approved):
Sure both patches look good.

PR #83, opened 2022-07-13
Update dependency clowdapp

akhil-jha:
`sources-api` was the older version of the service. It has been replaced by `sources-api-go`. 
`sources-api-go` is the active repo right now.
https://github.com/RedHatInsights/sources-api-go

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -73,7 +73,7 @@ objects:
         name: provisioning-backend
         version: 13
       dependencies:
-        - sources-api
+        - sources
 
 parameters:
   - description: ClowdEnv Name


akhil-jha:
Cannot add you all in `Reviewers`. So here we go @ezr-ondrej @mshriver

deploy/clowdapp.yaml
```diff
@@ -73,7 +73,7 @@ objects:
         name: provisioning-backend
         version: 13
       dependencies:
-        - sources-api
+        - sources-api-go
```
ezr-ondrej, line 76: this does not work for me locally. but using `bonfire deploy sources provisioning` I get the sources-api backed by the go code.

ezr-ondrej (changes requested):
It is the active repo, but the dependency is named still the same, so this trick does not work unfortunatelly.

deploy/clowdapp.yaml
mshriver, line 76:
```suggestion
        - sources
```

ezr-ondrej: I've updated the docs here: https://github.com/RHEnVision/provisioning-backend/pull/78/files ... you can use appsre and it still overrides the locally defined apps :)

ezr-ondrej: the dependencies have bit broader meaning, so we should keep it as is and make sure we deploy with sources app together by other means... as dependencies will make sure we will get URL of that service injected in the container where provisioning-backend runs :) and we need URL of sources-api not sources IMHO :)
Deploying together is better achieved by `bonfire deploy sources provisioning` is that possible for QE? :)

mshriver: that's still reading the service from `deploy/bonfire.yaml` config file, its not being read from appsre itself.

mshriver: I don't think its actually valid as-is, because `sources-api` is also not the name of a clowdapp. I was wrong to approve the previous PR with it.

The concern about getting the right service endpoint is very valid, but we should be able to get this from the sources-api component of the sources app.

To me this is a clear dependency as defined by ClowdAppSpec:
https://redhatinsights.github.io/clowder/clowder/dev/api_reference.html#k8s-api-github-com-redhatinsights-clowder-apis-cloud-redhat-com-v1alpha1-clowdappspec

In order for provisioning-backend app component to function, it requires the sources app to be deployed.  Anyone deploying `provisioning` app (or the individual provisioning-backend component) shouldn't have to separately specify the dependency on `sources`.

Looking at this call to resolve the endpoint, `clowder.DependencyEndpoints["sources-api"]["svc"];`, maybe it works as is, maybe its as simple as prefixing `['sources']` in the key chain.

ezr-ondrej: The docs seems to concurr with you, but I've tried (just once and maybe with wrong config) and it did not work with anything but `sources-api`, so it seems to me, that it actually needs components as dependencies, not the whole apps. Or it might be that every component is one ClowdApp (don't really know here).

Tho the DependencyEnpoints tell the story a bit, because if you take a look at the code it does not allow more nesting, so those need to be components already, but maybe if `sources` would be dependency, then there would be all endpoints for the app...
https://github.com/RedHatInsights/app-common-go/blob/6ca81c147285330f80af3384fa41397abfa9cdaa/pkg/api/v1/config.go#L65

I'm happy to talk and test something together? Maybe we can also try getting some insight from folks who've designed it.

mshriver: @ezr-ondrej @akhil-jha after some discussion this morning, though the app definition is sources, the resourceTemplates defined for sources fulfill the components available to be resolved for dependencies in the ClowdAppSpec.

So `sources-api` is correct here, and currently will deploy `sources-api-go`

PR #85, opened 2022-07-13
Updated source seed script to create DB entries

lzap:
SSIA, this now creates a very much needed authorization with a configured ARN string. The previous seed was not very useful, I am commenting this out for the moment until we need it, if we ever need it I am not sure.

Tested, works fine. I will do a followup PR with documentation on how to create such ARN.


adiabramovitch:
I guess it is necessary to write that this is an example.
the arn is supposed to be linked to a specific account otherwise it is useless.

scripts/sources.conf
```diff
@@ -18,6 +18,7 @@ export REDIS_CACHE_HOST=localhost
 export REDIS_CACHE_PORT=6379
 export SOURCES_API_HOST=http://localhost
 export SOURCES_API_PORT=$PORT
+export ARN_ROLE=arn:aws:iam::123456789:role/redhat-provisioning-1
```
ezr-ondrej, line 21: This needs to be defined by everyone separately as @adiabramovitch has wrote already :thinking:
Shall we keep blank/commented out and fail seeding if left empty (with message what to do)?

ezr-ondrej (commented):
I think we actually need to document how to define the ARN for each account and require it to be set by user prior running this.
In this state it wouldn't work for any other account then `13` (which probably doesn't exist, so it can't be used from our UI - where we use real accounts, thus everyone uses different one) and no other AWS account then yours I think, right?

scripts/sources.seed.sh
```diff
@@ -4,7 +4,36 @@
...
 BASEDIR=$(dirname $0)
 source $BASEDIR/sources.conf
-git clone https://github.com/MikelAlejoBR/sources-database-populator $BASEDIR/sources-database-populator
-pushd $BASEDIR/sources-database-populator
-go run main.go
-popd
+
+# create source for provisioning type with account 13/00013
+curl --location -g  --request POST "http://localhost:$PORT/api/sources/v3.1/bulk_create" \
+--header 'X-RH-Identity: eyJpZGVudGl0eSI6IHsidHlwZSI6ICJVc2VyIiwgImFjY291bnRfbnVtYmVyIjoiMTMiLCAiaW50ZXJuYWwiOnsib3JnX2lkIjoiMDAwMDEzIn19fQo=' \
```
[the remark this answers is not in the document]
lzap: I rewrote this to account 13/00013 actually.

Anyway, good point, this should be configurable.

scripts/sources.conf
lzap: Yeah I was just lazy.

lzap:
Right, to prevent confusing, I am adding a condition - if the variable was not set, then seed will error out and do nothing. Added also configuration values for org_id. I don't think these are important to redefine so I am not erroring out when they are not redefined - both account and tenant will be simply created in both apps automatically.

> In this state it wouldn't work for any other account then 13 (which probably doesn't exist, so it can't be used from our UI - where we use real accounts, thus everyone uses different one) and no other AWS account then yours I think, right?

Yes, sources will create account 13 as well as our app, in the README I am mentioning this, let me elaborate this a bit.

Rebased.

More about the ARN creation in: https://github.com/RHEnVision/provisioning-backend/pull/86

ezr-ondrej:
> Yes, sources will create account 13 as well as our app

I get that, but now when we have UI, we figured having random id is not good enough, because in UI testing we use real staging accounts, so the data needs to be seeded for those accounts in order to seed data visible in UI :)

So thanks a lot for making it customizable, it would not be useful in my case without this! :tada:

scripts/sources.seed.sh
```diff
@@ -4,7 +4,44 @@
...
+		{
+			"source_name": "Amazon source",
+			"application_type_name": "provisionin"
+		}
+	],
+	"authentications": [
+		{
+			"resource_type": "Application",
+			"resource_name": "provisioning",
+			"username": "$ARN_ROLE",
```
ezr-ondrej, line 37: isn't this enclosed in simple quotes for the -d `'` and in that case this variable would not be evaluated, right? :thinking:

ezr-ondrej (commented):
Looks great to me ^_^
Just one nit I do not understand, but apart of it great addition, it will be quite helpful :)

scripts/sources.seed.sh
```diff
@@ -4,7 +4,44 @@
...
+		{
+			"name": "Amazon source",
+			"source_type_name": "amazon",
+			"app_creation_workflow": "manual_configuration"
+		}
+	],
+	"applications": [
+		{
+			"source_name": "Amazon source",
+			"application_type_name": "provisionin"
```
adiabramovitch, line 30: A g is missinh

lzap: Ah good catch.

scripts/sources.seed.sh
```diff
@@ -4,7 +4,44 @@
...
-pushd $BASEDIR/sources-database-populator
-go run main.go
-popd
+
+if [[ -z "$ARN_ROLE" ]]; then
+	echo "ARN_ROLE must be defined in sources.local.conf!"
+	exit 1
+fi
+
+ITITY=$($BASEDIR/identity_header.sh $ACCOUNT_ID $ORG_ID)
```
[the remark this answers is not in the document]
lzap: Ah good catch.

scripts/sources.seed.sh
lzap: Ah good catch.

lzap:
Rebased.

scripts/sources.seed.sh
```diff
@@ -4,7 +4,46 @@
...
-go run main.go
-popd
+
+if [[ -z "$ARN_ROLE" ]]; then
+	echo "ARN_ROLE must be defined in sources.local.conf!"
+	exit 1
+fi
+
+echo "Creating $ARN_ROLE with account_id $ACCOUNT_ID org_id $ORG_ID"
+IDENTITY=$($BASEDIR/identity_header.sh $ACCOUNT_ID $ORG_ID)
```
adiabramovitch, line 14: I am getting an error, I am wondering if it just me or for someone else as well: 
```
./scripts/sources.seed.sh: line 14: /identity_header.sh: No such file or directory
{"errors":[{"detail":"Internal Server Error: account number or OrgId not present in x-rh-identity","status":"500"}]}
```

lzap: Hmmm it worked for me:

```
[lzap@nuc provisioning-backend]$ scripts/sources.seed.sh
Creating arn:aws:iam::12334455655:role/Provisioning1 with account_id 13 org_id 000013
{"sources":[{"id":"1","created_at":"2022-07-18T09:31:40+02:00","updated_at":"2022-07-18T09:31:40+02:00","name":"Amazon source","uid":"c770fbfd-daa7-453a-ac6e-8208cd2ed68a","app_creation_workflow":"manual_configuration","source_type_id":"4"}],"applications":[{"id":"1","created_at":"2022-07-18T09:31:40+02:00","updated_at":"2022-07-18T09:31:40+02:00","source_id":"1","application_type_id":"3"}],"endpoints":[],"authentications":[{"id":"1","authtype":"provisioning-arn","username":"arn:aws:iam::569221671538:role/Provisioning1","resource_type":"Application","resource_id":"1"}]}

[lzap@nuc provisioning-backend]$ scripts/sources.seed.sh
Creating arn:aws:iam::234234234524:role/Provisioning1 with account_id 13 org_id 000013
{"errors":[{"detail":"Internal Server Error: name already exists in tenant","status":"500"}]}
```

Can you investigate? It looks like it cannot find `identity_header.sh` in your case. The script tries to find it in the same directory as the script you executed. Therefore you need to have `scripts/identity_header.sh`. Does it also have executable permissions? Oh by the way do you have ACCOUNT_ID and ORG_ID variables set? You need to have them. They are set in the `sources.conf` by default.

adiabramovitch: So it is executable and I have copied the sources.conf to sources.local.conf and pasted my arn.

lzap: Can you paste the full error please?

adiabramovitch: Replied on Slack @lzap

lzap: For the record, sources.conf sources sources.local.conf and if you copied that file it causes endless loop of sourcing itself leading to OOM kill of bash.

lzap:
Fixed, it was designed badly I think. Config files should be plain, I moved all sourcing to scripts.

adiabramovitch (approved):

adiabramovitch:
@lzap Thanks!

PR #88, opened 2022-07-14
Fix prefix in http client environment

lzap:
SSIA


lzap:
I also added instance types request and parametrized source ID as a separate commit.

adiabramovitch (approved):

PR #89, opened 2022-07-18
HMSPROV-147: Add errors from sources client

adiabramovitch:
Sources client have JSON404 and JSON400 errors linked to their response. 
In this PR I wanted to include those errors in our error handling process to provide more information about errors that weren't caused on our side. 

It is a draft to pass the general idea, I am open to suggestions :)

### DIFF
--- a/internal/clients/sources/sources_errors.go
+++ b/internal/clients/sources/sources_errors.go
@@ -1,6 +1,7 @@
 package sources
 
 import (
+	"encoding/json"
 	"errors"
 )
 
@@ -8,6 +9,23 @@ var MoreThenOneAuthenticationForSourceErr = errors.New("more then one authentica
 var AuthenticationForSourcesNotFoundErr = errors.New("authentications for source weren't found in sources app")
 var SourcesClientErr = errors.New("sources client error")
 var ApplicationNotFoundErr = errors.New("application not found is sources app")
-var ApplicationTypesFetchUnsuccessful = errors.New("failed to fetch ApplicationTypes")
-var ApplicationTypeNotFound = errors.New("application type 'provisioning' has not been found in types supported by sources")
-var ApplicationTypeCacheFailed = errors.New("application type id failed to write to cache")
+var ApplicationTypesFetchUnsuccessfulErr = errors.New("failed to fetch ApplicationTypes")
+var ApplicationTypeNotFoundErr = errors.New("application type 'provisioning' has not been found in types supported by sources")
+var ApplicationTypeCacheFailedErr = errors.New("application type id failed to write to cache")
+var CantMarshalErr = errors.New("cant marshal array to json")
+
+func ParseErrorNotFoundToJSON(sourcesErrors ErrorNotFound) ([]byte, error) {
+	j, err := json.Marshal(sourcesErrors)
+	if err != nil {
+		return nil, CantMarshalErr
+	}
+	return j, nil
+}
+
+func ParseErrorBadRequestToJSON(sourcesErrors ErrorBadRequest) ([]byte, error) {
+	j, err := json.Marshal(sourcesErrors)
+	if err != nil {
+		return nil, CantMarshalErr
+	}
+	return j, nil
+}
--- a/internal/payloads/error_payload.go
+++ b/internal/payloads/error_payload.go
@@ -28,6 +28,15 @@ type ResponseError struct {
 	Context context.Context `json:"-"`
 }
 
+type ClientError struct {
+	Message string `json:"msg"`
+	Body    string `json:"body"`
+}
+
+func (e ClientError) Error() string {
+	return e.Message
+}
+
 func (e *ResponseError) Render(_ http.ResponseWriter, r *http.Request) error {
 	render.Status(r, e.HTTPStatusCode)
 	return nil
--- a/internal/services/sources_service.go
+++ b/internal/services/sources_service.go
@@ -20,28 +20,42 @@ func ListSources(w http.ResponseWriter, r *http.Request) {
 	}
 	appTypeId, err := client.GetProvisioningTypeId(r.Context(), AddIdentityHeader)
 	if err != nil {
-		renderError(w, r, payloads.SourcesClientError(r.Context(), "get provisioning app type", err, 500))
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "get provisioning type id", err, 500))
 		return
 	}
 	resp, err := client.ListApplicationTypeSourcesWithResponse(r.Context(), appTypeId, &sources.ListApplicationTypeSourcesParams{}, AddIdentityHeader)
 	statusCode := resp.StatusCode()
 	if err != nil {
-		renderError(w, r, payloads.SourcesClientError(r.Context(), "list sources", err, statusCode))
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "list application types", err, statusCode))
 		return
 	}
+
 	if parsing.IsHTTPNotFound(statusCode) {
-		renderError(w, r, payloads.SourcesClientError(r.Context(), "sources not found", err, statusCode))
+		errors, err := sources.ParseErrorNotFoundToJSON(*resp.JSON404)
+		if err != nil {
+			renderError(w, r, payloads.NewRenderError(r.Context(), "list sources", err))
+			return
+		}
+		sourcesErr := payloads.ClientError{Message: fmt.Sprintf("source not found %s", errors)}
+		renderError(w, r, payloads.NewNotFoundError(r.Context(), sourcesErr))
 		return
 	}
+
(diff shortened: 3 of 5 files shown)


internal/parsing/parsing.go
```diff
@@ -1,9 +1,32 @@
...
+	return j, nil
+}
+
+func ParseErrorBadRequestToJSON(errors sources.ErrorBadRequest) ([]byte, error) {
+	j, err := json.Marshal(errors)
+	if err != nil {
+		return nil, fmt.Errorf("cant marshel array to json")
+	}
+	return j, nil
+}
```
lzap, line 32: Since these errors are specific to sources, could you put them there? This package looks like generic parsing.

internal/parsing/parsing.go
```diff
@@ -1,9 +1,32 @@
...
+
+func ParseErrorBadRequestToJSON(errors sources.ErrorBadRequest) ([]byte, error) {
+	j, err := json.Marshal(errors)
+	if err != nil {
+		return nil, fmt.Errorf("cant marshel array to json")
```
lzap, line 29:
```suggestion
		return nil, fmt.Errorf("cant marshal array to json")
```

lzap: The linter will complain about this, use errors.New for this case.

internal/payloads/error_payload.go
```diff
@@ -28,6 +28,13 @@ type ResponseError struct {
 	Context context.Context `json:"-"`
 }
 
+type ClientResponseError struct {
+	ResponseError
+
+	// Response Body from Client
+	ClientErrors []byte `json:"-"`
```
lzap, line 35: Why plural? From what I can see, the client returns one error. It can probably contain wrapped error, but in that case the generic consensus in Go is to simply use singular.

Is there a particular reason why you do not want this to be marshaled to JSON? I mean, it can be pretty useful for API users to actually see the root cause.

internal/services/sources_service.go
```diff
@@ -52,59 +65,95 @@ func GetSource(w http.ResponseWriter, r *http.Request) {
 	resp, err := client.ShowSourceWithResponse(r.Context(), sources.ID(id), AddIdentityHeader)
 	statusCode := resp.StatusCode()
 	if err != nil {
-		renderError(w, r, payloads.SourcesClientError(r.Context(), "show source", err, statusCode))
+		renderError(w, r, payloads.SourcesClientError(r.Context(), "show source", err, statusCode, resp.Body))
```
lzap, line 68: I don't like mixing errors and body in a single argument (variable). Can't we simply create a new struct that will have both? Then you can set one, or the other accordingly.

lzap:
The intention is very good, it is extremely useful not to hide root errors coming from 3rd party systems. Nice.

Now, the implementation can be vastly improved - what it needs is a good encapsulation. See, you are passing byte slice around, everytime anyone has to work a byte slice there is the big question - what does it represent? It is too abstract. You can have a dedicated type for that. Not only that, those parsing functions that you struggle to find home for are great candidates to be declared on such type! They can actually be "constructors" (NewFunctions).

I think that new type should be actually an error (implementing error Go interface), because then you can take it even further and take advantage of error wrapping. Everytime you encouter a client error, you can simply create a new value of such type (e.g. `ClientError`), put the original error as well as body into it (using the parsing code) and continue. The final error handling code in `SourcesClientError` then can simply unwrap (via `errors.As` library function) the original client error and if it is present it can put it to the newly created fields in `ClientResponseError`.

I say fields, because your initial draft has a field named ClientErrors and then you put either error messages or whole body in there. I feel like both can be useful, so I suggest to create actually two fields: ClientError and Body. Both can be optional depending on the situation.

I hope it makes sense, this is a really great improvement. Good work so far!

internal/payloads/error_payload.go
adiabramovitch: https://github.com/RHEnVision/provisioning-backend/blob/70009fc58bfed0a287266135c9a9eac955678968/internal/clients/sources/client.gen.go#L624
I am getting from sources an array of errors and I am parsing all of them.

internal/services/sources_service.go
adiabramovitch: Do you mean to change the signature of `payloads.SourcesClientError `to something like that: 
```golang
func SourcesClientError(ctx context.Context, message string, err error, status int, errors []byte, body []byte) 
```
And pass nil when one or the other is not relevant?

adiabramovitch:
@ezr-ondrej I need your opinion here: 
I think I understood what lzap meant. 
I created this type which implements go Error interface: 
```golang
type ClientError struct {
	Message    string
	WrappedErr error
	Body       string // optional body
}
```

And I intend to pass this error instead of the err given here (and drop the msg variable because it is included in new err struct): 
https://github.com/RHEnVision/provisioning-backend/blob/71227736ff05e32c16ead1888e61510e933c9e34/internal/services/sources_service.go#L29

like that: 
```golang
sourcesErr := payloads.ClientError{
			Message:    "list sources",
			WrappedErr: err,  // the general error I got from the link above
			Body:       string(resp.Body),
		}
```
I am getting the errors from sources after I am parsing those to a []byte. 
I am getting from sources a response which has these fields (no matter the type of the response):
https://github.com/RHEnVision/provisioning-backend/blob/71227736ff05e32c16ead1888e61510e933c9e34/internal/clients/sources/client.gen.go#L3713 
and JSON404/400 are errors of this type: 
https://github.com/RHEnVision/provisioning-backend/blob/71227736ff05e32c16ead1888e61510e933c9e34/internal/clients/sources/client.gen.go#L624 
doesn't seem like they implement the err interface of go. 
So should I add another field to `ClientError` which holds a []byte of errors? 
```golang
type ClientError struct {
	Message    string
	WrappedErr error // General error
	Body       string // optional body
	Errors []byte or string // more information about the error from sources side
}
```

ezr-ondrej:
> I created this type which implements go Error interface:

good, I think we have good example here to follow (naming the WrappedErr as Err) https://github.com/RHEnVision/provisioning-backend/blob/71227736ff05e32c16ead1888e61510e933c9e34/internal/dao/dao_errors.go#L9

I wonder tho, what is the difference between Body and message.
I think that the whole purpose of wrapping the errors is to take message from the error you get and translate it into your own.


> So should I add another field to `ClientError` which holds a []byte of errors?

I think it also answers this, by: no, I think you should not copy the information from their structures, that's why we wrap it.
If those structures do not implement the error interface, then I'd simply not pass them at all.

Our intent should be extracting any useful information and construct a meaningful message to the user.
We can decide to join the errors.Details by `,` and just include them in the message, or just not pass them at all.

internal/clients/sources/sources_errors.go
```diff
@@ -1,13 +1,31 @@
...
 var AuthenticationForSourcesNotFoundErr = errors.New("authentications for source weren't found in sources app")
 var SourcesClientErr = errors.New("sources client error")
 var ApplicationNotFoundErr = errors.New("application not found is sources app")
-var ApplicationTypesFetchUnsuccessful = errors.New("failed to fetch ApplicationTypes")
-var ApplicationTypeNotFound = errors.New("application type 'provisioning' has not been found in types supported by sources")
-var ApplicationTypeCacheFailed = errors.New("application type id failed to write to cache")
+var ApplicationTypesFetchUnsuccessfulErr = errors.New("failed to fetch ApplicationTypes")
+var ApplicationTypeNotFoundErr = errors.New("application type 'provisioning' has not been found in types supported by sources")
+var ApplicationTypeCacheFailedErr = errors.New("application type id failed to write to cache")
+var CantMarshalErr = errors.New("cant marshal array to json")
```
adiabramovitch, line 15: I added Err to the errors, lzap said it is Go good practice

internal/payloads/error_payload.go
```diff
@@ -58,7 +67,7 @@ func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
 func SourcesClientError(ctx context.Context, message string, err error, status int) *ResponseError {
```
ezr-ondrej, line 67: couldn't we drop the status and just use NewNotFoundError for the 404?
I think we need to split notFound and badRequest, because those should get reported on differently to the user, 404 is something users can act on and we do not need to log, because support does not need to react, whereas badRequest is the opposite, we need to log it, but to user we just need to send info "Hey our service is broken, try again later"

lzap:
> I wonder tho, what is the difference between Body and message.

Yeah, my thinking was that if we are able to generate/parse a meaningful message it should be in message. But services often return 500 nonsense (e.g. even HTML code in case you hit a wrong URL by mis-configuration) and in these reasons the whole body in our log could be useful (message would be something like "unknown error").

ezr-ondrej:
> in these reasons the whole body in our log could be useful

Make sense, I guess that we should standardize the client errors and have just one, I believe that is still good thing to do, even tho we've moved the code :)

PR #90, opened 2022-07-18
HMSPROV-137: Get account from identity

ezr-ondrej:
Adds Account to the context if found by identity information present in the header.

Test is separate in #92 as it needs #91 first and I didn't want to do #91 in this PR

### DIFF
--- a/internal/ctxval/ctx_getters.go
+++ b/internal/ctxval/ctx_getters.go
@@ -3,6 +3,7 @@ package ctxval
 import (
 	"context"
 
+	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
@@ -36,4 +37,11 @@ func GetRequestId(ctx context.Context) string {
 
 func GetIdentity(ctx context.Context) identity.XRHID {
 	return identity.Get(ctx)
+}
+
+func GetAccount(ctx context.Context) *models.Account {
+	if ctx.Value(AccountCtxKey) == nil {
+		return nil
+	}
+	return ctx.Value(AccountCtxKey).(*models.Account)
 }
--- a/internal/dao/sqlx/account_dao.go
+++ b/internal/dao/sqlx/account_dao.go
@@ -2,6 +2,10 @@ package sqlx
 
 import (
 	"context"
+	"database/sql"
+	"errors"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	"github.com/RHEnVision/provisioning-backend/internal/db"
@@ -69,6 +73,28 @@ func (di *accountDaoSqlx) GetById(ctx context.Context, id int64) (*models.Accoun
 	return result, nil
 }
 
+func (di *accountDaoSqlx) GetOrCreateByIdentity(ctx context.Context, orgId string, accountNumber string) (*models.Account, error) {
+	var acc, err = di.GetByOrgId(ctx, orgId)
+	var norows *dao.NoRowsError
+	if err == nil {
+		return acc, nil
+	} else if errors.As(err, &norows) {
+		ctxval.GetLogger(ctx).Debug().Msgf("Account not found by org id %s", orgId)
+	} else {
+		return nil, err
+	}
+	acc, err = di.GetByAccountNumber(ctx, accountNumber)
+	if err == nil {
+		return acc, nil
+	} else if errors.As(err, &norows) {
+		ctxval.GetLogger(ctx).Debug().Msgf("Account not found by account number %s", accountNumber)
+	} else {
+		return nil, err
+	}
+	// TODO create HMSPROV-135
+	return nil, err
+}
+
 func (di *accountDaoSqlx) GetByAccountNumber(ctx context.Context, number string) (*models.Account, error) {
 	query := getAccountByAccountNumber
 	stmt := di.getByAccountNumber
@@ -76,7 +102,11 @@ func (di *accountDaoSqlx) GetByAccountNumber(ctx context.Context, number string)
 
 	err := stmt.GetContext(ctx, result, number)
 	if err != nil {
-		return nil, NewGetError(ctx, di, query, err)
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, NewNoRowsError(ctx, di, query, err)
+		} else {
+			return nil, NewGetError(ctx, di, query, err)
+		}
 	}
 	return result, nil
 }
@@ -88,7 +118,11 @@ func (di *accountDaoSqlx) GetByOrgId(ctx context.Context, orgId string) (*models
 
 	err := stmt.GetContext(ctx, result, orgId)
 	if err != nil {
-		return nil, NewGetError(ctx, di, query, err)
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, NewNoRowsError(ctx, di, query, err)
+		} else {
+			return nil, NewGetError(ctx, di, query, err)
+		}
 	}
 	return result, nil
 }
--- a/internal/middleware/account.go
+++ b/internal/middleware/account.go
@@ -1,0 +1,36 @@
+package middleware
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+)
+
+func AccountMiddleware(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		rhId := ctxval.GetIdentity(r.Context())
+		logger := ctxval.GetLogger(r.Context())
+		accDao, err := dao.GetAccountDao(r.Context())
+		if err != nil {
+			logger.Error().Err(err).Msg("Failed to initialize connection to fetch Account info")
+			http.Error(w, http.StatusText(500), 500)
+			return
+		}
+		acc, err := accDao.GetOrCreateByIdentity(r.Context(), rhId.Identity.OrgID, rhId.Identity.AccountNumber)
+		if err != nil {
+			logger.Error().Err(err).Msgf("Failed to fetch account by org_id=%s/account=%s", rhId.Identity.OrgID, rhId.Identity.AccountNumber)
+			// TODO switch to error out once we enable account creation in HMSPROV-148
+			//http.Error(w, http.StatusText(500), 500)
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		newLogger := logger.With().Str("org_id", acc.OrgID).Str("account_number", *acc.AccountNumber).Logger()
+		ctx := context.WithValue(r.Context(), ctxval.AccountCtxKey, acc)
+		ctx = context.WithValue(ctx, ctxval.LoggerCtxKey, newLogger)
+		next.ServeHTTP(w, r.WithContext(ctx))
+	}
+	return http.HandlerFunc(fn)
+}
--- a/internal/ctxval/ctx_keys.go
+++ b/internal/ctxval/ctx_keys.go
@@ -6,5 +6,6 @@ const (
 	LoggerCtxKey     CommonKeyId = iota
 	RequestIdCtxKey  CommonKeyId = iota
 	RequestNumCtxKey CommonKeyId = iota
+	AccountCtxKey    CommonKeyId = iota
 	ResourceCtxKey   CommonKeyId = iota
 )
--- a/internal/dao/dao_interfaces.go
+++ b/internal/dao/dao_interfaces.go
@@ -10,6 +10,7 @@ var GetAccountDao func(ctx context.Context) (AccountDao, error)
 
 type AccountDao interface {
 	GetById(ctx context.Context, id int64) (*models.Account, error)
+	GetOrCreateByIdentity(ctx context.Context, orgId string, accountNumber string) (*models.Account, error)
 	GetByAccountNumber(ctx context.Context, number string) (*models.Account, error)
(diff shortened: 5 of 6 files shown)


internal/middleware/account.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+)
+
+func AccountMiddleware(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		rhId := ctxval.GetIdentity(r.Context())
+		accDao, err := dao.GetAccountDao(r.Context())
+		if err != nil {
+			ctxval.GetLogger(r.Context()).Error().Err(err).Msg("Failed to initialize connection to fetch Account info")
+			// TODO should we fail hard?
```
adiabramovitch, line 17: Wondering why not? what needs to be considered here?

adiabramovitch (commented):
overall LGTM just few questions :)

internal/middleware/account.go
```diff
@@ -0,0 +1,33 @@
...
+			next.ServeHTTP(w, r)
+			return
+		}
+		acc, err := accDao.GetOrCreateByIdentity(r.Context(), rhId.Identity.OrgID, rhId.Identity.AccountNumber)
+		if err != nil {
+			ctxval.GetLogger(r.Context()).Warn().Err(err).Msg("Failed to fetch account by identity")
+			// TODO do something about it
```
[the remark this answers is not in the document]
ezr-ondrej: Yeah I guess, but for now we do not have create, so ATM it just lets it pass if the account is not in the database.
Bit confusing tho, agreed :thinking:

internal/ctxval/ctx_getters.go
```diff
@@ -37,3 +38,10 @@ func GetRequestId(ctx context.Context) string {
 func GetIdentity(ctx context.Context) identity.XRHID {
 	return identity.Get(ctx)
 }
+
+func GetAccount(ctx context.Context) *models.Account {
+	if ctx.Value(AccountCtxKey) == nil {
+		return nil
+	}
+	return ctx.Value(AccountCtxKey).(*models.Account)
+}
```
lzap, line 47: In the future, I think we should move also all "setters" here. Not as part of this PR, I actually introduced this and now I feel like it would be cleaner to have it here.

internal/dao/sqlx/account_dao.go
```diff
@@ -68,6 +72,27 @@ func (di *accountDaoSqlx) GetById(ctx context.Context, id int64) (*models.Accoun
...
+	}
+	acc, err = di.GetByAccountNumber(ctx, accountNumber)
+	if err == nil {
+		return acc, nil
+	} else if errors.As(err, &norows) {
+		ctxval.GetLogger(ctx).Info().Msgf("Account not found by account number %s", accountNumber)
+	} else {
+		return nil, err
+	}
+	// TODO create
```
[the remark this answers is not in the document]
ezr-ondrej: There is #84 which implements it, so I didn't want to step on someone's toes.
And this PR is adding value even without that, not complete tho, so I'd be ok if we decide to block until #84 is in.

internal/ctxval/ctx_getters.go
ezr-ondrej: Yeah, it deserves some discussion I believe, but yeah, having the setters there and making the keys private is probably a good path forward.

internal/dao/sqlx/account_dao.go
```diff
@@ -68,6 +72,27 @@ func (di *accountDaoSqlx) GetById(ctx context.Context, id int64) (*models.Accoun
...
+	} else if errors.As(err, &norows) {
+		ctxval.GetLogger(ctx).Info().Msgf("Account not found by org id %s", orgId)
+	} else {
+		return nil, err
+	}
+	acc, err = di.GetByAccountNumber(ctx, accountNumber)
+	if err == nil {
+		return acc, nil
+	} else if errors.As(err, &norows) {
+		ctxval.GetLogger(ctx).Info().Msgf("Account not found by account number %s", accountNumber)
```
lzap, line 89: I suggest to pull out GetLogger into a local variable if you use it twice. But since this is an error state, probably okay I am just wondering if copy-pasting the code would not create a misleading pattern.

internal/dao/sqlx/account_dao.go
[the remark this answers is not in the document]
lzap: Yeah, the same question :)

Can you drop a comment for this function that this is intentionally not wrapped in a SQL transaction for performance reasons (we do it on EVERY request). Just to know about it for the future. I think we might run into concurrency issues someday.

internal/dao/sqlx/account_dao.go
```diff
@@ -68,6 +72,27 @@ func (di *accountDaoSqlx) GetById(ctx context.Context, id int64) (*models.Accoun
 	}
 	return result, nil
 }
+func (di *accountDaoSqlx) GetOrCreateByIdentity(ctx context.Context, orgId string, accountNumber string) (*models.Account, error) {
+	var acc, err = di.GetByOrgId(ctx, orgId)
+	var norows *dao.NoRowsError
+	if err == nil {
+		return acc, nil
+	} else if errors.As(err, &norows) {
+		ctxval.GetLogger(ctx).Info().Msgf("Account not found by org id %s", orgId)
```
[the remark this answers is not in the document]
ezr-ondrej: I feel like it is quite interesting to log tho (maybe just logging account created tho)?
This puts light on an issue that `NoRowsError` automatically logs, so at this point we have two logs `Warn` about NoRow, and then this one... would love advice on what to do about that :thinking:

internal/middleware/account.go
ezr-ondrej: Probably nothing, I just will need to discuss how we fail in middlewares :)

lzap: Absolutely, if DB is down or there is a SQL syntax error, we simply cannot continue - that is 500 and call us to fix this.

lzap: Uh these multiple `GetLogger` calls are bugging me, but again, these are error states. If this was in the success code path, I would be against it but... lucky you :)

internal/middleware/account.go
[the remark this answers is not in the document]
lzap: Absolutely, if the identity is missing/corrupt then that's a 500.

internal/middleware/account.go
```diff
@@ -0,0 +1,33 @@
...
+			next.ServeHTTP(w, r)
+		} else {
+			ctxval.GetLogger(r.Context()).Info().Msgf("Current account: orgId '%s', number: '%s'", acc.OrgID, *acc.AccountNumber)
+			ctx := context.WithValue(r.Context(), ctxval.AccountCtxKey, acc)
+			next.ServeHTTP(w, r.WithContext(ctx))
+		}
```
lzap, line 30: Please keep the line of sight. In other words, drop the else block, it's the success code path otherwise we will eventually end up with 8 inner else blocks.

https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

internal/middleware/account.go
```diff
@@ -0,0 +1,33 @@
...
+			next.ServeHTTP(w, r)
+			return
+		}
+		acc, err := accDao.GetOrCreateByIdentity(r.Context(), rhId.Identity.OrgID, rhId.Identity.AccountNumber)
+		if err != nil {
+			ctxval.GetLogger(r.Context()).Warn().Err(err).Msg("Failed to fetch account by identity")
+			// TODO do something about it
+			next.ServeHTTP(w, r)
+		} else {
+			ctxval.GetLogger(r.Context()).Info().Msgf("Current account: orgId '%s', number: '%s'", acc.OrgID, *acc.AccountNumber)
```
lzap, line 27: I would use debug here, I will explain in a bit why.

lzap:
Looks good, I have one big request tho.

AccountID and OrgID both should be absolutely logged with every single log entry, this can be achieved simply by moving the line

	r.Use(m.LoggerMiddleware(&log.Logger))

after the line

	routes.SetupRoutes(r)

and updating the `LoggerMiddleware` to actually acquire account from the context putting it into the context logger as two string fields named `account_id` and `org_id`.

Then I think your success path log line should be debug, because we will see the account information with every single log line written as a custom field. More importantly, we will be able to see all logs for a particular account (e.g. when investigating a bug spread over multiple days).

internal/dao/sqlx/account_dao.go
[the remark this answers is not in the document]
lzap: See my remark about context logging update below. But if you insist on info, I have no problem with that. It's just quite detailed information to my taste at info level.

ezr-ondrej:
> and updating the `LoggerMiddleware` to actually acquire account from the context putting it into the context logger as two string fields named `account_id` and `org_id`.

I've instead went for updating the logger in the context, so the logger can stay ready early, just without the acc and org info, and it will be logging these further in the request lifecycle.

internal/middleware/account.go
ezr-ondrej: I've enhanced the logger, so I believe this is bit useless, this info will be debug logged by the dao (on not found and created).

internal/dao/sqlx/account_dao.go
```diff
@@ -68,6 +72,27 @@ func (di *accountDaoSqlx) GetById(ctx context.Context, id int64) (*models.Accoun
 	}
 	return result, nil
 }
+func (di *accountDaoSqlx) GetOrCreateByIdentity(ctx context.Context, orgId string, accountNumber string) (*models.Account, error) {
```
lzap, line 75: Missing a newline, no biggie.

internal/middleware/account.go
```diff
@@ -0,0 +1,34 @@
...
+		rhId := ctxval.GetIdentity(r.Context())
+		logger := ctxval.GetLogger(r.Context())
+		accDao, err := dao.GetAccountDao(r.Context())
+		if err != nil {
+			logger.Error().Err(err).Msg("Failed to initialize connection to fetch Account info")
+			panic(err)
+		}
+		acc, err := accDao.GetOrCreateByIdentity(r.Context(), rhId.Identity.OrgID, rhId.Identity.AccountNumber)
+		if err != nil {
+			logger.Warn().Err(err).Msg("Failed to fetch account by identity")
```
lzap, line 22:
```suggestion
			logger.Warn().Err(err).Msgf("Failed to fetch account by org_id=%s/account=%s", rhId.Identity.OrgID, rhId.Identity.AccountNumber)
```

internal/middleware/account.go
```diff
@@ -0,0 +1,34 @@
...
+			// TODO do something about it, or fail once we have create implemented
```
lzap, line 23: Please error out with 500 here, we must not continue at this point it's a fatal error and security related, so no TODOs please anymore :)

lzap (changes requested):
I am good, except that when identity cannot be fetched, we must return 500. I call for doing this right from the day one.

internal/middleware/account.go
ezr-ondrej: You know it will break the app now, right? So we need to wait for Create in that case :)

internal/dao/sqlx/account_dao.go
ezr-ondrej: Fixed :)

internal/dao/sqlx/account_dao.go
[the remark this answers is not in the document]
ezr-ondrej: The logging of the error is addressed in #93

internal/middleware/account.go
ezr-ondrej: I've improved the TODO message, and put a commentable implementation how I'd return 500 there.
Tho if you insist on this comment, we need to have `account.Create` in place prior merging this.
As also explained in https://github.com/RHEnVision/provisioning-backend/pull/90#discussion_r923295961

lzap:
Please rebase on top of the NoRowsError changes.

lzap (approved):
Thanks.

internal/middleware/account.go
```diff
@@ -0,0 +1,36 @@
...
+		acc, err := accDao.GetOrCreateByIdentity(r.Context(), rhId.Identity.OrgID, rhId.Identity.AccountNumber)
+		if err != nil {
+			logger.Error().Err(err).Msgf("Failed to fetch account by org_id=%s/account=%s", rhId.Identity.OrgID, rhId.Identity.AccountNumber)
+			// TODO switch to error out once we enable account creation
+			//http.Error(w, http.StatusText(500), 500)
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		newLogger := logger.With().Str("org_id", acc.OrgID).Str("account_number", *acc.AccountNumber).Logger()
```
adiabramovitch, line 30: Why are you creating a new loggger? :)

internal/dao/sqlx/account_dao.go
```diff
@@ -69,14 +73,40 @@ func (di *accountDaoSqlx) GetById(ctx context.Context, id int64) (*models.Accoun
...
 func (di *accountDaoSqlx) GetByAccountNumber(ctx context.Context, number string) (*models.Account, error) {
 	query := getAccountByAccountNumber
 	stmt := di.getByAccountNumber
 	result := &models.Account{}
 
 	err := stmt.GetContext(ctx, result, number)
 	if err != nil {
-		return nil, NewGetError(ctx, di, query, err)
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, NewNoRowsError(ctx, di, query)
```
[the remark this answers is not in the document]
ezr-ondrej: Yeah, needed to rebase on top of #93 :)

internal/middleware/account.go
ezr-ondrej: I'm adding `org_id` and `account_number` to the logger, so from this point on, it will be logged automatically with every log entry during the request. Neat, right? :stuck_out_tongue: (@lzap's invention)

adiabramovitch:
@ezr-ondrej LGTM thanks!

PR #91, opened 2022-07-18
Add Account DAO stubs

ezr-ondrej:


internal/dao/stubs/account_dao.go
```diff
@@ -0,0 +1,51 @@
...
+	return nil, nil
+}
+
+func (stub *AccountDaoStub) GetByOrgId(ctx context.Context, orgId string) (*models.Account, error) {
+	return nil, nil
+}
+
+func (stub *AccountDaoStub) List(ctx context.Context, limit, offset int64) ([]*models.Account, error) {
+	return stub.store, nil
+}
```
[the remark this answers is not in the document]
ezr-ondrej: So you mean add limit and offset here? this is very complete apart of that.

ezr-ondrej:
I agree "working" stub is needed, but I don't want to have account as the one, I'd like to keep our tests simple and for that I believe one static account is completely fine, so now I've created exactly that, no store just one account you can get or not, it errors out explicitly if you try anything else. So I think now it is fully funcioning DAO, it should not set example for other DAO stubs so I've made it kinda imposible to copy.

I'm happy to take on improvements for the other daos, but lets not make Account dao the one :)
Quite a good piece of funcionality would need to go in without test if we block on DAO improvements.

lzap:
> I don't want to have account as the one, I'd like to keep our tests simple and for that I believe one static account is completely fine

I can think of scenarios when this is limiting. For example, a test of type "create something in account A, make sure that account B does not see object from A". Do you have another reason to avoid implementing a fully working DAO stub? I mean, we know for fact we will need it, why not do it now? The current status of DAO stubs is already a bit ugly, I would rather see PRs which improve it than which just stacks another temporary commit on top of that. My concern is that we would start adding tests with poor stubs ending up with ugly tests.

Let me review the changes you made, at least the static stub should have all functions fully working and not returning nils.

internal/dao/stubs/account_dao.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return &stubAccount, nil
+}
+
+func (stub *AccountDaoStub) GetOrCreateByIdentity(ctx context.Context, orgId string, accountNumber string) (*models.Account, error) {
+	if orgId != "1" && accountNumber != "1" {
+		return nil, SingleAccountError
+	}
+	return &stubAccount, nil
+}
```
lzap, line 42: I am really looking forward seeing unit tests for the account creation middleware functionality we are merging in, let's say few hours. :-)

internal/dao/stubs/account_dao.go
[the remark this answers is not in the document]
lzap: No I meant all functions returning `nil`.

internal/dao/stubs/context_getters.go
```diff
@@ -0,0 +1,43 @@
...
+	ctx := context.WithValue(parent, accountCtxKey, &AccountDaoStub{})
+	return ctx
+}
+
+func getAccountDaoStub(ctx context.Context) (accdao *AccountDaoStub, err error) {
+	var ok bool
+	if accdao, ok = ctx.Value(accountCtxKey).(*AccountDaoStub); !ok {
+		err = contextReadError
+	}
+	return
```
lzap, line 42: Uh, oh, well, okay. Cannot say I love it! But it's fine.

For the record, this is a "hidden feature" of Go: https://go.dev/tour/basics/7

lzap (approved):
Mediocre patch.

internal/dao/stubs/account_dao.go
ezr-ondrej: Agreed, I got the real store in again.

internal/dao/stubs/account_dao.go
```diff
@@ -0,0 +1,57 @@
...
+
+var SingleAccountError = errors.New("We only have Account with id = 1 orgId = 1 and accountNumber = 1 available in stub DAO")
+
+var stubAccount = models.Account{
+	ID:            1,
+	OrgID:         "1",
+	AccountNumber: pointer.String("1"),
+}
+
+type AccountDaoStub struct{}
```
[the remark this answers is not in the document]
ezr-ondrej: Yeah I do not know why we are exporting the stubs, probably bad idea :+1:

internal/dao/stubs/context_getters.go
ezr-ondrej: Got rid of it :)

internal/dao/stubs/account_dao.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/pointer"
+)
+
+var notInStoreError = errors.New("The Account you've been looking for does not exist")
```
lzap: By capitalization I mean the string, not the variable ;) I think this actually must be exported. Also maybe you can move it to some common filename like `errors.go` or something like that because these should be shared across DAOs.

```suggestion
var StubNotFoundError = errors.New("DAO record does not exist")
```

internal/dao/stubs/context_getters.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type daoStubCtxKeyType int
+
+const (
+	accountCtxKey daoStubCtxKeyType = iota
+	pubkeyCtxKey  daoStubCtxKeyType = iota
+)
+
+var contextReadError = errors.New("the message")
```
lzap, line 17: Export this, perhaps move it to `errors.go` too or just rename this file to `common.go` or whatever and put all errors here.

```suggestion
var ContextReadError = errors.New("DAO stub not found in context, use WithXyzDao() to create one")
```

internal/dao/stubs/account_dao.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/pointer"
+)
+
+var SingleAccountError = errors.New("We only have Account with id = 1 orgId = 1 and accountNumber = 1 available in stub DAO")
```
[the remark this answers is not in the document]
ezr-ondrej: Oh I needed to reread to understand, it still feels weird not to capitalize for some reason. Fixed :)

internal/dao/dao_interfaces.go
```diff
@@ -10,8 +10,8 @@ var GetAccountDao func(ctx context.Context) (AccountDao, error)
 
 type AccountDao interface {
 	GetById(ctx context.Context, id int64) (*models.Account, error)
-	GetByAccountNumber(ctx context.Context, number string) (*models.Account, error)
 	GetByOrgId(ctx context.Context, orgId string) (*models.Account, error)
+	GetByAccountNumber(ctx context.Context, number string) (*models.Account, error)
 	List(ctx context.Context, limit, offset int64) ([]*models.Account, error)
```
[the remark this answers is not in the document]
ezr-ondrej: it's just a reorder ... I feel like I'm keep forgetting wich one is the one that will stay, so I've figured let's make it more important by order xD

internal/dao/stubs/account_dao.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/pointer"
+)
+
+var notInStoreError = errors.New("the Account you've been looking for does not exist")
```
lzap, line 12: Errors must be exported otherwise they are pretty useless. I suggest you move this to `errors.go` or elsewhere, this one can be reused by many DAO stubs.

```suggestion
var RecordNotFoundError = errors.New("DAO record does not exist")
```

internal/dao/stubs/context_getters.go
```diff
@@ -0,0 +1,45 @@
...
+
+func WithPubkeyDao(parent context.Context, init_store []*models.Pubkey) context.Context {
```
lzap, line 19:
```suggestion
func WithPubkeyDao(parent context.Context, initStore []*models.Pubkey) context.Context {
```

I still think that the DAO interface should be utilized to actually hide the implementation of stubs. What if we choose to use map instead of slice? Or some kind of in memory database. If you insist exposing the slice this way tho, I am okay.

lzap (changes requested):
Few errors nitpicks, other than that good.

internal/dao/stubs/account_dao.go
ezr-ondrej: I've gone ahead and reused the `NoRowsError` to create a feel of the DAO interface... I guess we are improving :stuck_out_tongue_closed_eyes:

internal/dao/stubs/context_getters.go
ezr-ondrej: moved to `stub_errors`

internal/dao/stubs/account_dao.go
ezr-ondrej: done :)

internal/dao/stubs/account_dao.go
```diff
@@ -0,0 +1,79 @@
...
+	return &accountDaoStub{
+		store: []*models.Account{
+			&models.Account{
+				ID:            1,
+				OrgID:         "1",
+				AccountNumber: pointer.String("1"),
+			},
+		},
+	}
+}
```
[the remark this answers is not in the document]
ezr-ondrej: Done ^_^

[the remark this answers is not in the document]
ezr-ondrej: So I'm back with dynamic, just in case, unless you think it's bad :)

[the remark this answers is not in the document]
adiabramovitch: Just to make sure I understood: 
Do you that ipreferableeable to keep it dynamically because we mperformeform changes on static accounts which can lead failures in other tests? so we want our data to be revoked each time as a "clean slate"?

internal/dao/stubs/context_getters.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type daoStubCtxKeyType int
+
+const (
+	accountCtxKey daoStubCtxKeyType = iota
+	pubkeyCtxKey  daoStubCtxKeyType = iota
+)
+
+func WithPubkeyDao(parent context.Context, init_store []*models.Pubkey) context.Context {
```
adiabramovitch, line 16: I think it is supposed to be `initStore` because go supports camelCase

internal/dao/stubs/account_dao.go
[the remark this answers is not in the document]
lzap: Very good point, this must be dynamically created. Performance is not the issue here but concurrency is - we can easily run tests in parallel with Go and having this as a global variable would mean we share it - that would be bad.

internal/dao/stubs/context_getters.go
ezr-ondrej: Agreed and also commented by @lzap, but I'm moving code here, so I'd like to move as is and address this as a whole thing and completely change the implementation of Pubkey stub in a follow-up

internal/dao/stubs/account_dao.go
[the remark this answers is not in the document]
lzap: Now that I am thinking about it, it is worth enabling concurrent unit tests from the day one so if we run into concurrency issues we would solve them now rather than later with many tests.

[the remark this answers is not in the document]
ezr-ondrej: yes, exactly! :) Having always the same data at the beginning of the test is quite important for predictable test outcomes :)

adiabramovitch:
I am good here
@lzap any opinion?

lzap (approved):
No comments, thanks.

PR #92, opened 2022-07-18
Test account middleware

ezr-ondrej:
Tests for #90, but need to have #91 in first, so creating separate draft PR.

### DIFF
--- a/internal/middleware/account_test.go
+++ b/internal/middleware/account_test.go
@@ -1,0 +1,36 @@
+package middleware_test
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/dao/stubs"
+	"github.com/RHEnVision/provisioning-backend/internal/middleware"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAccountMiddleware(t *testing.T) {
+	ctx := identity.WithIdentity(t, context.Background())
+	ctx = stubs.WithAccountDaoOne(ctx)
+
+	req, err := http.NewRequestWithContext(ctx, "GET", "/api/provisioning/v1/pubkeys", nil)
+	if err != nil {
+		assert.Nil(t, err, fmt.Sprintf("Error creating a new request: %v", err))
+	}
+
+	rr := httptest.NewRecorder()
+
+	isAccInNext := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var acc = ctxval.GetAccount(r.Context())
+		assert.NotNil(t, acc, "account was not set")
+	})
+
+	handler := middleware.AccountMiddleware(isAccInNext)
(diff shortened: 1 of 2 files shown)


lzap:
LGTM

Just to let you know, there is no reason to create such a patch in a separate PR. You can simply have that with the code, or as a separate commit in the same PR. I am fine with both. But if you feel like you want to play around with branches and rebasing - go ahead :-)

ezr-ondrej:
> Just to let you know, there is no reason to create such a patch in a separate PR. You can simply have that with the code, or as a separate commit in the same PR. I am fine with both. But if you feel like you want to play around with branches and rebasing - go ahead :-)

This should ideally go with #90 together, or before even, but it would fail without #91, so I've opened a separate PR just for visibility, if we merge #91 first, I'll incorporate this into #91 :+1:

adiabramovitch:
I have merged #90, can you rebase? :)

ezr-ondrej:
> I have merged #90, can you rebase? :)

It needs #91 as well, without it tests will keep failing :)

lzap:
So #91 is in, please rebase. I am not starting any new reviews.

ezr-ondrej:
Rebased, ready for review, I've updated the OrgID and accNumber for test identity header to 1/1 so it matches the 1/1 in the stub and is the basic account we use in tests

internal/middleware/account_test.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func TestAccountMiddleware(t *testing.T) {
+	ctx := identity.WithIdentity(t, context.Background())
+	ctx = stubs.WithAccountDaoOne(ctx)
+
+	req, err := http.NewRequestWithContext(ctx, "GET", "/api/provisioning/v1/pubkeys", nil)
+	if err != nil {
+		panic(err)
+	}
```
adiabramovitch, line 23: Maybe it is better to be consistent in our error handling? 
https://github.com/RHEnVision/provisioning-backend/blob/72c72ff2e09978d6b926ae2f0916a64295fa3054/internal/services/pubkey_service_test.go#L37
Not sure what is the better option

internal/middleware/account_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	rr := httptest.NewRecorder()
+
+	isAccInNext := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var acc = ctxval.GetAccount(r.Context())
+		assert.NotNil(t, acc, "account was not set")
+	})
```
adiabramovitch, line 30: Can you pls explain why is it necessary to check `AccountMiddleware` function? 
Is it because we are expecting the account to be set here: https://github.com/ezr-ondrej/provisioning-backend/blob/5da17892cd4e228de458787bffecad7272f9684f/internal/middleware/account.go#L31 ?

ezr-ondrej: Yeah, basically :)
This test is testing whether the AccountMiddleware "had set the account into the context"

Given context is only passed to the next middleware in the stack, we create a function that is actually checking the value in the context and we add this function as a next function in the stack and thus there are two functions in the middleware stack: 1) AccountMiddleware 2) the check function

So AccountMiddleware should set account to context and send the context to the following middleware, which in here is the test function, so this fnc can verify what is actually in the context :)

adiabramovitch (approved):

PR #93, opened 2022-07-18
NoRowsError wrap the original error

ezr-ondrej:
NoRowsError shall wrap the sql.NoRowsErr so we can test for it in inheritance.

This also levels down the error level of logging this error,
it might be expected by the  user.


lzap (approved):
TY

PR #94, opened 2022-07-18
Pass errors by value copy

ezr-ondrej:
THIS IS A PROPOSAL. I do not feel this is necessary, but IMHO it reads better :)

We are passing errors as pointers, but errors shall be static structures.
No changes should be made to it post creation.

Given performance impact is very small, we can do this to assure correct use of error structures.

### DIFF
--- a/internal/dao/dao_errors.go
+++ b/internal/dao/dao_errors.go
@@ -25,22 +25,22 @@ type MismatchAffectedError struct {
 	Context context.Context
 }
 
-func (e *Error) Error() string {
+func (e Error) Error() string {
 	return fmt.Sprintf("DAO error: %s: %s", e.Message, e.Err.Error())
 }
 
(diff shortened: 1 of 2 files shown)


lzap (approved):
https://go.dev/tour/methods/8

It is something that is absolutely allowed even by Go authors and I have no problems with. I tend to use pointers in receivers because at later stage if we decide to modify the object, then debugging these kind of issues (function operates on a copy) can be tricky. 

In fact, in some cases this allows the compiler to optimize and store the values on stack instead of allocating it on heap saving some GC cycles. But in our case (and most of our app) it won't matter as we do not do heavy calculations in loops. So this is not an issue.

Feel free to rebase conflicts if you want this in.

ezr-ondrej:
Rebased :)

lzap (approved):

PR #95, opened 2022-07-19
Add context setters to wrap adding values

ezr-ondrej:
To add values to the context, we should abstract callers.
Caller does not need to know under what key the value lives.

Also stores pointer to the logger and not the logger itself.

Fixes HMSPROV-118


lzap:
Damn I actually did the same in https://github.com/RHEnVision/provisioning-backend/pull/98 plus making sure the logger is stored as a pointer (which makes more sense since we want to share this across the whole app).

lzap:
Okay my patch also removes Get prefixes. Your patch use WithXYZ instead of SetXY, I think With is nicer that is how Go team names the functions in the library. I can rename to With if you like my patch more.

ezr-ondrej:
Closing in favor of #98

PR #96, opened 2022-07-19
Use versioned endpoint for sources

ezr-ondrej:
We have forgotten a version in the config from clowder.


ezr-ondrej:
[a89dd7c](pull quay.io/envision/provisioning-backend:a89dd7c)

adiabramovitch (approved):

PR #98, opened 2022-07-20
Encapsulate context getters and setters

lzap:
SSIA

I suggest to review per commit as this is split into separate commits for better readability.


ezr-ondrej (commented):
I'm happy to close #95, but I'd use the `With` prefix if you'd agree, IMHO it is less confusing.

adiabramovitch:
> I'm happy to close #95, but I'd use the `With` prefix if you'd agree, IMHO it is less confusing.

Maybe `Add` is also an option?

lzap:
I like With as it is aligned with the Go standard library function names. I will rename. Any other comments?

lzap:
Pushed as an extra commit, please squash on merge.

ezr-ondrej (approved):
Works for me :+1:

PR #99, opened 2022-07-20
Use ptr package from AWS smithy project

lzap:
SSIA

We already have that dependency because AWS SDK use it anyway.


ezr-ondrej (approved):
Thanks @lzap ! nice change :+1:

PR #100, opened 2022-07-20
HMSPROV-135: Add Create for Account DAO

ezr-ondrej:
Fixes HMSPROV-135


internal/dao/sqlx/account_dao.go
```diff
@@ -14,6 +15,7 @@ import (
 )
 
 const (
+	createAccount             = `INSERT INTO accounts (account_number, org_id) VALUES ($1, $2) ON CONFLICT DO UPDATE RETURNING id`
```
avitova, line 18: @ lzap [asked me](https://github.com/RHEnVision/provisioning-backend/pull/84#discussion_r920834394) to drop this, but I am not sure what was the idea behind the 'DO UPDATE' in the ticket.

internal/dao/sqlx/account_dao.go
```diff
@@ -91,8 +109,11 @@ func (di *accountDaoSqlx) GetOrCreateByIdentity(ctx context.Context, orgId strin
 	} else {
 		return nil, err
 	}
-	// TODO create HMSPROV-135
-	return nil, err
+	acc = &models.Account{OrgID: orgId, AccountNumber: ptr.String(accountNumber)}
```
avitova, line 112: Is this reachable? I feel like this method is going to end after the first if-else statement.

avitova (commented):
I might be missing some conversation about this and the GetOrCreateByIdentity method, but I also feel like it does not check if there is multiple results which we talked about in the previous PR.

internal/dao/sqlx/account_dao.go
avitova: Okay, nevermind, it is reachable. But it might not check multiple results for two diffferent accounts - one matching the orgId, the other one matching account number, right?

internal/dao/sqlx/account_dao.go
ezr-ondrej: Oh yeah, that is basically wrong, I see, I'll look into that, I guess we can `DO NOTHING` for now :thinking: but erroring out and letting UI/User try again is the best way I guess.

internal/dao/sqlx/account_dao.go
ezr-ondrej: Yeah, that case is completely ignored in this implementation.
Although I feel like we should not reconcile on it, rather make sure it can never happen (by making the orgId actually unique in database)

ezr-ondrej: Also that brings up the thought of arranging this code in a way it would be clear it continues after thos if/else statements, I'll try to [keep line of sight](https://github.com/RHEnVision/provisioning-backend/pull/90#discussion_r923312186) :)

ezr-ondrej: I've improved the line of sight to make it bit more readable

internal/dao/sqlx/account_dao.go
ezr-ondrej: I just left it out, so we error out and let the User try again.

ezr-ondrej:
@avitova @adiabramovitch I've rebased and tested what happens on duplicate account, I've went with erroring out and letting User try again, given this is usually just request done too fast, I think it is actually the best approach here.

Tho once we have dao test suits, I'd love to have tests for that, until then I guess this is best effort :)

adiabramovitch:
except for my question LGTM 😃

internal/middleware/account_test.go
```diff
@@ -11,26 +11,50 @@ import (
...
 	"github.com/RHEnVision/provisioning-backend/internal/middleware"
 	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
+	"github.com/aws/smithy-go/ptr"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestAccountMiddleware(t *testing.T) {
-	ctx := identity.WithIdentity(t, context.Background())
-	ctx = stubs.WithAccountDaoOne(ctx)
+	t.Run("existing account", func(t *testing.T) {
```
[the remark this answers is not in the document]
ezr-ondrej: so I am bit out of good docs on it, but basically it is a grouping mechanism, it is quite simmilar to something like `TestAccountMiddlewareWithExistingAccount` +  `TestAccountMiddlewareWithNewAccount`, but this subgrouping is quite useful IMHO, it helps if you want to share setup for example, but mainly makes it bit more readable, one Test method for single go method, with multiple sub-test by `Run` ... if we keep this approach I think it will be easier to lookup the appropriate tests in the future.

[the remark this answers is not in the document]
adiabramovitch: So each Run method is a separate test? 
 Can we share the setup between a few Runs? 
 that is possible when the setup is out of the function scope?  (function with` func(t *testing.T)` signature)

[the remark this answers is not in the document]
ezr-ondrej: it is certainly possible, but need to be approached with care, because only very static data should be shared. If you share dynamic data, tests could start interfering one with another, and that's quite bad ^_^
So they can share data, but should do with care. Grouping is much better usecase for this.

adiabramovitch (approved):

avitova (approved):

ezr-ondrej:
Nice ! :tada:

PR #102, opened 2022-07-20
Improve pubkey DAO stub

ezr-ondrej:
Implements all DAO functions properly.
Implements pubkey generation.
Hides out the DAO stub structure, so it can't be manipulated from outside.

### DIFF
--- a/internal/dao/stubs/pubkey_dao.go
+++ b/internal/dao/stubs/pubkey_dao.go
@@ -28,17 +28,42 @@ func getPubkeyDao(ctx context.Context) (dao.PubkeyDao, error) {
 	return getPubkeyDaoStub(ctx)
 }
 
-func (mock *PubkeyDaoStub) Create(ctx context.Context, pk *models.Pubkey) error {
-	pk.ID = mock.lastId + 1
-	mock.store = append(mock.store, pk)
-	mock.lastId++
+func (stub *PubkeyDaoStub) Create(ctx context.Context, pk *models.Pubkey) error {
+	pk.ID = stub.lastId + 1
+	stub.store = append(stub.store, pk)
+	stub.lastId++
 	return nil
 }
-func (*PubkeyDaoStub) Update(ctx context.Context, pk *models.Pubkey) error { return nil }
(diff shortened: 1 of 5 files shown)


internal/dao/stubs/pubkey_dao.go
```diff
@@ -27,15 +28,42 @@ func getPubkeyDao(ctx context.Context) (dao.PubkeyDao, error) {
 	return getPubkeyDaoStub(ctx)
 }
 
-func (mock *PubkeyDaoStub) Create(ctx context.Context, pk *models.Pubkey) error {
-	mock.store = append(mock.store, pk)
+func (stub *pubkeyDaoStub) Create(ctx context.Context, pk *models.Pubkey) error {
+	pk.ID = stub.lastId + 1
```
adiabramovitch, line 32: Why are we changing the pk id and didn't we do that before?
I guess we want the pk ids to match our store order?

internal/services/pubkey_service_test.go
```diff
@@ -9,29 +9,19 @@ import (
...
-	}}
-}
-
 func TestListPubkeysHandler(t *testing.T) {
 	ctx := identity.WithIdentity(t, context.Background())
-	ctx = dao_stubs.WithPubkeyDao(ctx, buildPkStore())
+	ctx = stubs.WithPubkeyDao(ctx)
+	_, err := stubs.GeneratePubkey(ctx, nil)
+	assert.Nil(t, err, fmt.Sprintf("Error while generating a pubkey: %v", err))
+	_, err = stubs.GeneratePubkey(ctx, nil)
```
adiabramovitch, line 23: Why have you chosen to send nil and create pubkey this way?

ezr-ondrej: On the other hand not allowing the shorthand would simplify the return to just error, which would be neat :thinking:

adiabramovitch: I think it is cleaner to pass an "empty" pub key

ezr-ondrej: oki :+1: :)

adiabramovitch (approved):

PR #103, opened 2022-07-21
Fixes UpdatePubkey

avitova:
I've run into this while testing.


adiabramovitch (approved):

ezr-ondrej:
Thanks ! I think that's exatly the reason why we need DAO tests :stuck_out_tongue_closed_eyes:

PR #104, opened 2022-07-22
Integration Test for Pubkeys DAO

avitova:
I have few thoughts on this:
1) Is it safe to check with values from seed files? Are we not going to change the values in the process of development?
2) Should we come up with more detailed errors for failed tests?

### DIFF
--- a/cmd/dao_tests/pubkey_test.go
+++ b/cmd/dao_tests/pubkey_test.go
@@ -5,16 +5,37 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"testing"
 
+	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
-	"github.com/stretchr/testify/assert"
-
-	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/db"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
+	"github.com/stretchr/testify/assert"
 )
+
+func createPk() *models.Pubkey {
+	return &models.Pubkey{
+		ID:        1,
+		AccountID: 1,
+		Name:      "lzap-ed25519-2021",
+		Body:      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap",
+	}
+}
+
+func InitTestEnvironment() error {
+	config.Initialize()
+
+	err := db.Initialize()
+	if err != nil {
+		panic(fmt.Errorf("database setup had failed: %v", err))
+	}
+	return nil
+}
 
 func Setup(t *testing.T, s string) (dao.PubkeyDao, context.Context, error) {
 	err := db.Seed("dao_test")
@@ -22,7 +43,7 @@ func Setup(t *testing.T, s string) (dao.PubkeyDao, context.Context, error) {
 		t.Errorf("Error purging the database: %v", err)
 		return nil, nil, err
 	}
-	ctx := context.Background()
+	ctx := identity.WithTenant(t, context.Background())
 	pkDao, err := dao.GetPubkeyDao(ctx)
 	if err != nil {
 		t.Errorf("%s test had failed: %v", s, err)
@@ -32,15 +53,7 @@ func Setup(t *testing.T, s string) (dao.PubkeyDao, context.Context, error) {
 }
 
 func CleanUpDatabase(t *testing.T) {
-	config.Initialize()
-
-	err := db.Initialize()
-	if err != nil {
-		t.Errorf("Error initializing database: %v", err)
-		return
-	}
-
-	err = db.Seed("drop_all")
+	err := db.Seed("drop_all")
 	if err != nil {
 		t.Errorf("Error purging the database: %v", err)
 		return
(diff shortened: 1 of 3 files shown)


ezr-ondrej:
> 1. Is it safe to check with values from seed files? Are we not going to change the values in the process of development?

Seed values in tests (or fixtures) are meant to set a default state for tests, those should be kept as minimal as possible, to reduce boilerplate with creating the data that every (or most) of the tests use.
So that means if we follow these rules and put only really necessary fixtures in, those should stay quite static and almost never change.

 

> 2\. Should we come up with more detailed errors for failed tests?

Error messages (any) needs to be as short as possible to give the target reader enough information what had failed.
So no more details then necessary, these tests are targeted at developers, if there is message that can guide developers to the right conclusion, its fine :+1:

internal/db/seeds/dao_pubkey.sql
```diff
@@ -0,0 +1,10 @@
+--
+-- Testing seed data, to execute this file during migration set DB_SEED variable to "dao_pubkey".
+--
+BEGIN;
+
+INSERT INTO pubkeys(id, account_id, name, body)
+VALUES (1, 1, 'firstkey', 'sha-rsa body')
```
ezr-ondrej, line 7: I'd try to use a valid public key here.
I'm not entirely against the idea of pubkey in the fixtures, but it is a bit of a stretch already (as not all the tests will need to have it in the DB)

cmd/dao_tests/pubkey.go
```diff
@@ -0,0 +1,208 @@
...
+
+	pubkeysAfter, err := pkDao.List(ctx, 10, 0)
+	if err != nil {
+		t.Errorf("Delete pubkey test had failed")
+		return
+	}
+	assert.Equal(t, len(pubkeys)-1, len(pubkeysAfter), "Delete Pubkey error.")
+}
+
+func main() {
```
ezr-ondrej, line 196: Why haven't we used https://pkg.go.dev/testing@go1.18.4#hdr-Main I think then we would need to run `go test cmd/dao_tests` but not sure, did not test, but according to the docs, it is the function to be used for these cases, so I wonder why we have chosen to use internals of the testing library ([the docs for Main](https://pkg.go.dev/testing@go1.18.4#Main) say it stays exported as best effort, but no promises are made to keep it)

cmd/dao_tests/pubkey.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	err = db.Migrate()
+	if err != nil {
+		t.Errorf("Error running migration")
+		return
+	}
+}
+
+func CreatePubkeyTest(t *testing.T) {
```
ezr-ondrej, line 56: Go convention is to name Test functions with Test prefix, not suffix, is there a reason we are using suffix instead here?

cmd/dao_tests/pubkey.go
```diff
@@ -0,0 +1,208 @@
...
+func updatePk() []*models.Pubkey {
+	return []*models.Pubkey{{
+		ID:        1,
+		AccountID: 1,
+		Name:      "updated-key",
+		Body:      "updated body",
+	}}
+}
+
+func TearDown(t *testing.T) {
```
ezr-ondrej, line 34: We call this TearDown at the beggining of every test, I'd love if we could use some wrapper for all tests that would make sure TearDown happens after the run of every test and Setup is called prior the run of every test.

cmd/dao_tests/pubkey.go
```diff
@@ -0,0 +1,208 @@
...
+func createPk() []*models.Pubkey {
+	return []*models.Pubkey{{
+		ID:        10,
+		AccountID: 2,
+		Name:      "testkey",
+		Body:      "sha-rsa body",
+	}}
+}
+
+func updatePk() []*models.Pubkey {
```
ezr-ondrej, line 25: same as `createPk()`

cmd/dao_tests/pubkey.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/db"
+)
+
+func createPk() []*models.Pubkey {
```
ezr-ondrej, line 16: This is used in a single test, having it a function seems not necessary and might confuse ppl who are adding tests, that they are supposed to use it, is there a reason this is not a local variable in the test that needs it?

cmd/dao_tests/pubkey.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		t.Errorf("Update pubkey test had failed. %s", err)
+		return
+	}
+	assert.Equal(t, updatePk()[0].Name, pubkeys[0].Name, "Update pubkey test had failed.")
+}
+
+func GetPubkeyByIdTest(t *testing.T) {
+	TearDown(t)
+	err := db.Seed("dev_small")
```
ezr-ondrej, line 138: I do not think we want to seed dev_small in tests, that is subject to a change, I thought that we were talking about test fixtures in description, but this seed data can change quite frequently and should not be used in tests.

ezr-ondrej (commented):
Hi! good start most certainly! It looks quite nice and does exactly what it is supposed to do, but I have few worries, sorry for rasing some concerns that might have came up from initial design that I was not part of, but if it is the case, could we share some design decisions in the PR descriptions, why we have chosen given solution.
I'm also sorry for nitpicking, but I believe this one will be copied over, so we should choose the best solutions we can, so we copy the right thing :)

Note for the seeds: I see you're adding two seeds, but none of them is used, and we are using dev_small instead, I'd not do that. Ideally we should have one seed data for tests, that will serve as fixture set for all the tests IMHO

cmd/dao_tests/pubkey_test.go
```diff
@@ -10,19 +10,30 @@ import (
...
 	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/db"
 )
 
+func createPk() []*models.Pubkey {
```
ezr-ondrej, line 21: Why are we returning an array and then dereferencing it in every use?
Wouldn't it be easire to just return the pubkey?

internal/db/seeds/dao_test.sql
```diff
@@ -4,7 +4,7 @@
 BEGIN;
 
 INSERT INTO accounts(id, account_number, org_id)
-VALUES (1, '13', '000013')
+VALUES (1, '13', '1')
```
ezr-ondrej, line 7: Although WithTenant does not fail on this today, it should in the future.
We have DefaultOrgId = "1" and default_account_number = "1", now we are doing it wrong and looking just by orgId, but hopefully we will improve on it and then test would start to fail :)

```suggestion
VALUES (1, '1', '1')
```

cmd/dao_tests/pubkey_test.go
```diff
@@ -63,6 +74,124 @@ func TestSimple(t *testing.T) {
...
+	pubkeysAfter, err := pkDao.List(ctx, 10, 0)
+	if err != nil {
+		t.Errorf("Delete pubkey test had failed")
+		return
+	}
+	assert.Equal(t, len(pubkeys)-1, len(pubkeysAfter), "Delete Pubkey error.")
+}
+
 func TestMain(t *testing.M) {
 	exitVal := t.Run()
```
ezr-ondrej, line 196: Can we extract `config.Initialize()` and `db.Initialize` from the CleanUpDatabase method into `InitTestEnvironment()` method and run it only once here, in the TestMain?

ezr-ondrej (commented):
Great progress! I found few nitpicks, but overall it looks great!

ezr-ondrej (commented):
One nitpick on the panic, and I also wonder if we should remove the `TestSimle` now, that we have more meaningful tests, it doesn't make much sense to keep it around IMHO :)

cmd/dao_tests/pubkey_test.go
```diff
@@ -63,7 +75,126 @@ func TestSimple(t *testing.T) {
...
+		return
+	}
+	err = pkDao.Create(ctx, createPk())
+	if err != nil {
+		t.Errorf("Delete pubkey test had failed. %s", err)
+		return
+	}
+	pubkey, err := pkDao.GetById(ctx, 1)
+	if err != nil {
+		t.Errorf("Get pubkey test had failed.")
```
ezr-ondrej, line 158:
```suggestion
		t.Errorf("Get pubkey test had failed: %v", err)
```

cmd/dao_tests/pubkey_test.go
```diff
@@ -63,7 +75,126 @@ func TestSimple(t *testing.T) {
...
+	assert.Equal(t, "lzap-ed25519-2021", pubkey.Name, "Get Pubkey error: pubkey name does not match.")
+	assert.Equal(t, "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap", pubkey.Body, "Get Pubkey error: pubkey body does not match.")
+
+}
+
+func TestDeletePubkeyById(t *testing.T) {
+	CleanUpDatabase(t)
+	pkDao, ctx, err := Setup(t, "Delete pubkey")
+	if err != nil {
+		t.Errorf("Database setup had failed")
```
ezr-ondrej, line 170:
```suggestion
		t.Errorf("Database setup had failed: %v", err)
```

cmd/dao_tests/pubkey_test.go
```diff
@@ -63,7 +75,126 @@ func TestSimple(t *testing.T) {
...
+		return
+	}
+	err = pkDao.Create(ctx, createPk())
+	if err != nil {
+		t.Errorf("Delete pubkey test had failed. %s", err)
```
ezr-ondrej, line 175:
```suggestion
		t.Errorf("Delete pubkey test had failed: %v", err)
```

cmd/dao_tests/pubkey_test.go
```diff
@@ -63,7 +75,126 @@ func TestSimple(t *testing.T) {
...
+	assert.Equal(t, "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap", pubkey.Body, "Get Pubkey error: pubkey body does not match.")
+
+}
+
+func TestDeletePubkeyById(t *testing.T) {
+	CleanUpDatabase(t)
+	pkDao, ctx, err := Setup(t, "Delete pubkey")
+	if err != nil {
+		t.Errorf("Database setup had failed")
+		return
```
ezr-ondrej, line 171: I'm not sure, but shouldn't `t.Errorf` stop the test execution? Do we need to do it manually by `return`?

ezr-ondrej (commented):
Eh, few more formatting nitpicks, but those are really minor

ezr-ondrej (approved):
I'm good here, we can improve formatting in follow-ups, once we figure out how to properly log failures and maybe make the logger bit less verbose :)

ezr-ondrej:
Awesome addition @avitova ! Thanks, that's a big leap forward for the project :+1:

avitova:
Great, thanks! :fireworks:

PR #105, opened 2022-07-22
HMSPROV-157: Guard Pubkey DAO by tenant

ezr-ondrej:
Fixes HMSPROV-157

TODO: do the same for stubbed dao, once #102 is in.

### DIFF
--- a/internal/dao/stubs/pubkey_dao.go
+++ b/internal/dao/stubs/pubkey_dao.go
@@ -28,17 +28,31 @@ func getPubkeyDao(ctx context.Context) (dao.PubkeyDao, error) {
 	return getPubkeyDaoStub(ctx)
 }
 
-func (stub *pubkeyDaoStub) Create(ctx context.Context, pk *models.Pubkey) error {
-	pk.ID = stub.lastId + 1
-	stub.store = append(stub.store, pk)
+func (stub *pubkeyDaoStub) Create(ctx context.Context, pubkey *models.Pubkey) error {
+	if pubkey.AccountID == 0 {
+		pubkey.AccountID = ctxAccountId(ctx)
+	}
+	if pubkey.AccountID != ctxAccountId(ctx) {
+		return dao.WrongTenantError
+	}
+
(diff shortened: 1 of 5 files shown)


adiabramovitch:
looks great!
Do you add the same for the stubbed dao in this PR? 
I merged #102

ezr-ondrej:
> Do you add the same for the stubbed dao in this PR?

Done :)

ezr-ondrej:
I'm hitting an issue in tests, where I do not have account in context, which should have been solved when I've been working on the account, but it was not, so #108

ezr-ondrej:
The test failure brings up interesting points, that I want to address:

- Should we filter out the accountID from the request? I'd propose yes in this PR
- Should we fill the nil AccountID on creation by the current tenant? I'll also propose yes

Converting to draft until those are addressed.

internal/dao/stubs/pubkey_dao.go
```diff
@@ -28,39 +28,54 @@ func getPubkeyDao(ctx context.Context) (dao.PubkeyDao, error) {
...
+	for _, pk := range stub.store {
+		if pk.AccountID == ctxAccountId(ctx) && pk.ID == id {
+			return pk, nil
 		}
 	}
 	return nil, NewRecordNotFoundError(ctx, "Pubkey")
 }
 
 func (stub *pubkeyDaoStub) List(ctx context.Context, limit, offset int64) ([]*models.Pubkey, error) {
+	// not tenant filtered, should it?
```
adiabramovitch, line 72: I guess it should but to which account id do we compare?

adiabramovitch (approved):
except for the question I dropped, LGTM :)

internal/dao/stubs/pubkey_dao.go
ezr-ondrej: I'll add it, it is just me being lazy... lets not be lazy :stuck_out_tongue:

ezr-ondrej: Alright, done :+1:

PR #106, opened 2022-07-22
HMSPROV-158: Add validations to pubkeys

ezr-ondrej:
Adds general validations for models.
Adds pubkeys validations on Body and Name.

TODO:

- [x] Add the same validations to DAO, once it is in


internal/models/validator.go
```diff
@@ -0,0 +1,23 @@
...
+	validate = validator.New()
+
+	_ = validate.RegisterValidation("sshPubkey", func(fl validator.FieldLevel) bool {
+		_, err := ssh.ParsePublicKey(fl.Field().Bytes())
+		return err == nil
+	})
+}
+
+func Validate(ctx context.Context, model interface{}) error {
+	return validate.StructCtx(ctx, model)
```
adiabramovitch, line 22: I see that we are validating a model using this method. 
According to the documentation, we are validating the exposed fields and by passing the ctx contextual validation information happens. but against what do the validations occur?

ezr-ondrej: I've added the `validate` tags, on the model structure fields, so there we define what rule should be used to validate given field, the validator library automatically extracts that tag and uses those rules. I dig its golang native feel, but there are libraries that allow more explicit validations and I'm happy to talk about those if we do not like this :)

adiabramovitch: My question is, what rules do the tags validate?

ezr-ondrej: This PR is adding `sshPubkey` for which you can read the definition in the method, rest is defined in the library here: https://github.com/go-playground/validator#baked-in-validations
I'm using just required in this PR, but we can obviously use any validation in the future :)

ezr-ondrej: yup, you are just making the list of validations in this manner: https://github.com/RHEnVision/provisioning-backend/pull/106/files#diff-645fe10c68e235c13074b5db21d4033bbf06828ebd1fa00ad5fb1c336f52e24fR15

adiabramovitch, line 22: Should we use errors.new or create a type implementing Error interface?

ezr-ondrej: the validate.StructCtx alreay returns type that impements error interface, with some additional goodies, but it already is an error, we could state clearly what we return here tho, as we know it won't be a generic error

internal/dao/stubs/stub_errors.go
```diff
@@ -25,3 +28,16 @@ func NewCreateError(ctx context.Context, stubName string) dao.Error {
...
+	for _, ve := range validationErr {
+		errors = append(errors, ve.Error())
+	}
+	msg := strings.Join(errors, ", ")
+
+	if logger := ctxval.Logger(ctx); logger != nil {
+		logger.Info().Msg(msg)
+	}
+	return dao.ValidationError{Context: ctx, Message: msg, Err: validationErr, Model: model}
+}
```
adiabramovitch, line 43: I see it is identical to the function in sqlx_errors.go. 
I guess you prefer to repeat code while having a logical separation?

adiabramovitch (approved):
just a small question, other than that I'm good :)

internal/dao/stubs/stub_errors.go
ezr-ondrej: yeah, I guess I was lazy, because this gets the stubName as a string, but we can improve that and make the Daos adhere the name for errors interface :)

ezr-ondrej: Done :+1:

internal/models/validator.go
```diff
@@ -0,0 +1,31 @@
...
+		return err == nil
+	})
+}
+
+func Validate(ctx context.Context, model interface{}) validator.ValidationErrors {
+	err := validate.StructCtx(ctx, model)
+	var validationError validator.ValidationErrors
+	if err != nil && !errors.As(err, &validationError) {
+		// dependency cycle
+		//ctxval.Logger(ctx).Fatal().Msgf("Invalid model passed for validation: %v", model)
```
adiabramovitch, line 27: Don't you want to use the logger here?

internal/dao/stubs/stub_errors.go
```diff
@@ -12,16 +12,16 @@ var ContextReadError = errors.New("missing variable in context")
...
+func NewRecordNotFoundError(ctx context.Context, stubName dao.NamedForError) dao.NoRowsError {
 	return dao.NoRowsError{
-		Message: fmt.Sprintf("%s DAO record does not exist", stubName),
+		Message: fmt.Sprintf("%s DAO record does not exist", stubName.NameForError()),
 		Context: ctx,
 	}
 }
 
-func NewCreateError(ctx context.Context, stubName string) dao.Error {
+func NewCreateError(ctx context.Context, stubName dao.NamedForError) dao.Error {
```
adiabramovitch, line 22: Why did you decide to do that change? does that make sense to pass the type? why not pass directly the result NameForError()?

internal/models/validator.go
ezr-ondrej: I'd love to, but we've got a depency cycle if we do, I'd like to rething how we approach `ctxval` package, but as of now we can't use logger in models, because ctxval package references with `Account`, which lives in models package and thus dependency cycle.

internal/dao/stubs/stub_errors.go
ezr-ondrej: It is bit cleaner as if we decide to change the NamedForError to something else and provide more context, it will be easier to change. But I did not decide to do that, I've just went with what we have in `dao/sqlx/sqlx_errors`

PR #107, opened 2022-07-24
HMSPROV-154: Add data for aws launch instance

adiabramovitch:
This is a DRAFT 😄 
seems like the reservation was created correctly
image
image

### DIFF
--- a/internal/db/migrations/005_reservations.sql
+++ b/internal/db/migrations/005_reservations.sql
@@ -16,6 +16,10 @@ CREATE TABLE aws_reservation_details
   reservation_id BIGINT NOT NULL,
   provider INTEGER NOT NULL DEFAULT provider_type_aws(),
   pubkey_id BIGINT REFERENCES pubkeys(id),
+  source_id INTEGER NOT NULL,
+  instance_type TEXT NOT NULL,
+  amount INTEGER NOT NULL,
+  image_id INTEGER NOT NULL,
 
   FOREIGN KEY (reservation_id, provider) REFERENCES reservations(id, provider) ON DELETE CASCADE,
   CHECK (provider = provider_type_aws())
--- a/internal/payloads/reservation_payload.go
+++ b/internal/payloads/reservation_payload.go
@@ -28,6 +28,8 @@ type AWSReservationRequest struct {
 		NewName    *string `json:"new_name"`
 		NewBody    *string `json:"new_body"`
 	}
+	// We have pubkey here as a struct and in the AWSReservation
+	*models.AWSReservation
 }
 
 type AWSReservationResponse struct {
@@ -63,6 +65,7 @@ func NewReservationResponse(reservation *models.Reservation) render.Renderer {
 }
 
 func NewAWSReservationResponse(reservation *models.AWSReservation) render.Renderer {
+	// TODO: this needes to be changed to Reservation?
 	return &AWSReservationResponse{AWSReservation: reservation}
 }
 
--- a/internal/dao/sqlx/reservation_dao.go
+++ b/internal/dao/sqlx/reservation_dao.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	createReservation       = `INSERT INTO reservations (provider, account_id, status) VALUES ($1, $2, $3) RETURNING *`
-	createAwsDetail         = `INSERT INTO aws_reservation_details (reservation_id, pubkey_id) VALUES ($1, $2)`
+	createAwsDetail         = `INSERT INTO aws_reservation_details (reservation_id, pubkey_id, source_id, instance_type, amount, image_id) VALUES ($1, $2, $3, $4, $5, $6)`
 	updateReservationStatus = `UPDATE reservations SET status = $2 WHERE id = $1 RETURNING *`
 	finishReservationStatus = `UPDATE reservations SET status = $2, success = $3, finished_at = now() WHERE id = $1 RETURNING *`
 	deleteReservationById   = `DELETE FROM reservations WHERE id = $1`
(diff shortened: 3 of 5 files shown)


internal/db/migrations/005_reservations.sql
```diff
@@ -16,6 +16,10 @@ CREATE TABLE aws_reservation_details
   reservation_id BIGINT NOT NULL,
   provider INTEGER NOT NULL DEFAULT provider_type_aws(),
   pubkey_id BIGINT REFERENCES pubkeys(id),
+  source_id INTEGER NOT NULL,
+  instance_type TEXT NOT NULL,
+  amount INTEGER NOT NULL,
+
```
adiabramovitch, line 22: Not sure what attributes we should add to those columns
Does all that info required to be saved in the reservation model?
If so, is it because we want a reservation to be saved and has all the info required to start a job?

internal/payloads/reservation_payload.go
```diff
@@ -28,6 +28,11 @@ type AWSReservationRequest struct {
 		NewName    *string `json:"new_name"`
 		NewBody    *string `json:"new_body"`
 	}
+	LaunchInstance struct {
```
ezr-ondrej, line 31: Do we need this substructure? I think `AWSReservation` = Reserve an instance ( I agree that the choice of the name is not great, but `launch` and reserve sound kinda same, so I'm not sure if we need nested structure for it.

internal/db/migrations/005_reservations.sql
ezr-ondrej: I don't know what attributes are we talking about, but I believe the datatypes seems correct and I don't think it needs more :)
And yes, that is exactly it, we are passing these to the goroutine that executes the actual job and also it might be useful if it fails, because we can easily re-run if required, so it is good to keep the data about the lunch in the DB for a while

internal/payloads/reservation_payload.go
adiabramovitch: I am not sure it is for launch/reserve an instance, do we need the pubkey for launching an instance? 
Why did lzap add the `Pubkey` struct there? 
I see that he used  `AWSReservation` for the response and not the request 
```golang
type AWSReservationResponse struct {
	*models.AWSReservation
}

```

internal/db/migrations/005_reservations.sql
adiabramovitch: I mean does "NOT NULL" sufficient here or there are more to be added in your opinion?

internal/payloads/reservation_payload.go
ezr-ondrej: oh yeah, that is probably wrong, I guess it should have been for a Request.

the Pubkey struct wraps interchangeble information of how to get `pubkey_id`, I feel this is bit different case :thinking:

internal/db/migrations/005_reservations.sql
ezr-ondrej: NOT NULL sounds good enough :)

internal/db/migrations/005_reservations.sql
```diff
@@ -16,6 +16,10 @@ CREATE TABLE aws_reservation_details
   reservation_id BIGINT NOT NULL,
   provider INTEGER NOT NULL DEFAULT provider_type_aws(),
   pubkey_id BIGINT REFERENCES pubkeys(id),
+  source_id INTEGER NOT NULL,
+  instance_type TEXT NOT NULL,
+  amount INTEGER NOT NULL,
```
ezr-ondrej, line 21: I'm missing the AMI/image_id information we've talked about today! :thinking:
Should we add one or the other? :)

ezr-ondrej (changes requested):
I think we need an image identification :thinking:

internal/db/migrations/005_reservations.sql
adiabramovitch: yes, you are right. 
I am just wondering which info is supposed to be saved for a reservation and which for a job 
because in order to perform a job I don't need the source id so it is not needed there, but for a reservation do we want to save all relevant info?

internal/payloads/reservation_payload.go
adiabramovitch:
```golang
type AWSReservationRequest struct {
	AWSReservation AWSReservation
    Pubkey struct {
		    ExistingID *int64  `json:"existing_id"`
		    NewName    *string `json:"new_name"`
		    NewBody    *string `json:"new_body"`
	    }
}
```
This is what you think?

internal/db/migrations/005_reservations.sql
ezr-ondrej: Yeah, I'd save all relevant information :thinking:
I think we will make use of the source_id in the future for sure, we need to assume role for actually running the instance on AWS right?

internal/payloads/reservation_payload.go
ezr-ondrej: doesn't AWSReservation contain the Pubkey already, I think if we embedd AWSReservation, we can get rid of the Pubkey :)

internal/db/migrations/005_reservations.sql
adiabramovitch: good point, maybe in order to run ec2.RunInstance I would need to assume role again :D

internal/payloads/reservation_payload.go
adiabramovitch: I would imagine not having pubkey in AWSReservation because it is an attribute specific to one job. 
and then it makes sense to use my suggestion, but maybe I am missing something? 
why we want pubkey in AWSReservation?
why not create a type per job?

ezr-ondrej (approved):

internal/payloads/reservation_payload.go
```diff
@@ -28,6 +28,8 @@ type AWSReservationRequest struct {
 		NewName    *string `json:"new_name"`
 		NewBody    *string `json:"new_body"`
 	}
+	// We have pubkey here as a struct and in the AWSReservation
+	*models.AWSReservation
```
lzap, line 32: This will expose the whole reservation which is fine, but we need to ensure that users are unable to set things like created_at, state or finished_at. Since the Create DAO method does not allow setting them explicitly, it is probably fine since there is no update endpoint, but I think it is worth documenting this in the OpenAPI (that some fields must not be set).

PR #108, opened 2022-07-25
Add Account to test context

ezr-ondrej:
Adds helper to put the correct Tenant info in the context in one method.
Improves the tests initialization a bit.
Utilizes the _test package for test izolation
Exports default testing tenant ids.

Just a simple improvement, to make our lives with tests easier


internal/services/pubkey_service_test.go
```diff
@@ -1,4 +1,4 @@
-package services
+package services_test
```
adiabramovitch, line 1: How does that work? don't you need to change all of the test files to be in `services_test` package?

ezr-ondrej: you do not need to, but it should be the goal.
The main gist of this package is go allows these to exists, within the same directory (would not be the case for any other name diffrent from the dir name), and they behave like a different package, so you actually need to test only the exported interfaces (which is good practice as you do not test internal functions of packages).

but `go test` reads `_test.go` files, so both works at the same time, thus it is not as important

That being said, the other test in services is testing internal functions, so that would require a rewrite, which would bloat a scope of this PR.
But ideal end goal is to put all tests in `*_test` packages, yes, tho not necessary to do for all at the same time.

ezr-ondrej: tho if you'd prefer to do it at the same time, I can drop it from this PR and we can do it properly in a follow-up

adiabramovitch: DO you mean that the sources_service_test can be just changed because we are checking filterSourceAuthentications fiction which is an internal one? 

if so, it can be refactored in a different PR and I would change this change.

ezr-ondrej: Yeah, exactly we are testing internal function there.
ok, agreed, I'll leave it for another PR, there is no rush with that, I just want to push in that direction, but I can create a ticket :)

adiabramovitch: just for the sources_service_test, you can leave the change here

ezr-ondrej: And the change is gone :+1: I've created HMSPROV-163 to track the `_test` package conversion

adiabramovitch (approved):

adiabramovitch:
@ezr-ondrej Thanks :)

PR #109, opened 2022-07-25
Db tests simple

avitova:

**Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,3 +102,9 @@ update-deps:
 
 .PHONY: update-test-deps
 update-test-deps: update-deps test
+
+.PHONY: integration-test
+integration-test:
+	go test -c -tags=integration -o dao_tests.test ./cmd/dao_tests
+	./dao_tests.test
+
--- a/cmd/dao_tests/pubkey_test.go
+++ b/cmd/dao_tests/pubkey_test.go
@@ -1,0 +1,69 @@
+//go:build integration
+// +build integration
+
+package main
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/db"
+)
+
+func Setup(t *testing.T, s string) (dao.PubkeyDao, context.Context, error) {
+	err := db.Seed("dao_test")
+	if err != nil {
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -102,3 +102,8 @@ update-deps:
 
 .PHONY: update-test-deps
 update-test-deps: update-deps test
+
+.PHONY: integration-test
+integration-test:
+	go build ./cmd/dao_tests/pubkey_test.go
```
ezr-ondrej, line 108: I think I've misslead you, the proper line to take example from IB was the next line:
```suggestion
	go test -c -tags=integration -o dao_tests.test ./cmd/dao_tests
```

cmd/dao_tests/pubkey_test.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	return pkDao, ctx, nil
+}
+
+func TearDown() {
+	config.Initialize()
+
+	err := db.Initialize()
+	if err != nil {
+		fmt.Errorf("Error initializing database: %v", err)
```
ezr-ondrej, line 40: fmt.Errorf only creates the error, but does not do anything with it, should we do t.Errorf instead?
Or if we move this to TestMain, then we can keep this, but panic(fmt....)

cmd/dao_tests/pubkey_test.go
```diff
@@ -0,0 +1,66 @@
...
+		return
+	}
+
+	err = db.Seed("drop_all")
+	if err != nil {
+		fmt.Errorf("Error purging the database: %v", err)
```
ezr-ondrej, line 46: This should be `t.Errorf` IMHO.

ezr-ondrej (commented):
Few improvement ideas, let me know what you think :)

cmd/dao_tests/pubkey_test.go
```diff
@@ -0,0 +1,66 @@
...
+		return
+	}
+}
+
+func TestSimple(t *testing.T) {
+	assert.Nil(t, nil)
+}
+
+func TestMain(t *testing.M) {
+	TearDown()
```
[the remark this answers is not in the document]
ezr-ondrej: oh yeah, probably not, Setup we should call at the beggining of the test, right?

cmd/dao_tests/pubkey_test.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func CleanUpDatabase(t *testing.T) {
+	config.Initialize()
+
+	err := db.Initialize()
+	if err != nil {
+		t.Errorf("Error initializing database: %v", err)
+		return
+	}
```
ezr-ondrej, line 41: I'd still like to do these just once, but we can do this in follow-ups :+1:

PR #110, opened 2022-07-25
HMSPROV-115: Update makefile, add pr_checks and build_deploy

mshriver:
These will be used by appsre CI for building the main branch image and
pushing to quay.io/cloudservices, and for executing PR checks against
all PRs opened in this repo.

pr_checks will call CI scripts from a shared repo to build an image with
the PR content and run iqe tests against it

build_deploy will execute against the main branch after PR merges.

### Testing

Executed build_deploy locally, producing and pushing this image:
https://quay.io/repository/mshriver/provisioning-backend?tab=tags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
 TEST_TAGS?=test
+CONTAINER_BUILD_OPTS ?= --build-arg=quay_expiration=2d
+CONTAINER_IMAGE ?= provisioning-backend
 
 .PHONY: build
 build: build-pbapi build-pbmigrate
@@ -23,7 +25,7 @@ clean:
 build-podman:
 	# remote podman build has problem with -f option, so we link the file as workaround
 	ln -f build/Dockerfile Containerfile
-	podman build --build-arg quay_expiration=2d -t provisioning-backend .
+	podman build $(CONTAINER_BUILD_OPTS) -t $(CONTAINER_IMAGE) .
 
 .PHONY: prep
 prep:
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+set -e
+
+IMAGE="${IMAGE:-quay.io/cloudservices/provisioning-backend}"
+IMAGE_TAG="$(git rev-parse --short=7 HEAD)"
+SMOKE_TEST_TAG="latest"
+
+if [[ -z "$QUAY_USER" || -z "$QUAY_TOKEN" ]]; then
+    echo "QUAY_USER and QUAY_TOKEN must be set"
+    exit 1
+fi
+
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -19,11 +19,15 @@ strip: build
 clean:
 	rm pbapi pbmigrate
 
+
+DOCKER_BUILD_OPTS ?= "--build-arg quay_expiration=2d"
+DOCKER_IMAGE ?= "provisioning-backend"
```
ezr-ondrej, line 24: Could we move these up the file with a comment for what they are used?
Lukas wants to split up the make file a bit in the future (in the manner Content has it) but in the meantime I'd like to have Vars at the top.

ezr-ondrej, line 24: Also, given we don't build by docker, should we name them `CONTAINER_*`
```suggestion
CONTAINER_BUILD_OPTS  ?= "--build-arg quay_expiration=2d"
CONTAINER_IMAGE_NAME ?= "provisioning-backend"
```

build_deploy.sh
```diff
@@ -0,0 +1,37 @@
...
+    registry.redhat.io
+
+# build and push
+make build-podman \
+    DOCKER_BUILD_OPTS=--no-cache \
+    # TODO opts for image base/tag 
+
+# push to logged in registries and tag for SHA
+docker tag "${IMAGE}:${IMAGE_TAG}" "${IMAGE}:${SMOKE_TEST_TAG}"
+docker push "${IMAGE}:${SMOKE_TEST_TAG}"
```
ezr-ondrej, line 37: can we use podman here? I'd stick with one tool and podman is bit lighter dependency, so we wanted to use podman if at all possible.

mshriver: Absolutely - thanks for the quick review! Will make all this consistent.

ezr-ondrej (approved):
Looks ready to me, do we want for it to be useful, or shall we merge? I'm ok with both :)

mshriver:
@ezr-ondrej you saying my PR isn't useful?? :) Doesn't hurt either way, but its not needed until I've got the app-interface followup with the jenkins-config and saas definitions.

Happy to take more review input/suggestions in the meantime.

mshriver:
@ezr-ondrej latest commit means this should be merged after the app-interface PR containing the build and deploy configs. Opening that now.

ezr-ondrej:
Yeah the repos are created, but I guess we do not have permissions to push there... :thinking:

mshriver:
@ezr-ondrej note my last commit changes the clowdapp template name to match the resourceTemplate in the SAAS spec.

ezr-ondrej:
> note my last commit changes the clowdapp template name to match the resourceTemplate in the SAAS spec.

Yeah, those make sense I believe, I'm not sure where should be app name and where component name, so I'll trust you on that one ^_^

PR #112, opened 2022-07-26
HMSPROV-164: Change sources client implementation

adiabramovitch:
The changes here:

- Changed the location of FetchARN and FilterSourceAuthentications to be as part of sources_impl and not sources_service 
- Added FetchARN and FilterSourceAuthentications to be a part of sources_interface
- Changed the naming of init_store to sources 
- Added authentication stub 
- Added test for FetchARN
- Extract AddIdentityHeader function from the services packages so it will be reachable in sources packages (cycle importing)

### DIFF
--- a/internal/clients/sources/sources_impl.go
+++ b/internal/clients/sources/sources_impl.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/RHEnVision/provisioning-backend/internal/headers"
+
 	"github.com/RHEnVision/provisioning-backend/internal/cache"
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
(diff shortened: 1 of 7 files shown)


internal/services/sources_service_test.go
```diff
@@ -91,8 +103,30 @@ func TestShowSourceHandler(t *testing.T) {
 	assert.Equal(t, "1", *s.Id, "expected source with id = 1")
 }
 
+func TestFetchARN(t *testing.T) {
```
ezr-ondrej, line 106: Should we move this test into `clients/sources/sources_impl_test.go`?

internal/services/sources_service_test.go
```diff
@@ -91,8 +103,30 @@ func TestShowSourceHandler(t *testing.T) {
...
+
+	}
+	arn, err := client.FetchARN(ctx, "1")
+	if err != nil {
+		t.Errorf("could not fetch arn for source id 1: %v", err)
+	}
+	assert.Equal(t, "arn:aws:iam::230934684733:role/Test", arn, "expected arn equal to arn:aws:iam::230934684733:role/Test")
+}
+
 func TestFilterSourceAuthentications(t *testing.T) {
```
ezr-ondrej, line 121: Ditto ^^

internal/clients/sources/sources_impl.go
```diff
@@ -65,3 +67,62 @@ func loadAppId(ctx context.Context, client *ClientWithResponses) (string, error)
...
+	}
+
+	if *res.JSON200.ApplicationTypeId == appTypeId {
+		return *auth.Username, nil
+
+	}
+	return "", fmt.Errorf("cannot find authentication linked to source id %s and to the provisioning app: %w", sourceId, err)
+}
+
+func (client *ClientWithResponses) FilterSourceAuthentications(authentications *[]AuthenticationRead) (AuthenticationRead, error) {
```
ezr-ondrej, line 115: Do we need to export this method, given FetchArn is the only user of it now?

PR #114, opened 2022-07-26
Organize clients directory structure

ezr-ondrej:
Clients do not expose the interfaces on the main package, which is probably the best approach here.
WDYT @adiabramovitch ?


adiabramovitch:
I don't have a problem changing that. but I am wondering why do you prefer to have a file with all the clients' interfaces and not one per client?

ezr-ondrej:
> I don't have a problem changing that. but I am wondering why do you prefer to have a file with all the clients' interfaces and not one per client?

In ideal state I'd like to import only client package and decide on the implementation in the run, so it is easily interchangable... like we do for dao... at this point it is not possible, but I'm hoping to get there

adiabramovitch:
> > I don't have a problem changing that. but I am wondering why do you prefer to have a file with all the clients' interfaces and not one per client?
> 
> In ideal state I'd like to import only client package and decide on the implementation in the run, so it is easily interchangable... like we do for dao... at this point it is not possible, but I'm hoping to get there

sure, sounds good. 
 but maybe we should finish the transition to source client v2 and then add this change?

adiabramovitch:
Can we close this PR? 
we have #132

ezr-ondrej:
Yup, this has been implemented in #132, closing

PR #115, opened 2022-07-26
Implement GetArn and ListProvisioningSources

adiabramovitch:


internal/clients/sources/sources_client.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, &ListApplicationTypeSourcesParams{}, headers.AddIdentityHeader)
+	if err != nil {
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Failed to fetch ApplicationTypes from sources")
+		return nil, fmt.Errorf("failed to get ApplicationTypes: %w", err)
+	}
+	statusCode := resp.StatusCode()
+
+	if parsing.IsHTTPNotFound(statusCode) {
+		return nil, NewNotFoundError(ctx, err, "sources not found")
```
adiabramovitch, line 46: If we have decided to use a generic NewNotFoundError
we shouldn't pass here err because it is nil, should we create a custom dynamic one?

internal/services/sources_service.go
```diff
@@ -6,44 +6,28 @@ import (
...
-	if parsing.IsHTTPNotFound(statusCode) {
-		renderError(w, r, payloads.SourcesClientError(r.Context(), "sources not found", err, statusCode))
-		return
-	}
-	if !parsing.IsHTTPStatus2xx(statusCode) {
-		renderError(w, r, payloads.SourcesClientError(r.Context(), "sources client error", err, statusCode))
-		return
-	}
-	if err := render.RenderList(w, r, payloads.NewListSourcesResponse(resp.JSON200.Data)); err != nil {
+	sources, err := client.ListProvisioningSources(r.Context())
```
adiabramovitch, line 22: @ezr-ondrej I need to add here error handling, should we use SourcesClientError? should I pass from ListProvisioningSources the statusCode as well? (we add SourcesClientError a status code parameter)

ezr-ondrej: I'd use SourcesClientError here, yes :+1: 
we can pass the status code, but it is not strictly necessary, if there is an error, we just return 500 with something like "hey there was an unexpected error in internal communication" ... so the internal statusCode is not too important here.

adiabramovitch: but the error can be 404, we don't want to pass that if that is the case?

ezr-ondrej: well list should not return 404, right? it should return empty list. So even 404 is weird in this case

adiabramovitch:
```golang		
renderError(w, r, payloads.SourcesClientError(r.Context(), "sources client error", err, 500))
```
?

ezr-ondrej: looks sane to me :+1:

internal/clients/sources/sources_client.go
```diff
@@ -0,0 +1,153 @@
...
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Failed to fetch ApplicationTypes from sources")
+		return nil, fmt.Errorf("failed to get ApplicationTypes: %w", err)
+	}
+	statusCode := resp.StatusCode()
+
+	if parsing.IsHTTPNotFound(statusCode) {
+		return nil, NewNotFoundError(ctx, "sources not found")
+	}
+	if !parsing.IsHTTPStatus2xx(statusCode) {
+		ctxval.Logger(ctx).Warn().Msgf("Sources replied with unexpected status while fetching sources of application type: %v", statusCode)
```
ezr-ondrej, line 49: %d would probably be a better fit given we know it's an integer, but this is ok as well :)
```suggestion
		ctxval.Logger(ctx).Warn().Msgf("Sources replied with unexpected status while fetching sources of application type: %d", statusCode)
```

adiabramovitch: the linter didn't like %d for some reason so I changed it to %v

ezr-ondrej: ok, I thought we had %s there before... but if %d doesn't work let's go with %v :+1:

ezr-ondrej (approved):
Cool thanks @adiabramovitch ! :+1:

PR #116, opened 2022-07-26
add image builder service for AMI

amirfefer:
- [x] Fetch image status
- [x] verify image status
- [x] verify aws type
- [x] Add stub
- [ ] Verify with image builder stage env
- [x] move  imageBuilder service to a client

### DIFF
--- a/internal/clients/image_builder/image_client.go
+++ b/internal/clients/image_builder/image_client.go
@@ -1,0 +1,73 @@
+package image_builder
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/headers"
+	"github.com/RHEnVision/provisioning-backend/internal/parsing"
+)
+
+type ImageBuilderClient struct {
+	client *ClientWithResponses
+}
+
+func init() {
+	clients.GetImageBuilderClient = newImageBuilderClient
+}
+
+func newImageBuilderClient(ctx context.Context) (clients.ImageBuilder, error) {
+	c, err := NewClientWithResponses(config.ImageBuilder.URL)
+	if err != nil {
+		return nil, err
+	}
+	return &ImageBuilderClient{client: c}, nil
+}
+
+func (c *ImageBuilderClient) GetAWSAmi(ctx context.Context, composeID string) (string, error) {
+	ctxval.Logger(ctx).Info().Msgf("Getting AMI of image %v", composeID)
+	imageStatus, err := c.fetchImageStatus(ctx, composeID)
+	if err != nil {
+		return "", err
+	}
+	if imageStatus.Type != UploadTypesAws {
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Image is not AWS type")
+		return "", BadImageTypeErr
+	}
+	awsStatus := imageStatus.Options.(AWSUploadStatus)
+	return awsStatus.Ami, nil
+}
+
+func (c *ImageBuilderClient) fetchImageStatus(ctx context.Context, composeID string) (*UploadStatus, error) {
+	ctxval.Logger(ctx).Info().Msgf("Fetching image status %v", composeID)
+	resp, err := c.client.GetComposeStatusWithResponse(ctx, composeID, headers.AddIdentityHeader)
+	if err != nil {
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Failed to fetch image status from image builder")
+		return nil, fmt.Errorf("cannot get compose status: %w", err)
+	}
+	statusCode := resp.StatusCode()
+	if parsing.IsHTTPNotFound(statusCode) {
+		return nil, ComposeNotFoundErr
+	}
+	if !parsing.IsHTTPStatus2xx(statusCode) {
+		ctxval.Logger(ctx).Warn().Msgf("Image builder replied with unexpected status while fetching image status: %v", statusCode)
+		return nil, ClientErr
+	}
+	err = verifyImage(resp.JSON200)
+	if err != nil {
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Image status in not ready")
+		return nil, err
+	}
+	return resp.JSON200.ImageStatus.UploadStatus, nil
+
+}
+
+func verifyImage(compose *ComposeStatus) error {
+	if compose.ImageStatus.Status != ImageStatusStatusSuccess {
+		return ImageStatusErr
+	}
+	return nil
+}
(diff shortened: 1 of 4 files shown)


ezr-ondrej (commented):
The code looks good to me :+1: just few questions for how we follow this PR up:

Are we planning adding some tests?
I know it is not really a thing in the sources client, so I'm just asking, but would be good to at least talk about a strategy for doing testing, as this will go mostly untested until lunched completly with image builder.

We were also talking about having some tooling to run image-builder locally, is that still a plan?

internal/clients/image_builder/stubs/image_builder_stub.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+type imageBuilderCtxKeyType string
+
+var imageBuilderCtxKey imageBuilderCtxKeyType = "image-builder-interface"
+
+type ImageBuilderClientStub struct{}
+
+func init() {
+	// We are currently using SourcesClientV2Stub
```
adiabramovitch, line 16: Can you remove this comment? :)

internal/clients/image_builder/image_client.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Image status in not ready")
+		return &UploadStatus{}, err
+	}
+	return resp.JSON200.ImageStatus.UploadStatus, nil
+
+}
+
+func verifyImage(compose *ComposeStatus) error {
+	if compose.ImageStatus.Status != ImageStatusStatusSuccess {
```
adiabramovitch, line 69: Didn't see the definition for `ImageStatusStatusSuccess`, where is it? :)

amirfefer: It's an image builder type, defined in the generated client, it's just a "success" string :) though comparing it this way can prevent future regressions I guess

ezr-ondrej (changes requested):
I've got few minor comments
The errors should start with lower-case letter, because you can chain go errors - I've corrected it, but to be clear :)

amirfefer:
> Are we planning adding some tests?
I know it is not really a thing in the sources client, so I'm just asking, but would be good to at least talk about a strategy for doing testing, as this will go mostly untested until lunched completly with image builder.

IMHO, I don't see much of a benefit in unit-testing an external client, it makes sense to add a stub so other modules that use it will get a mocked `AMI`, I'd prefer to test the external clients' integration by E2E testing.

> We were also talking about having some tooling to run image-builder locally, is that still a plan?

I tried to install image builder service locally, but it looks redundant to me for our development needs, the service run over [osbuild-composer](https://github.com/osbuild/osbuild-composer) and it's not a trivial local installation. how about mocking the AMI for dev or using the stage environment (like we use in our frontend dev environment)

ezr-ondrej:
sure, I guess stubbing is fine in dev :) anything that allows us running dev is fine for now, we can figure out the best way once we know what we want to test :)

amirfefer:
Thanks, @ezr-ondrej and @adiabramovitch  for your review :)
updated and rebased

adiabramovitch (approved):

adiabramovitch:
@ezr-ondrej LGTM, any other comments?

internal/clients/image_builder/image_client.go
```diff
@@ -0,0 +1,73 @@
...
+	c, err := NewClientWithResponses(config.ImageBuilder.URL)
+	if err != nil {
+		return nil, err
+	}
+	return &ImageBuilderClient{client: c}, nil
+}
+
+func (c *ImageBuilderClient) GetAWSAmi(ctx context.Context, composeID string) (string, error) {
+	ctxval.Logger(ctx).Info().Msgf("Getting AMI of image %v", composeID)
+	imageStatus, err := fetchImageStatus(ctx, c, composeID)
```
ezr-ondrej, line 32: why this is a function and not a method of ImageBuilderClient?

internal/clients/image_builder/image_client.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	if imageStatus.Type != UploadTypesAws {
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Image is not AWS type")
+		return "", BadImageTypeErr
+	}
+	awsStatus := imageStatus.Options.(AWSUploadStatus)
+	return awsStatus.Ami, nil
+}
+
+func fetchImageStatus(ctx context.Context, c *ImageBuilderClient, composeID string) (*UploadStatus, error) {
```
ezr-ondrej, line 44:
```suggestion
func (c *ImageBuilderClient) fetchImageStatus(ctx context.Context, composeID string) (*UploadStatus, error) {
```

internal/clients/image_builder/stubs/image_builder_stub.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+)
+
+type imageBuilderCtxKeyType string
+
+var imageBuilderCtxKey imageBuilderCtxKeyType = "image-builder-interface"
```
ezr-ondrej, line 11: minor comment: we are not switching to using `iota` for context keys.

internal/clients/image_builder/image_client.go
```diff
@@ -0,0 +1,73 @@
...
+	err = verifyImage(resp.JSON200)
+	if err != nil {
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Image status in not ready")
+		return nil, err
+	}
+	return resp.JSON200.ImageStatus.UploadStatus, nil
+
+}
+
+func verifyImage(compose *ComposeStatus) error {
```
ezr-ondrej, line 68: this one is good example of function - it makes sense as function as it does not need the structure :)

internal/clients/image_builder/image_errors.go
```diff
@@ -0,0 +1,10 @@
+package image_builder
+
+import (
+	"errors"
+)
+
+var ClientErr = errors.New("image builder client error")
+var ComposeNotFoundErr = errors.New("image has not found")
```
ezr-ondrej, line 8:
```suggestion
var ComposeNotFoundErr = errors.New("image was not found")
```

ezr-ondrej (approved):
Looks good, I'd like to tidy the method versus function, as I believe this is bit bad practice, but apart of it :+1:

internal/clients/image_builder/image_client.go
amirfefer: It was a method before the` clients` refactoring, after I moved the `ImageBuilderClient` interface to the `clients` package, I had to define a `UploadStatus` struct in the clients model (as we do with `Source`) 
The `fetchImageStatus` is used internally and only by the client itself  (like a helper), therefore I removed it from the interface, and transformed it into a function implementation.

1. We can define `UploadStatus` struct in our client's models, but it's a bit complex, because it's dynamically changed according to the image type
2. We can create our own structure for the image status, though it might fail in a future regression
3. other suggestions? 🤔

ezr-ondrej (approved):
LGTM :+1:
Thanks @amirfefer :+1:

PR #117, opened 2022-07-27
Add SourcesClientV2Stub, remove and fix tests

adiabramovitch:


internal/clients/sources/stubs/sources_stub.go
```diff
@@ -26,6 +32,46 @@ func (m *contextReadError) Error() string {
...
+
+func getSourcesClientV2Stub(ctx context.Context) (si sources.ClientV2, err error) {
+	var ok bool
+	if si, ok = ctx.Value(sourcesCtxKey).(*SourcesClientV2Stub); !ok {
+		err = &contextReadError{}
+	}
+	return si, err
+}
+func (mock *SourcesClientV2Stub) GetArn(ctx context.Context, sourceId sources.ID) (string, error) {
+	return "", nil
```
ezr-ondrej, line 49: should we return valid ARN here?

internal/clients/sources/stubs/sources_stub.go
```diff
@@ -26,6 +32,46 @@ func (m *contextReadError) Error() string {
...
+}
+
+func (mock *SourcesClientV2Stub) GetProvisioningTypeId(ctx context.Context) (string, error) {
+	return "", nil
```
ezr-ondrej, line 53:
```suggestion
	return "1", nil
```

Or any number, really, we can be `42`

internal/clients/sources/stubs/sources_stub.go
```diff
@@ -26,6 +32,46 @@ func (m *contextReadError) Error() string {
...
+}
+
+func (mock *SourcesClientV2Stub) ListProvisioningSources(ctx context.Context) (*[]sources.Source, error) {
+	var TestSourceData = []sources.Source{
```
ezr-ondrej, line 57: I love usage of static data here :orange_heart:

internal/clients/sources/stubs/sources_stub.go
```diff
@@ -26,6 +32,46 @@ func (m *contextReadError) Error() string {
 	return "failed to find or convert dao stored in testing context"
 }
 
+// SourcesClientV2
+func WithSourcesClientV2(parent context.Context, init_store *[]sources.Source) context.Context {
+	ctx := context.WithValue(parent, sourcesCtxKey, &SourcesClientV2Stub{store: init_store})
```
ezr-ondrej, line 37: I do not see the store being used anywhere, imho we can just get rid of it (no store) and keep just the static data.
Our code should not change data of sources, so I don't see a point in passing it dynamically.

ezr-ondrej (commented):
I'd just get rid of buildingSources in the test suite, we can have static fixtures as you have already instead.

internal/clients/sources/stubs/sources_stub.go
adiabramovitch: Sure, I didn't "implement" it because those functions are not used yes :D 
but adding :)

internal/clients/sources/stubs/sources_stub.go
```diff
@@ -26,6 +30,46 @@ func (m *contextReadError) Error() string {
...
+		err = &contextReadError{}
+	}
+	return si, err
+}
+func (mock *SourcesClientV2Stub) GetArn(ctx context.Context, sourceId sources.ID) (string, error) {
+	return "arn:aws:iam::230214684733:role/Test", nil
+}
+
+func (mock *SourcesClientV2Stub) GetProvisioningTypeId(ctx context.Context) (string, error) {
+	return "11", nil
```
ezr-ondrej, line 51: No `42`? :( xD

ezr-ondrej (approved):
LGTM, thanks @adiabramovitch ! :+1:

PR #120, opened 2022-07-27
Db tests tenant

avitova:

**Checklist**

### DIFF
--- a/cmd/dao_tests/pubkey_test.go
+++ b/cmd/dao_tests/pubkey_test.go
@@ -5,16 +5,26 @@ package main
 
 import (
 	"context"
-	"os"
+	"fmt"
 	"testing"
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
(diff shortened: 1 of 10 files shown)


cmd/dao_tests/pubkey_test.go
```diff
@@ -63,7 +75,120 @@ func TestSimple(t *testing.T) {
 	assert.Nil(t, nil)
 }
 
-func TestMain(t *testing.M) {
```
ezr-ondrej, line 66: Why is the TestMain gone? :(

avitova: I could move it back but since there is no Setup or DatabaseCleanup outside the tests, it is not needed anymore. Or do you think it would be better to run it with TestMain?

ezr-ondrej: Well if it is in, it is clear entry point to the program. We can later on decide to structure the tests into multiple files and then it might be useful to keep the clear entry point, but you're right we do not need it at this point.

Tho I still belive that initialization (configuration and DB connection) should happen just once exactly in that function and not on every DB cleanup, but as long as it works :)

PR #122, opened 2022-07-28
Use EC2 client after assuming role

adiabramovitch:


internal/jobs/pubkey_upload_aws_job.go
```diff
@@ -16,6 +19,7 @@ type PubkeyUploadAWSTaskArgs struct {
 	AccountID     int64 `json:"account_id"`
 	ReservationID int64 `json:"reservation_id"`
 	PubkeyID      int64 `json:"pubkey_id"`
+	SourceID      int64 `json:"source_id"`
```
adiabramovitch, line 22: Is there another place we should add the sourceId as a parameter?

ezr-ondrej (approved):
I guess we should think about how we log the errors we return here, for better observability and debugging, but that is a topic on it's own. This make sense, great stuff :+1:

PR #126, opened 2022-07-29
Add podman push for short SHA

mshriver:
Builds are only pushing latest tag, missed this as the make rule for container build doesn't include a push.


mshriver:
didn't see #125, thanks @ezr-ondrej

PR #127, opened 2022-07-29
Db tests account

avitova:

**Checklist**

### DIFF
--- a/cmd/dao_tests/account_test.go
+++ b/cmd/dao_tests/account_test.go
@@ -1,0 +1,164 @@
+//go:build integration
+// +build integration
+
+package main
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
+	"github.com/stretchr/testify/assert"
+)
+
+func createAccount() *models.Account {
+	return &models.Account{
+		OrgID:         "2",
+		AccountNumber: sql.NullString{String: "100", Valid: true},
+	}
+}
+
+func createAccountWithNullAccountNumber() *models.Account {
+	return &models.Account{
+		OrgID:         "2",
+		AccountNumber: sql.NullString{},
+	}
+}
+
+func setupAccount(t *testing.T) (dao.AccountDao, context.Context) {
+	setup()
+	ctx := identity.WithTenant(t, context.Background())
+	accDao, err := dao.GetAccountDao(ctx)
+	if err != nil {
+		panic(err)
+	}
+	return accDao, ctx
+}
+
+func teardownAccount(_ *testing.T) {
+	teardown()
+}
+
+func TestCreateAccount(t *testing.T) {
+	accDao, ctx := setupAccount(t)
+	defer teardownAccount(t)
+	acc := createAccount()
+	err := accDao.Create(ctx, acc)
+	if err != nil {
+		panic(err)
+	}
+	account, err := accDao.GetById(ctx, 2)
+	if err != nil {
+		panic(err)
+	}
+
+	assert.Equal(t, acc.OrgID, account.OrgID)
+	assert.Equal(t, acc.AccountNumber.String, account.AccountNumber.String)
+	assert.Equal(t, acc.AccountNumber.Valid, account.AccountNumber.Valid)
+}
+
+func TestCreateAccountWithNullAccountNumber(t *testing.T) {
+	accDao, ctx := setupAccount(t)
+	defer teardownAccount(t)
+	acc := createAccountWithNullAccountNumber()
+	err := accDao.Create(ctx, acc)
+	if err != nil {
+		panic(err)
+	}
+	account, err := accDao.GetById(ctx, 2)
+	if err != nil {
+		panic(err)
+	}
+
+	assert.Equal(t, acc.OrgID, account.OrgID)
+	assert.Equal(t, acc.AccountNumber.String, account.AccountNumber.String)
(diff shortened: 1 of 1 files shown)


lzap:
I think you could utilize Suite from testify package (dependency which we already have): 

https://pkg.go.dev/github.com/stretchr/testify/suite

avitova:
> I think you could utilize Suite from testify package (dependency which we already have):
> 
> https://pkg.go.dev/github.com/stretchr/testify/suite

Interesting. Do you feel like this would be better for all of the DAO tests?

ezr-ondrej:
> I think you could utilize Suite from testify package

We could, but it actually runs as a single go test and creates the suite from that single test, which I did not love, so I did like the go native approch bit better, but if suite looks better, lets use it :)

cmd/dao_tests/account_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+func TestCreateAccount(t *testing.T) {
+	accDao, ctx := setupAccount(t)
+	defer teardownAccount(t)
+	acc := createAccount()
+	err := accDao.Create(ctx, acc)
+	if err != nil {
+		t.Errorf("create failed: %v", err)
+		return
+	}
```
lzap, line 54: Just a nitpick: For dao errors, you can simply call panic because it will only happen when imports are incorrect.

cmd/dao_tests/account_test.go
```diff
@@ -0,0 +1,169 @@
...
+	assert.Equal(t, acc.AccountNumber.Valid, account.AccountNumber.Valid, "Create account error: AccountNumber does not match.")
+}
+
+func TestCreateAccountWithNullAccountNumber(t *testing.T) {
+	accDao, ctx := setupAccount(t)
+	defer teardownAccount(t)
+	acc := createAccountWithNullAccountNumber()
+	err := accDao.Create(ctx, acc)
+	if err != nil {
+		t.Errorf("create failed: %v", err)
```
lzap, line 72: You use tons of the "something failed" or "test xyz failed" messages. You do not need to create these, for example in this case it's pretty clear that Create function failed, you can tell from reading the code base. So you can get away with this just with:

```suggestion
		t.Error(err)
```

Messages do make sense when the situation is complex and you want to explain to the person who will be fixing such failed test 5 years from now.

cmd/dao_tests/account_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	account, err := accDao.GetById(ctx, 2)
+	if err != nil {
+		t.Errorf("getById failed: %v", err)
+		return
+	}
+
+	assert.Equal(t, acc.OrgID, account.OrgID, "CreateAccountWithNullAccountNumber error: account OrgID does not match.")
+	assert.Equal(t, acc.AccountNumber.String, account.AccountNumber.String, "CreateAccountWithNullAccountNumber error: AccountNumber does not match.")
+	assert.Equal(t, acc.AccountNumber.Valid, account.AccountNumber.Valid, "CreateAccountWithNullAccountNumber error: AccountNumber does not match.")
```
lzap, line 83: The same case, when it's obvious these values must be equal, you simply skip the messages:

```suggestion
	assert.Equal(t, acc.AccountNumber.Valid, account.AccountNumber.Valid)
```

I would only use it for example in the Pubkey.Fingerprint field which is auto generated and there it makes sense to explain what the value should contain.

I am just trying to save you a lot of typing and possibly ease burden of maintaining those messages (e.g. after rename of field etc).

cmd/dao_tests/account_test.go
avitova: Great!:) I tried to be consistent with the rest of tests, but this makes sense.

cmd/dao_tests/account_test.go
```diff
@@ -0,0 +1,169 @@
...
+		return
+	}
+	accountsAfter, err := accDao.List(ctx, 100, 0)
+	if err != nil {
+		t.Errorf("list failed: %v", err)
+		return
+	}
+
+	assert.Equal(t, len(accountsBefore)+1, len(accountsAfter), "GetOrCreateByIdentity error.")
+}
```
lzap, line 169: This could also assert that the returned value has 2/100 fields set correctly.

avitova:
Thank you @lzap for useful review, everything mentioned should be fixed now. :+1:

cmd/dao_tests/account_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+func TestGetOrCreateByIdentityAccountCreate(t *testing.T) {
+	accDao, ctx := setupAccount(t)
+	defer teardownAccount(t)
+	accountsBefore, err := accDao.List(ctx, 100, 0)
+	if err != nil {
+		panic(err)
+	}
+	_, err = accDao.GetOrCreateByIdentity(ctx, "2", "100")
```
lzap, line 147: What this is supposed to return? If it does return an account, I would also assert the result.

avitova: This should return the account (found if it already exists, created if no such record is in database). GetByOrgId is used here to check if the record was created. We could assert if the return value of this method is correct, but I would say that it does not belong to dao tests, does it?

ezr-ondrej:
@avitova can you pls undraft this ? :)

ezr-ondrej (approved):
Thanks @avitova ! :+1:

cmd/dao_tests/account_test.go
lzap: The main purpose of most tests is to execute various code paths, here we could test the path when identity did not exist and the function actually had to construct new record and return a valid model value. I would definitely not throw the value away because GetByOrgId call below tests a different part of the code and I was interested of testing correctness of this one:

https://github.com/RHEnVision/provisioning-backend/blob/51e26b1c3c7ffeab20657d2b6dc50b5adbaecf01/internal/dao/sqlx/account_dao.go#L112-L121

Not big deal, the PR was merged. But asserts typically cost nothing, especially if you do not put long messages into them :-)

PR #128, opened 2022-07-29
Update pr_check component

mshriver:
should have removed this with the fix of the clowdapp template

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -29,8 +29,7 @@ source $CICD_ROOT/build.sh
 source $CICD_ROOT/deploy_ephemeral_env.sh
 
 # Run smoke tests using a ClowdJobInvocation and iqe-tests
-COMPONENT_NAME="provisioning"  # name of app used in deploy-iqe-cji
 source $CICD_ROOT/cji_smoke_test.sh
 
 # Post a comment with test run IDs to the PR
-# source $CICD_ROOT/post_test_results.sh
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
IMHO, we should first merge #129

ezr-ondrej (approved):
I'd love to see the effect on the check (by merging 129 first), but I do not insist.

mshriver:
the `ci.int` check for #129 is actually failing *because* this isn't merged. `bonfire deploy-iqe-cji` is failing because COMPONENT_NAME needs to match what's in app-interface for this repo.

pr_check.sh
```diff
@@ -29,8 +29,7 @@ source $CICD_ROOT/build.sh
 source $CICD_ROOT/deploy_ephemeral_env.sh
 
 # Run smoke tests using a ClowdJobInvocation and iqe-tests
-COMPONENT_NAME="provisioning"  # name of app used in deploy-iqe-cji
```
ezr-ondrej, line 32: So should this be `provisioning-api`/`provisioning-backend`? what will `COMPONENT_NAME` be if we do not set it here?

mshriver: It's set at the top of the file to `provisioning-backend`, and should match whatever the component/clowdapp name is. This is what's used to do endpoint lookups in clowder envs for CJI pipelines.

ezr-ondrej (approved):
Thanks @mshriver ! :+1:

PR #130, opened 2022-08-01
HMSPROV-153: Add launch instance aws job

adiabramovitch:
If you have a better naming suggestions I am happy to hear 😄 
- [x] DAO tests for reservation_instances 
- [x] Adding reservation_instances table for keeping the instances id that were created in aws
- [x] Save aws_reservation id to aws_reservation_details table 
- [x] Add ListInstances for easy DAO testing

### DIFF
--- a/internal/clients/ec2/ec2_client.go
+++ b/internal/clients/ec2/ec2_client.go
@@ -120,3 +120,32 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
 
 	return resp.InstanceTypes, nil
 }
+
+func (c *Client) RunInstances(ctx context.Context, amount int32, instanceType types.InstanceType, AMI string, keyName string) ([]*string, *string, error) {
+	log.Trace().Msg("Run AWS EC2 instance")
+	input := &ec2.RunInstancesInput{
+		MaxCount:     aws.Int32(amount),
+		MinCount:     aws.Int32(amount),
+		InstanceType: instanceType,
+		ImageId:      aws.String(AMI),
+		KeyName:      &keyName,
+	}
+	resp, err := c.ec2.RunInstances(ctx, input)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot run instances: %w", err)
+	}
+	instances := c.parseRunInstancesResponse(resp)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot ParseRunInstancesResponse: %w", err)
+	}
+	return instances, resp.ReservationId, nil
+}
+
+func (c *Client) parseRunInstancesResponse(respAWS *ec2.RunInstancesOutput) []*string {
+	instances := respAWS.Instances
+	list := make([]*string, 0, len(instances))
+	for _, instance := range instances {
+		list = append(list, instance.InstanceId)
+	}
+	return list
+}
--- a/internal/db/migrations/002_utility_functions.sql
+++ b/internal/db/migrations/002_utility_functions.sql
@@ -62,6 +62,8 @@ BEGIN
               AND table_type = 'BASE TABLE'
               AND table_name !~* '^(schema_version|jobs|job_dependencies|heartbeats)$'
               AND table_name !~* '_reservation_details$'
+              AND table_name !~* 'instance_reservation$'
+
     LOOP
       EXECUTE format(
         'SELECT setval(pg_get_serial_sequence(''"%s"'', ''id''), (SELECT COALESCE(MAX("id"), 1) from "%s"))', tn, tn);
--- a/internal/db/migrations/005_reservations.sql
+++ b/internal/db/migrations/005_reservations.sql
@@ -11,6 +11,12 @@ CREATE TABLE reservations
   UNIQUE (id, provider)
 );
 
+CREATE TABLE instance_reservation
+(
+  reservation_id BIGINT NOT NULL REFERENCES reservations(id) ON DELETE CASCADE,
+  instance_id TEXT NOT NULL UNIQUE
+);
+
 CREATE TABLE aws_reservation_details
 (
   reservation_id BIGINT NOT NULL,
@@ -19,7 +25,8 @@ CREATE TABLE aws_reservation_details
   source_id INTEGER NOT NULL,
   instance_type TEXT NOT NULL,
   amount INTEGER NOT NULL,
-  image_id INTEGER NOT NULL,
+  ami TEXT NOT NULL,
+  aws_reservation_id TEXT,
 
   FOREIGN KEY (reservation_id, provider) REFERENCES reservations(id, provider) ON DELETE CASCADE,
   CHECK (provider = provider_type_aws())
--- a/internal/jobs/launch_instance_aws.go
+++ b/internal/jobs/launch_instance_aws.go
@@ -1,0 +1,128 @@
+package jobs
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients/ec2"
+	"github.com/RHEnVision/provisioning-backend/internal/clients/sts"
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/lzap/dejq"
+)
+
+type LaunchInstanceAWSTaskArgs struct {
+	AccountID     int64 `json:"account_id"`
+	PubkeyID      int64 `json:"pubkey_id"`
+	ReservationID int64 `json:"reservation_id"`
+	// AWS AMI
+	AMI string `json:"ami"`
+	// Amount of instances to launch
+	Amount int32 `json:"amount"`
+	// Amazon EC2 Instance Type
+	InstanceType string `json:"instance_type"`
+	// The ARN fetched fron Sources which is linked to a specific source
+	ARN string `json:"arn"`
+}
+
+func EnqueueLaunchInstanceAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) error {
+	logger := ctxval.Logger(ctx)
+	logger.Debug().Interface("arg", args).Msgf("Enqueuing launch instance AWS job: %+v", args)
+
+	pj := dejq.PendingJob{
+		Type: TypeLaunchInstanceAws,
+		Body: args,
+	}
+	err := Queue.Enqueue(ctx, pj)
+	if err != nil {
+		return fmt.Errorf("unable to enqueue: %w", err)
+	}
+
+	return nil
+}
+
+func HandleLaunchInstanceAWS(ctx context.Context, job dejq.Job) error {
+	ctxLogger := ctxval.Logger(ctx)
+	ctxLogger.Debug().Msg("Started launch instance AWS job")
+
+	args := LaunchInstanceAWSTaskArgs{}
+	err := job.Decode(&args)
+	if err != nil {
+		ctxLogger.Error().Err(err).Msg("unable to decode arguments")
+		return fmt.Errorf("unable to decode args: %w", err)
+	}
+
+	ctx = ctxval.WithAccountId(ctx, args.AccountID)
+	logger := ctxLogger.With().Int64("reservation", args.ReservationID).Logger()
+	logger.Info().Interface("args", args).Msg("Processing launch instance AWS job")
+
+	client := ec2.NewEC2Client(ctx)
+	stsClient, err := sts.NewSTSClient(ctx)
+	if err != nil {
+		return fmt.Errorf("cannot initialize sts client: %w", err)
+	}
+
+	crd, err := stsClient.AssumeRole(args.ARN)
+	if err != nil {
+		return fmt.Errorf("cannot assume role: %w", err)
+	}
+
+	newEC2Client, err := client.CreateEC2ClientFromConfig(crd)
+	if err != nil {
+		return fmt.Errorf("cannot create new ec2 client from config: %w", err)
+	}
+
+	pkD, err := dao.GetPubkeyDao(ctx)
+	if err != nil {
+		return fmt.Errorf("cannot get pubkey dao: %w", err)
+	}
+
+	pk, err := pkD.GetById(ctx, args.PubkeyID)
+	if err != nil {
+		return fmt.Errorf("cannot get pubkey by id: %w", err)
+	}
+
+	logger.Info().Msg("Starting running instances on AWS")
+	instances, awsReservationId, err := newEC2Client.RunInstances(ctx, args.Amount, types.InstanceType(args.InstanceType), args.AMI, pk.Name)
+	if err != nil {
+		return fmt.Errorf("cannot run instances: %w", err)
+	}
+
+	resD, err := dao.GetReservationDao(ctx)
+	if err != nil {
+		return fmt.Errorf("cannot GetReservationDao: %w", err)
+	}
+
+	// For each instance that was created in AWS, add it as a DB record
+	for _, instanceId := range instances {
+		err = resD.CreateInstance(ctx, &models.InstancesReservation{
+			ReservationID: args.ReservationID,
+			InstanceID:    *instanceId,
+		})
+		if err != nil {
+			return fmt.Errorf("cannot create InstancesReservation for id %d: %w", instanceId, err)
+		}
+		logger.Info().Str("instance_id", *instanceId).Msgf("Created new instance via AWS reservation %s", *awsReservationId)
+	}
+
+	logger.Info().Str("aws_reservation_id", *awsReservationId).Msg("Adding aws reservation id")
+	// Save the AWS reservation id in aws_reservation_details table
+	err = resD.UpdateReservationIDForAWS(ctx, args.ReservationID, *awsReservationId)
+	if err != nil {
+		return fmt.Errorf("cannot UpdateReservationIDForAWS: %w", err)
+	}
+
+	// mark the reservation as finished
+	rDao, err := dao.GetReservationDao(ctx)
+	if err != nil {
+		return fmt.Errorf("cannot get reservation DAO: %w", err)
+	}
+	err = rDao.UpdateStatus(ctx, args.ReservationID, "Finished")
+	if err != nil {
+		return fmt.Errorf("cannot update reservation status: %w", err)
+	}
+
+	return nil
+}
--- a/internal/jobs/pubkey_upload_aws_job.go
+++ b/internal/jobs/pubkey_upload_aws_job.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strconv"
 
-	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/RHEnVision/provisioning-backend/internal/clients/ec2"
 	"github.com/RHEnVision/provisioning-backend/internal/clients/sts"
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
@@ -16,10 +14,10 @@ import (
 )
 
 type PubkeyUploadAWSTaskArgs struct {
-	AccountID     int64 `json:"account_id"`
-	ReservationID int64 `json:"reservation_id"`
-	PubkeyID      int64 `json:"pubkey_id"`
-	SourceID      int64 `json:"source_id"`
+	AccountID     int64  `json:"account_id"`
+	ReservationID int64  `json:"reservation_id"`
+	PubkeyID      int64  `json:"pubkey_id"`
+	ARN           string `json:"arn"`
 }
 
 func EnqueuePubkeyUploadAWS(ctx context.Context, args *PubkeyUploadAWSTaskArgs) error {
@@ -48,6 +46,8 @@ func HandlePubkeyUploadAWS(ctx context.Context, job dejq.Job) error {
 		ctxLogger.Error().Err(err).Msg("unable to decode arguments")
 		return fmt.Errorf("unable to decode args: %w", err)
 	}
+
+	ctx = ctxval.WithAccountId(ctx, args.AccountID)
 	logger := ctxLogger.With().Int64("reservation", args.ReservationID).Logger()
 	logger.Info().Interface("args", args).Msg("Processing pubkey upload AWS job")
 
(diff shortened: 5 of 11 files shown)


internal/jobs/pubkey_upload_aws_job.go
```diff
@@ -48,6 +46,20 @@ func HandlePubkeyUploadAWS(ctx context.Context, job dejq.Job) error {
...
+		return fmt.Errorf("failed to initialize connection to fetch Account info: %w", err)
+	}
+
+	acc, err := accDao.GetById(ctx, args.AccountID)
+	if err != nil {
+		ctxLogger.Error().Err(err).Msg("Failed to get Account by id")
+		return fmt.Errorf("failed to get Account by id: %w", err)
+	}
+
+	ctx = ctxval.WithAccount(ctx, acc)
```
ezr-ondrej, line 62:
```suggestion
	ctx = ctxval.WithAccountId(ctx, args.AccountID)
```

internal/clients/ec2/ec2_client.go
```diff
@@ -120,3 +120,34 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
 
 	return resp.InstanceTypes, nil
 }
+
+func (c *Client) RunInstances(ctx context.Context, amount int32, instanceType types.InstanceType, imageId string, keyName string) ([]*string, error) {
+	log.Trace().Msg("Run AWS EC2 instance")
+	input := &ec2.RunInstancesInput{
+		MaxCount:     aws.Int32(amount),
+		MinCount:     aws.Int32(amount),
```
adiabramovitch, line 128: There isn't an `amount` field only min and max, and both are required. this is why it is written that way.

internal/clients/ec2/ec2_client.go
```diff
@@ -120,3 +120,34 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
...
+		InstanceType: instanceType,
+		ImageId:      aws.String(imageId),
+		KeyName:      &keyName,
+	}
+
+	resp, err := c.ec2.RunInstances(ctx, input)
+	if err != nil {
+		return nil, fmt.Errorf("cannot run instances: %w", err)
+	}
+	// TODO: Do we want to parse the response? or is it redundant?
```
adiabramovitch, line 138: @ezr-ondrej Should we Parse the response? how can we save this info in the job?

internal/jobs/launch_instance_aws.go
```diff
@@ -0,0 +1,91 @@
...
+		return fmt.Errorf("cannot assume role: %w", err)
+	}
+
+	newEC2Client, err := client.CreateEC2ClientFromConfig(crd)
+	if err != nil {
+		return fmt.Errorf("cannot create new ec2 client from config: %w", err)
+	} //get the info needed to launch and instance and pass it
+	//TODO: NEEDS THE RESOURCE NAME LIKE "lzap-rsa-2010-{{$randomInt}}"
+	// Should be added to args.PubkeyHandle, the name is not saved?
+	_, err = newEC2Client.RunInstances(ctx, args.Amount, types.InstanceType(args.InstanceType), args.ImageID, "lzap-rsa-2010-{{$randomInt}}")
```
amirfefer, line 75: the `imageID` from the reservation request is an inner identifier of the image in image-builder service,
we should [fetch](https://github.com/RHEnVision/provisioning-backend/pull/116/files#diff-84c4846afa93893d6a3a76ddb49781b320530975ebb9c6e781a320f09ca97cd6R30) the ami from image-builder via its client

internal/clients/ec2/ec2_client.go
amirfefer: it looks like it is a range, if the account can't run the max amount it will launch
the largest possible number of instances above MinCount

internal/jobs/launch_instance_aws.go
adiabramovitch: the image Id written here is the AWS image identifier like in the example: 
"ami-0cff7528ff583bf9a"

internal/clients/ec2/ec2_client.go
adiabramovitch: I don't think that it is the functionality that we want for now. 
For now, we want a user to choose a specific amount of instances to provision. 
This is the way I found for now to enable that, but I guess that there is another one - I'll check it out.

internal/clients/ec2/ec2_client.go
```diff
@@ -120,3 +120,34 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
 
 	return resp.InstanceTypes, nil
 }
+
+func (c *Client) RunInstances(ctx context.Context, amount int32, instanceType types.InstanceType, imageId string, keyName string) ([]*string, error) {
+	log.Trace().Msg("Run AWS EC2 instance")
```
lzap, line 125: Thanks for using trace level, I feel like we already have too many debug messages. It would be useful to log also via `Msgf` and add all the arguments.

One field is extra important that is the Instance id returned by Amazon, I think this is worth logging at info level in a separate logging field like `instance_id` so we can search for that easily.

internal/clients/ec2/ec2_client.go
lzap: My guess is that this is for spot instances, for these types you define a price you are willing to pay and perhaps a range?

adiabramovitch: After looking around didn't see any other options people used so maybe you are right and it is based on a budget range.

internal/clients/ec2/ec2_client.go
adiabramovitch: I am parsing the response from Amazon, wondering if it is necessary, if we want to display it, or just log it.

internal/clients/ec2/ec2_client.go
lzap: We should store it in the AWS reservation detail table, the field is not there yet. We will only need the instance id until we confirm that the host was registered to the inventory, then we can delete the reservation from the database. We do not want to do any kind of post-provisioning management, there are other services to do that @ezr-ondrej

internal/jobs/launch_instance_aws.go
ezr-ondrej: Should we name it AMI then? and make use of Amir's pr in the service I guess (simiar to what we do with ARN)

internal/jobs/launch_instance_aws.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		ctxLogger.Error().Err(err).Msg("unable to decode arguments")
+		return fmt.Errorf("unable to decode args: %w", err)
+	}
+
+	ctx = ctxval.WithAccountId(ctx, args.AccountID)
+	logger := ctxLogger.With().Int64("reservation", args.ReservationID).Logger()
+	logger.Info().Interface("args", args).Msg("Processing launch instance AWS job")
+
+	// upload to cloud with a tag
```
lzap, line 57: This comment can be dropped, we are not uploading anything in here.

internal/clients/ec2/ec2_client.go
lzap: That should be done in the background job tho, not here.

internal/jobs/launch_instance_aws.go
```diff
@@ -0,0 +1,91 @@
...
+		return fmt.Errorf("cannot create new ec2 client from config: %w", err)
+	} //get the info needed to launch and instance and pass it
+	//TODO: NEEDS THE RESOURCE NAME LIKE "lzap-rsa-2010-{{$randomInt}}"
+	// Should be added to args.PubkeyHandle, the name is not saved?
+	_, err = newEC2Client.RunInstances(ctx, args.Amount, types.InstanceType(args.InstanceType), args.ImageID, "lzap-rsa-2010-{{$randomInt}}")
+	if err != nil {
+		return fmt.Errorf("cannot run instances: %w", err)
+	}
+
+	// mark the reservation as finished
```
lzap, line 80: So, this is the tricky part. In order to mark the reservation as finished, we need to confirm the instance has been launched. I haven't looked into RunInstances response yet, so let's do it now.

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html

Looks like the call returns a reservation id, which is perhaps a good handle to store in the AWS reservation detail table for debugging purposes (in case the customer needs to open up a case for AWS Support).

Then it returns an array of instances: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Instance.html

This means we cannot store a single instance id in AWS reservation detail, we need a table relationship for these instance IDs.

internal/jobs/launch_instance_aws.go
ezr-ondrej: I believe we should send image_id to the API, but I'd resolve that into AMI before queing the job, but happy to be convinced that we should push these to the job, but we are missing the auth header in the context, so we just resolved `source_id -> ARN` in the service and I guess this is the same case.
But I'd agree that it's kinda being lazy and pushing too much to the service can make it too slow.

internal/clients/ec2/ec2_client.go
ezr-ondrej: Agreed, we fire notifications, make sure inventory has all info we can provide and we are done.

To the original question, I think we are saying we want to "parse" the response and return the reservation_id from this method on success, right?

adiabramovitch: So after talking with @lzap  we decided to create a new table, saving the instance id that was created. 
so yes, we are parsing the response to extract valuable info to store in the DB.

internal/jobs/launch_instance_aws.go
adiabramovitch: Not sure what did you mean.
I don't use image_id from image builder, I just use the AMI. 
Don't mind changing it to AMI instead of image_id if we all agree.

internal/clients/ec2/ec2_client.go
lzap: Yeah and since this will be associated table with CASCADE DELETE we can only create those records, so no extra DAO file is necessary similarly to "detail" DAO function.

internal/jobs/launch_instance_aws.go
lzap: I agree that all jobs should only need to access our database and the cloud, therefore resolving the imagebuilder ID to AMI would be the best to do before the job is enqueued.

internal/clients/ec2/ec2_client.go
lzap: I would only log this. For each instance ID I would create a new log entry with "Created new instance via reservation XXX" (reservation id returned from AWS). And each log entry would have a field named "instance_id" with the instance id included. This way we can easily search the logs and find the whole transaction responsible for doing the request in case something went wrong.

internal/clients/ec2/ec2_client.go
```diff
@@ -120,3 +120,34 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
...
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot run instances: %w", err)
+	}
+	instances := c.parseRunInstancesResponse(resp)
+	if err != nil {
+		return nil, nil, fmt.Errorf("cannot ParseRunInstancesResponse: %w", err)
+	}
+	return instances, resp.ReservationId, nil
+}
```
lzap, line 144: Please add some logging as I described in other comment, I feel like it is very important because sooner or later a customer will come with "hey my instance id was spawned incorrectly".

internal/db/migrations/002_utility_functions.sql
```diff
@@ -62,6 +62,8 @@ BEGIN
               AND table_type = 'BASE TABLE'
               AND table_name !~* '^(schema_version|jobs|job_dependencies|heartbeats)$'
               AND table_name !~* '_reservation_details$'
+              AND table_name !~* 'reservation_instances$'
```
lzap, line 65: This could have been a single line:

```
AND table_name !~* '_reservation_(details|instances)$'
```

internal/jobs/launch_instance_aws.go
```diff
@@ -0,0 +1,124 @@
...
+
+type LaunchInstanceAWSTaskArgs struct {
+	AccountID     int64  `json:"account_id"`
+	PubkeyID      int64  `json:"pubkey_id"`
+	ReservationID int64  `json:"reservation_id"`
+	ImageID       string `json:"image_id"`
+	Amount        int32  `json:"amount"`
+	InstanceType  string `json:"instance_type"`
+	ARN           string `json:"arn"`
+}
```
lzap, line 24: Maybe add comments to each field? Specifically that ImageID is actually AMI and InstanceType is the textual representation of instance type.

internal/jobs/launch_instance_aws.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot run instances: %w", err)
+	}
+
+	resD, err := dao.GetReservationDao(ctx)
+	if err != nil {
+		return fmt.Errorf("cannot GetReservationDao: %w", err)
+	}
+
+	logger.Info().Interface("instances", instances).Msg("Creating DB instances records")
```
lzap, line 94: This is unfortunately not very searchable on CloudWatch, as I mentioned, can you simply log each record per each instance? This will create 40 log events for 40 instances but it is easily searchable. Simply add the log string field so we can easily search it.

You can keep "Creating DB instances records" of course if you like, I would do it at debug level tho. Each job should only log 1-2 info level logs and the rest should be debug otherwise we will have hundreds of info level messages per operation.

internal/jobs/launch_instance_aws.go
```diff
@@ -0,0 +1,124 @@
...
+	// For each instance that was created in AWS, add it as a DB record
+	for _, instanceId := range instances {
+		err = resD.CreateInstances(ctx, &models.InstancesReservation{
+			ReservationID: args.ReservationID,
+			InstanceID:    *instanceId,
+		})
+		if err != nil {
+			return fmt.Errorf("cannot create InstancesReservation for id %d: %w", instanceId, err)
+		}
+	}
```
lzap, line 104: This is fine for now, postgres also supports batch insert and it is _relatively_ easy to do. You could have passed the whole array to the DAO function and do it that way if you want. It is a bit complicated so tests are needed tho, perhaps as a separate PR:

https://github.com/jackc/pgx/issues/764#issuecomment-685249471

internal/jobs/launch_instance_aws.go
```diff
@@ -0,0 +1,124 @@
...
+
+	logger.Info().Interface("aws_reservation_id", *awsReservationId).Msg("Adding aws reservation id")
```
lzap, line 106: Use interface only for complex fields, JSON is typed so we can take advantage of types which is the unique feature of zerolog:

```suggestion
	logger.Info().Str("aws_reservation_id", *awsReservationId).Msg("Adding aws reservation id")
```

Similarly use `Int` or `Float` etc.

internal/models/reservation_model.go
```diff
@@ -48,8 +48,16 @@ type AWSReservation struct {
 	InstanceType string `db:"instance_type" json:"instance_type"`
 
 	// Amount of instances to provision of type: Instance type.
-	Amount int64 `db:"amount" json:"amount"`
+	Amount int32 `db:"amount" json:"amount"`
 
 	// The ID of the image from which the instance is created.
-	ImageID int64 `db:"image_id" json:"image_id"`
+	ImageID string `db:"image_id" json:"image_id"`
+}
```
lzap, line 55: Please update the comment so it's obvious this is an AMI not IB ID.

internal/services/aws_reservation_service.go
```diff
@@ -139,7 +163,27 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
...
+		AccountID:     accountId,
+		ReservationID: reservation.ID,
+		PubkeyID:      pk.ID,
+		ImageID:       reservation.ImageID,
+		ARN:           arn,
+		Amount:        reservation.Amount,
+		InstanceType:  reservation.InstanceType,
+		//TODO: NEEDS THE RESOURCE NAME LIKE "lzap-rsa-2010-{{$randomInt}}"
+		// PubkeyHandle:  pkr.Name,
+	}
```
lzap, line 178: Yeah see my patch https://github.com/RHEnVision/provisioning-backend/pull/135

You can simply pass Pubkey.Name here now, its not unique now but it will be once my PR is merged.

internal/services/aws_reservation_service.go
```diff
@@ -139,7 +163,27 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 		logger.Debug().Msgf("Found existing pubkey resource %d, upload not enqueued", pkr.ID)
 	}
 
+	logger.Debug().Msgf("Enqueuing launch instance job for pubkey %d and source %d", pk.ID, reservation.SourceID)
+
+	argsLaunch := &jobs.LaunchInstanceAWSTaskArgs{
+		AccountID:     accountId,
+		ReservationID: reservation.ID,
+		PubkeyID:      pk.ID,
+		ImageID:       reservation.ImageID,
```
lzap, line 172: Image ID for now is AMI so the API needs AMI too. We can add resolving later I guess.

scripts/rest_examples/reservation-create-aws.http
```diff
@@ -1,10 +1,15 @@
...
 
 {
+
+  "source_id": 9, 
+  "image_id": "ami-0cff7528ff583bf9a",
+  "amount": 1,
+  "instance_type": "t1.micro",
   "pubkey": {
-    "existing_id": 1
+    "existing_id": 2
```
lzap, line 13: Can you keep this on 1? Because if you seed the database it is the only key that is present (my key). Feel free to add more keys into the seed tho. Then make this a `{variable}` so it can be configured locally.

lzap (commented):
Looks great.

internal/services/aws_reservation_service.go
adiabramovitch: This comment is outdated, I have deleted it.
But I am not sure we should store the pubkey name in addition to the pubkey id. 
Now I am fetching pubkey by id from the DB. 
What do you think is better?

internal/models/reservation_model.go
adiabramovitch: I thought of changing it to AMI, What do you think?

internal/services/aws_reservation_service.go
adiabramovitch: here we only need the AMI, so I thought of changing from ImageID to AMI

internal/clients/ec2/ec2_client.go
adiabramovitch: @lzap Should I add the logging here? 
I wanted to add it after creating the record: 
```golang
	// For each instance that was created in AWS, add it as a DB record
	for _, instanceId := range instances {
		err = resD.CreateInstance(ctx, &models.InstancesReservation{
			ReservationID: args.ReservationID,
			InstanceID:    *instanceId,
		})
		if err != nil {
			return fmt.Errorf("cannot create InstancesReservation for id %d: %w", instanceId, err)
		}
		logger.Info().Str("instance_id", *instanceId).Msgf("Created new instance via AWS reservation %s", *awsReservationId)
	}
```

internal/db/migrations/002_utility_functions.sql
adiabramovitch: so it didn't work because we don't have a prefix to the reservation_instances table (like aws_). 
Do you know how can we include that as well?

internal/jobs/launch_instance_aws.go
ezr-ondrej: yeah, agreed, so I'd name it `ami` to be clear

internal/clients/ec2/ec2_client.go
lzap: You got it right, I just commented here, at the wrong place.

internal/models/reservation_model.go
lzap: Whatever you like, the argument in the AWS SDK is called ImageID actually so both are valid.

internal/services/aws_reservation_service.go
lzap: Yeah I thought about it too, then I deleted my comment to keep it simple. Since you started...

Both are doable, but I would probably keep just passing the ID and fetching it later, because if a customer will hit a problem that an already existing key has a name that cannot be used in our database, we will probably do some renaming routine that will name the key something like "my_key-2" and we would need to store such name in PubkeyResource. So fetching would be the only option anyway to support these renamed keys.

internal/models/reservation_model.go
lzap: If you still want to hear my opinion, I would go for the rename, AMI is definitely widely popular term and it is crystal clear for everyone.

internal/db/migrations/002_utility_functions.sql
ezr-ondrej: Tho I do not feel like being less verbose is better here, so I'd just keep it as is.
```
AND table_name !~* '_?reservation_(details|instances)$'
```

internal/clients/ec2/ec2_client.go
```diff
@@ -120,3 +120,34 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
...
+	log.Trace().Msg("Run AWS EC2 instance")
+	input := &ec2.RunInstancesInput{
+		MaxCount:     aws.Int32(amount),
+		MinCount:     aws.Int32(amount),
+		InstanceType: instanceType,
+		ImageId:      aws.String(AMI),
+		KeyName:      &keyName,
+	}
+	resp, err := c.ec2.RunInstances(ctx, input, func(o *ec2.Options) {
+		o.Region = config.AWS.Region
```
ezr-ondrej, line 134: does this do something? aren't we initializing the client with region already?
I know we will probably need it in the future, but now it just looks funny to override one static value with the same one.

adiabramovitch: yes you are right, it is not needed :)

scripts/rest_examples/reservation-create-aws.http
```diff
@@ -1,9 +1,14 @@
 // @no-log
-POST http://{{hostname}}:{{port}}/{{prefix}}/reservations/aws HTTP/1.1
+POST http://localhost:8005/api/provisioning/v1/reservations/aws HTTP/1.1
 Content-Type: application/json
 X-Rh-Identity: {{identity}}
 
 {
+
+  "source_id": 9, 
+  "ami": "ami-0cff7528ff583bf9a",
```
ezr-ondrej, line 9: I think this should not be ami, but image_id...
can we just drop change to this file for now and add it only once we are addressing the endpoint itself and have #116 in? :)

adiabramovitch: So should I change this to image_id? just the for the reservation  model

ezr-ondrej: I'd just drop the change of this file... we need to change the endpoint first anyway :)

ezr-ondrej (approved):
I believe this is great progress, not yet done, but good way forward :+1:

scripts/rest_examples/reservation-create-aws.http
ezr-ondrej: second thought, lets just improve on this file when we are addressing the endpoint itself :+1:

internal/db/migrations/005_reservations.sql
```diff
@@ -11,6 +11,12 @@ CREATE TABLE reservations
   UNIQUE (id, provider)
 );
 
+CREATE TABLE instance_reservation
+(
+  reservation_id BIGINT NOT NULL REFERENCES reservations(id) ON DELETE CASCADE,
+  instance_id TEXT NOT NULL UNIQUE
+);
+
```
lzap, line 19: We have missed a small detail: all our tables are plural, this is singular. Fixed in: https://github.com/RHEnVision/provisioning-backend/pull/144

PR #131, opened 2022-08-01
Keep only account id in the context

ezr-ondrej:
Keeping the whole account in the context is bit troublesome.


lzap:
Yeah I never liked it, I think we can store accountNumber/orgId separately too if we need to (probably we won't need it).

PR #132, opened 2022-08-01
Define source model ourselves

ezr-ondrej:
Define our source model and use that instead of generated one.

### DIFF
--- a/internal/clients/sources/stubs/sources_stub.go
+++ b/internal/clients/sources/stubs/sources_stub.go
@@ -17,11 +17,11 @@ type SourcesIntegrationStub struct {
 	store           *[]sources.Source
 	authentications *[]sources.AuthenticationRead
 }
-type SourcesClientV2Stub struct{}
+type SourcesClientStub struct{}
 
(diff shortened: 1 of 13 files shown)


internal/clients/sources/stubs/sources_stub.go
```diff
@@ -21,7 +21,7 @@ type SourcesClientV2Stub struct{}
 
 func init() {
 	// We are currently using SourcesClientV2Stub
-	sources.GetSourcesClientV2 = getSourcesClientV2Stub
+	clients.GetSourcesClient = getSourcesClientV2Stub
```
adiabramovitch, line 24: `getSourcesClientV2Stub` Maybe this should be changed as well?

ezr-ondrej: Done :)

adiabramovitch (approved):
Checked that /sources & /instance_types/<source_id> didn't break 
LGTM :) 
Thanks @ezr-ondrej

PR #133, opened 2022-08-02
Clean up unused methods from Sources API client

ezr-ondrej:
Rename client to the API client and clean up methods that were move


internal/clients/sources/sources_api.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+)
+
+func getSourcesAPIClient(ctx context.Context) (APIClient, error) {
+	return NewClientWithResponses(config.Sources.URL)
+}
+
+func init() {
+	GetSourcesAPIClient = getSourcesAPIClient
+}
```
ezr-ondrej, line 15: Not so visible, but this file has been renamed and methods removed, these methods are already in sources_client.

adiabramovitch: Did you change the name of sources_impl to sources_api?

ezr-ondrej: Yeah, exactly ... impl sounded bit ambigous now that we have two implementations

adiabramovitch (approved):

PR #134, opened 2022-08-02
HMSPROV-79 Reservation dao test

avitova:

**Checklist**

### DIFF
--- a/cmd/dao_tests/reservation_test.go
+++ b/cmd/dao_tests/reservation_test.go
@@ -1,0 +1,223 @@
+//go:build integration
+// +build integration
+
+package main
+
+import (
+	"context"
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
+	"github.com/RHEnVision/provisioning-backend/internal/db"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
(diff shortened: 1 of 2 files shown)


ezr-ondrej:
Can you pls remove the `dao_tests.test` from the PR and add it to `.gitignore`?

avitova:
> Can you pls remove the `dao_tests.test` from the PR and add it to `.gitignore` so we don't accidentally commit it again? :)

I did not notice that, it should be okay now!

.gitignore
```diff
@@ -13,3 +13,4 @@ TAGS
 /scripts/sources-*/
 /scripts/sources.local.conf
 /scripts/rest_examples/*.local.http
+/dao_tests.test
```
lzap, line 16: I have refactored the Makefile so this is not necessary and dropped this in: https://github.com/RHEnVision/provisioning-backend/pull/135

You can drop this from here.

lzap: Or I can drop it from my PR so this can be done in a separate PR if you like.

avitova: > Or I can drop it from my PR so this can be done in a separate PR if you like.

No problem, I will just drop it from this PR, leave it in yours.

ezr-ondrej (approved):
I feel like we could simplify by some fixtures, but lets discuss that somewhere else :)
Also I don't want to block on changes from #138, as discussed offline, lets go with style we have until we finish discussion over the change.
Thanks @avitova ! :+1:

PR #135, opened 2022-08-02
HMSPROV-179: Make pubkey name and body unique per account

lzap:
To my surprise, @adiabramovitch showed me that AWS API does not use ssh key-pair AWS id when creating instances but its name (which must be unique as well as key-pair body). This means we also need to enforce this on our side. This patch does exactly that.


lzap:
You know what, I just noticed we now import SSH library https://pkg.go.dev/golang.org/x/crypto/ssh which already contains fingerprint functions. I will add a new fingerprint column which is much better way to test the body because it can be different in the suffix but with the same key. Unique constraint on fingerprint will catch these cases.

internal/db/migrations/004_pubkeys.sql
```diff
@@ -3,7 +3,10 @@ CREATE TABLE pubkeys
   id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
   account_id BIGINT NOT NULL REFERENCES accounts(id),
   name TEXT NOT NULL CHECK (NOT empty(name)),
-  body TEXT NOT NULL CHECK (NOT empty(body))
+  body TEXT NOT NULL CHECK (NOT empty(body)),
+
+  UNIQUE(name, account_id),
```
ezr-ondrej, line 8: Should this instead be `UNIQUE(pubkey_resources, handle)`?

ezr-ondrej (commented):
Agreed with the fingerprinting :+1: It not only feels cool, but also feels like correct way (unless we want to be trimming comments out of the pubkeys).

internal/db/migrations/004_pubkeys.sql
adiabramovitch: IMO no.
we found out that in order to RunInstances we need the **keyName** - the name of the uploaded key. 
So in order to use the name we want to make sure it is unique.

ezr-ondrej: ok, I'm fine with that, I just thought handle is the name we use as AWS keyName, but I'm not sure if I got that correctly.

adiabramovitch: no, the handle is the unique id AWS gives to a key - it looks something like `key-488030` (like the AMI identifier), the key name is saved in Pubkey 😃

ezr-ondrej: interesting... so it's really a name not an handle we need to pass? Interesting :thinking:
Ok then, ignore me! :+1:

lzap: Not sure what you mean by this, pubkey must be unique in the main pubkeys table already, even before it gets deployed. In my working branch, there are two unique checks now: name, account + fingerprint, account.

lzap:
So as @adiabramovitch already mentioned, we have found that public keys (ssh key pairs in AWS terms) is identified by key name, not by key aws id. I was not expecting that when I was writing the code.

Therefore this patch actually drops uploading keys under generated names and uses the name that user provided. The name must be unique now for each account.

More than that, sice AWS also enforces that keys are unique (body), I am including a new library which is recommended by the golang validate library (https://github.com/go-playground/validator#complementary-software): *meld* library. It has the very similar design as validate, but it is used for transforming fields or whole structs. This is used to automatically generate key fingerprint (SHA256 string) that is then also added as a SQL constraint check.

To summarize:

* Public keys must be unique by name and by body (thus fingerprint) per account.
* A friendly message is thrown an dupe insert ("pubkey with such name or fingerprint already exists for this account").
* The error message currently does not show which is the duplicit entry, we can add that later.

The patch also removes the accidentally commited binary file and prevents from commiting it again. To run tests from the cmd/dao_test folder, a symlink was created so we don't need to utilize the hack in the Makefile.

.gitignore
```diff
@@ -13,3 +13,4 @@ TAGS
 /scripts/sources-*/
 /scripts/sources.local.conf
 /scripts/rest_examples/*.local.http
+/dao_tests.test
```
lzap, line 16: Just in case others still have it.

Makefile
```diff
@@ -107,5 +107,4 @@ update-test-deps: update-deps test
 
 .PHONY: integration-test
 integration-test:
-	go test -c -tags=integration -o dao_tests.test ./cmd/dao_tests
-	./dao_tests.test
+	go test -tags=integration ./cmd/dao_tests
```
lzap, line 110: I created a symlink which allows running tests just like that, looks cleaner to me.

cmd/dao_tests/pubkey_test.go
```diff
@@ -88,6 +88,23 @@ func TestCreatePubkey(t *testing.T) {
...
+	pk := createPk()
+	err = pkDao.Create(ctx, pk)
+	if err != nil {
+		t.Errorf("Create pubkey test had failed: %v", err)
+		return
+	}
+
+	assert.Equal(t, "SHA256:gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk", pk.Fingerprint, "Generate fingerprint failed.")
+}
+
```
lzap, line 107: This is already covered with a unit test, I might drop this.

internal/payloads/error_payload.go
```diff
@@ -47,7 +47,21 @@ func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
...
 	}
 	return &ResponseError{
-		HTTPStatusCode: 500,
+		HTTPStatusCode: 400,
+		Message:        msg,
+		RequestId:      ctxval.RequestId(ctx),
+		Err:            err,
+		Context:        ctx,
+	}
+}
```
lzap, line 56: This was invalid request error with 500, but it should be 400.

internal/db/error_codes.go
```diff
@@ -0,0 +1,17 @@
...
+
+// https://www.postgresql.org/docs/current/errcodes-appendix.html
+
+func IsUniqueConstraintError(err error) error {
+	var pgErr *pgconn.PgError
+	if err != nil && errors.As(err, &pgErr) && pgErr.Code == "23505" {
+		return pgErr
+	}
+	return nil
+}
```
lzap, line 17: Let's put all postgres error code detection here so we can utilize this for meaningful error messages.

internal/services/pubkey_service.go
```diff
@@ -4,18 +4,26 @@ import (
...
 	if err := render.Bind(r, payload); err != nil {
 		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), err))
 		return
 	}
 
+	// validate payload
+	if payload.Fingerprint != "" {
+		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), FingerprintReadOnlyError))
+	}
+
```
lzap, line 26: We do not allow setting the fingerprint explicitly.

ezr-ondrej: is this conflicting with the validations approach? I feel like we are scattering validations across the codebase :thinking:

lzap:
Resolved OpenAPI spec error.

internal/services/pubkey_service.go
lzap: Damn sure I just forgot we have that now!

lzap: You know what, this check is useless, everytime we create new record it will be overwritten anyways. And we do not allow updating of keys.

lzap: Removed.

lzap:
All green except our QA tests, looks like some sources setup problem.

internal/models/transformer.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	return nil
+}
+
+func generateFingerprint(ctx context.Context, sl mold.StructLevel) error {
+	logger := ctxval.Logger(ctx)
+	pk := sl.Struct().Interface().(Pubkey)
+	pkey, _, _, _, err := ssh.ParseAuthorizedKey([]byte(pk.Body))
+	if err != nil {
+		return fmt.Errorf("uanble to calculate fingerprint for %s: %w", pk.Name, err)
```
adiabramovitch, line 33: typo - unable

lzap:
Rebased.

Makefile
ezr-ondrej: I'm bit lost here, why is this change even here, it feels like we are trying to address bit too much for a single PR.
I'm ok with the change, but it seems like bit out of scope of the PR :thinking:

internal/services/pubkey_service.go
```diff
@@ -4,11 +4,14 @@ import (
 	"net/http"
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/db"
 	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 	"github.com/go-chi/render"
 	"github.com/pkg/errors"
 )
 
+var FingerprintReadOnlyError = errors.New("fingerprint is a read-only field")
```
ezr-ondrej, line 13: Looks like a dead code.
I was thinking we can directly [filter it out](https://github.com/go-chi/chi/blob/9db25e5b381fbd6350583b857e74ecaffe355819/_examples/rest/main.go#L328) in the request payload, wouldn't it be a good strategy?
That way we would ignore the field completly

adiabramovitch:
LGTM
I would prefer next time separating unrelated code in different PRS/commits :)

internal/db/error_codes.go
```diff
@@ -0,0 +1,17 @@
...
+	"errors"
+
+	"github.com/jackc/pgconn"
+)
+
+// https://www.postgresql.org/docs/current/errcodes-appendix.html
+
+func IsUniqueConstraintError(err error) error {
+	var pgErr *pgconn.PgError
+	if err != nil && errors.As(err, &pgErr) && pgErr.Code == "23505" {
```
ezr-ondrej, line 13: Could `23505` be a constant?

internal/models/transformer_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+var examplePubkey = Pubkey{
+	Name: "lzap-ed25519-2021",
+	Body: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap",
+}
+
+func TestPubkeyFingerprintGeneration(t *testing.T) {
```
ezr-ondrej, line 15: I don't have a strong feeling about this, but I feel like this is testing Pubkey transformation, not transform method itself...
But I guess two birds... just wondering if we add another transformation, would we test it all in this test file?

internal/payloads/error_payload.go
```diff
@@ -47,7 +47,21 @@ func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
...
+
+func PubkeyAlreadyExistsError(ctx context.Context, err error) *ResponseError {
+	msg := "pubkey with such name or fingerprint already exists for this account"
+	if logger := ctxval.Logger(ctx); logger != nil {
+		logger.Error().Msg(msg)
+	}
+	return &ResponseError{
+		HTTPStatusCode: 400,
```
ezr-ondrej, line 64: Rails uses [`422`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) on validation errors, should we do that as well, I'm fine with whatever, but being it a concious choice, so @lzap is that choice of general 400 intentional, or you just didn't want 500?

internal/services/pubkey_service.go
```diff
@@ -24,7 +27,11 @@ func CreatePubkey(w http.ResponseWriter, r *http.Request) {
 
 	err = pkDao.Create(r.Context(), payload.Pubkey)
 	if err != nil {
-		renderError(w, r, payloads.NewDAOError(r.Context(), "create pubkey", err))
+		if db.IsUniqueConstraintError(err) != nil {
+			renderError(w, r, payloads.PubkeyAlreadyExistsError(r.Context(), err))
```
ezr-ondrej, line 31: Should we wrap it in DAO and return a nicer error? I mean this is fine, but I feel like letting pg errors bubble up feels bit wrong (eg, it's hard to stub them)

internal/services/pubkey_service.go
lzap: Absolutely makes sense, this is exactly what Bind/Render functions are for! It was just one-liner. Dropped this unused line too.

internal/db/error_codes.go
lzap: Oh yeah, good point. It looks much better now.

internal/models/transformer_test.go
lzap: Yeah I thought about this too, but the general practice is to create xyz_test.go for each xyz.go file. And it looks like you set the standard in validator to put all the code into the single file. Thus my intention is to keep this - all transformation functions to be in transformer.go thus the tests belong to transformer_test.go. I feel like this is okay since we only have a single function for each, we can break this down later.

internal/payloads/error_payload.go
lzap: Nah I was just lazy, 422 it is.

internal/services/pubkey_service.go
lzap: Fact is, it already is a "nice error". We are taking advantage of error wrapping here, the `IsPostgresError` function actually drills down the wrapped errors until it finds the root cause (pg error). But what you actually have is `sqlx.NewCreateError` (our own type from dao/sqlx).

api/openapi.gen.yaml
```diff
@@ -173,6 +175,8 @@ components:
       properties:
         body:
           type: string
+        fingerprint:
+          type: string
         id:
           format: int64
           type: integer
```
lzap, line 182: QA: We are adding new field "fingerprint" but it is read-only, if you try to set it to any value it will be ignored.

Also the description describes that pubkeys now must be unique: both name and pubkey itself (the public key body). If it's not, error 422 is returned.

akhil-jha:
@lzap I am getting 500 error on POST /pubkeys
Is there some requirements to have prior doing a POST call? 

```
6:42AM ERR invalid request: EOF account_number=<account number> bytes_in=0 hostname=provisioning-backend-service-6b946d7864-vpxwv method=POST org_id=<ord id> remote_ip=<remote ip> rid=cbpl6sv240flhi1om690 rn=147 url=/api/provisioning/v1/pubkeys
6:42AM INF Completed POST request /api/provisioning/v1/pubkeys in 64ms ms with 500 bytes_in=0 bytes_out=67 hostname=provisioning-backend-service-6b946d7864-vpxwv latency_ms=63.933536 method=POST remote_ip=<remote ip> rid=cbpl6sv240flhi1om690 rn=147 status=500 url=/api/provisioning/v1/pubkeys
```

internal/models/transformer_test.go
ezr-ondrej: right, make sense, lets see how it goes once we actually add another transformations/validations :)

ezr-ondrej (approved):
Code looks good, just drop the Makefile change pls as I believe we already have that in main :)

ezr-ondrej:
> Is there some requirements to have prior doing a POST call?

I don't think there is any requirement for this call :thinking:
Is that a true positive error @lzap ? :thinking:

lzap:
> Is that a true positive error @lzap ?

Absolutely, the implementation is easy it just validates the data and stores it in our DB. Not sure why it fails for you. We can investigate.

lzap:
Rebased.

lzap:
My wild guess is your POST has no data because: `invalid request: EOF`

You need to provide JSON in the body, also make sure to set the Content-Type:

```
// @no-log
POST http://{{hostname}}:{{port}}/{{prefix}}/pubkeys/ HTTP/1.1
X-Rh-Identity: {{identity}}

{
  "account_id": 1,
  "name": "lzap-rsa-2010-{{$randomInt}}",
  "body": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC8w6DONv1qn3IdgxSpkYOClq7oe7davWFqKVHPbLoS6+dFInru7gdEO5byhTih6+PwRhHv/b1I+Mtt5MDZ8Sv7XFYpX/3P/u5zQiy1PkMSFSz0brRRUfEQxhXLW97FJa7l+bej2HJDt7f9Gvcj+d/fNWC9Z58/GX11kWk4SIXaKotkN+kWn54xGGS7Zvtm86fP59Srt6wlklSsG8mZBF7jVUjyhAgm/V5gDFb2/6jfiwSb2HyJ9/NbhLkWNdwrvpdGZqQlYhnwTfEZdpwizW/Mj3MxP5O31HN45aE0wog0UeWY4gvTl4Ogb6kescizAM6pCff3RBslbFxLdOO7cR17 lzap+rsakey+{{$randomInt}}@redhat.com"
}
```

lzap:
@ezr-ondrej can we merge DAO tests refactoring first? Then I will rebase this on top of that, there are test changes in here.

lzap:
https://github.com/RHEnVision/provisioning-backend/pull/149

lzap:
Rebased.

api/openapi.gen.json
```diff
@@ -61,6 +61,9 @@
           "body": {
             "type": "string"
           },
+          "fingerprint": {
```
akhil-jha, line 64: Do you think the spec also needs the following in POST /pubkeys ?
```
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/v1.Pubkey"
              }
            }
          }
```

lzap: We have an error in the paths.yaml so the payload is currently incorrect. This fixes it: https://github.com/RHEnVision/provisioning-backend/pull/161

lzap: If we merge both, you should be fine.

akhil-jha: I think merging #161 and rebasing this PR would be correct.

lzap: Sure, we need an ACK from you there :)

lzap:
Rebased.

api/openapi.gen.json
mshriver: I don't think #161 alone addresses it. @akhil-jha is adding schema to the requestBody for POST, 161 only addresses the response body

lzap: Oh right, asked @ezr-ondrej to fix this properly.

lzap: Rebased on top of the input body change in OpenAPI.

lzap:
You know what, I slightly elaborated the documentation of the POST body :-)

lzap:
All is passing, no other changes than the description of the body field in OpenAPI.

api/openapi.gen.json
ezr-ondrej: @akhil-jha did it help, are we able to test this now? :)

ezr-ondrej (approved):

api/openapi.gen.json
akhil-jha: I'm creating the iqe image now. But i think this should work :)
I'll update it here.

akhil-jha:
@lzap 
Working:
Able to create pubkeys and get any pubkey by the {ID} :)

Not working:
**Able to create duplicates.**
```
$ curl -L --request POST --user <username>:<password> https://<hostname>/api/provisioning/v1/pubkeys --data '{"name": "akjha-fake-key", "body": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"}'

{"id":8,"name":"akjha-fake-key","body":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"}

$ curl -L --request POST --user <username>:<password> https://<hostname>/api/provisioning/v1/pubkeys --data '{"name": "akjha-fake-key", "body": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"}'

{"id":9,"name":"akjha-fake-key","body":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"}

$ curl -L --request POST --user <username>:<password> https://<hostname>/api/provisioning/v1/pubkeys --data '{"name": "akjha-fake-key", "body": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"}'

{"id":10,"name":"akjha-fake-key","body":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"}
```
**Listing the pubkeys**
```
$ curl -L --request GET --user <username>:<password> https://<hostname>/api/provisioning/v1/pubkeys 
[{"id":8,"name":"akjha-fake-key","body":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"},
{"id":9,"name":"akjha-fake-key","body":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"},
{"id":10,"name":"akjha-fake-key","body":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEbKq5U57fhzQ3SBbs3NVmgY2ouYZfPhc6cXBNEFpRT3T100fnbkYw+EHi76nwsp+uGxk08kh4GG881DrgotptrJj2dJxXpWp/SFdVu5S9fFU6l6dCTC9IBYYCCV8PvXbBZ3oDZyyyJT7/vXSaUdbk3x9MeNlYrgItm2KY6MdHYEg8R994Sspn1sE4Ydey5DfG/WNWVrzFCI0sWI3yj4zuCcUXFz9sEG8fIYikD9rNuohiMenWjkj6oLTwZGVW2q4wRL0051XBkmfnPD/H6gqOML9MbZQ8D6/+az0yF9oD61SkifhBNBRRNaIab/Np7XD61siR8zNMG/vCKjFGICnp"}]
```

lzap:
Hey @akhil-jha can you check if you really test against this PR? Because it works just fine here. Also, in your GET output when listing pubkeys, I do not see a new field that has been added named "fingerprint". See my output:

```
http://localhost:8000/api/provisioning/v1/pubkeys/

HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8
Date: Thu, 18 Aug 2022 08:36:08 GMT
Content-Length: 720

[
  {
    "id": 1,
    "name": "lzap-ed25519-2021",
    "body": "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap",
    "fingerprint": "SHA256:gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk"
  },
  {
    "id": 2,
    "name": "lzap-rsa-2010-405",
    "body": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC8w6DONv1qn3IdgxSpkYOClq7oe7davWFqKVHPbLoS6+dFInru7gdEO5byhTih6+PwRhHv/b1I+Mtt5MDZ8Sv7XFYpX/3P/u5zQiy1PkMSFSz0brRRUfEQxhXLW97FJa7l+bej2HJDt7f9Gvcj+d/fNWC9Z58/GX11kWk4SIXaKotkN+kWn54xGGS7Zvtm86fP59Srt6wlklSsG8mZBF7jVUjyhAgm/V5gDFb2/6jfiwSb2HyJ9/NbhLkWNdwrvpdGZqQlYhnwTfEZdpwizW/Mj3MxP5O31HN45aE0wog0UeWY4gvTl4Ogb6kescizAM6pCff3RBslbFxLdOO7cR17 lzap+rsakey+679@redhat.com",
    "fingerprint": "SHA256:ENShRe/0uDLSw9c+7tc9PxkD/p4blyB/DTgBSIyTAJY"
  }
]
```

akhil-jha:
> Hey @akhil-jha can you check if you really test against this PR? Because it works just fine here. Also, in your GET output when listing pubkeys, I do not see a new field that has been added named "fingerprint". See my output:
> 
> ```
> http://localhost:8000/api/provisioning/v1/pubkeys/
> 
> HTTP/1.1 200 OK
> Content-Type: application/json; charset=utf-8
> Date: Thu, 18 Aug 2022 08:36:08 GMT
> Content-Length: 720
> 
> [
>   {
>     "id": 1,
>     "name": "lzap-ed25519-2021",
>     "body": "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap",
>     "fingerprint": "SHA256:gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk"
>   },
>   {
>     "id": 2,
>     "name": "lzap-rsa-2010-405",
>     "body": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC8w6DONv1qn3IdgxSpkYOClq7oe7davWFqKVHPbLoS6+dFInru7gdEO5byhTih6+PwRhHv/b1I+Mtt5MDZ8Sv7XFYpX/3P/u5zQiy1PkMSFSz0brRRUfEQxhXLW97FJa7l+bej2HJDt7f9Gvcj+d/fNWC9Z58/GX11kWk4SIXaKotkN+kWn54xGGS7Zvtm86fP59Srt6wlklSsG8mZBF7jVUjyhAgm/V5gDFb2/6jfiwSb2HyJ9/NbhLkWNdwrvpdGZqQlYhnwTfEZdpwizW/Mj3MxP5O31HN45aE0wog0UeWY4gvTl4Ogb6kescizAM6pCff3RBslbFxLdOO7cR17 lzap+rsakey+679@redhat.com",
>     "fingerprint": "SHA256:ENShRe/0uDLSw9c+7tc9PxkD/p4blyB/DTgBSIyTAJY"
>   }
> ]
> ```

Hmm....Let me create new images then.

ezr-ondrej:
Now it worked I believe and posted new tag `pr-135-3d0ed7b`, right?

lzap:
FYI I just added version info into the HTTP headers and logs via this patch which should ease to spot these situations in the future: https://github.com/RHEnVision/provisioning-backend/pull/168

akhil-jha:
> FYI I just added version info into the HTTP headers and logs via this patch which should ease to spot these situations in the future: #168
> 
> If that's merged and I rebase this on top of that, you can tell from the HTTP header what SHA commit are you actually testing.

That would be very helpful!!

Also, yes, I was using wrong image for testing. :disappointed: 
It's working now. 
Giving an ACK and removing the qe label here.
If you want to rebase or merge, please do it :)

akhil-jha (approved):

lzap:
Okay thanks, so I see @adiabramovitch had a review previously. Can you do the final blessing and merge?

PR #136, opened 2022-08-02
Fix AccountNumber type to NullString

avitova:


internal/dao/sqlx/account_dao.go
```diff
@@ -76,7 +74,7 @@ func (di *accountDaoSqlx) Create(ctx context.Context, account *models.Account) e
 	query := createAccount
 	stmt := di.create
 
-	err := stmt.GetContext(ctx, account, account.AccountNumber, account.OrgID)
+	err := stmt.GetContext(ctx, account, account.AccountNumber.String, account.OrgID)
```
avitova, line 77: Should we validate for x.Valid?

lzap: Not here, NULL is valid here.

internal/dao/sqlx/account_dao.go
```diff
@@ -113,7 +111,7 @@ func (di *accountDaoSqlx) GetOrCreateByIdentity(ctx context.Context, orgId strin
 	} // else it is NoRows and we just continue
 	ctxval.Logger(ctx).Debug().Msgf("Account not found by account number %s", accountNumber)
 
-	acc = &models.Account{OrgID: orgId, AccountNumber: ptr.String(accountNumber)}
+	acc = &models.Account{OrgID: orgId, AccountNumber: sql.NullString{String: accountNumber, Valid: true}}
```
lzap, line 114: Here this code needs to actually make use of NULL:

* When accountNumber is blank (https://github.com/RedHatInsights/identity/blob/main/identity.go#L48 - it is not a pointer so the only option is blank) it needs to set Valid to false.
* Otherwise it needs to create new struct with Valid = true.

internal/dao/sqlx/account_dao.go
lzap: But the problem is, you need to pass the AccountNumber struct, not string. The sql/sqlx library will take care of storing either NULL or value.

internal/dao/stubs/account_dao.go
```diff
@@ -88,7 +88,7 @@ func (stub *accountDaoStub) GetByOrgId(ctx context.Context, orgId string) (*mode
 
 func (stub *accountDaoStub) GetByAccountNumber(ctx context.Context, number string) (*models.Account, error) {
 	for _, acc := range stub.store {
-		if *acc.AccountNumber == number {
+		if acc.AccountNumber.Valid && acc.AccountNumber.String == number {
 			return acc, nil
 		}
 	}
```
lzap, line 94: Can you add a test when account number is actually NULL and test that? For both insert and select statements.

lzap (changes requested):
We need more tests.

internal/dao/stubs/account_dao.go
avitova: Tests are in this PR: https://github.com/RHEnVision/provisioning-backend/pull/127/files
I could rebase and make one PR out of these two, but I thought maybe it might be better idea to keep these two separated.

ezr-ondrej (approved):
LGTM, tests in the other PR make sense, given we do not have a test file for accounts yet, I'd not add all tests in this PR.

ezr-ondrej:
oh yeah, API spec does not understand the new type... but given the account endpoint is definitelly not user facing, I'd just go with it.
So could you pls run `make generate-spec` and commit the change?

api/openapi.gen.yaml
```diff
@@ -156,8 +156,7 @@ components:
   schemas:
     v1.Account:
       properties:
-        account_number:
-          type: string
+        account_number: {}
```
lzap, line 159: This is not right, we want to either expose the whole String/Valid struct or preferably make our own marhsaler because JSON does support null values natively. Let me explain in the review comment.

internal/dao/stubs/account_dao.go
lzap: I would rather see unit tests executing all edge cases (nulls, empty values etc) and integration tests to focus on SQL statements and data. Also, I would like to ensure this patch is correct as part of the PR.

lzap (changes requested):
Ok looks good except JSON handling.

Those SQL types do not have JSON marhsaling implemented. But there is a very nice way how to do it. It is explained here:

https://medium.com/aubergine-solutions/how-i-handled-null-possible-values-from-database-rows-in-golang-521fb0ee267

Can you copy and paste the marhaling types from the article (the updated gist is here: https://gist.github.com/rsudip90/022c4ef5d98130a224c9239e0a1ab397). I think int64, bool, string and time are enough (we won't use float for sure). Put it into internal/db/json_null.go or something like that.

And then only change the types from `sql.NullString` to `db.NullString` and regenerate OpenAPI. It should again be `string` type but this time it will actually work with JSON `null` values too. That's exactly what we want.

internal/models/account_model.go
```diff
@@ -9,5 +13,5 @@ type Account struct {
 	OrgID string `db:"org_id" json:"org_id"`
 
 	// EBS account number. Can be NULL but not blank.
-	AccountNumber *string `db:"account_number" json:"account_number"`
+	AccountNumber db.NullString `db:"account_number" json:"account_number"`
```
lzap, line 16: Can you try this:

```suggestion
	AccountNumber db.NullString `db:"account_number" json:"account_number,string"`
```

Then regenerate the OpenAPI to see if it picks up the string type here. Then it should work, but my expectations are low :)

avitova: This does not help either.

lzap: Damn, okay, apologies for this extra work I really hoped it will work out.

Please remove embedding of models.Account in AccountPayload and simply copy ID/OrgID there, use `AccountNumber *string` in the payload type and `AccountNumber sql.NullString` in the model type. Move json go struct tags from the model to the payload as model will no longer be serializable. And finally, update NewAccountResponse and NewAccountListResponse functions to convert between the two types.

Converting sql.NullString to *string will be reused, so create payloads/null_convert.go file with a helper function that can be used in other payloads as well.

internal/db/json_null.go
```diff
@@ -0,0 +1,53 @@
...
+
+type NullInt64 struct {
+	sql.NullInt64
+}
+
+func (ni *NullInt64) MarshalJSON() ([]byte, error) {
+	if !ni.Valid {
+		return []byte("null"), nil
+	}
+	return json.Marshal(ni.Int64)
```
adiabramovitch, line 18: I think you should address the error returning from the Marshel function :)

internal/db/json_null.go
```diff
@@ -0,0 +1,53 @@
...
+
+type NullBool struct {
+	sql.NullBool
+}
+
+func (nb *NullBool) MarshalJSON() ([]byte, error) {
+	if !nb.Valid {
+		return []byte("null"), nil
+	}
+	return json.Marshal(nb.Bool)
```
adiabramovitch, line 29: ditto

internal/db/json_null.go
```diff
@@ -0,0 +1,53 @@
...
+
+type NullString struct {
+	sql.NullString
+}
+
+func (ns *NullString) MarshalJSON() ([]byte, error) {
+	if !ns.Valid {
+		return []byte("null"), nil
+	}
+	return json.Marshal(ns.String)
```
adiabramovitch, line 40: ditto

internal/middleware/account_test.go
```diff
@@ -59,7 +58,56 @@ func TestAccountMiddleware(t *testing.T) {
 			assert.Equal(t, "124", acc.OrgID)
 		})
 
-		handler := middleware.AccountMiddleware(isAccInNext)
+		handler := AccountMiddleware(isAccInNext)
+		handler.ServeHTTP(rr, req)
+
+		count, err := stubs.AccountStubCount(ctx)
+		assert.Nil(t, err, fmt.Sprintf("Error creating a new request: %v", err))
```
ezr-ondrej, line 65: Why are we adding these lines? I do not think it is correct

avitova: You are right, these should not be here.

internal/middleware/account_test.go
```diff
@@ -1,15 +1,15 @@
-package middleware_test
+package middleware
```
ezr-ondrej, line 1: I'd love to keep this and not discuss usage of `_test` in this PR, there is a ticket where we can discuss usage of this https://issues.redhat.com/browse/HMSPROV-163

lzap: I am good with this, I only requested this because I thought for a moment this is causing the issue.

internal/payloads/account_payload.go
```diff
@@ -9,7 +9,14 @@ import (
 )
 
 type AccountPayload struct {
-	*models.Account
+	// Required auto-generated PK.
+	ID int64 `db:"id" json:"id"`
```
lzap, line 13: This struct will never be stored in the db, so don't tag it please:

```suggestion
	ID int64 `json:"id"`
```

The same applies below.

internal/payloads/null_convert.go
```diff
@@ -0,0 +1,11 @@
+package payloads
+
+import "database/sql"
+
+func NullConvert(s sql.NullString) *string {
```
lzap, line 5: Perhaps this is more future-proof:

```suggestion
func SqlNullToStringPtr(s sql.NullString) *string {
```

We will be likely adding few more.

internal/payloads/null_convert.go
```diff
@@ -0,0 +1,11 @@
...
+	str := ""
+	if s.Valid {
+		str = s.String
+	}
+	return &str
+}
```
lzap, line 11: So the idea is to return `nil` when the SQL value is `NULL` not empty string. It prehaps makes sense for string, but once we add integers, we simply cannot use 0 as a value for `NULL`. To be consistent across types and because you use string pointer (not string) in the whole patch, this needs to return `nil` for invalid SQL value.

lzap (changes requested):
One major thing is that we need to return `nil` not empty string on `NULL` values, that's why you actually use string pointer everywhere. This will map to JSON "null" value.

During parsing in the middleware, please respect the following cases as a SQL `NULL` value:

* empty string (the current version)
* JSON "null" value (this is a native null not a string containing "null")
* missing key (when the hash is simply missing)

lzap (approved):
Perfect! This was challenging, great work.

```json
[
  {
    "id": 1,
    "org_id": "000013",
    "account_number": "13"
  },
  {
    "id": 2,
    "org_id": "000015",
    "account_number": "15"
  },
  {
    "id": 3,
    "org_id": "000042",
    "account_number": null
  },
  {
    "id": 4,
    "org_id": "000077",
    "account_number": null
  },
  {
    "id": 5,
    "org_id": "000016",
    "account_number": "6089719"
  }
]
```

PR #137, opened 2022-08-03
Use arbitrary path in middleware test

ezr-ondrej:
Middleware tests do not use the request path.
Having meaningful pathcan be confusing at times.
Also uses t.Errorf for erroring the test out

### DIFF
--- a/internal/middleware/account_test.go
+++ b/internal/middleware/account_test.go
@@ -2,7 +2,6 @@ package middleware_test
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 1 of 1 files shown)


internal/middleware/account_test.go
```diff
@@ -21,9 +20,9 @@ func TestAccountMiddleware(t *testing.T) {
 		ctx := stubs.WithAccountDaoOne(context.Background())
 		ctx = identity.WithTenant(t, ctx)
 
-		req, err := http.NewRequestWithContext(ctx, "GET", "/api/provisioning/v1/pubkeys", nil)
+		req, err := http.NewRequestWithContext(ctx, "GET", "/test", nil)
```
adiabramovitch, line 23: If it is an account test I would change it `/account_test` but not so important 😄

ezr-ondrej: well I wanted to be clear that the url has no meaning, `/account_test` could still imply we have some endpoint set up for testing account, so I'd keep it general to really underline it does nothing at all.

ezr-ondrej: Unless we want `/useless_path`, but I thought that being bit on the nose xD

adiabramovitch (approved):
LGTM

PR #138, opened 2022-08-03
Improve configuration for integration tests

lzap:
This patch fixes how DAO tests are configured. Instead loading the all-commented configs/default.yaml this is now removed. Tests now start with default values (which will try to connect to localhost). To override this behavior (e.g. when you have remote pg host like me), simply create `./cmd/dao_tests/local.yaml` with custom settings.

Moreover, the test suite now runs in "integration" schema, just for safety reasons in case someone runs the tests against the development database. It will not remove the public schema so all development data will be kept untouched.

There is another commit which should serve as an inspiration how to use testify/suite to write a little bit nicer tests. Feel free to elaborate this into something more robust. I can think about creating a common DAOSuite that will have the common setup/teardown functions and embedding this into specific DAO tests (pubkeys, accounts) - like an OOP approach. Or you can copy and paste the suite for all DAO interfaces.

@ezr-ondrej @avitova


cmd/dao_tests/local.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
ezr-ondrej, line 1: did you mean to commit this file? :thinking:

README.md
```diff
@@ -17,7 +17,7 @@ make build
 
 ## Configuration
 
-Configuration is done via configuration files in `configs/` directory, see [defaults.yaml](configs/defaults.yaml) file for list of options with documentation. Environmental variables override the values from configuration files and always take precedence.
+Configuration is done via configuration files in `configs/` directory, see [defaults.yaml](configs/local_example.yaml) file for list of options with documentation. Environmental variables override the values from configuration files and always take precedence.
```
ezr-ondrej, line 20:
```suggestion
Configuration is done via configuration files in `configs/` directory, see [local_example.yaml](configs/local_example.yaml) file for list of options with documentation. Environmental variables override the values from configuration files and always take precedence.
```

cmd/dao_tests/local.yaml
lzap: Oh no, luckily it only contains my hostname and database name :)

internal/config/5_real_config.go
```diff
@@ -27,17 +27,15 @@ func init() {
 	} else {
 		viper.SetDefault("featureFlags.environment", "development")
 
+		// load from main subdirectory
 		viper.AddConfigPath("./configs")
-		viper.SetConfigName("defaults")
+		// load from working directory for integration tests
+		viper.AddConfigPath(".")
```
ezr-ondrej, line 33: I do not see the link added in this PR, so I do not know for sure, but I think we do not need this if I'm correct.

ezr-ondrej (commented):
I see part of the config change here and part of it in https://github.com/RHEnVision/provisioning-backend/pull/135 I believe it works together, but it is very hard to check being in two PRs.

If someone has more context, feel free to merge, but reading changes cross PR is just too hard for me :(

Also splitting the files in integration tests trips up the linter, which does not see the entry point.

cmd/dao_tests/pubkey_test.go
```diff
@@ -1,191 +1,121 @@
...
 	"testing"
 
-	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
-	"github.com/RHEnVision/provisioning-backend/internal/db"
-	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
```
ezr-ondrej, line 15: I feel this package is purely implemented and TestMain is superior tool for the job (except the Suite context), can you share briefly why have you decided to go this way?

avitova (commented):
By the way, the Create method does not return anything in it's normal flow, which does not really help testing (if we will be using SELECT reset_sequences(); IDs are going to be incremented even if we are going to delete the records in between the tests).
How would you approach this? We need the id at least for GetById methods. We could always List the database and get the ID from the list, but there might be a better solution (maybe we should reset the id sequence?).

cmd/dao_tests/pubkey_data.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+func pubkeyAvitova1() *models.Pubkey {
+	return &models.Pubkey{
+		ID:        1,
+		AccountID: 1,
+		Name:      "avitova-ed25519-2021",
+		Body:      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN avitova",
+	}
+}
```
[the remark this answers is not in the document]
ezr-ondrej: I like this :+1:

lzap:
I rebased and split the test suite, this PR only contains the improved configuration.

cmd/dao_tests/pubkey_test.go
```diff
@@ -29,8 +31,9 @@ func createPk() *models.Pubkey {
 
 func InitTestEnvironment() error {
 	config.Initialize()
+	logging.InitializeStdout()
```
lzap, line 34: This line actually silences the logging depending on the setting in `local.yaml`. You can enable or disable it, by default it is disabled tho which is much cleaner on CI.

In the other PR I want to take a look on saving all logs into buffer and showing it only on failures.

lzap:
Can you share how fast is this for you? My database is remote through gigabyte ethernet so it's I think much slower than on localhost:

```
=== RUN   TestCreatePubkey
--- PASS: TestCreatePubkey (0.44s)
=== RUN   TestListPubkey
--- PASS: TestListPubkey (0.36s)
=== RUN   TestUpdatePubkey
--- PASS: TestUpdatePubkey (0.36s)
=== RUN   TestGetPubkeyById
--- PASS: TestGetPubkeyById (0.38s)
=== RUN   TestDeletePubkeyById
--- PASS: TestDeletePubkeyById (0.40s)
=== RUN   TestCreateNoop
--- PASS: TestCreateNoop (0.46s)
=== RUN   TestCreateAWS
--- PASS: TestCreateAWS (0.49s)
=== RUN   TestCreateInstance
--- PASS: TestCreateInstance (0.51s)
=== RUN   TestListReservation
--- PASS: TestListReservation (0.45s)
=== RUN   TestUpdateReservationIDForAWS
--- PASS: TestUpdateReservationIDForAWS (0.53s)
=== RUN   TestUpdateStatusReservation
--- PASS: TestUpdateStatusReservation (0.47s)
=== RUN   TestDeleteReservation
--- PASS: TestDeleteReservation (0.47s)
=== RUN   TestFinishReservation
--- PASS: TestFinishReservation (0.53s)
PASS
```

ezr-ondrej:
I'm getting the logger output about migrations, not test names with times :thinking:

lzap:
Oh in that case just do this:

```
go test -tags=integration ./cmd/dao_tests
ok  	github.com/RHEnVision/provisioning-backend/cmd/dao_tests	6.089s
```

I will update the Makefile, it is no longer needed to compile the binary because of the working directory, my change actually fixed this problem.

lzap:
> Create method does not return anything in it's normal flow,

The idea is to modify the argument you pass in instead of returning a value. If you want to see the database ID, just check the ID field on the value you passed in. That is how SQL/SQLX libraries do it.

configs/local_example.yaml
```diff
@@ -0,0 +1,62 @@
...
+  # 1-none, 2-error, 3-warn, 4-info, 5-debug, 6-trace
+  loglevel: 1
+
+logging:
+  # Overall log level (-1 trace, 0 debug, 1 info ... 5 panic)
+  level: 1
+  # Write logs to standard output in human-readable form
+  stdout: false
+
+AWS provisioning account (EC2, STS...)
```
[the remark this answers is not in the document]
lzap: Thanks good catch.

cmd/dao_tests/pubkey_test.go
```diff
@@ -1,191 +1,121 @@
...
+func (suite *PubkeySuite) TearDownTest() {
+	commonTearDownTest()
+}
 
-	err = db.Migrate()
-	if err != nil {
-		t.Errorf("Error running migration: %v", err)
-		return
-	}
+func TestDAOSuite(t *testing.T) {
```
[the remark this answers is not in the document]
lzap: Yes, I actually removed the commit from this PR, let's discuss this elsewhere. I will add an example with two files to see how we can approach this.

lzap:
> I'm getting the logger output about migrations, not test names with times

Can you try to create `cmd/dao_tests/local.yaml`:

```yaml
---
logging:
  stdout: false
```

lzap:
I am pushing a small change that makes stdout logger output much more readable, this will improve CI as well. It is no longer JSON but text with key=value formatting and long values are truncated not to be longer than 40 characters.

ezr-ondrej (commented):
so this basically adds two ways of separate DB for tests?
1. by different config file
2. by different schema

Just clarifying :)

I've had an issue with this, because of the missing extension `uuid-ossp`, I've managed to fix it by reinstalling the `postgres-contrib` and running `CREATE EXTENSION IF NOT EXISTS "uuid-ossp" schema pg_catalog version "1.1";` but I wonder if we shouldn't do this in the migrations?
I suspect that the issue was having new schema, where the extension was not added, but might be different issue. I don't even think we shoud block on it, but just making sure it is not something trivial we forgotten to address.

lzap:
Yeah the config solves the issue and the schema is an extra protection in case you don't drop custom config and let the tests to run in your dev database.

Good point, both me and @avitova saw this weird uuid extension error which disappeared after the first migration. I am not sure why this happens tho. We have the statement in `006_jobqueue.sql` and it loads it into the current schema (either public or integration). I will try to reproduce it again by deleting it.

lzap:
Ok to my surprise I have learned today that postgres extensions can be only installed once, meaning that we have two options:

1) Install it into `pg_catalog` which is by default in the search path (it's hidden but it's always there) therefore all schemas will see the installed functions.

2) The proper way of creating special schema and then setting search path for each connection. This should be possible (https://github.com/jackc/pgx/issues/1013) but I find it too much of burden to carry just because we will need one or two extensions.

So I vote for (1), this is how exactly @ezr-ondrej solved the problem. I rebased, please try again. The only drawback is that all developers must now drop the database or issue this command:

```
ALTER EXTENSION "uuid-ossp" SET SCHEMA "pg_catalog";
```

In order to move the extension to the correct schema (it is currently in `public`). Only after this command is executed, the tests will work again.

I did not see this error because I use a separate database for integration tests, there is no issue there because the database is dedicated to integration tests therefore the extension was installed in the correct schema.

ezr-ondrej (approved):
Thanks for the investigation and writup @lzap, thanks for testing @adiabramovitch ! :+1:

PR #142, opened 2022-08-04
Move CICD files into .rhcicd dir

mshriver:
Consistency with frontend app, and nice to collect them and not have them in root


mshriver:
ci.int pipeline is going to fail without this being merged, because the pipeline is using the existing config. The app-interface change and this PR need to be merged together

ezr-ondrej:
Can we link from old file location to the new one for now and remove the link once we get the app-interface change merged? ;)

mshriver:
app-interface MR is up and should be merged pretty quickly, don't think its worth linking. 

https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/44645/diffs

ezr-ondrej:
> and should be merged pretty quickly

I think that is a brave statement :)
I believe link is cheap and it is not an issue to keep it around for some time if we forget it ;)

mshriver:
3hrs later, starting to sweat.  pushed a commit with symlinks.

ezr-ondrej (approved):
I hope symlinks will do for the bot, but it should, let see :)

PR #143, opened 2022-08-08
Remove pubkey from reservation endpoint

lzap:
Also move pubkey upload to a job.

I am currently having issues with my Role/STS setup (getting permission denied). Please test the patch carefully :)

### DIFF
--- a/cmd/dao_tests/pubkey_test.go
+++ b/cmd/dao_tests/pubkey_test.go
@@ -73,19 +73,20 @@ func TestCreatePubkey(t *testing.T) {
 		t.Errorf("Database setup had failed: %v", err)
 		return
 	}
-	err = pkDao.Create(ctx, createPk())
+	pk := createPk()
+	err = pkDao.Create(ctx, pk)
 	if err != nil {
 		t.Errorf("Create pubkey test had failed: %v", err)
 		return
 	}
 
-	pubkeys, err := pkDao.List(ctx, 10, 0)
+	pk2, err := pkDao.GetById(ctx, pk.ID)
 	if err != nil {
 		t.Errorf("Create pubkey test had failed: %v", err)
 		return
 	}
 
-	assert.Equal(t, 1, len(pubkeys), "Create pubkey test had failed.")
+	assert.Equal(t, pk.Name, pk2.Name, "Create pubkey test had failed.")
 }
 
 func TestListPubkey(t *testing.T) {
@@ -101,7 +102,7 @@ func TestListPubkey(t *testing.T) {
 		t.Errorf("List pubkey test had failed: %v", err)
 		return
 	}
-	assert.Equal(t, 1, len(pubkeys), "List Pubkey error.")
+	assert.Equal(t, 2, len(pubkeys), "List Pubkey error.")
 }
 
 func TestUpdatePubkey(t *testing.T) {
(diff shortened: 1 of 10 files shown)


ezr-ondrej (commented):
I'm good code wise :+1: linter ain't tho, could you make linter happy so we can test it out? :)

internal/db/seeds/dao_test.sql
```diff
@@ -7,4 +7,9 @@ INSERT INTO accounts(id, account_number, org_id)
 VALUES (1, '1', '1')
   ON CONFLICT DO NOTHING;
 
+INSERT INTO pubkeys(id, account_id, name, body)
+VALUES (1, 1, 'lzap-ed25519-2021',
+        'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap')
+ON CONFLICT DO NOTHING;
+
```
ezr-ondrej, line 14: We need to reset sequences here if we want to add the key.
But do we? Shouldn't we figure out how to use some factories?

lzap: Oh of course, good catch. While I think static IDs might be useful in tests even more than in dev environment, I will keep what we have today and we can add this later if needed. This could possibly slow down seeding quite a bit when we have like 20 tables in the end.

lzap: You know what, let's do reset_sequences here and if DAO tests are slow then let's use implicit IDs instead. Since we will use IDs (typically only 1) everywhere, I feel like being explicit is more clean.

ezr-ondrej: I'm fine with that :+1:

cmd/dao_tests/pubkey_test.go
```diff
@@ -101,7 +102,7 @@ func TestListPubkey(t *testing.T) {
 		t.Errorf("List pubkey test had failed: %v", err)
 		return
 	}
-	assert.Equal(t, 1, len(pubkeys), "List Pubkey error.")
+	assert.Equal(t, 2, len(pubkeys), "List Pubkey error.")
```
lzap, line 105: I have some comments about these tests, many of them do not necessary assert correct things. For example here, the function could store record with all NULLs (that would be incorrect) but this would pass since it only test that "something" was added. I think the correct test is to literally take the listed elements, find the entry and compare it to the inserted entry literally using `==`.

But let's not tackle this in this PR, these tests need more changes.

lzap:
Finally it passes :)

ezr-ondrej (approved):
I think in the future we could go even further and wrap enqueing both of the jobs under `jobs.EnqueueAWSProvisioning` (or whatever the name).
But this is a great progress and the service method looks much cleaner now :+1:

scripts/rest_examples/reservation-create-aws.http
```diff
@@ -1,15 +1,13 @@
...
-  "source_id": 9, 
+  "source_id": 1,
   "ami": "ami-0cff7528ff583bf9a",
   "amount": 1,
   "instance_type": "t1.micro",
-  "pubkey": {
-    "existing_id": 1
-  }
+  "pubkey_id": 1
 }
```
lzap, line 13: All: Do this change in your .http files. Sorry! :)

internal/db/migrations/005_reservations.sql
```diff
@@ -21,7 +21,7 @@ CREATE TABLE aws_reservation_details
 (
   reservation_id BIGINT NOT NULL,
   provider INTEGER NOT NULL DEFAULT provider_type_aws(),
-  pubkey_id BIGINT REFERENCES pubkeys(id),
+  pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
```
lzap, line 24: Remember to drop and remigrate the database! Sorry! :)

PR #145, opened 2022-08-09
Generalize testing factories

ezr-ondrej:
Factories should be usable in unit and DAO tests alike.

Adding `testing/factories` package, that should host model generation methods for tests.

### DIFF
--- a/internal/services/pubkey_service_test.go
+++ b/internal/services/pubkey_service_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao/stubs"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/factories"
 	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
 	"github.com/stretchr/testify/assert"
 )
@@ -19,12 +20,25 @@ func TestListPubkeysHandler(t *testing.T) {
 	ctx := stubs.WithAccountDaoOne(context.Background())
 	ctx = identity.WithTenant(t, ctx)
 	ctx = stubs.WithPubkeyDao(ctx)
-	err := stubs.GeneratePubkey(ctx, models.Pubkey{})
-	assert.Nil(t, err, fmt.Sprintf("Error while generating a pubkey: %v", err))
-	err = stubs.GeneratePubkey(ctx, models.Pubkey{})
-	assert.Nil(t, err, fmt.Sprintf("Error while generating a pubkey: %v", err))
+	err := stubs.AddPubkey(ctx, &models.Pubkey{
+		Name: factories.GetSequenceName("pubkey"),
+		Body: factories.GenerateRSAPubKey(t),
+	})
+	if err != nil {
+		t.Fatalf("failed to add stubbed key: %v", err)
+	}
+	err = stubs.AddPubkey(ctx, &models.Pubkey{
+		Name: factories.GetSequenceName("pubkey"),
+		Body: factories.GenerateRSAPubKey(t),
+	})
+	if err != nil {
+		t.Fatalf("failed to add stubbed key: %v", err)
+	}
 
 	req, err := http.NewRequestWithContext(ctx, "GET", "/api/provisioning/pubkeys", nil)
+	if err != nil {
+		t.Fatalf("Error creating a new request: %v", err)
+	}
 	assert.Nil(t, err, fmt.Sprintf("Error creating a new request: %v", err))
 
 	rr := httptest.NewRecorder()
--- a/internal/testing/factories/factories.go
+++ b/internal/testing/factories/factories.go
@@ -1,0 +1,31 @@
+package factories
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"fmt"
+	"sync/atomic"
+	"testing"
+
+	"golang.org/x/crypto/ssh"
+)
+
+var nameSequence uint64 = 1
+
+func GetSequenceName(prefix string) string {
+	return fmt.Sprintf("%s %d", prefix, atomic.AddUint64(&nameSequence, 1))
+}
+
+// GenerateRSAPubKey generates pubkey for use in tests
+func GenerateRSAPubKey(t *testing.T) string {
+	t.Helper()
+	rsaKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		t.Fatalf("Failed to generate pubky: %v", err)
+	}
+	pubkey, err := ssh.NewPublicKey(&rsaKey.PublicKey)
+	if err != nil {
+		t.Fatalf("Failed to generate pubky: %v", err)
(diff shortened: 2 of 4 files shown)


adiabramovitch (approved):

adiabramovitch:
LGTM, but @lzap  Do you have any comments?

internal/testing/factories/factories.go
```diff
@@ -0,0 +1,50 @@
+package factories
+
+import (
+	"context"
+	"crypto/rand"
```
lzap, line 5: This is an overkill, you can use just rand package which gives pseudo random numbers instead. Tests could deplete good entropy just for nothing and a CI instance might get short of good entropy for no reason. The key generator works with any reader.

internal/testing/factories/factories.go
```diff
@@ -0,0 +1,50 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"golang.org/x/crypto/ssh"
+)
+
+var RSAGenerationError = errors.New("rsa key generation failed")
+
+var nameSequence = 1
```
lzap, line 16: Although we do not have parallel tests right now, I would definitely encourage to do defensive programming and use atomic integer here. I am all for running unit tests in parallel as soon as possible so we avoid bigger refactoring later.

internal/testing/factories/factories.go
```diff
@@ -0,0 +1,50 @@
...
+	if pubkey.Name == "" {
+		pubkey.Name = nextSequenceName("pubkey")
+	}
+	if pubkey.Body == "" {
+		if pubkey.Body, err = generateRandomRSAPubKey(ctx); err != nil {
+			return pubkey, err
+		}
+	}
+	return pubkey, nil
+}
```
lzap, line 33: The comment for this function does not appear to be correct, I don't see anything being stored "in the context stub".

More importantly, I don't like this function at all - it takes a pubkey as an argument by value, then changes some fields and then returns it by value again. What is the purpose of this function? I can't tell neither from the name nor from the body.

The other function `AddPubkey` has the argument named `template` so it looks like this was the intention, tho I find this weird. This function could easily be designed in a way it takes a reference to pubkey and only setting the Body appropriately.

internal/testing/factories/factories.go
```diff
@@ -0,0 +1,50 @@
...
+var RSAGenerationError = errors.New("rsa key generation failed")
+
+var nameSequence = 1
+
+// GeneratePubkey generates a pubkey and stores it in the context stub
+func GeneratePubkey(ctx context.Context, pubkey models.Pubkey) (models.Pubkey, error) {
+	var err error
+	if pubkey.AccountID == 0 {
+		pubkey.AccountID = 1
+	}
```
lzap, line 23: This is a side effect I would not expect and it will bite someone's ass pretty soon.

lzap (changes requested):
I don't like this. I get the point, but I think this is overengineered.

What do we want? Easily and consistently create records in stubs. A record is a struct, like everything else in Go. Structs can be easily created via `models.MyModel{}` and this is a really powerful syntax with default values (booleans are false, strings are empty etc), some fields can be initialized others not.

I would rather create a helper function something like `generateRandomPubkeyBody` and `generateUniqueName` that can be used easily:

```go
models.Pubkey{
  Name: generateUniqueName("pubkey"),
  Body: generateRandomPubkeyBody(),
  AccountID: 1,
}
```

The `generateUniqueName(prefix string)` can be shared across all models, it just needs to generate some increasing numbers and be thread safe.

This gives the best flexibility to the caller, yet it is very readable and idiomatic Go.

internal/testing/factories/factories.go
ezr-ondrej: could you pls point me to that package? I need a reader, which math/rand does not have and I think that crypto does not use enthropy by default, or does it?

internal/services/pubkey_service_test.go
```diff
@@ -19,12 +20,25 @@ func TestListPubkeysHandler(t *testing.T) {
...
 	ctx = identity.WithTenant(t, ctx)
 	ctx = stubs.WithPubkeyDao(ctx)
-	err := stubs.GeneratePubkey(ctx, models.Pubkey{})
-	assert.Nil(t, err, fmt.Sprintf("Error while generating a pubkey: %v", err))
-	err = stubs.GeneratePubkey(ctx, models.Pubkey{})
-	assert.Nil(t, err, fmt.Sprintf("Error while generating a pubkey: %v", err))
+	err := stubs.AddPubkey(ctx, &models.Pubkey{
+		Name: factories.GetSequenceName("pubkey"),
+		Body: factories.GenerateRSAPubKey(t),
+	})
```
ezr-ondrej, line 26: it's bit wordy just to generate a pubkey, but I guess it is fine :)

ezr-ondrej (commented):
@lzap I don't know what package you meant for random, but I've addressed all other comments.
It seems bit wordy to me, but looks more like Go for sure ^_^

internal/testing/factories/factories.go
lzap: I thought this would be sufficient: https://pkg.go.dev/math/rand

I *think* it has the same interface, if not, keep using cryptographic (e.g. much stronger) generator. It's a nitpick.

internal/testing/factories/factories.go
```diff
@@ -0,0 +1,31 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"fmt"
+	"sync/atomic"
+	"testing"
+
+	"golang.org/x/crypto/ssh"
+)
+
+var nameSequence uint32 = 1
```
lzap, line 13: Super nitpick - 32bit numbers are slower on 64bit systems while they take the same amount of memory. But no biggie 32bit ought to be enough for everybody (TM). ;)

lzap:
This looks WAY better, thanks! Dropped a comment about the crypto number.

Just for the record, if you don't initialize the math random generator (which you should not do for tests), it will always create the same sequence. I believe this is not an issue, if you really want a random sequence we can initialize it in https://github.com/RHEnVision/provisioning-backend/pull/159

internal/testing/factories/factories.go
ezr-ondrej: well, most HW we run on will be probably 64, so probably better, lets go with 64 ^_^

internal/testing/factories/factories.go
ezr-ondrej: yeah I've tried that one, but it does not have the function [`Read`](https://pkg.go.dev/crypto/rand@go1.19#Read) I actually use unfortunatelly... the interfaces are not the same (I'd guess as purpose is not the same, they just don't try).
If that's ok with you, I'd keep crypto

ezr-ondrej:
Fixed the uint64 :)

internal/testing/factories/factories.go
lzap: ALL of the hardware. My God do not play with fire! :)

lzap (approved):

PR #146, opened 2022-08-09
Change SourcesClientError to ClientError

adiabramovitch:
A suggestion to make the error more general so we could use it for image builder client and others

### DIFF
--- a/internal/payloads/error_payload.go
+++ b/internal/payloads/error_payload.go
@@ -55,8 +55,8 @@ func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
 	}
 }
 
-func SourcesClientError(ctx context.Context, message string, err error, status int) *ResponseError {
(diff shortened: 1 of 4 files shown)


ezr-ondrej (approved):
@adiabramovitch I think this is great addition and I'm ok to merge as-is, but it got me thinking:

Do you think that it would make sense to create general error on the client.sources side? maybe directly in the `clients` package, so we would have one error, with custom message instead of many errors, for which we need to branch by `error.Is` and only change the message, we could instead have one error type for any client related errors, where we woud just change the error message (similar to what we do in dao)

It is completely fine to do it as a follow-up :)

lzap (approved):

lzap:
> directly in the `clients` package

That is the ideal state, currently we do use `fmt.Errorf` which at least wraps all errors as it is required by the linter, but ideally we should define our own types with Error and Wrapper interfaces.

PR #147, opened 2022-08-09
HMSPROV-127 Delete Pubkey Resource

avitova:
This is not tested at all and is just a draft. There are no tests for DELETE. How should we test this?

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -391,15 +391,8 @@
           }
         ],
         "responses": {
-          "200": {
-            "content": {
-              "application/json": {
-                "schema": {
-                  "$ref": "#/components/schemas/v1.PubkeyResponse"
-                }
-              }
-            },
-            "description": "Returned on success."
+          "204": {
+            "description": "The Pubkey was deleted successfully."
           },
           "404": {
             "$ref": "#/components/responses/NotFound"
--- a/internal/db/migrations/011_delete_cascade.sql
+++ b/internal/db/migrations/011_delete_cascade.sql
@@ -1,0 +1,14 @@
+ALTER TABLE aws_reservation_details
+DROP CONSTRAINT aws_reservation_details_pubkey_id_fkey,
+ADD CONSTRAINT aws_reservation_details_pubkey_id_fkey
+FOREIGN key (pubkey_id) REFERENCES pubkeys(id) ON DELETE CASCADE;
+
+ALTER TABLE gcp_reservation_details
+DROP CONSTRAINT gcp_reservation_details_pubkey_id_fkey,
+ADD CONSTRAINT gcp_reservation_details_pubkey_id_fkey
+FOREIGN key (pubkey_id) REFERENCES pubkeys(id) ON DELETE CASCADE;
+
+ALTER TABLE pubkey_resources
+DROP CONSTRAINT pubkey_resources_pubkey_id_fkey,
+ADD CONSTRAINT pubkey_resources_pubkey_id_fkey
+FOREIGN key (pubkey_id) REFERENCES pubkeys(id) ON DELETE CASCADE;
--- a/internal/services/pubkey_service.go
+++ b/internal/services/pubkey_service.go
@@ -3,8 +3,11 @@ package services
 import (
 	"net/http"
 
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	"github.com/RHEnVision/provisioning-backend/internal/db"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 	"github.com/go-chi/render"
 	"github.com/pkg/errors"
@@ -74,6 +77,13 @@ func GetPubkey(w http.ResponseWriter, r *http.Request) {
 }
 
 func DeletePubkey(w http.ResponseWriter, r *http.Request) {
+	logger := ctxval.Logger(r.Context())
+	sourcesClient, err := clients.GetSourcesClient(r.Context())
+	if err != nil {
+		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "sources client v2", err))
+		return
+	}
+
 	id, err := ParseInt64(r, "ID")
 	if err != nil {
 		renderError(w, r, payloads.NewURLParsingError(r.Context(), "ID", err))
@@ -81,6 +91,51 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
 	}
 
 	pubkeyDao := dao.GetPubkeyDao(r.Context())
+
+	pubkey, err := pubkeyDao.GetById(r.Context(), id)
+	if err != nil {
+		renderError(w, r, payloads.NewNotFoundError(r.Context(), err))
+		return
+	}
+
+	resources, err := pubkeyDao.UnscopedListResourcesByPubkeyId(r.Context(), pubkey.ID)
+	if err != nil {
+		renderError(w, r, payloads.NewDAOError(r.Context(), "delete pubkey", err))
+		return
+	}
+
+	for _, res := range resources {
+		if res.Provider == models.ProviderTypeAWS {
+			if res.Handle != "" {
+				logger.Info().Msgf("Deleting pubkey resource ID %v with handle %s", res.ID, res.Handle)
+				authentication, errAuth := sourcesClient.GetAuthentication(r.Context(), res.SourceID)
+				if errAuth != nil {
+					if errors.Is(err, clients.NotFoundErr) {
+						renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources: application not found", errAuth, 404))
+						return
+					}
+					renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources", errAuth, 500))
+					return
+				}
+
+				ec2Client, errEc2 := clients.GetCustomerEC2Client(r.Context(), authentication, res.Region)
+				if errEc2 != nil {
+					renderError(w, r, payloads.NewAWSError(r.Context(), "failed to establish ec2 connection", errEc2))
+					return
+				}
+
+				errDelete := ec2Client.DeleteSSHKey(r.Context(), res.Handle)
+				if errDelete != nil {
+					renderError(w, r, payloads.NewAWSError(r.Context(), "can't delete public key", errDelete))
+					return
+				}
+			} else {
+				logger.Warn().Msgf("Pubkey resource with empty handle: resource ID %d", res.ID)
+			}
+		} else {
+			renderError(w, r, payloads.NewInvalidRequestError(r.Context(), ProviderTypeNotImplementedError))
+		}
+	}
 
 	err = pubkeyDao.Delete(r.Context(), id)
 	if err != nil {
--- a/api/openapi.gen.yaml
+++ b/api/openapi.gen.yaml
@@ -52,12 +52,8 @@ paths:
             type: integer
             format: int64
       responses:
-        "200":
-          description: 'Returned on success.'
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/v1.PubkeyResponse'
+        "204":
+          description: The Pubkey was deleted successfully.
         "404":
           $ref: "#/components/responses/NotFound"
         "500":
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -52,12 +52,8 @@ paths:
             type: integer
             format: int64
       responses:
-        "200":
-          description: 'Returned on success.'
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/v1.PubkeyResponse'
+        "204":
+          description: The Pubkey was deleted successfully.
         "404":
           $ref: "#/components/responses/NotFound"
         "500":


internal/services/pubkey_service.go
```diff
@@ -94,6 +129,12 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	err = DeleteResource(w, r, id)
```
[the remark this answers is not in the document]
avitova: Each time I need to get different dao, if I am not mistaken. I split it in two functions so it is a bit more readable. The nesting in the for loop was a bit scary but there was no way to prevent that, so I decided to do at least this.
But if you think it looks clear even in one function, then that is great, I can merge it into one function. :)

avitova:
Okay, so I tested this functionality in dao tests, and the resource is deleted after deletion of the pubkey, nothing seems to fail. I am not sure about the deletion from aws.

internal/services/pubkey_service.go
```diff
@@ -94,6 +98,29 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+
+	resources, err := pubkeyResource.ListByPubkeyId(r.Context(), id)
+	if err != nil {
+		renderError(w, r, payloads.NewDAOError(r.Context(), "delete pubkey", err))
+		return
+	}
+
+	for _, res := range resources {
+		if res.Handle != "" {
+			err = ec2Client.DeleteSSHKey(res.Handle)
```
ezr-ondrej, line 115: This will not work IMHO, we need to get ARN and assume role and get the sts client, because ec2 deals with our account, but the key is in the customer account.

IMHO we might also want to log with Info lvl, that we are deleting that key :thinking:
It's an action on the customer account, so there should be a trace for such action.

ezr-ondrej (commented):
I think we are missing the assume role step. Which I now realized will be hard, because we do not know what account the pubkey is associated with :thinking:
It should do:
1. Get ARN (where we need the source_id, which we do not have), example: [aws_reservation_service](https://github.com/RHEnVision/provisioning-backend/blob/cf5da03a5729cf1f6a1e2ce98dd6e10b09b5c084/internal/services/aws_reservation_service.go#L78)
2. Assume role and move deleting the pubkey to sts client, example: [aws_deploy_job](https://github.com/RHEnVision/provisioning-backend/blob/cf5da03a5729cf1f6a1e2ce98dd6e10b09b5c084/internal/jobs/launch_instance_aws.go#L51)

Also I was thinking it might be a good idea to offload this deletion into a job, but I'm fine doing it as a follow-up, not in this PR :)

ezr-ondrej:
I guess we will need to start storing the source_id, under which this handle was created [here](https://github.com/RHEnVision/provisioning-backend/blob/cf5da03a5729cf1f6a1e2ce98dd6e10b09b5c084/internal/db/migrations/004_pubkeys.sql#L9) and also make it part of the unique key on the table.

lzap:
> because we do not know what account the pubkey is associated with

What do you mean? Pubkey.AccountID hello? :) I guess what you meant was:

> I guess we will need to start storing the source_id

Yeah, good point. Definitely something we need to store for a successful delete.

> Also I was thinking it might be a good idea to offload this deletion into a job, but I'm fine doing it as a follow-up, not in this PR :)

I knew we should name the main endpoint `/tasks` and not `/reservations`! :) See, the problem is that there is no API to get status of any other background job than reservation. Maybe we should rename this while we still can? Something like:

* /tasks/launch_noop
* /tasks/launch_aws
* /tasks/delete_pubkey

internal/services/pubkey_service.go
lzap: I guess we should implement a check - if any operation other than assume role is executed on the service account, error out immediately. Just an idea...

adiabramovitch:
@avitova Joining to lzap suggestion. you can take a look at my PRs #130 #107  to see which changes are supposed to happen. 
If you need help, ping me :)

ezr-ondrej:
>> because we do not know what account the pubkey is associated with
> What do you mean? Pubkey.AccountID hello?

Yeah, agreed, terminology not sufficient there, I meant we do not know which AWS account it is associated with :) (lets call them provider/hypervizor accounts? I do not like to call them sources too much :thinking: xD )

avitova:
SourceID PR should now be in #158. Let me know if that is what we need. :+1:

ezr-ondrej:
> SourceID PR should now be in #158. Let me know if that is what we need.

I believe that is exactly it :)

ezr-ondrej:
So now we can finally asume the role and actually delete this xD

ezr-ondrej:
Linter :police_officer: ain't happy @avitova :(

internal/services/pubkey_service.go
```diff
@@ -99,6 +117,48 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+
+	resources, err := pubkeyResource.ListByPubkeyId(r.Context(), id)
+	if err != nil {
+		renderError(w, r, payloads.NewDAOError(r.Context(), "delete pubkey", err))
+		return
+	}
+
+	for _, res := range resources {
+		if res.Handle != "" {
+			arn, err := sourcesClient.GetArn(r.Context(), res.SourceID)
```
ezr-ondrej, line 134: I believe this would fix the issue:
```suggestion
			var arn string
			arn, err = sourcesClient.GetArn(r.Context(), res.SourceID)
```

But I'd love to see a better solution :) @lzap ? :stuck_out_tongue:

internal/services/pubkey_service.go
```diff
@@ -99,6 +117,48 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+					return
+				}
+				renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources", err, 500))
+				return
+			}
+			crd, err := stsClient.AssumeRole(arn)
+			if err != nil {
+				renderError(w, r, payloads.ClientError(r.Context(), "Sources", "cannot assume role: %w", err, 500))
+				return
+			}
```
lzap, line 151: It looks like we have a common pattern across our codebase, there are three other places where we create client, get arn from sources and assume role. I believe we should refactor this into a single function, not here, not in this PR, but this would be a nice PR for you to work on.

The question is where to put it, since this operation uses both AWS client and sources client. I guess putting it into AWS is reasonable, this will create a dependency (AWS needs Sources package) but I think it is reasonable. @ezr-ondrej ?

internal/services/pubkey_service.go
```diff
@@ -99,6 +117,48 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+				if errors.Is(err, sources.AuthenticationForSourcesNotFoundErr) {
+					renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources: authentication not found", err, 404))
+					return
+				}
+				renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources", err, 500))
+				return
+			}
+			crd, err := stsClient.AssumeRole(arn)
+			if err != nil {
+				renderError(w, r, payloads.ClientError(r.Context(), "Sources", "cannot assume role: %w", err, 500))
```
lzap, line 149: The ClientError is reserved only for errors coming from HTTP clients (sources, image builder). Of all other errors, find an suitable error in the payloads package, or create a new one.

```suggestion
				renderError(w, r, payloads.NewAWSError(r.Context(), "assume role sts", err))
```

internal/services/pubkey_service.go
```diff
@@ -99,6 +117,48 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	pubkeyResource, err := dao.GetPubkeyResourceDao(r.Context())
+	if err != nil {
+		renderError(w, r, payloads.NewDAOError(r.Context(), "delete pubkey", err))
```
lzap, line 122: Ok errors are confusing, I have to say. Not sure how to improve this, @ezr-ondrej? Anyway, when you need to acquire DAO reference, use the InitializeDaoError:

```suggestion
		renderError(w, r, payloads.NewInitializeDAOError(r.Context(), "pubkey resource DAO", err))
```

internal/services/pubkey_service.go
```diff
@@ -99,6 +117,48 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	resources, err := pubkeyResource.ListByPubkeyId(r.Context(), id)
+	if err != nil {
+		renderError(w, r, payloads.NewDAOError(r.Context(), "delete pubkey", err))
```
lzap, line 128: For normal DAO operations, there's the DAOError, this is correct.

internal/services/pubkey_service.go
```diff
@@ -99,6 +117,48 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+			crd, err := stsClient.AssumeRole(arn)
+			if err != nil {
+				renderError(w, r, payloads.ClientError(r.Context(), "Sources", "cannot assume role: %w", err, 500))
+				return
+			}
+			err = client.DeleteSSHKey(*crd.AccessKeyId)
+			if err != nil {
+				renderError(w, r, payloads.ClientError(r.Context(), "Sources", "cannot delete SSH key: %w", err, 500))
```
lzap, line 154: And for AWS client errors we do have AWSError type.

```suggestion
				renderError(w, r, payloads.NewAWSError(r.Context(), "can't delete public key", err))
```

internal/services/pubkey_service.go
lzap: Well, the alternative is to use a different name, like `errArn` or `arnErr`. That's what I tend to use.

We do not like shadowing of variables because there can be some nasty bugs sometimes.

lzap:
It's almost there, please fix errors and this is ready for the final review.

I commented on future refactoring since this PR adds now 4 (fourth) place where we have the very same code (create client, talk to sources, assume role) and it is definitely the time to fix this. Feel free to do it as a separate commit in this PR, or a separate PR completely.

lzap:
Remember to hit Ready for review button above when you feel this is ready for review. Also it's good idea to drop a comment because GitHub will not notify us when you hit that button (I think).

avitova:
Okay, I forgot this was a draft before. Now this should be ready for review, lint is failing, but I am not sure why. I feel like there is no problem with linting rn.

internal/services/pubkey_service.go
```diff
@@ -99,6 +117,48 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+					return
+				}
+				if errors.Is(err, sources.AuthenticationForSourcesNotFoundErr) {
+					renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources: authentication not found", err, 404))
+					return
+				}
+				renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources", err, 500))
+				return
+			}
+			crd, err := stsClient.AssumeRole(arn)
```
lzap, line 147: This is the reason, the linter does not allow redeclaration of err from the parent context. Rename `err` to something else too.

lzap (dismissed):
Handle and loop comments.

lzap:
So I understood that you don't have full AWS development setup yet (assume role needs a service account), I talked to @ezr-ondrej and he will get it for you. In the meantime, I can test this patch for you if you want (changes are still needed, my comments are still valid). Or if you choose to, just get back to this once you have full access and ability to test the stuff.

internal/services/pubkey_service.go
```diff
@@ -99,6 +117,48 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+			if err != nil {
+				renderError(w, r, payloads.NewAWSError(r.Context(), "assume role sts", err))
+				return
+			}
```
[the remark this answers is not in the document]
ezr-ondrej: We've to assume every resource has been created under different account and later on even under different cloud provider, which makes this a necessity, but we can now `clients.GetCustomerAccount(ctx, arn, region)` we do not have the region, which kinda sucks, but I'd be ok having static region for now, but that means it will be necessary to also keep track of which region the key has been created in.

avitova:
Okay, this deletes reservation while deleting pubkey. Do we want this approach? Or should we do something else with reservation after deleting the pubkey?

lzap:
Hmmm I forgot about this completely. Deleting reservations is not good, there might be running worker. The same problem is when we set those records to `NULL` (database can do this for you automatically when you tell it to do so).

The safest approach is to simply deny deletion if there are reservations, but this is the least user friendly approach. I can see users trying to create many keys during our launch period and then deleting them only to find out they cannot delete those until we clean old reservations.

For this reason, I think actually this could be the most sane approach:

* In a single transaction, check if all reservations for the given account and public key are finished (Succes is not NULL).
* If that's the case, perform the deletion. Configure the `ON DELETE SET NULL`.
* Otherwise create a custom error (`var PubkeyDeletePendingReservationsErr = errors.New("cannot delete pubkey: there are pending reservations")`) and return it if there are unfinished reservations.

We could actually do all of this in a SQL constraint statement and catch the error, I believe `pgx` driver would allow us to see exactly the constraint name. That's also an option, but I guess everyone working on this project would prefer doing this in Go :-) This is up to you.

Here is an example how to do deletion in a transaction:

```go
func (di *pubkeyDaoSqlx) Delete(ctx context.Context, id int64) error {
	err := dao.WithTransaction(ctx, func(tx *sqlx.Tx) error {
		// reservation check and deletion code
		return nil
	})
	if err != nil {
		return NewTransactionError(ctx, err)
	}
	return nil
}
```

@ezr-ondrej opinions? I think it's okay to acquire a DAO object from within another DAO, I remember we talked about it. If not, then let's do the SQL constraint...

ezr-ondrej:
> I remember we talked about it

Yeah we did and I think we've said it's ok, but might be also interesting to think about the DAO for pubkeys and pubkey resources as being only one dao and thus you'd not have to acquire another DAO, but both solutions feels fine :)

ezr-ondrej:
I totaly agree with @lzap's approach tho, checking any of the reservation's is in progress and if not, nullify the pubkeys.

ezr-ondrej:
Another (bit crazy) solution would be to have `deleted_at` column and never actually delete the pubkey from our database, but it is quite crazy, because then we can't use DB uniqueness on fingerprint easily and a lot of other issues, so lets keep it simple and just nullify the references :+1:

lzap:
I like merging DAOs together: Pubkey and PubkeyResource into just single Pubkey DAO interface. We do the same for Reservations already. In that case, I suggest to do this in a separate PR @avitova so this PR won't get too complicated.

lzap:
Just a reminder for @RHEnVision/qe - the app currently does not allow deleting of PubkeyResources, this leads to inability to actually delete a Pubkey. The error message is `update or delete on table \"pubkeys\" violates foreign key constraint \"pubkey_resources_pubkey_id_fkey\" on table \"pubkey_resources\"`. This will happen once a pubkey is uploaded to AWS (when you launch at least one instance).

akhil-jha:
> Just a reminder for @RHEnVision/qe - the app currently does not allow deleting of PubkeyResources, this leads to inability to actually delete a Pubkey. The error message is `update or delete on table \"pubkeys\" violates foreign key constraint \"pubkey_resources_pubkey_id_fkey\" on table \"pubkey_resources\"`. This will happen once a pubkey is uploaded to AWS (when you launch at least one instance).

ACK. Yes. it was mentioned before as well. That's why I changed the scope of HMSPROV-256. Earlier it was CRD, but it's just DELETE /pubkey/{ID} now. 
POST and GET test are in iqe right now.

avitova:
I have rebased after changes and updated parts that were outdated. This is tested and is now ready for review again.

internal/services/pubkey_service.go
```diff
@@ -99,6 +109,50 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	pubkey, err := dao.GetPubkeyDao(r.Context())
+	if err != nil {
+		renderError(w, r, payloads.NewInitializeDAOError(r.Context(), "pubkey resource DAO", err))
+		return
+	}
+
+	resources, err := pubkey.UnscopedListByPubkeyId(r.Context(), id)
```
ezr-ondrej, line 118: the `Unscoped` prefix means this method does not validate the tenant (which would allow any Customer to delete another Customer's pubkeys, which might be quite dangerous), so we should fetch the pubkey itself first and return 404 if not found.
Also, we can now use the `pubkeyDao`

```suggestion
	resources, err := pubkey.UnscopedListByPubkeyId(r.Context(), pubkey.id)
```

internal/services/pubkey_service.go
```diff
@@ -99,6 +109,50 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	pubkey, err := dao.GetPubkeyDao(r.Context())
```
ezr-ondrej, line 112: I believe this is not necessary, we have th `pubkeyDao` already and it is the same dao now (after the merge of daos).
We can fetch the pubkey by ID here (we should also render 404 if the pubkey is not found).
```suggestion
	pubkey, err := pubkeyDao.GetByID(r.Context(), id)
```

internal/services/pubkey_service.go
```diff
@@ -99,6 +109,50 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+				}
+				if errors.Is(err, pkgErr.AuthenticationForSourcesNotFoundErr) {
+					renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources: authentication not found", errArn, 404))
+					return
+				}
+				renderError(w, r, payloads.ClientError(r.Context(), "Sources", "can't fetch arn from sources", errArn, 500))
+				return
+			}
+
+			ec2Client, errEc2 := clients.GetCustomerEC2Client(r.Context(), arn, "us-east-1")
```
ezr-ondrej, line 140: we can now use the region from the resouce I believe:
```suggestion
			ec2Client, errEc2 := clients.GetCustomerEC2Client(r.Context(), arn, res.Region)
```

internal/services/pubkey_service.go
```diff
@@ -3,6 +3,9 @@ package services
 import (
 	"net/http"
 
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	pkgErr "github.com/RHEnVision/provisioning-backend/internal/clients/http"
```
ezr-ondrej, line 7: NIT: I'd like a more meaningful alias here ^_^
Maybe `httpClients` or something along those lines?

ezr-ondrej (changes requested):
Nice, I believe we're almost there, I've got just mostly nits and clean-ups.
One find might bite us in the foot and should really get fixed (fetching the pubkey itself first)

ezr-ondrej:
> Can it be addressed before merging this?

Right, good catch, the OpenAPI response is incorrect, we are returning NoContent, @avitova can you pls fix it in
https://github.com/RHEnVision/provisioning-backend/blob/8d8230918db5df8eb898d4476690c442daa4fe96/cmd/spec/path.yaml#L55
to something like:
```yaml
      responses:
        '204':
          description: The Pubkey was deleted successfully.
```

This comes from knowing we are using [`NoContent`](https://github.com/go-chi/render/blob/996350ff55de76c0d4a7d5d7e06469031ff93149/responder.go#L138) and swagger docs: https://swagger.io/docs/specification/describing-responses/
Swagger is mostly the same as our OpenAPI 3.0, so it's good docs to take a look at, but pls take it with grain of salt, OpenAPI changes thing or two :)

akhil-jha:
@ezr-ondrej / @avitova , any issues with enabling auto-merge on QE ACK?
I'd want @mayurilahane to get familiar on testing it and approve after the changes have been made.

ezr-ondrej:
Sure, it should be available once we make changes to the OpenAPI :) So once Anna is done with the changes, I'll ack with auto merge :+1:

internal/services/pubkey_service.go
avitova: What do you mean by this suggestion? This variable is for pubkeyDao and does not have an 'id' field.

internal/services/pubkey_service.go
```diff
@@ -99,6 +109,50 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	pubkey, err := pubkeyDao.GetById(r.Context(), id)
+	if err != nil {
+		renderError(w, r, payloads.NewInitializeDAOError(r.Context(), "pubkey resource DAO", err))
```
ezr-ondrej, line 114: probably a leftover? :)
```suggestion
		renderError(w, r, payloads.NewNotFoundError(r.Context(), err))
```

or ideally 
```suggestion
		if errors.Is(err, dao.NoRowsError) {
		  renderError(w, r, payloads.NewNotFoundError(r.Context(), err))
		} else {
		  renderError(w, r, payloads.NewDAOError(r.Context(), "pubkey DAO", err))
		}
```

ezr-ondrej (changes requested):
We've forgot to update the error message I believe. There should be 404 if the pubkey is not found.
The static region is bit of a shame (given we already have the region, but not a big deal)

avitova:
One more overlook almost slipped through. It should be ok now.

ezr-ondrej (approved):
Looks good to me :+1:
And @lzap's feedback is also addressed from what I see, so aproving and enabling the auto-merge.
It will merge once @mayurilahane acks :) (or any @RHEnVision/qe )

api/openapi.gen.json
```diff
@@ -346,6 +346,9 @@
             },
             "description": "Returned on success."
           },
+          "204": {
```
akhil-jha, line 349: If we are returning 204, then I believe we should remove 200?

ezr-ondrej: oh yeah, I've missed we kept it in, nice catch.
@avitova could you remove the 200 response from `path.yml` pls?

internal/db/migrations/005_reservations.sql
```diff
@@ -24,7 +24,7 @@ CREATE TABLE aws_reservation_details
 (
   reservation_id BIGINT NOT NULL,
   provider INTEGER NOT NULL DEFAULT provider_type_aws(),
-  pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
+  pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id) ON DELETE CASCADE,
   source_id TEXT NOT NULL,
   image_id TEXT NOT NULL,
   aws_reservation_id TEXT,
```
lzap, line 30: We cannot do this. This will never be executed on stage anymore.

lzap: Unfortunately you need to ALTER TABLE DROP CASCADE CREATE CASCADE, I think this all can be done in a single statement.

internal/services/pubkey_service.go
```diff
@@ -99,6 +109,50 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	resources, err := pubkeyDao.UnscopedListByPubkeyId(r.Context(), pubkey.ID)
+	if err != nil {
+		renderError(w, r, payloads.NewDAOError(r.Context(), "delete pubkey", err))
+		return
+	}
+
+	for _, res := range resources {
```
lzap, line 124: I would appreciate a debug or info level log line in the loop something like "Deleting pubkey resource ID xyz with handle XYZ".

internal/db/migrations/005_reservations.sql
lzap: I wonder why the migration check did not error out.

lzap: Ah found it, fixed it hopefully: https://github.com/RHEnVision/provisioning-backend/pull/242

After you rebase this PR on top of this, it should fail because you touched existing SQL files...

avitova: There should now be a new migration, and it works for me. I would be happy if you checked that. In the PR, there is no editing of an old migration, so hopefully, the tests will pass.

ezr-ondrej (commented):
We still have changes in migrations.

ezr-ondrej (approved):
Nice :+1: so now we are ready to get @mayurilahane's ack ^_^

internal/db/migrations/011_delete_cascade.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE reservation_instances
+DROP CONSTRAINT reservation_instances_reservation_id_fkey,
+ADD CONSTRAINT reservation_instances_reservation_id_fkey
+FOREIGN key (reservation_id) REFERENCES reservations(id) ON DELETE CASCADE;
+
+ALTER TABLE aws_reservation_details
+DROP CONSTRAINT aws_reservation_details_pubkey_id_fkey,
+ADD CONSTRAINT aws_reservation_details_pubkey_id_fkey
+FOREIGN key (pubkey_id) REFERENCES pubkeys(id) ON DELETE CASCADE;
```
lzap, line 9: Missing newline (small nit).

But wait with merging please, we also have GCP detail table and I want to check it.

internal/db/migrations/011_delete_cascade.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE reservation_instances
+DROP CONSTRAINT reservation_instances_reservation_id_fkey,
+ADD CONSTRAINT reservation_instances_reservation_id_fkey
+FOREIGN key (reservation_id) REFERENCES reservations(id) ON DELETE CASCADE;
```
lzap, line 4: This foreign key already has delete cascade in my database, so we might not need this at all.

lzap (changes requested):
Please drop the change for `reservation_instances` table, it already has cascade deletion:

https://github.com/RHEnVision/provisioning-backend/blob/main/internal/db/migrations/005_reservations.sql#L19

_(Hint: GoLand Database plugin can actually show you all of this in a nice way so you don't have to dig in the codebase for it. Or any other good PostgreSQL client will do.)_

Also, we need to do the same change for the `gcp_reservation_details` it suffers from the very same problem.

lzap:
Just for clarity, here is the current database DDL schema (generated via GoLand "Generate DDL to clipboard" function) so you can doublecheck all the foreign keys you are interested in:

```sql
create table schema_version
(
    version integer not null
);

alter table schema_version
    owner to lzap;

create trigger track_applied_migrations
    after update
    on schema_version
    for each row
execute procedure track_applied_migration();

create table schema_migrations_history
(
    id         serial
        primary key,
    version    bigint                                 not null,
    applied_at timestamp with time zone default now() not null
);

alter table schema_migrations_history
    owner to lzap;

create table accounts
(
    id             bigint generated by default as identity
        primary key,
    account_number text
        unique
        constraint accounts_account_number_check
            check (NOT empty(account_number)),
    org_id         text not null
        unique
        constraint accounts_org_id_check
            check (NOT empty(org_id))
);

alter table accounts
    owner to lzap;

create table pubkeys
(
    id          bigint generated by default as identity
        primary key,
    account_id  bigint not null
        references accounts,
    name        text   not null
        constraint pubkeys_name_check
            check (NOT empty(name)),
    body        text   not null
        constraint pubkeys_body_check
            check (NOT empty(body)),
    fingerprint text   not null
        constraint pubkeys_fingerprint_check
            check (NOT empty(fingerprint)),
    unique (name, account_id),
    unique (fingerprint, account_id)
);

alter table pubkeys
    owner to lzap;

create table pubkey_resources
(
    id        bigint generated by default as identity
        primary key,
    tag       text default ''::text not null,
    pubkey_id bigint                not null
        references pubkeys,
    provider  integer               not null
        constraint pubkey_resources_provider_check
            check (valid_provider(provider)),
    source_id text                  not null,
    handle    text                  not null,
    region    text
);

alter table pubkey_resources
    owner to lzap;

create unique index pubkey_resources_pubkey_id_provider
    on pubkey_resources (pubkey_id, provider);

create trigger prevent_tag_update_on_pubkey_resources
    before update
    on pubkey_resources
    for each row
execute procedure prevent_tag_update();

create table reservations
(
    id          bigint generated by default as identity
        primary key,
    provider    integer                             not null
        constraint reservations_provider_check
            check (valid_provider(provider)),
    account_id  bigint                              not null
        references accounts,
    created_at  timestamp default CURRENT_TIMESTAMP not null,
    steps       smallint                            not null,
    step        smallint  default 0                 not null,
    status      text                                not null
        constraint reservations_status_check
            check (NOT empty(status)),
    error       text      default ''::text          not null,
    finished_at timestamp,
    success     boolean,
    unique (id, provider)
);

alter table reservations
    owner to lzap;

create table reservation_instances
(
    reservation_id bigint not null
        references reservations
            on delete cascade,
    instance_id    text   not null
        unique
);

alter table reservation_instances
    owner to lzap;

create table aws_reservation_details
(
    reservation_id     bigint                              not null,
    provider           integer default provider_type_aws() not null
        constraint aws_reservation_details_provider_check
            check (provider = provider_type_aws()),
    pubkey_id          bigint                              not null
        references pubkeys,
    source_id          text                                not null,
    image_id           text                                not null,
    aws_reservation_id text,
    detail             jsonb,
    foreign key (reservation_id, provider) references reservations (id, provider)
        on delete cascade
);

alter table aws_reservation_details
    owner to lzap;

create table jobs
(
    id          uuid                                  not null
        primary key,
    token       uuid
        constraint token_is_never_uuid_nil
            check ((token IS NULL) OR (token <> uuid_nil())),
    type        varchar                               not null,
    args        jsonb,
    result      jsonb,
    queued_at   timestamp                             not null,
    started_at  timestamp,
    finished_at timestamp,
    canceled    boolean default false                 not null,
    channel     varchar default ''::character varying not null,
    constraint not_finished_when_not_started
        check ((finished_at IS NULL) OR (started_at IS NOT NULL)),
    constraint not_finished_when_canceled
        check ((canceled = false) OR (finished_at IS NULL)),
    constraint chronologic_started_at
        check ((started_at IS NULL) OR (queued_at <= started_at)),
    constraint chronologic_finished_at
        check ((finished_at IS NULL) OR (started_at <= finished_at)),
    constraint token_is_set_when_started
        check ((started_at IS NULL) OR (token IS NOT NULL))
);

alter table jobs
    owner to lzap;

create table job_dependencies
(
    job_id        uuid
        constraint job_dependencies_dependency_id_fkey
            references jobs
            on delete cascade,
    dependency_id uuid
        constraint job_dependencies_job_id_fkey
            references jobs
            on delete cascade
);

alter table job_dependencies
    owner to lzap;

create table heartbeats
(
    token     uuid      not null
        primary key
        constraint token_is_never_uuid_nil
            check (token <> uuid_nil()),
    id        uuid
        references jobs
            on delete cascade,
    heartbeat timestamp not null
);

alter table heartbeats
    owner to lzap;

create table gcp_reservation_details
(
    reservation_id     bigint                              not null,
    provider           integer default provider_type_gcp() not null
        constraint gcp_reservation_details_provider_check
            check (provider = provider_type_gcp()),
    pubkey_id          bigint                              not null
        references pubkeys,
    source_id          text                                not null,
    image_id           text                                not null,
    gcp_operation_name text,
    detail             jsonb,
    foreign key (reservation_id, provider) references reservations (id, provider)
        on delete cascade
);

alter table gcp_reservation_details
    owner to lzap;
```

internal/services/pubkey_service.go
```diff
@@ -99,6 +109,51 @@ func DeletePubkey(w http.ResponseWriter, r *http.Request) {
...
+			if errEc2 != nil {
+				renderError(w, r, payloads.NewAWSError(r.Context(), "failed to establish ec2 connection", errEc2))
+				return
+			}
+
+			errDelete := ec2Client.DeleteSSHKey(r.Context(), res.Handle)
+			if errDelete != nil {
+				renderError(w, r, payloads.NewAWSError(r.Context(), "can't delete public key", errDelete))
+				return
+			}
```
lzap, line 151: The whole ec2 client deletion must be in a switch-case statement, see above.

lzap (changes requested):
Since we have added support for GCP and Azure into sources, the code must be generic now. It's an easy change tho.

ezr-ondrej:
> Since we have added support for GCP and Azure into sources, the code must be generic now. It's an easy change tho.

I'd disagree here, I think we can do this change as part of adding support for creating pubkeys for Azure/GCP.
I'd like to stop bloating the scope here, lets get this in, unless it's breaking something and it's a complete change, that is a step in the right direction.

Until there is no way to add a GCP key, I think there is no reason to push this PR to count on its existence.

lzap:
While I appreciate the sentiment, problem is, the moment we add GCP nobody will ever think about pubkey deletion and that it will break it. All I am asking is to create a switch or if statement that will let the code to error out if authorization type does not match.

But I don't insist. If it breaks, well, then it breaks.

internal/services/pubkey_service.go
lzap: Doesn't have to be a switch statement, just make sure that the EC2 code is only executed when AWS authorization type is returned, otherwise raise an error. That's all.

lzap:
Thanks, our linter insist of always using all cases, so in that case you can just use if and else I guess. That's fine for now.

avitova:
Okay, I am sorry. I think that we should ALTER pubkey_resources table as well, right? I realized it was not there, and it works for me now, is it okay like this, or should that not be there? @lzap

lzap:
Correct, I don't see it in my DB. Thanks, looking into this.

lzap (approved):
Perfect.

This needs to be tested, will take a look tomorrow morning.

lzap:
One more thing: we should check if public key does not have pending reservation. Therefore, we need:

* New DAO function `Pubkey.ListReservations(ctx, id)`
* In this new code call this new function for given pubkey and iterate over list of reservations, they all must be finished (Success must not be nil).
* Ideally the function could have a parameter `finished` and do the filtering in SQL so it's faster.

We should not use any SQL transactions as external calls are needed, therefore transactions are not possible.

Now, since QA team is blocked by this, we can implement this later. Let's create a ticket for that, it's not urgent.

lzap:
Enabled auto merge, @akhil-jha feel free to rereview thus merge.

Created: https://issues.redhat.com/browse/HMSPROV-305

lzap:
@akhil-jha ping, this was a long PR so wanted to check up if this is on your agenda thanks :-)

lzap:
We have a new enforced commit message checker, can you rebase on top of `main` and reword the commit message to:

`feat: pubkey resource deletion`

Thanks!

mayurilahane:
https://github.com/RHEnVision/provisioning-backend/pull/257#issuecomment-1275068872
seems like the app is broken on stage.

ezr-ondrej:
@avitova can you rebase this pls? :)

lzap:
This looks weird: `failed on setup with "AttributeError: 'IQEVaultReference' object has no attribute 'openapi_types'"` can someone take a look @RHEnVision/qe ? Thanks.

akhil-jha:
> This looks weird: `failed on setup with "AttributeError: 'IQEVaultReference' object has no attribute 'openapi_types'"` can someone take a look @RHEnVision/qe ? Thanks.

Oops. I updated the vault secrets and forgot updating the sources about letting it know the new vault path! Sorry!

mayurilahane:
Hey,
am getting this error after removing the pubkey
`(Pdb) application.provisioning.rest_client.default_api.remove_pubkey_by_id(8)
INFO:iqe.base.rest_client:REST: METHOD=DELETE, request_id=None, params=[]
2022-10-13 15:25:46 INFO REST: METHOD=DELETE, request_id=None, params=[]
*** iqe_provisioning_api.exceptions.ApiTypeError: Invalid type for variable 'received_data'. The required value type is V1PubkeyResponse and the passed type was str at ['received_data']
`
but I can see the pubkey gets removed and the response is incorrect!

akhil-jha (approved):

PR #148, opened 2022-08-09
Use GetAWSAmi of image builder client

adiabramovitch:
Depends on #153 #150

### DIFF
--- a/internal/db/migrations/005_reservations.sql
+++ b/internal/db/migrations/005_reservations.sql
@@ -25,7 +25,7 @@ CREATE TABLE aws_reservation_details
   source_id INTEGER NOT NULL,
   instance_type TEXT NOT NULL,
   amount INTEGER NOT NULL,
-  ami TEXT NOT NULL,
+  image_id TEXT NOT NULL,
   aws_reservation_id TEXT,
 
   FOREIGN KEY (reservation_id, provider) REFERENCES reservations(id, provider) ON DELETE CASCADE,
--- a/internal/services/aws_reservation_service.go
+++ b/internal/services/aws_reservation_service.go
@@ -5,6 +5,8 @@ import (
 	"net/http"
 	"strconv"
 
+	_ "github.com/RHEnVision/provisioning-backend/internal/clients/image_builder"
+
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/RHEnVision/provisioning-backend/internal/clients/sources"
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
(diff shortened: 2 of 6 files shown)


internal/db/migrations/005_reservations.sql
```diff
@@ -25,7 +25,7 @@ CREATE TABLE aws_reservation_details
   source_id INTEGER NOT NULL,
   instance_type TEXT NOT NULL,
   amount INTEGER NOT NULL,
-  ami TEXT NOT NULL,
+  image_id TEXT NOT NULL,
```
ezr-ondrej, line 28: I think that here we want to store AMI, we want to exchange image_id before we create aws_reservation_details IMHO :thinking:

lzap: Just an idea: since its the third time we are changing this, why not to keep both?

scripts/rest_examples/ib_list.http
```diff
@@ -0,0 +1,5 @@
+// @no-log
+GET https://console.stage.redhat.com/api/image-builder/v1/composes HTTP/1.1
+Content-Type: application/json
+Authorization: Basic user:password
+
```
lzap, line 5: FYI this will only work with the HTTP proxy. Unfortunately, HTTP proxy cannot be configured via this format. Maybe drop a comment explaining that.

internal/db/migrations/005_reservations.sql
lzap: Data can be stored both in reservation tables and job arguments (JSON), duplicity is not an issue. However, important thing to realize is that if we do anything in the service level, it means we can error out early so the reservation request won't even start.

I am not sure in this case tho what is the benefit of resolving AMI before creating a job. Everything can fail (pubkey upload, aws launch) so in this light AMI is not any different. Why you think it is worth doing before @ezr-ondrej ?

scripts/rest_examples/ib_list.http
adiabramovitch: Yes, I configured my proxy on my IDE, we should add it to the request, I'll add it on my next push.

internal/services/aws_reservation_service.go
```diff
@@ -99,14 +99,28 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 		},
 	}
 
+	// Get Image builder client
+	IBClient, err := clients.GetImageBuilderClient(r.Context())
```
lzap, line 103: This will likely panic as the GetImageBuilderClient is nil, this is missing an import.

scripts/rest_examples/ib_list.http
amirfefer: you can configure a proxy in vscode easily, in settings -> `http.proxy` 
extensions (like http clients) automatically use it

internal/db/migrations/005_reservations.sql
ezr-ondrej: well just because it is easier.
We were barely able to pass the accountId to jobs, now we would need to reconstruct the Auth header so we could pass it to IB (I mean we should, I just did not want to go there)

adiabramovitch: I don't think it is easier. 
I think we should be consistent:  fetching the ami should be as fetching the arn, this is why I chose this approach.

ezr-ondrej: oh! I get it now, we are fething it in the service, but database holds the original data, makes sense, sorry! :)
And it also will ease the transition when we decide to fetch AMI and ARN in the job itself

scripts/rest_examples/ib_list.http
lzap: Yeah I know I was just crying that the .http format does not have support for that. I wish.

amirfefer:
#150 is just POC to verify that our clients work well in stage,  I guess you can remove #150 commit, rebase and this PR is ready for merge :)

ezr-ondrej (approved):
LGTM :+1:

PR #149, opened 2022-08-10
Use setup/teardown for DAO tests, truncate tables instead of seeding

lzap:
There are two commits in this PR.

The first one improves the DAO tests in general:

* Generic functions were moved into main_test.go and are not exported.
* There are now global `setup` and `teardown` functions and `setupPubkey` etc functions in each file for more specific setup code.
* All setup functions were renamed and broken down into `dbDrop`, `dbMigrate` and `dbSeed` to better reflect what they do.
* Error handling was vastly simplified - for fatal error states (cannot connect to DB, cannot migrate or seed) using `panic` is completely fine because this is a fatal state anyway. Setup functions themselves no longer return errors which makes the code longer due to use of panic keyword.
* Teardown functions are called via `defer`.

The second commit then switches from migrating the database for each test to more straightforward approach - the database is seeded just once and then every test all tables are truncated. This makes DAO tests faster, for me it dropped from 6 seconds down to 4.3 seconds. It's not much, but this feels cleaner.

### DIFF
--- a/cmd/dao_tests/pubkey_test.go
+++ b/cmd/dao_tests/pubkey_test.go
@@ -1,24 +1,16 @@
 //go:build integration
 // +build integration
 
-// To override application configuration for integration tests, copy local.yaml into this directory.
-
 package main
 
 import (
 	"context"
-	"fmt"
-	"os"
 	"testing"
 
-	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
-	"github.com/RHEnVision/provisioning-backend/internal/db"
-	"github.com/RHEnVision/provisioning-backend/internal/logging"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
-	"github.com/rs/zerolog/log"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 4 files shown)


avitova (changes requested):
I can't test this right now, if this works, I agree. Thank you for refactoring reservation tests. We just have to make sure that we are consistent in error messages, so that we are consistent.

avitova:
Update pubkey test panics with
```
panic: runtime error: slice bounds out of range [:-1]
	panic: runtime error: slice bounds out of range [:-1]

```

lzap:
Please rebase on top of main, I fixed this overlook in the logging function. I will shortly rebase this PR.

lzap:
Rebased!

cmd/dao_tests/pubkey_test.go
```diff
@@ -111,41 +68,37 @@ func TestListPubkey(t *testing.T) {
...
 		Name:      "avitova-ed25519-2021",
 		Body:      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN avitova",
 	}
-	pkDao, ctx, err := Setup(t, "Update pubkey")
+	pkDao, ctx := setupPubkey(t)
+	defer teardownPubkey(t)
+	err := pkDao.Create(ctx, createPk())
 	if err != nil {
-		t.Errorf("Database setup had failed. %s", err)
+		t.Errorf("Create pubkey test failed. %s", err)
```
adiabramovitch, line 81: You are checking UpdatePubkey, so maybe change the error msg?

cmd/dao_tests/pubkey_test.go
```diff
@@ -111,41 +68,37 @@ func TestListPubkey(t *testing.T) {
...
 	updatePk := &models.Pubkey{
 		ID:        1,
 		AccountID: 1,
 		Name:      "avitova-ed25519-2021",
 		Body:      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN avitova",
 	}
-	pkDao, ctx, err := Setup(t, "Update pubkey")
+	pkDao, ctx := setupPubkey(t)
+	defer teardownPubkey(t)
+	err := pkDao.Create(ctx, createPk())
```
adiabramovitch, line 79: and I guess this is supposed to be update?

adiabramovitch (approved):

PR #150, opened 2022-08-10
POC: Add proxy client

amirfefer:
This adds a hardcoded proxy client for communicating to external clients via stage env during development

## How to use

1. add proxy url in `proxy_client` file
2. add stage user:password encoded to base 64 in `AddBasicAuth` function

### DIFF
--- a/internal/clients/image_builder/image_client.go
+++ b/internal/clients/image_builder/image_client.go
@@ -20,7 +20,8 @@ func init() {
 }
 
 func newImageBuilderClient(ctx context.Context) (clients.ImageBuilder, error) {
-	c, err := NewClientWithResponses(config.ImageBuilder.URL)
+	proxiedClient, err := clients.GetProxiedClient()
+	c, err := NewClientWithResponses(config.ImageBuilder.URL, WithHTTPClient(proxiedClient))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


adiabramovitch:
I think we should merge it just for testing ib for now. 
What do you think? 
@lzap @ezr-ondrej

lzap:
This looks incomplete, but overall I am good except hardcoding is not an option - this will never work well with git, you will constantly run into conflicts/merge issues.

I suggest to simply implement configuration for this. It needs to be per-service tho, because typically we will only use this for image builder and not for sources (which are trivial to run locally):

```yaml
restEndpoints:
  imageBuilder:
    url: http://localhost
    proxy_url: http://xyz:1234
    proxy_auth: user:password
```

I can implement this for you if this is something that is blocking you.

lzap:
The implementation should be done via HttpRequestDoer interface, it can be explicitly set in our code and this will work both for sources and image builder. This draft does not use this approach, I am not sure how this is supposed to work actually.

amirfefer:
I wanted to avoid changing the config for preventing any production leaks and errors, we can just use this PR for stage verification or merge it as a separate branch.

lzap:
Here is my patch: https://github.com/RHEnVision/provisioning-backend/pull/154

amirfefer:
Closing thanks to #154

PR #152, opened 2022-08-10
Fix incorrect string truncation in logger

lzap:
Copy pasted from stack overflow without unit tests!

Oh my me.

### DIFF
--- a/internal/logging/zerolog.go
+++ b/internal/logging/zerolog.go
@@ -3,8 +3,8 @@ package logging
 import (
 	"fmt"
 	"os"
-	"strings"
 	"time"
+	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


internal/logging/zerolog.go
```diff
@@ -40,7 +45,6 @@ func InitializeStdout() zerolog.Logger {
 	return decorate(log.Output(zerolog.ConsoleWriter{
 		Out:        os.Stdout,
 		TimeFormat: time.Kitchen,
-		NoColor:    true,
```
lzap, line 43: Also logger is now colored again, sorry this was not intentional.

lzap:
WAIT!

lzap:
Ok now it's fixed, I also wanted to keep the three dots and double quote. RTM

PR #153, opened 2022-08-10
Fix IB client

amirfefer:
During testing the IB client with #150, it ended with panic due to a failure in a casting.
Verified with ImageBuilder stage

### DIFF
--- a/internal/clients/image_builder/image_client.go
+++ b/internal/clients/image_builder/image_client.go
@@ -33,12 +33,16 @@ func (c *ImageBuilderClient) GetAWSAmi(ctx context.Context, composeID string) (s
 	if err != nil {
 		return "", err
 	}
+	ctxval.Logger(ctx).Debug().Msgf("Verifying AWS type")
 	if imageStatus.Type != UploadTypesAws {
 		ctxval.Logger(ctx).Warn().Err(err).Msg("Image is not AWS type")
 		return "", BadImageTypeErr
 	}
-	awsStatus := imageStatus.Options.(AWSUploadStatus)
-	return awsStatus.Ami, nil
+	awsStatus, ok := imageStatus.Options.(map[string]interface{})
+	if !ok {
+		return "", BadImageTypeErr
+	}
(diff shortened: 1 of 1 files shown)


internal/clients/image_builder/image_client.go
```diff
@@ -33,12 +33,16 @@ func (c *ImageBuilderClient) GetAWSAmi(ctx context.Context, composeID string) (s
...
 		ctxval.Logger(ctx).Warn().Err(err).Msg("Image is not AWS type")
 		return "", BadImageTypeErr
 	}
-	awsStatus := imageStatus.Options.(AWSUploadStatus)
-	return awsStatus.Ami, nil
+	awsStatus, ok := imageStatus.Options.(map[string]interface{})
+	if !ok {
+		return "", BadImageTypeErr
+	}
+	return awsStatus["ami"].(string), nil
```
amirfefer, line 45: we can use the `ok` pattern here as well to prevent an unwanted panic

internal/clients/image_builder/image_client.go
```diff
@@ -56,7 +60,8 @@ func (c *ImageBuilderClient) fetchImageStatus(ctx context.Context, composeID str
 		ctxval.Logger(ctx).Warn().Msgf("Image builder replied with unexpected status while fetching image status: %v", statusCode)
 		return nil, ClientErr
 	}
-	err = verifyImage(resp.JSON200)
+	ctxval.Logger(ctx).Debug().Msgf("Fetching image status was finished %+v\n", resp.JSON200.ImageStatus.UploadStatus)
```
lzap, line 63: This is something you used for debugging, not sure how useful is this from the operational perspective.

lzap:
LGTM I don't have IB setup yet.

adiabramovitch (approved):
Works for me 👏🏻

ezr-ondrej:
@amirfefer can you just improve the logging so we can merge pls?

amirfefer:
cleaned the logs 👍

PR #154, opened 2022-08-11
Client proxy support with basic auth for stage

lzap:
This patch is continuation of https://github.com/RHEnVision/provisioning-backend/pull/150

The idea is to introduce HTTP client configurable proxy per individual service (sources, ib). This is meant for development use against stage. Unfortunately, RHID header cannot be used in this scenario, therefore optional basic auth was added. This is only meant for development, in test/stage/prod environments the basic auth will be blank therefore RHID will be passed instead.

### DIFF
--- a/internal/clients/image_builder/image_client.go
+++ b/internal/clients/image_builder/image_client.go
@@ -20,11 +20,37 @@ func init() {
 }
 
 func newImageBuilderClient(ctx context.Context) (clients.ImageBuilder, error) {
-	c, err := NewClientWithResponses(config.ImageBuilder.URL)
+	c, err := NewClientWithResponses(config.ImageBuilder.URL, func(c *Client) error {
+		if config.ImageBuilder.Proxy.URL != "" {
+			if config.Features.Environment != "development" {
+				return clients.ClientProxyProductionUseErr
+			}
+			client, err := clients.NewProxyDoer(ctx, config.ImageBuilder.Proxy.URL)
+			if err != nil {
+				return fmt.Errorf("cannot create proxy doer: %w", err)
+			}
+			c.Client = client
+		}
+		return nil
+	})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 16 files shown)


adiabramovitch:
Do we want to add the new routes to the spec file?

internal/clients/image_builder/image_client.go
```diff
@@ -20,13 +20,36 @@ func init() {
 }
 
 func newImageBuilderClient(ctx context.Context) (clients.ImageBuilder, error) {
-	c, err := NewClientWithResponses(config.ImageBuilder.URL)
+	c, err := NewClientWithResponses(config.ImageBuilder.URL, func(c *Client) error {
+		if config.ImageBuilder.Proxy.URL != "" {
```
ezr-ondrej, line 24: It makes me bit unease to not have gate for this by not being in prod environment :thinking:

lzap: So let me translate: you would like to have a prod gating? I can do that.

lzap:
Rebased.

lzap:
> Do we want to add the new routes to the spec file?

I would not add them so there is no API commitment. It's just for us.

adiabramovitch:
> > Do we want to add the new routes to the spec file?
> 
> I would not add them so there is no API commitment. It's just for us.

So LGTM  😃 , @ezr-ondrej do you have any comments?

internal/clients/image_builder/image_client.go
ezr-ondrej: yeah, I know how to write unreadable sentences, don't I xD
That was exactly what I meat, thanks :+1:

ezr-ondrej (approved):
Code LGTM, thanks @lzap good stuff :+1:
Did not test tho :)

adiabramovitch (approved):
Works as expected! 👍🏻 
thanks @lzap

PR #155, opened 2022-08-15
Add reservation to OpenApi

adiabramovitch:
@lzap Can you take a look at the descriptions I wrote? I think they lack information :)

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -33,6 +33,59 @@
       }
     },
     "schemas": {
+      "v1.AWSReservationRequest": {
+        "properties": {
+          "amount": {
+            "format": "int32",
+            "type": "integer"
+          },
+          "image_id": {
+            "type": "string"
+          },
+          "instance_type": {
+            "type": "string"
+          },
+          "pubkey_id": {
+            "format": "int64",
+            "type": "integer"
+          },
+          "source_id": {
+            "format": "int64",
+            "type": "integer"
+          }
+        },
+        "type": "object"
+      },
+      "v1.AWSReservationResponse": {
+        "properties": {
+          "amount": {
+            "format": "int32",
+            "type": "integer"
+          },
+          "aws_reservation_id": {
+            "type": "string"
+          },
+          "image_id": {
+            "type": "string"
+          },
+          "instance_type": {
+            "type": "string"
+          },
+          "pubkey_id": {
+            "format": "int64",
+            "type": "integer"
+          },
+          "reservation_id": {
+            "format": "int64",
+            "type": "integer"
+          },
+          "source_id": {
+            "format": "int64",
+            "type": "integer"
+          }
+        },
+        "type": "object"
+      },
       "v1.Account": {
         "properties": {
           "account_number": {
@@ -56,6 +109,15 @@
         },
         "type": "object"
       },
+      "v1.NoopReservationResponse": {
+        "properties": {
+          "reservation_id": {
+            "format": "int64",
+            "type": "integer"
+          }
+        },
+        "type": "object"
+      },
       "v1.Pubkey": {
         "properties": {
           "body": {
@@ -67,6 +129,32 @@
           },
           "name": {
             "type": "string"
+          }
+        },
+        "type": "object"
+      },
+      "v1.ReservationResponse": {
+        "properties": {
+          "created_at": {
+            "format": "date-time",
+            "type": "string"
+          },
+          "finished_at": {
+            "format": "date-time",
+            "type": "string"
+          },
+          "id": {
+            "format": "int64",
+            "type": "integer"
+          },
+          "provider": {
+            "type": "integer"
+          },
+          "status": {
+            "type": "string"
+          },
+          "success": {
+            "type": "boolean"
           }
         },
         "type": "object"
@@ -256,6 +344,81 @@
           },
           "404": {
             "$ref": "#/components/responses/NotFound"
+          },
+          "500": {
+            "$ref": "#/components/responses/InternalError"
+          }
+        }
+      }
+    },
+    "/reservations": {
+      "get": {
+        "description": "A reservation is a way to activate a job, keeps all data needed for a job to start. This operation returns list of all reservations for particular account.\n",
+        "operationId": "getReservationsList",
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "items": {
+                    "$ref": "#/components/schemas/v1.ReservationResponse"
+                  },
+                  "type": "array"
+                }
+              }
+            },
+            "description": "Returned on success."
+          },
+          "500": {
+            "$ref": "#/components/responses/InternalError"
+          }
+        }
+      }
+    },
+    "/reservations/aws": {
+      "post": {
+        "description": "A reservation is a way to activate a job, keeps all data needed for a job to start. An AWS reservation is a reservation created for an AWS job.\n",
+        "operationId": "createAwsReservation",
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/v1.AWSReservationRequest"
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/v1.AWSReservationResponse"
+                }
+              }
+            },
+            "description": "Returned on success."
+          },
+          "500": {
+            "$ref": "#/components/responses/InternalError"
+          }
+        }
+      }
+    },
+    "/reservations/noop": {
+      "post": {
+        "description": "A reservation is a way to activate a job, keeps all data needed for a job to start. A Noop reservation actually does nothing and immediately finish background job. This reservation has no input payload\n",
+        "operationId": "createNoopReservation",
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/v1.NoopReservationResponse"
+                }
+              }
+            },
+            "description": "Returned on success."
           },
           "500": {
             "$ref": "#/components/responses/InternalError"
--- a/api/openapi.gen.yaml
+++ b/api/openapi.gen.yaml
@@ -133,6 +133,59 @@ paths:
             $ref: "#/components/responses/NotFound"
           '500':
             $ref: "#/components/responses/InternalError"
+  /reservations:
+    get:
+      operationId: getReservationsList
+      description: >
+        A reservation is a way to activate a job, keeps all data needed for a job to start.
+        This operation returns list of all reservations for particular account.
+      responses:
+        '200':
+          description: 'Returned on success.'
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/v1.ReservationResponse'
+        "500":
+          $ref: '#/components/responses/InternalError'
+  /reservations/aws:
+    post:
+      operationId: createAwsReservation
+      description: >
+        A reservation is a way to activate a job, keeps all data needed for a job to start.
+        An AWS reservation is a reservation created for an AWS job.
+      requestBody: 
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/v1.AWSReservationRequest'
+      responses:
+        '200':
+          description: 'Returned on success.'
+          content:
+            application/json:
+              schema:
+                  $ref: '#/components/schemas/v1.AWSReservationResponse'
+        "500":
+          $ref: '#/components/responses/InternalError'
+  /reservations/noop:
+    post:
+      operationId: createNoopReservation
+      description: >
+        A reservation is a way to activate a job, keeps all data needed for a job to start.
+        A Noop reservation actually does nothing and immediately finish background job.
+        This reservation has no input payload
+      responses:
+        '200':
+          description: 'Returned on success.'
+          content:
+            application/json:
+              schema:
+                  $ref: '#/components/schemas/v1.NoopReservationResponse'
+        "500":
+          $ref: '#/components/responses/InternalError'
 components:
   responses:
     BadRequest:
@@ -154,6 +207,43 @@ components:
             $ref: '#/components/schemas/v1.ResponseError'
       description: The specified resource was not found
   schemas:
+    v1.AWSReservationRequest:
+      properties:
+        amount:
+          format: int32
+          type: integer
+        image_id:
+          type: string
+        instance_type:
+          type: string
+        pubkey_id:
+          format: int64
+          type: integer
+        source_id:
+          format: int64
+          type: integer
+      type: object
+    v1.AWSReservationResponse:
+      properties:
+        amount:
+          format: int32
+          type: integer
+        aws_reservation_id:
+          type: string
+        image_id:
+          type: string
+        instance_type:
+          type: string
+        pubkey_id:
+          format: int64
+          type: integer
+        reservation_id:
+          format: int64
+          type: integer
+        source_id:
+          format: int64
+          type: integer
+      type: object
     v1.Account:
       properties:
         account_number:
@@ -169,6 +259,12 @@ components:
         id:
           type: string
       type: object
+    v1.NoopReservationResponse:
+      properties:
+        reservation_id:
+          format: int64
+          type: integer
+      type: object
     v1.Pubkey:
       properties:
         body:
@@ -178,6 +274,24 @@ components:
           type: integer
         name:
           type: string
+      type: object
+    v1.ReservationResponse:
+      properties:
+        created_at:
+          format: date-time
+          type: string
+        finished_at:
+          format: date-time
+          type: string
+        id:
+          format: int64
+          type: integer
+        provider:
+          type: integer
+        status:
+          type: string
+        success:
+          type: boolean
       type: object
     v1.ResponseError:
       properties:
--- a/cmd/spec/main.go
+++ b/cmd/spec/main.go
@@ -53,6 +53,10 @@ func main() {
 	gen.addSchema("v1.ResponseError", &payloads.ResponseError{})
 	gen.addSchema("v1.Source", &payloads.SourceResponse{})
 	gen.addSchema("v1.InstanceType", &payloads.InstanceTypeResponse{})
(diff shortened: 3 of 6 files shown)


api/openapi.gen.yaml
```diff
@@ -133,6 +133,57 @@ paths:
...
+                items:
+                  $ref: '#/components/schemas/v1.Reservation'
+        "500":
+          $ref: '#/components/responses/InternalError'
+  /reservations/aws:
+    post:
+      operationId: createAwsReservation
+      description: >
+        A reservation is a way to activate a job, keeps all data needed for a job to start.
+        an AWS reservation is a reservation created for an AWS job.
```
lzap, line 158:
```suggestion
        An AWS reservation is a reservation created for an AWS job. 
```

api/openapi.gen.yaml
```diff
@@ -133,6 +133,57 @@ paths:
...
+                items:
+                  $ref: '#/components/schemas/v1.AWSReservation'
+        "500":
+          $ref: '#/components/responses/InternalError'
+  /reservations/noop:
+    post:
+      operationId: createNoopReservation
+      description: >
+        A reservation is a way to activate a job, keeps all data needed for a job to start.
+        a Noop reservation actually does nothing and immediately finish background job.
```
lzap, line 175:
```suggestion
        A Noop reservation actually does nothing and immediately finish background job. 
```

cmd/spec/path.yaml
```diff
@@ -133,3 +133,54 @@ paths:
...
+        '200':
+          description: 'Returned on success.'
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/v1.NoopReservation'
+        "500":
+          $ref: '#/components/responses/InternalError'
```
lzap, line 186: Damn I commented on the wrong file, do the edits here not there :)

lzap:
Other than that, it's fine.

I would really, really like to find a better name for this. These are tasks, not reservations. But Ondra and Marek really did not like tasks for some reason.

I think the endpoint should be named tasks and it should contain filtering mechanism so only reservations (or launches or whatever term we end up using in the UI) could be filtered. The same way we could filter other tasks like SSH pubkey deletion.

ezr-ondrej:
> I think the endpoint should be named tasks

I believe that it is overgeneralization :thinking: I do not have any data to proove it at hand, but I don't like too general endpoints, pages as those get overloaded and too complex. I'd rather see endpoint /reservations/xxx and /pubkeys/deletions ... or whatever the names would be. And also have pubkey list, where we would show delete in progress. And reservations page, where we would have reservation status.
Tho this is just gut feeling and would need to get discussed (probably with UX) :)

lzap:
That's a decent solution as well, although I like /tasks more as it is more consistent I won't insist.

Sorry dragging this elsewhere, @adiabramovitch it looks good!

lzap (approved):
Thanks.

lzap:
This needs approval from @RHEnVision/qe thanks. We are adding a new endpoint, use the noop operation to test the job queue.

cmd/spec/path.yaml
```diff
@@ -133,3 +133,54 @@ paths:
...
+      description: >
+        A reservation is a way to activate a job, keeps all data needed for a job to start.
+        A Noop reservation actually does nothing and immediately finish background job. 
+      responses:
+        '200':
+          description: 'Returned on success.'
+          content:
+            application/json:
+              schema:
+                type: array
```
adiabramovitch, line 182: is that supposed to be array? @lzap @ezr-ondrej

ezr-ondrej: I guess we should respond with empty body, but does not matter much, best would be to check what we actually return in the request and just go with that.

adiabramovitch: this is what we are returning an AWSReservation

lzap: Oh I totally forgot, this is the problem that @avitova had to solve because `sql.NullXYZ` does not nicely map to JSON `null`. Schema generator does not support that, so it actually generates invalid schema (empty maps). The solution was to split model and payload and we need to do the same here, unless someone else figures out better way.

api/openapi.gen.json
```diff
@@ -33,6 +33,48 @@
...
+            "type": "integer"
+          },
+          "aws_reservation_id": {
+            "type": "string"
+          },
+          "created_at": {
+            "format": "date-time",
+            "type": "string"
+          },
+          "finished_at": {},
```
lzap, line 49: This is a problem I overlooked earlier, the OpenAPI generator does not support `sql.NullXYZ` structs, I mean it can't.

But this can be mapped nicely because JSON does support `null` natively. We tried to implement our own marshallers but it did not work out so we ended up using separate Payload struct with simply string pointer (here it would be timestamp pointer - perhaps integer) where Go `nil` maps to JSON `null`.

See here:

```go
type AccountPayload struct {
	// Required auto-generated PK.
	ID int64 `json:"id"`

	// Organization ID. Required.
	OrgID string `json:"org_id"`

	// EBS account number. Can be NULL but not blank.
	AccountNumber *string `json:"account_number"`
}
type AccountRequest AccountPayload
type AccountResponse AccountPayload
```

lzap (changes requested):
There is a problem with SQL NULL types.

api/openapi.gen.json
```diff
@@ -33,6 +33,48 @@
...
+            "type": "integer"
+          },
+          "source_id": {
+            "format": "int64",
+            "type": "integer"
+          },
+          "status": {
+            "type": "string"
+          },
+          "success": {}
```
adiabramovitch, line 74: I guess the same goes for success. @lzap ?

api/openapi.gen.json
adiabramovitch: And it means I need to create a new payload for the was reservation and the reservation

akhil-jha:
@adiabramovitch , noop and listing the reservations is working fine.
I don't think aws reservation is still ready yet, right?

adiabramovitch:
> 

I guess it is not supposed to work if you won't have an AWS account configured

akhil-jha:
> > 
> 
> I guess it is not supposed to work if you won't have an AWS account configured

ah. It is not yet. No reason to keep this PR in blocking from qe end then. 
Removing the `Needs qe ack` label and ACKing.

akhil-jha (approved):

api/openapi.gen.json
lzap: That reads like an unfinished sentence, but that is unfortunately the solution we landed on with @avitova, yes. I do not like it either, but this is exactly why we do actually have payload types - I was expecting this at some point. Came little bit earlier tho :-)

api/openapi.gen.json
adiabramovitch: I have changed it but it doesnt seem to work for noop and was, do you know why?
I tried to remove the usage of &models.SOME_MODEL but that all the fields were removed for some reason

internal/payloads/reservation_payload.go
```diff
@@ -2,33 +2,58 @@ package payloads
...
+
+	// Textual status of the reservation or error when there was a failure
+	Status string `json:"status"`
+
+	// Time when reservation was finished or nil when it's still processing.
+	FinishedAt time.Time `json:"finished_at"`
+
+	// Flag indicating success, error or unknown state (NULL). See Status for the actual error.
+	Success bool `json:"success"`
 }
```
lzap, line 34: All the fields which you want to be NULLable in JSON must be pointers. Only then they will be marshalled as JSON "null" values (when "nil" in Go).

internal/payloads/reservation_payload.go
```diff
@@ -2,33 +2,58 @@ package payloads
...
+
+type ReservationResponse ReservationPayload
+
+type NoopReservationRequest ReservationRequest
+
+type NoopReservationResponse NoopReservationPayload
+
+type AWSReservationRequest AWSReservationPayload
+
+type AWSReservationResponse AWSReservationPayload
```
lzap, line 56: I think this gives us opportunity to split request and response payloads and customize them a bit. I can imagine that for both types, requests do not need the following fields to be present:

* AccountID (this one can be completely deleted from all payloads)
* ID (because this is generated on save)
* Provider (because that is part of the URL)
* Created
* Status
* Finished
* Success (all the four are relevant only for response)

This all applies to both AWS and Noop. For the latter, it will probably be a completely empty request payload. We want to remove them, because then they will not be generated in the OpenAPI and it will not confuse API users.

Responses are a different story, there we need all of the above. As I said, NULLable fields must be pointers tho.

api/openapi.gen.yaml
```diff
@@ -179,6 +273,24 @@ components:
...
+        id:
+          format: int64
+          type: integer
+        provider:
+          type: integer
+        status:
+          type: string
+        success:
+          type: boolean
+      type: object
```
lzap, line 293: So success was generated here correctly (ignoring the fact this should be a pointer). Not sure why it does not generate correctly for AWS/Noop yet.

cmd/spec/main.go
```diff
@@ -53,6 +53,9 @@ func main() {
 	gen.addSchema("v1.ResponseError", &payloads.ResponseError{})
 	gen.addSchema("v1.Source", &payloads.SourceResponse{})
 	gen.addSchema("v1.InstanceType", &payloads.InstanceTypeResponse{})
+	gen.addSchema("v1.Reservation", &payloads.ReservationResponse{})
+	gen.addSchema("v1.NoopReservation", &payloads.NoopReservationResponse{})
+	gen.addSchema("v1.AWSReservation", &payloads.AWSReservationResponse{})
```
lzap, line 58: I see one flaw with this code, those v1 names do not include Request or Response words. For some endpoints we need to differentiate between the two in the specification. Like in this case. This needs to be fixed, let us do it only for reservations in this PR tho:

```
    gen.addSchema("v1.ReservationResponse", &payloads.ReservationResponse{})
	gen.addSchema("v1.NoopReservationRequest", &payloads.NoopReservationRequest{})
	gen.addSchema("v1.NoopReservationResponse", &payloads.NoopReservationResponse{})
    // the same for AWS
```

lzap (changes requested):
Hmm not sure why it does not generate it correctly for embedded fields, can you try pointers if that helps? Might be a bug in the OpenAPI generator we use.

cmd/spec/main.go
lzap: FYI I am filing a draft to refactor the existing ones: https://github.com/RHEnVision/provisioning-backend/pull/167

ezr-ondrej:
Can you fix the failure pls? We forgot to use the new payload name on some places

api/openapi.gen.json
lzap: Found the reason, you embed the models into payloads. They do overwrite the fields with the sql.Null ones. You should not use embedding, just copy and paste those fields and make sure you leave only json tags in payloads and db tags in the model:

```diff
        // Flag indicating success, error or unknown state (NULL). See Status for the actual error.
-       Success sql.NullBool `db:"success" json:"success"`
+       Success sql.NullBool `db:"success"`
 }
```

internal/payloads/reservation_payload.go
```diff
@@ -2,38 +2,77 @@ package payloads
...
+	FinishedAt *time.Time `json:"finished_at"`
+
+	// Flag indicating success, error or unknown state (NULL). See Status for the actual error.
+	Success *bool `json:"success"`
 }
 
-type AWSReservationRequest struct {
+type AWSReservationResponsePayload struct {
+	ReservationResponsePayload
 	*models.AWSReservation
```
lzap, line 35: This line breaks it, instead of struct embedding just copy and paste the fields from the model:

```
	// Pubkey ID.
	PubkeyID int64 `db:"pubkey_id" json:"pubkey_id"`

	// Source ID.
	SourceID int64 `db:"source_id" json:"source_id"`

	//AWS Instance type.
	InstanceType string `db:"instance_type" json:"instance_type"`

	// Amount of instances to provision of type: Instance type.
	Amount int32 `db:"amount" json:"amount"`

	ImageID string `db:"image_id" json:"image_id"`

	// The ID of the aws reservation which was created.
```

Make sure to remove all the db tags from here, and in the model remove json tags.

Alternatively, if you still want to use embedding, you need to split the model into two structs and only embed the detail part:

```go
type AWSReservationDetail struct {
	// Pubkey ID.
	PubkeyID int64 `db:"pubkey_id" json:"pubkey_id"`

	// Source ID.
	SourceID int64 `db:"source_id" json:"source_id"`

	//AWS Instance type.
	InstanceType string `db:"instance_type" json:"instance_type"`

	// Amount of instances to provision of type: Instance type.
	Amount int32 `db:"amount" json:"amount"`

	ImageID string `db:"image_id" json:"image_id"`

	// The ID of the aws reservation which was created.
}

type AWSReservation struct {
	Reservation
	AWSReservationDetail
}
```

In that case keep both db and json tags. You can only use this when there are no NULLable fields tho. It is up to you how to approach it, I think the split should be better.

lzap:
Validation error.

api/openapi.gen.json
```diff
@@ -33,6 +33,32 @@
...
+            "format": "int64",
+            "type": "integer"
+          },
+          "source_id": {
+            "format": "int64",
+            "type": "integer"
+          }
+        },
+        "type": "object"
+      },
```
lzap, line 61: I am missing one very important field - reservation id. That is really needed for doing followup updates on the status.

api/openapi.gen.json
```diff
@@ -71,6 +123,32 @@
...
+          },
+          "status": {
+            "type": "string"
+          },
+          "success": {
+            "type": "boolean"
+          }
+        },
+        "type": "object"
+      },
```
lzap, line 151: Finally, this looks good :-) One idea: maybe we could name the schema name as v1.GenericReservationResponse to really differentiate it. But that is just an idea.

cmd/spec/main.go
```diff
@@ -53,6 +53,10 @@ func main() {
 	gen.addSchema("v1.ResponseError", &payloads.ResponseError{})
 	gen.addSchema("v1.Source", &payloads.SourceResponse{})
 	gen.addSchema("v1.InstanceType", &payloads.InstanceTypeResponse{})
+	gen.addSchema("v1.ReservationResponse", &payloads.ReservationResponse{})
+	gen.addSchema("v1.NoopReservationResponse", &payloads.NoopReservationResponse{})
+	gen.addSchema("v1.AWSReservationResponse", &payloads.AWSReservationResponse{})
+	gen.addSchema("v1.AWSReservationResponse", &payloads.AWSReservationResponse{})
```
lzap, line 59: Two lines are the same, I think you wanted to say Request and Response there.

cmd/spec/path.yaml
```diff
@@ -133,3 +133,54 @@ paths:
...
+              schema:
+                  $ref: '#/components/schemas/v1.AWSReservationResponse'
+        "500":
+          $ref: '#/components/responses/InternalError'
+  /reservations/noop:
+    post:
+      operationId: createNoopReservation
+      description: >
+        A reservation is a way to activate a job, keeps all data needed for a job to start.
+        A Noop reservation actually does nothing and immediately finish background job.
```
lzap, line 177: Maybe it is worth adding: _This reservation has no input payload._

internal/payloads/reservation_payload.go
```diff
@@ -2,38 +2,85 @@ package payloads
...
+	FinishedAt *time.Time `json:"finished_at"`
+
+	// Flag indicating success, error or unknown state (NULL). See Status for the actual error.
+	Success *bool `json:"success"`
 }
 
-type AWSReservationRequest struct {
-	*models.AWSReservation
+type AWSReservationResponsePayload struct {
+	Reservation ReservationResponsePayload
```
lzap, line 34: This is a typo - you actually add a new field named Reservation which is not marshalled because it is missing the JSON tag. Since we only want to show reservation ID, I would simply copy the ID here and copy it manually in the function, no need to make things more complex then they need. Do not embed this is what I mean.

internal/payloads/reservation_payload.go
```diff
@@ -2,38 +2,85 @@ package payloads
...
+	ImageID string `json:"image_id"`
+
+	// The ID of the aws reservation which was created.
+	AWSReservationID string `json:"aws_reservation_id"`
 }
 
-type AWSReservationResponse struct {
-	*models.AWSReservation
+type NoopReservationResponsePayload struct {
+	ReservationResponsePayload
```
lzap, line 56: Here as well, only ID needs to be shown.

internal/payloads/reservation_payload.go
```diff
@@ -2,38 +2,85 @@ package payloads
...
+type AWSReservationResponsePayload struct {
+	Reservation ReservationResponsePayload
+
+	// Pubkey ID.
+	PubkeyID int64 `json:"pubkey_id"`
+
+	// Source ID.
+	SourceID int64 `json:"source_id"`
+
+	//AWS Instance type.
```
lzap, line 42: Space.

internal/payloads/reservation_payload.go
```diff
@@ -2,38 +2,85 @@ package payloads
...
+	ImageID string `json:"image_id"`
 }
 
+type AWSReservationRequest AWSReservationRequestPayload
+
+type NoopReservationResponse NoopReservationResponsePayload
+
+type AWSReservationResponse AWSReservationResponsePayload
+
+type ReservationResponse ReservationResponsePayload
```
lzap, line 82: I do not understand these aliases, we simply need the following types:

* AWSReservationRequestPayload
* AWSReservationResponsePayload
* NoopReservationRequestPayload
* NoopReservationResponsetPayload
* GenericReservationResponsePayload (was: ReservationResponsePayload)

Nothing else, no aliases or embedding is needed, these are all unique payloads.

api/openapi.gen.json
adiabramovitch: the struct includes ReservationResponsePayload which has the id. 
I see that none of the ReservationResponsePayload fields are shown..

adiabramovitch: added :)

ezr-ondrej (approved):
LGTM, some descriptions might be bit improved (jobs -> provisioning jobs), but we can improve as we go, let's expose the main endpoint :+1:

lzap (approved):
Well, this was harder than we all expected. Kudos! Merging :)

PR #157, opened 2022-08-15
Add aws reservation service test

adiabramovitch:


internal/services/aws_reservation_service_test.go
```diff
@@ -0,0 +1,57 @@
...
+		t.Fatal("unable to marshal values to json")
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", "/api/provisioning/reservations/aws", bytes.NewBuffer(json_data))
+	req.Header.Add("Content-Type", "application/json")
+	assert.Nil(t, err, fmt.Sprintf("Error creating a new request: %v", err))
+
+	rr := httptest.NewRecorder()
+	handler := http.HandlerFunc(CreateAWSReservation)
+	handler.ServeHTTP(rr, req)
```
adiabramovitch, line 50: I seem to have a problem with an uninitialized job queue (https://github.com/RHEnVision/provisioning-backend/blob/42110d4edf1ac22a6fa4cfa110e21916b08ade2a/internal/services/aws_reservation_service.go#L133). 
 Do you any suggestion how should I stub it or initialize it?  @lzap , @ezr-ondrej

lzap: I am not sure what level of integration do you expect in your test. I would say, that for unit tests, we should stub everything (database, REST clients, queue). So you want to only test the creation of job, the actual processing shall be tested separately in another test.

For this, you can utilize an implementation called "memory" which is asynchronous memory-only. You could initialize a single shared implementation in an `init()` function that would be common for all tests, with all handlers registered to "do nothing". It will silently swallow everything, maybe there could be some kind of counter that would increase or better (for concurrency reasons) providing a way to retrieve enqueued jobs.

You know what let me do a small prototype, I need to test my ideas.

lzap: See in here: https://github.com/RHEnVision/provisioning-backend/pull/159

internal/services/pubkey_service_test.go
```diff
@@ -19,9 +19,9 @@ func TestListPubkeysHandler(t *testing.T) {
 	ctx := stubs.WithAccountDaoOne(context.Background())
 	ctx = identity.WithTenant(t, ctx)
 	ctx = stubs.WithPubkeyDao(ctx)
-	err := stubs.GeneratePubkey(ctx, models.Pubkey{})
+	err := stubs.GeneratePubkey(ctx, &models.Pubkey{})
```
ezr-ondrej, line 22: FYI #145 is now addressing this, so once merged, we can drop this change.

internal/services/aws_reservation_service_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"net/http/httptest"
+	"testing"
+
+	ibClientStub "github.com/RHEnVision/provisioning-backend/internal/clients/image_builder/stubs"
+	sourcesClientStub "github.com/RHEnVision/provisioning-backend/internal/clients/sources/stubs"
+	"github.com/RHEnVision/provisioning-backend/internal/dao/stubs"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
+	"github.com/stretchr/testify/assert"
+)
```
lzap, line 18: This is missing the import, the "go test" command picks tests in random order (or unknown order to me) and this can possibly fail it it gets executed as the first test:

    _ "github.com/RHEnVision/provisioning-backend/internal/testing/initialization"

lzap (changes requested):
It needs an import. If you get dependency cycle, rename the package to "services_test".

internal/services/aws_reservation_service_test.go
adiabramovitch: Added :)

lzap (approved):
TY

PR #158, opened 2022-08-15
Add Source ID

avitova:
Please, let me know if we need to change Source ID in updatePubkeyResource. :) I was not sure about that one. Both integration and unit test are passing.

### DIFF
--- a/internal/dao/sqlx/pubkey_resource_dao.go
+++ b/internal/dao/sqlx/pubkey_resource_dao.go
@@ -12,8 +12,8 @@ import (
 )
 
 const (
-	createPubkeyResource            = `INSERT INTO pubkey_resources (pubkey_id, provider, handle, tag) VALUES ($1, $2, $3, $4) RETURNING id, tag`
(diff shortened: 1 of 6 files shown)


ezr-ondrej:
> Please, let me know if we need to change Source ID in updatePubkeyResource.

Good question, hard to answer, because we are not using that method, so I'm of an opinion we should not have such a method, but given it's there, lets just allow changing everything. I'd argue for deleting it, but thats out of scope :)

ezr-ondrej (approved):
Needs db drop and remigrate, so I'll wait for everyone to see we are adding this.
@lzap @adiabramovitch @amirfefer can you pls ack? :)

lzap:
I vote for deleting the unused update function. If you can squeeze that in, not a blocker tho.

adiabramovitch:
Sure we just need a follow-up PR for that - to add the assume role part in the handler itself.

adiabramovitch (approved):

ezr-ondrej:
> I vote for deleting the unused update function

I'll do that in a follow-up to have a discussion over unused functions in general :)

PR #159, opened 2022-08-15
Job queue stub and test for noop job

lzap:
Testing of services which enqueue new jobs was not possible because job queue was not initialized in tests. This was found in https://github.com/RHEnVision/provisioning-backend/pull/157. To be able to initialize it to "memory" implementation, there must be a package that does this on import. This patch creates such package, therefore all unit tests which need to call services (or any code) that enqueue any job must have this:

```go
_ "github.com/RHEnVision/provisioning-backend/internal/jobs/stub"
```

It makes a lot of sense to also have a good logging for tests, therefore in this patch I extracted that into separate package too, so tests should import all the packages in order to prevent nil pointer errors:

```go
	// Initialize logging (must be kept the first)
	_ "github.com/RHEnVision/provisioning-backend/internal/logging/testing"
	// HTTP client stub implementations
	_ "github.com/RHEnVision/provisioning-backend/internal/clients/image_builder/stubs"
	_ "github.com/RHEnVision/provisioning-backend/internal/clients/sources/stubs"
	// DAO stub implementation
	_ "github.com/RHEnVision/provisioning-backend/internal/dao/stubs"
	// Job queue stub
	_ "github.com/RHEnVision/provisioning-backend/internal/jobs/stub"
```

The patch introduces new package internal/testing/initialization which all tests should import in order to get the benefits. We do not need to add this into all tests, small unit tests that only test a single function do not benefit from this, but all DAO and service tests should import it.

I will add imports to all relevant tests once we agree on the solution.


internal/services/noop_reservation_service_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	err := jobs.Enqueue(context.Background(), pj)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func TestCreateNoopReservation(t *testing.T) {
+	// TODO full service test
+}
```
lzap, line 29: Cannot implement this test without: https://github.com/RHEnVision/provisioning-backend/pull/157

internal/services/noop_reservation_service_test.go
```diff
@@ -0,0 +1,29 @@
+package services
+
+import (
+	"context"
+	"testing"
+
+	_ "github.com/RHEnVision/provisioning-backend/internal/testing/initialization"
```
lzap, line 7: This import needs to be added in most our tests for a consistent experience (no nil pointer exceptions, all tests do logging in human-readable form).

internal/testing/initialization/test_init.go
```diff
@@ -0,0 +1,20 @@
...
+	_ "github.com/RHEnVision/provisioning-backend/internal/dao/stubs"
+	// Job queue stub
+	_ "github.com/RHEnVision/provisioning-backend/internal/jobs/stub"
+
+	"github.com/rs/zerolog/log"
+)
+
+func init() {
+	log.Logger.Debug().Msg("initialized logging, job queue for tests")
+}
```
lzap, line 20: This should perhaps read: "initialized logging, HTTP client(s), DAO and job queue stubs"

Will fix in the next rebase.

internal/jobs/queue.go
```diff
@@ -49,8 +50,37 @@ func Initialize(ctx context.Context, logger *zerolog.Logger) error {
...
+	ctx := context.Background()
+	queue, err = mem.NewClient(ctx, zerologr.New(&log.Logger))
+	if err != nil {
+		panic(err)
+	}
+	queue.RegisterHandler(TypeNoop, HandleStub)
+	queue.RegisterHandler(TypePubkeyUploadAws, HandleStub)
+	queue.RegisterHandler(TypeLaunchInstanceAws, HandleStub)
+	queue.DequeueLoop(ctx)
+	log.Logger.Debug().Msg("initialized job queue stub")
```
lzap, line 85: Unfortunately, we need to keep this list of handlers up to date with the production list above. I can perhaps implement a flag in the dejq memory implementation to silently ignore when a handler is not registered, that would do it. Opinions?

internal/jobs/queue.go
```diff
@@ -49,8 +50,37 @@ func Initialize(ctx context.Context, logger *zerolog.Logger) error {
...
+	queue.DequeueLoop(ctx)
+}
+
+func StopDequeueLoop() {
+	queue.Stop()
+}
+
+func HandleStub(_ context.Context, _ dejq.Job) error {
+	return nil
+}
```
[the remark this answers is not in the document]
lzap: Oh the implementation actually silently skips if there is no handler available! I can delete this completely, great. Will do in a rebase.

https://github.com/lzap/dejq/blob/master/mem/mem_client.go#L75

internal/jobs/queue.go
ezr-ondrej: I believe it should not be as bad to keep the list, we have the same for dao, if you wanna test something that needs enque something, you need to stub it's dependencies, one line dependency is ok IMHO :shrug:

ezr-ondrej (approved):
I wonder whether it makes sense to split the initialization methods into their files and call them on init() as init sideffect, but thats completely minor stuff.
:heart: this and personally I'm for merging asap as it is huge leap forward to be able to test queues :+1:

internal/jobs/queue.go
lzap: Actually, this was not valid comment. I figured out (see above) that the memory implementation silently skips unregistered handlers so I am going to remove these registrations and `HandleStub` completely! https://github.com/lzap/dejq/blob/master/mem/mem_client.go#L75

lzap:
One thing that appeared once I started adding the single-line import to other tests is cycle dependency:

```
# github.com/RHEnVision/provisioning-backend/internal/models
package github.com/RHEnVision/provisioning-backend/internal/models
	imports github.com/RHEnVision/provisioning-backend/internal/dao/stubs
	imports github.com/RHEnVision/provisioning-backend/internal/testing/initialization: import cycle not allowed in test
FAIL	github.com/RHEnVision/provisioning-backend/internal/models [setup failed]
```

When I add the import to `models/pubkey_resource_test.go` it imports the `initialization` which also imports `dao/stubs` which imports `models` and that is a cycle which is not allowed. Now, look what I've learned today:

If I change the package of the test from `models` to `models_test` it is no longer a cycle. This solves the problem.

lzap:
Rebased.

internal/services/aws_reservation_service.go
```diff
@@ -130,7 +130,7 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Enqueue all jobs
-	err = jobs.Queue.Enqueue(r.Context(), uploadPubkeyJob, launchJob)
+	err = jobs.Enqueue(r.Context(), uploadPubkeyJob, launchJob)
```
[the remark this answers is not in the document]
lzap: Initially I thought I would hide the implementation, the `Queue`, behind an interface so I can provide a stub implementation. But I realized that actually a single function is in use: `Enqueue` and I dropped the effort. I think I should complete this and follow the same pattern which we use elsewhere - defining a getXXXDao function that is set through `init()` functions.

Very good point, this is not nice, I will do that.

lzap:
FYI these are the functions I would need for the NoopServiceTest: https://github.com/RHEnVision/provisioning-backend/pull/157/files#diff-c13b43c97569ba22359f52f1409603261444e3c5ad46faf2dfd0e4d31946e33aR40

adiabramovitch:
> FYI these are the functions I would need for the NoopServiceTest: https://github.com/RHEnVision/provisioning-backend/pull/157/files#diff-c13b43c97569ba22359f52f1409603261444e3c5ad46faf2dfd0e4d31946e33aR40

Oh! I got it so we can implement that later :)

lzap:
So I pushed new commit with refactoring, it now uses "abstract factory function" (if that's a design pattern no idea). Basically the same approach we use in DAO and HTTP clients as well. **Squash on merge**.

internal/services/aws_reservation_service.go
```diff
@@ -118,7 +118,7 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 
 	logger.Debug().Msgf("Enqueuing launch instance job for source %d", reservation.SourceID)
 	launchJob := dejq.PendingJob{
-		Type: jobs.TypeLaunchInstanceAws,
+		Type: queue.TypeLaunchInstanceAws,
```
adiabramovitch, line 121: It is not supposed to be dejqQueue now?

lzap: Services must only use the interface, that is what's defined in `queue_interface.go` nothing else.

adiabramovitch (approved):
Tested it and it works! 
Thanks @lzap

PR #160, opened 2022-08-16
Add fifth missing arg to createPubkeyResource query

adiabramovitch:


avitova (approved):

lzap (approved):
Well, turns out that `TestCreatePubkey` is not correct :) Correction, the test is not supposed to test PubkeyResource! :)

PR #162, opened 2022-08-16
Change source id from string to int

adiabramovitch:
DO NOT MERGE :)

### DIFF
--- a/internal/clients/sources/sources_client.go
+++ b/internal/clients/sources/sources_client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	"github.com/RHEnVision/provisioning-backend/internal/cache"
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
@@ -41,13 +42,17 @@ func newSourcesClient(ctx context.Context) (clients.Sources, error) {
 	return &SourcesClient{client: c}, nil
 }
 
-func copySource(src Source) clients.Source {
-	return clients.Source{
-		Id:           src.Id,
+func copySource(src Source) (*clients.Source, error) {
+	srcId, err := strconv.ParseInt(*src.Id, 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create proxy doer: %w", err)
+	}
+	return &clients.Source{
+		Id:           srcId,
 		Name:         src.Name,
 		SourceTypeId: src.SourceTypeId,
 		Uid:          src.Uid,
-	}
+	}, nil
 }
 
 type appType struct {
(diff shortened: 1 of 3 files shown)


ezr-ondrej:
:thinking: I'm not sure we want that xD and if we can change the type of ID :)

adiabramovitch:
> thinking I'm not sure we want that xD and if we can change the type of ID :)

We want to enable the triggering from the UI. 
We saw that the triggering fails because we are sending the source id as a string and not as an int. 
I am wondering if we need that change or if should we do simple casing on the client side.

internal/clients/sources/sources_client.go
```diff
@@ -41,13 +42,17 @@ func newSourcesClient(ctx context.Context) (clients.Sources, error) {
 	return &SourcesClient{client: c}, nil
 }
 
-func copySource(src Source) clients.Source {
-	return clients.Source{
-		Id:           src.Id,
+func copySource(src Source) (*clients.Source, error) {
+	srcId, err := strconv.ParseInt(*src.Id, 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create proxy doer: %w", err)
```
ezr-ondrej, line 48: probably not the correct error message xD

ezr-ondrej (commented):
Oh right... well, in that case, this makes sense... although, I guess we can do it for all the ids, not just source id?

adiabramovitch:
> Oh right... well, in that case, this makes sense... although, I guess we can do it for all the ids, not just source id?

I am not sure we need to. 
The ones we need to check are the ones from the clients, Ib is a uuid so a string is valid and sources came up after we saw that they have a problem with their generated client

akhil-jha:
Would this require apispec update as well?

ezr-ondrej:
> I am not sure we need to. The ones we need to check are the ones from the clients, Ib is a uuid so a string is valid and sources came up after we saw that they have a problem with their generated client

I was more thinking about doing it for all the sources IDs we have other IDs in `Source`, but I'm fine as is, if this is fixing what we have, I'd go ahead and try to get this in (by polishing, making the OpenAPI change and fixing the checks)

lzap:
I am wondering if we should actually do it. I mean all PKs do have a separate schema that literally says it should be an integer formatted string: https://github.com/RedHatInsights/sources-api-go/blob/e3149080121894fb68a76c756192c27e87f0bb42/public/openapi-3-v3.1.json#L2090

It looks like this was a deliberate choice, maybe they are planning moving to UUIDs later. Although I would not do this and use PKs, I can understand it. What exactly is the problem on our side?

lzap:
Just for you to let know I asked the guys on Slack and the reply was that this likely looks like something from legacy version (Ruby I guess). They could consider moving to integer in the future. So I guess I would hold on for now and this could be solved with a client upgrade in the future.

adiabramovitch:
The reason why I suggested doing that because we need to cast in the UI. 
we are receiving the id as a string from the server 
in order to send that id in POST reservation/aws we are casting from string to int so the request will succeed. 
If we want to skip that, I am okay with it just think you should be aware from the UI perspective.

ezr-ondrej:
> aware from the UI perspective.

If it's just casting for the trigger request, that sounds sane to me :thinking:

PR #163, opened 2022-08-17
HMSPROV-79 Resource tests

avitova:
I think we might have a copy paste mistake in pubkey_resource.go, but I did not want to make a PR just for that.

### DIFF
--- a/cmd/dao_tests/pubkey_resource_test.go
+++ b/cmd/dao_tests/pubkey_resource_test.go
@@ -1,0 +1,112 @@
+//go:build integration
+// +build integration
+
+package main
+
+import (
+	"context"
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/factories"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func createPubkeyResourceNoop() *models.PubkeyResource {
+	return &models.PubkeyResource{
+		Tag:      "tag1",
+		PubkeyID: 1,
+		Provider: models.ProviderTypeNoop,
+		Handle:   factories.GetSequenceName("handle"),
+	}
+}
+
+func createPubkeyResourceAzure() *models.PubkeyResource {
+	return &models.PubkeyResource{
+		Tag:      "tag1",
+		PubkeyID: 1,
+		Provider: models.ProviderTypeAzure,
+		Handle:   factories.GetSequenceName("handle"),
+	}
+}
+
+func setupResource(t *testing.T) (dao.PubkeyResourceDao, context.Context) {
+	setup()
+	ctx := identity.WithTenant(t, context.Background())
+	resourceDao, err := dao.GetPubkeyResourceDao(ctx)
+	if err != nil {
+		panic(err)
+	}
+	return resourceDao, ctx
+}
+
+func teardownPubkeyResource(_ *testing.T) {
+	teardown()
+}
+
+func TestCreateResource(t *testing.T) {
+	resourceDao, ctx := setupResource(t)
+	defer teardownPubkeyResource(t)
+	resource := createPubkeyResourceNoop()
+	err := resourceDao.Create(ctx, resource)
+	require.NoError(t, err)
+	resources, err := resourceDao.ListByPubkeyId(ctx, resource.PubkeyID)
+	require.NoError(t, err)
+
+	assert.Equal(t, 1, len(resources))
+	assert.Equal(t, resource, resources[0])
+}
+
+func TestGetResourceByProviderType(t *testing.T) {
+	resourceDao, ctx := setupResource(t)
+	defer teardownPubkeyResource(t)
+	resource := createPubkeyResourceNoop()
+	err := resourceDao.Create(ctx, resource)
+	require.NoError(t, err)
+	createdResource, err := resourceDao.GetResourceByProviderType(ctx, resource.PubkeyID, resource.Provider)
+	require.NoError(t, err)
+
+	assert.Equal(t, resource, createdResource)
+}
+
+func TestListByPubkeyIdResource(t *testing.T) {
+	resourceDao, ctx := setupResource(t)
+	defer teardownPubkeyResource(t)
+	pkId := int64(1)
+	resourcesBefore, err := resourceDao.ListByPubkeyId(ctx, pkId)
+	require.NoError(t, err)
+	noopResource := createPubkeyResourceNoop()
+	err = resourceDao.Create(ctx, noopResource)
+	require.NoError(t, err)
+	azureResource := createPubkeyResourceAzure()
+	err = resourceDao.Create(ctx, azureResource)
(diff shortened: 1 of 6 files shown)


cmd/dao_tests/resource_test.go
```diff
@@ -0,0 +1,165 @@
+//go:build integration
+// +build integration
```
lzap, line 2: Please rename the filename as `pubkey_resource_test.go`.

cmd/dao_tests/resource_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
+	"github.com/stretchr/testify/assert"
+)
+
+func createPkResource() *models.PubkeyResource {
```
lzap, line 17: All functions in this package are shared, we need to name them very verbosely to prevent conflicts.

```suggestion
func createPubkeyResourceNoop() *models.PubkeyResource {
```

cmd/dao_tests/resource_test.go
```diff
@@ -0,0 +1,165 @@
...
+	return &models.PubkeyResource{
+		Tag:      "tag1",
+		PubkeyID: 1,
+		Provider: models.ProviderTypeNoop,
+		Handle:   "handle1",
+	}
+}
+
+func createPkResource2() *models.PubkeyResource {
```
lzap, line 26:
```suggestion
func createPubkeyResourceAzure() *models.PubkeyResource {
```

cmd/dao_tests/resource_test.go
```diff
@@ -0,0 +1,165 @@
...
+	return &models.PubkeyResource{
+		Tag:      "tag1",
+		PubkeyID: 1,
+		Provider: models.ProviderTypeAzure,
+		Handle:   "handle1",
```
lzap, line 31: I think handles might be unique in our database:

```suggestion
		Handle:   factories.GetSequenceName("handle"),
```

The same for other fixtures.

cmd/dao_tests/resource_test.go
```diff
@@ -0,0 +1,165 @@
...
+	teardown()
+}
+
+func TestCreateResource(t *testing.T) {
+	resourceDao, ctx := setupResource(t)
+	defer teardownPubkeyResource(t)
+	resource := createPkResource()
+	err := resourceDao.Create(ctx, resource)
+	if err != nil {
+		panic(err)
```
lzap, line 55: It is okay to use panic for GetXXXDao function because that is typically a configuration error. But for all other function calls, please fail the test properly so it will not stop the suite. You can take advantage that errors are also strings I believe, there is no reason to use formatted message and spend time writing "test failed blah blah":

```
		t.Fatal(err)
```

What is even better and I would highly encourage you to refactor not only this DAO test file, but all of them, is to use require package from the testify library we use:

```suggestion
		require.NoError(t, err)
```

This will immediately fail the test if there was an error effectively shortening the test code from three lines to one.

cmd/dao_tests/resource_test.go
```diff
@@ -0,0 +1,165 @@
...
+	resources, err := resourceDao.ListByPubkeyId(ctx, resource.PubkeyID)
+	if err != nil {
+		panic(err)
+	}
+
+	assert.Equal(t, 1, len(resources))
+	assert.Equal(t, resource.ID, resources[0].ID)
+	assert.Equal(t, resource.Provider, resources[0].Provider)
+	assert.Equal(t, resource.Handle, resources[0].Handle)
+	assert.Equal(t, resource.Tag, resources[0].Tag)
```
lzap, line 66: When you comparing the whole struct, then you can take advantage of Go equality contract: two structs are equal if all fields are equal:

```suggestion
	assert.Equal(t, resource, resources[0])
```

cmd/dao_tests/resource_test.go
```diff
@@ -0,0 +1,165 @@
...
+	err = resourceDao.Create(ctx, createPkResource2())
+	if err != nil {
+		panic(err)
+	}
+	resourcesAfter, err := resourceDao.ListByPubkeyId(ctx, pkId)
+	if err != nil {
+		panic(err)
+	}
+
+	assert.Equal(t, len(resourcesBefore)+2, len(resourcesAfter))
```
lzap, line 109: I would suggest to also assert that the returned list actually contains both resources. You can take advantage of `require.Contains` which works with slices I think.

lzap (changes requested):
Nice, I have some improvement suggestions. I would appreciate if you could implement them also on existing DAO test files too if you don't have objections.

cmd/dao_tests/pubkey_resource_test.go
```diff
@@ -0,0 +1,137 @@
+//go:build integration
+// +build integration
+
+package main
+
+import (
+	"context"
+	"github.com/RHEnVision/provisioning-backend/internal/testing/factories"
```
ezr-ondrej, line 8: I believe go imports wants this to be part of the other github imports.

cmd/dao_tests/pubkey_resource_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/dao"
+	_ "github.com/RHEnVision/provisioning-backend/internal/dao/sqlx"
```
ezr-ondrej, line 12: Should we move this one into `main_test`? Just a thought. Not important.

ezr-ondrej (commented):
Just failing import checks and I believe we are good to merge? :)

cmd/dao_tests/pubkey_resource_test.go
lzap: Yeah all side-effect imports should be in the `main.go` so you don't have to think about it.

cmd/dao_tests/pubkey_resource_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func setupResource(t *testing.T) (dao.PubkeyResourceDao, context.Context) {
+	setup()
+	ctx := identity.WithTenant(t, context.Background())
+	resourceDao, err := dao.GetPubkeyResourceDao(ctx)
+	if err != nil {
+		panic(err)
+	}
```
lzap, line 43: Just for the record: I just made a proposal on our Slack to actually move away from returning an error for all dao.GetXXX functions and simply panicking right there. This will make our life much easier.

lzap (approved):
Yeah looks great, I have no comments!

cmd/dao_tests/pubkey_resource_test.go
avitova: Great. I will have just removed this everywhere and moved it to main.go.

avitova:
Imports should be sorted properly now.

lzap (changes requested):
Oh one more thing to simplify.

I believe you could do this change also on other tests in the future (not in this PR please). This dramatically improve readability.

avitova:
Great point! Now this should be fixed and ready.

lzap (approved):
Thanks! Merging.

PR #164, opened 2022-08-17
Update docs for ephemeral deploy

ezr-ondrej:
With us being in the app-interface, we no longer need local config for ephemeral deployments.


lzap (approved):
Haven't tried this yet tho.

PR #165, opened 2022-08-17
Add requestbody to POST /pubkeys

akhil-jha:
Please let me know if I have to add anything anywhere else too!


adiabramovitch:
@akhil-jha if you are adding an endpoint you should follow the instructions here: 
https://github.com/RHEnVision/provisioning-backend/blob/main/docs/openapi.md

You are missing the change in path.yml, after that run ,make generate-spec and those files will be generated automatically

ezr-ondrej (approved):
Thanks ! :+1: I believe we should distinguish between request and response, as `id` is not allowed in request, but that's a story for another day ^_^

ezr-ondrej:
Force merging, given the contribution is from QE.... weird it still requires an ack :thinking:

akhil-jha:
@lzap could you rebase #135 ? :)

PR #166, opened 2022-08-17
Add image builder as clowder dependency

ezr-ondrej:
We now have IB in ephemeral, so adding the `image-builder` as optional dependency, not to break stage (where image-builder is still not deployed with clowder).

This will find IB config through clowder if there is one, but will not force it's presence.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -163,6 +163,8 @@ objects:
       inMemoryDb: true
       dependencies:
         - sources-api
+      optionalDependencies:
+        - image-builder
 
 # possible application ENV variables are in config/api.env.example
 parameters:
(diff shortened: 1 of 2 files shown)


lzap:
We have a new enforced commit message checker, can you rebase on top of `main` and reword the commit message to:

`type: message`

Thanks!

akhil-jha (commented):
Are you able you deploy it? 
I added `--optional-deps-method all` but no luck :(

lzap (approved):

ezr-ondrej:
It is not deploying automatically for me, I still need to mark it for deployment manually, but when deployed, it passes the service location to the clowder config.

ezr-ondrej:
Gonna merge this one so once IB can build images in ephemeral, we can actually deploy those :)

PR #168, opened 2022-08-18
Add version information to header and logs

lzap:
I just experienced what I believe was a testing of an incorrect build artifact. It is worth adding version information (sha commit, build date and time) into HTTP header as well as logs. It should be safe to show this for every request, but I will confirm on Slack.

This patch adds exactly that, all log lines contain "version" string which holds the commit sha. All responses contain new X-Rh-Version header which contains the same plug date and time (in UTC) of comiplation.

When compiling this via IDE (e.g. not via Makefile) no version information will be provided, thus the header will not be present. The same for log field.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,15 @@
 TEST_TAGS?=test
 CONTAINER_BUILD_OPTS ?= --build-arg=quay_expiration=2d
 CONTAINER_IMAGE ?= provisioning-backend
+PACKAGE_BASE = github.com/RHEnVision/provisioning-backend/internal
+LDFLAGS = "-X $(PACKAGE_BASE)/version.BuildCommit=$(shell git rev-parse --short HEAD) -X $(PACKAGE_BASE)/version.BuildTime=$(shell date +'%Y-%m-%d_%T')"
 
 .PHONY: build
 build: build-pbapi build-pbmigrate
 
 .PHONY: build-pbapi
 build-pbapi:
-	CGO_ENABLED=0 go build -o pbapi ./cmd/pbapi
+	CGO_ENABLED=0 go build -ldflags $(LDFLAGS) -o pbapi ./cmd/pbapi
 
 .PHONY: build-pbmigrate
 build-pbmigrate:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,13 +1,15 @@
...
+PACKAGE_BASE = github.com/RHEnVision/provisioning-backend/internal
+LDFLAGS = "-X $(PACKAGE_BASE)/version.BuildCommit=$(shell git rev-parse --short HEAD) -X $(PACKAGE_BASE)/version.BuildTime=$(shell date +'%Y-%m-%d_%T')"
 
 .PHONY: build
 build: build-pbapi build-pbmigrate
 
 .PHONY: build-pbapi
 build-pbapi:
-	CGO_ENABLED=0 go build -o pbapi ./cmd/pbapi
+	CGO_ENABLED=0 go build -ldflags $(LDFLAGS) -o pbapi ./cmd/pbapi
```
lzap, line 13: We do not have a target for pbworker, I will add it.

ezr-ondrej: should we add it in a separate PR, I believe we should also put that in the container, right?

lzap: Well, I have no intentions working on the container right now. A target makes sense even if we don't have a container yet.

lzap: Tho isn't this just about putting the binary into the container? That is like literally one line.

lzap: Hmm there's more than that, the problem is entrypoint. I have no idea how to solve at the moment.

I am adding only Makefile target, as a separate commit in this PR if you don't mind.

lzap:
Rebased, adding pbworker into the container is definitely material for a different PR. I think we should remove the entrypoint from the container specification and add it into clowdapp.yaml (?) or somewhere else, depending on which component we want to start.

ezr-ondrej:
> Rebased, adding pbworker into the container is definitely material for a different PR. I think we should remove the entrypoint from the container specification and add it into clowdapp.yaml (?) or somewhere else, depending on which component we want to start.

The entrypoint is just a default, that can always be overriden, so I guess it's fine, as we can override it for the worker

lzap:
> The entrypoint is just a default, that can always be overriden, so I guess it's fine, as we can override it for the worker

Yeah, maybe it would be better to remove it from Dockerfile and add it explicitly to ___________ (no idea where)? :)

ezr-ondrej (approved):
Security doesn't seem to have a big issue with this, so lets do it :+1:

PR #169, opened 2022-08-18
Filter instance types suited for RHEL

adiabramovitch:
- [ ] UNIT TEST


internal/clients/client_models.go
```diff
@@ -16,3 +16,14 @@ type Source struct {
...
+type InstanceType struct {
+	// The name of the instance type
+	Name string `json:"name,omitempty"`
+
+	// The number of cores for the instance type
+	CPUcores int32 `json:"cpu_cores,omitempty"`
+
+	// The size of the memory, in MiB.
+	Memory int64 `json:"memory,omitempty"`
+}
```
lzap, line 29: I am bit confused, I think this just needs a rebase...

internal/supported_types/supported_types_interface.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+)
+
+type SupportedInstanceTypes struct {
+	Aws struct {
+		Ec2 []string `yaml:"ec2"`
+	} `yaml:"aws"`
+}
+
+var supported SupportedInstanceTypes
```
lzap, line 15: Why these internal structures are exported? I would totally not export those, the only way of getting any results would be IMHO just calling the FilterInstanceTypes function.

internal/supported_types/supported_instance_types.yml
```diff
@@ -0,0 +1,447 @@
...
+      - r6id.large
+      - r6id.xlarge
+      - r6id.2xlarge
+      - r6id.4xlarge
+      - r6id.8xlarge
+      - r6id.12xlarge
+      - r6id.16xlarge
+      - r6id.24xlarge
+      - r6id.32xlarge
+      - r6id.metal
```
lzap, line 447: Missing newline.

internal/clients/client_models.go
lzap, line 29: Does this need to be rebased on top of the other PR?

internal/supported_types/aws_supported_types.go
```diff
@@ -0,0 +1,18 @@
+package supported_types
```
lzap, line 1: If anyone has a better name, please submit. Idiomatic Go is to have short packages (e.g. fmt instead of format). Maybe "instance" or "inst"?

lzap (commented):
Looks good!

internal/supported_types/embed.go
```diff
@@ -0,0 +1,27 @@
...
+		panic(err)
+	}
+
+	if err := yaml.Unmarshal(it, &supported); err != nil {
+		panic(err)
+	}
+
+	for _, i := range supported.Aws.Ec2 {
+		awsSupported[i] = nil
+	}
```
ezr-ondrej, line 26: ugh? ^_^

adiabramovitch: the map I was referring to :D
we can change it to a slice with binary search

internal/supported_types/supported_types_interface.go
```diff
@@ -0,0 +1,20 @@
...
+		Ec2 []string `yaml:"ec2"`
+	} `yaml:"aws"`
+}
+
+var supported supportedInstanceTypes
+
+type SupportedTypes interface {
+	// FilterUnsupportedTypes remove instances which are not supported
+	FilterInstanceTypes(ctx context.Context, is *[]clients.InstanceType) (*[]clients.InstanceType, error)
+}
```
ezr-ondrej, line 20: this interface is unused :)

internal/supported_types/aws_supported_types.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"context"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+)
+
+func FilterInstanceTypes(_ context.Context, it *[]clients.InstanceType) (*[]clients.InstanceType, error) {
```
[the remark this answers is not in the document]
adiabramovitch: So why do we need the interface?

ezr-ondrej (commented):
I believe it is quite nice, but I've commented few nits.
But I like the main idea of having separate package for this.
Tho is it good to have within root (`internal`) packages @lzap?

internal/supported_types/embed.go
ezr-ondrej: ah, right, got it now... `i` is very missleading, I thought it is the index and thus got confused.

internal/supported_types/aws_supported_types.go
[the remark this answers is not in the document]
ezr-ondrej: Oh! I think I know what Lukas wanted... with the interface we would be able to generalize better, but I do not think it is necessary ATM?

lzap:
> Tho is it good to have within root (`internal`) packages @lzap?

Let's find a better place than. I struggle to find one to be honest, I think this could live in the `internal/clients/itypes` or something perhaps?

internal/supported_types/supported_types_interface.go
lzap: This is my fault, I talked so much about interfaces that it left an impression we need one here. Not needed actually, we have a package with one exported function actually :)

We could if we would have three different implementation that would be different, but in this case it is just more simple to have a single YAML with three maps there.

internal/supported_types/supported_types_interface.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"context"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+)
+
+type supportedInstanceTypes struct {
+	Aws struct {
+		Ec2 []string `yaml:"ec2"`
```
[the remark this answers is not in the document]
lzap: I thought that having a single YAML for all clouds would be the easiest approach.

internal/supported_types/embed.go
lzap: Lists are more human-friendly then maps in YAML, the alternative was to sort out the slices and use bisearch, which Adi obviously did not like. I thought it would be more coding fun but I am not doing this feature, sadly :)

internal/supported_types/aws_supported_types.go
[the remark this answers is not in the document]
lzap: Well, I thought that the type will be shared, I mean, Azure has the very same parameters: name, vcpus, cores, memory... The same for GCE. Then it could be actually generic. Maybe I am wrong, I don't mind "AWS" in the function name it can be removed once we add Azure...

..which is like next Monday ;-)

internal/supported_types/supported_types_interface.go
adiabramovitch: I think that the interface is needed only if we want to have one - ListInstanceType function and to have a dispatch pattern to revoke the function for a specific cloud provider.

internal/payloads/instance_types_payload.go
```diff
@@ -3,12 +3,12 @@ package payloads
...
 
-	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/go-chi/render"
 )
 
 type InstanceTypeResponse struct {
-	Identifier string `json:"id"`
+	clients.InstanceType
 }
```
lzap, line 12: Nice, we are back to embedding :)

internal/payloads/instance_types_payload.go
```diff
@@ -19,13 +19,10 @@ func (s *InstanceTypeResponse) Render(_ http.ResponseWriter, _ *http.Request) er
...
 
-func NewListInstanceTypeResponse(sl *[]types.InstanceTypeInfo) []render.Renderer {
-	sList := *sl
-	list := make([]render.Renderer, 0, len(sList))
-	for i := range sList {
-		list = append(list, &InstanceTypeResponse{
-			Identifier: string(sList[i].InstanceType),
-		})
+func NewListInstanceTypeResponse(sl *[]clients.InstanceType) []render.Renderer {
+	list := make([]render.Renderer, 0)
```
lzap, line 23: This will be a tad faster:

```suggestion
	list := make([]render.Renderer, 0, len(sl))
```

internal/services/instance_types_service.go
```diff
@@ -61,8 +63,33 @@ func ListInstanceTypes(w http.ResponseWriter, r *http.Request) {
...
+		list = append(list, clients.InstanceType{
+			Name:   string(types[i].InstanceType),
+			VCPUs:  types[i].VCpuInfo.DefaultVCpus,
+			Cores:  types[i].VCpuInfo.DefaultCores,
+			Memory: *types[i].MemoryInfo.SizeInMiB,
+		})
+	}
+	return list
+
+}
```
lzap, line 95: Extra empty line.

I think that we have settled on the type and that it should live in the client package, I do believe this conversion function should be at the same place too. This is a service package, but we might need the conversion maybe in other places too.

It should be also exported, one idea for a name would be (trying to mimic Idiomatic Go not sure about it): `NewInstanceType`.

internal/supported_types/aws_supported_types.go
```diff
@@ -0,0 +1,18 @@
...
+	var filteredInstanceTypes []clients.InstanceType
+	for _, i := range *it {
+		_, ok := awsSupported[i.Name]
+		if ok {
+			filteredInstanceTypes = append(filteredInstanceTypes, i)
+		}
+	}
+	return &filteredInstanceTypes, nil
+}
```
lzap, line 18: Here is a crazy idea: what if this function was declared on the instance type itself? Then you could simply write:

```go
types := client.NewInstanceTypes(resultFromAws)
types = types.FilterSupportedTypes()
```

With this syntax, I can even see that the name could be even improved to something like: `types.OnlySupported()`. Also see that context is not passed in, for trivial functions which do not need logging, passing in cotext is perhaps not needed.

internal/supported_types/embed.go
lzap: I am fine with this, no need to do the search. By the way binary search is not the only option, there is plethora of other more complicated ones :)

internal/supported_types/supported_types_interface.go
lzap: Yeah, definitely get rid of the interface for now, also move this into a single file.

lzap (commented):
Looks good. I dropped this idea of actually declaring the new function on the type! :)

lzap:
Tech discussion:

* Let's add Architecture field in the instance type struct
* Let's add Supported boolean field in the instance type struct
* Always return all instance types, filtering can be done more flexibly in the UI

internal/supported_types/aws_supported_types.go
```diff
@@ -0,0 +1,20 @@
...
+		_, ok := awsSupported[i.Name]
+		if ok {
+			// If the Instance type is present in the yml it is supported
+			i.Supported = true
+		}
+	}
+	// Returns all the Instance types received from aws
+	// if the instance type is supported the flag supported will be true otherwise false
+	return it, nil
+}
```
lzap, line 20: This method should no longer return anything, you correctly use pointer to slice which means you can modify the slice as needed. Not even error is needed I think. Since it no longer filters anything, it could be named differently.

Please extract the `awsSupported[i.Name]` line and surrounding statements into a separate exported function something like `IsSupported`. We will need this in reservation endpoint check. This is ok to add in this PR as well, but we can do it separately if you prefer.

internal/supported_types/embed.go
```diff
@@ -0,0 +1,35 @@
...
+
+var awsSupported = make(map[string]interface{})
+
+type supportedInstanceTypes struct {
+	Aws struct {
+		Ec2 []string `yaml:"ec2"`
+	} `yaml:"aws"`
+}
+
+var supported supportedInstanceTypes
```
lzap, line 17: Can you move this into the init function? It does not need to be at package level. This will save some memory since at the end of the init function it can be cleaned up by gc.

internal/services/instance_types_service.go
```diff
@@ -61,8 +63,38 @@ func ListInstanceTypes(w http.ResponseWriter, r *http.Request) {
...
+			Name:                   string(types[i].InstanceType),
+			VCPUs:                  types[i].VCpuInfo.DefaultVCpus,
+			Cores:                  types[i].VCpuInfo.DefaultCores,
+			Memory:                 *types[i].MemoryInfo.SizeInMiB,
+			SupportedArchitectures: arch,
+			Supported:              false,
+		})
+	}
+	return list
+}
```
lzap, line 100: I suggest to move this function next to `FilterAWSInstanceTypes` and name it in Go way: `NewInstanceTypes` since it is really a "constructor" type of function.

internal/clients/client_models.go
```diff
@@ -30,4 +30,9 @@ type InstanceType struct {
 
 	// The size of the memory, in MiB.
 	Memory int64 `json:"memory,omitempty"`
+
+	// Does the instance type supports RHEL
+	Supported bool `json:"supported,omitempty"`
+
+	SupportedArchitectures []string `json:"supported_architectures,omitempty"`
```
lzap, line 37: As we discussed on Slack, let's break multiple architectures into multiple InstanceType values each having it's own architecture.

I suggest that you create constant for each architecture in this package and validate them (e.g. in switch-case with a default branch with error), because different clouds might have different architecture names (e.g. x86-64 vs x86_64). We should only use one, pick something and insist on it (convert), return errors on unknown architectures (we need to manually add support for new ones).

lzap:
Looks good, I still don't like pretty long package name `supported_types`. Since `types` is also misleading and we already have this package in AWS client, how about moving the function to the `clients` package where the type lives? I mean, it is pretty appropriate place, we will likely use the filtering and validation mainly in the client code, I think this solves the package name problem.

adiabramovitch:
> Looks good, I still don't like pretty long package name `supported_types`. Since `types` is also misleading and we already have this package in AWS client, how about moving the function to the `clients` package where the type lives? I mean, it is pretty appropriate place, we will likely use the filtering and validation mainly in the client code, I think this solves the package name problem.

I dont have a problem moving this dir to clients package but we will still need to decide on a package name for the new directory I have created (supported_types)

ezr-ondrej:
> I dont have a problem moving this dir to clients package but we will still need to decide on a package name for the new directory I have created (supported_types)

Can't the embed live in the `clients/ec2` and `FilterXXX` in `clients` package? :thinking:

adiabramovitch:
> > I dont have a problem moving this dir to clients package but we will still need to decide on a package name for the new directory I have created (supported_types)
> 
> Can't the embed live in the `clients/ec2` and `FilterXXX` in `clients` package? thinking

I guess it can

adiabramovitch:
@lzap @ezr-ondrej , I think it is a bit messier like that, what do you think?

internal/clients/ec2/aws_supported_types.go
```diff
@@ -0,0 +1,45 @@
...
+			if err != nil {
+				return nil, payloads.ClientError(ctx, "Instance type", "", err, 500)
+			}
+			list = append(list, clients.InstanceType{
+				Name:         string(types[i].InstanceType),
+				VCPUs:        types[i].VCpuInfo.DefaultVCpus,
+				Cores:        types[i].VCpuInfo.DefaultCores,
+				Memory:       *types[i].MemoryInfo.SizeInMiB,
+				Architecture: string(arch),
+				Supported:    false,
```
ezr-ondrej, line 40: We don't need the Filter method, if we do:
```suggestion
				Supported:    IsSupported(types[i].name),
```

internal/clients/embed.go
```diff
@@ -0,0 +1,54 @@
+package clients
```
ezr-ondrej, line 1: I'd move this one in `ec2` as well :thining:

internal/clients/embed.go
```diff
@@ -0,0 +1,54 @@
...
+var AwsSupported = make(map[string]interface{})
+var ErrArchitectureNotSupported = errors.New("architecture is not supported")
+
+const x86_64 = "x86_64"
+const arm64 = "arm"
+const i386 = "i386"
+
+type supportedInstanceTypes struct {
+	Aws struct {
+		Ec2 []string `yaml:"ec2"`
```
ezr-ondrej, line 21: I still don't get this nesting btw :)

lzap:
Looks like from the UX research that users would like to also see storage info. I think the most important aspect of storage is how much instance storage it has. Some types are are EBS-only so we should set this value to 0 and in the UI show something like "EBS-only" too. Since we want the type to be ideally common for all providers, I suggest to use a different term here: SAN (Storage Area Network) which is exactly what EBS is as other vendors have different meanings of the same while SAN is the standard name for this.

I do not want to make this PR complicated than it already is, but if you want, you can implement this it should be as easy as adding a new field (integer perhaps) and setting it to zero if an instance is EBS-only. I think we should definitely avoid adding and kind of SSD/HDD info as vendors have a TON of various configurations and also SAN storage have many categories depending on the connection (fibre channel, iSCSI etc). Let's not spend any time on this for now.

If you feel like it's too much, let me know and let's create an issue for that.

lzap:
> Can't the embed live in the clients/ec2 and FilterXXX in clients package? 

Why this would be under ec2/ ? I was hoping that we push towards having a single type for all vendors. I really do not see any reason not to share it across EC2/Azure/GCE at this point, all the vendors have the very same fields. I think we should keep this simply in clients/.

internal/clients/embed.go
lzap: I commented already below, why?

I was thinking we would have a common type with a common function(s) for all types. And a single YAML with all the lists. Why you think this should go into ec2 again?

internal/clients/embed.go
lzap: Yeah this is weird, I suggest to always have some root map, and then just use `ec2`. Something like:


```yaml
allowed_types:
  ec2:
    - x1.super-tiny
    - x9.huge-xxxxl
  azure:
    - z13.windows.laptop
```

internal/clients/client_models.go
```diff
@@ -30,4 +30,9 @@ type InstanceType struct {
 
 	// The size of the memory, in MiB.
 	Memory int64 `json:"memory,omitempty"`
+
+	// Does the instance type supports RHEL
+	Supported bool `json:"supported,omitempty"`
+
+	Architecture string `json:"architecture,omitempty"`
 }
```
lzap, line 38: This could be a type alias: `type Architecture string` with constant strings for both two architectures: x86_64 and ARM. But I think this could create a problem when marhsalling to JSON and you would need to implement `MarshalJSON() ([]byte, error)` function that converts it to string (which is pretty easy). Added benefit would be that no one could ever assign anything else than the two constants here. But feel free to keep it as a string.

As I commented above, I think one last field which is worth adding is `InstanceStorageMiB integer` (name and type would depend on what AWS library uses). It should be directly readable and for EBS-only types we would set this to zero. Not insisting tho.

Please always add code comments for important types like this one. Here, I would definitely drop a comment about what is the expected format of architecture.

internal/clients/client_models.go
```diff
@@ -30,4 +30,9 @@ type InstanceType struct {
 
 	// The size of the memory, in MiB.
 	Memory int64 `json:"memory,omitempty"`
```
lzap, line 32: Consider including unit of measurement in the name, I do this all the time (e.g. Sec for seconds). This is inspired by systemd and I think it is a reasonable thing because so many configurations and inputs don't do this and don't even document things properly.

```suggestion
	MemoryMiB int64 `json:"memory,omitempty"`
```

internal/clients/embed.go
lzap: Oh I get it, keeping only ec2 related functions there and common code in clients. Yeah, that makes sense!

internal/clients/ec2/aws_supported_types.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func FilterAWSInstanceTypes(_ context.Context, it *[]clients.InstanceType) {
+	for _, i := range *it {
+		if IsSupported(i) {
+			// If the Instance type is present in the yml it is supported
+			i.Supported = true
+		}
+	}
+}
```
lzap, line 18: Kind of don't like the function name, it's not filtering but it's ok.

But why this is exported? I would expect this to be called from `NewInstanceType` and nowhere else.

internal/clients/ec2/aws_supported_types.go
```diff
@@ -0,0 +1,45 @@
...
+
+func IsSupported(i clients.InstanceType) bool {
+	_, ok := clients.AwsSupported[i.Name]
+	return ok
+}
```
lzap, line 23: Please move this function to the clients package. Also unexport `AwsSupported` variable, it's definitely something that callers should not see.

internal/clients/embed.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"embed"
+	"errors"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+var AwsSupported = make(map[string]interface{})
```
lzap, line 12: This should not be exported, move `IsSupported` function into this package to create a contract.

internal/clients/embed.go
```diff
@@ -0,0 +1,54 @@
...
+var ErrArchitectureNotSupported = errors.New("architecture is not supported")
+
+const x86_64 = "x86_64"
+const arm64 = "arm"
+const i386 = "i386"
```
lzap, line 17: My God they still provide i386? Okay.

internal/clients/embed.go
```diff
@@ -0,0 +1,54 @@
...
+
+type supportedInstanceTypes struct {
+	Aws struct {
```
lzap, line 20: The root name should be something generic:

```suggestion
	SupportedTypes struct {
```

internal/clients/embed.go
```diff
@@ -0,0 +1,54 @@
...
+	switch {
+	case strings.Contains(arch, "i386"):
+		return i386, nil
+	case strings.Contains(arch, "86"):
+		return x86_64, nil
+	case strings.Contains(arch, "arm"):
+		return arm64, nil
+	}
+	return "", ErrArchitectureNotSupported
+}
```
lzap, line 54: I believe this mapping is ec2 only, so move this into ec2/ package.

internal/clients/embed.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+var AwsSupported = make(map[string]interface{})
+var ErrArchitectureNotSupported = errors.New("architecture is not supported")
+
+const x86_64 = "x86_64"
+const arm64 = "arm"
```
lzap, line 16: I suggest to use "arm64" which is more appropriate than just "arm" although AWS does not use that.

adiabramovitch:
> Looks like from the UX research that users would like to also see storage info. I think the most important aspect of storage is how much instance storage it has. Some types are are EBS-only so we should set this value to 0 and in the UI show something like "EBS-only" too. Since we want the type to be ideally common for all providers, I suggest to use a different term here: SAN (Storage Area Network) which is exactly what EBS is as other vendors have different meanings of the same while SAN is the standard name for this.
> 
> I do not want to make this PR complicated than it already is, but if you want, you can implement this it should be as easy as adding a new field (integer perhaps) and setting it to zero if an instance is EBS-only. I think we should definitely avoid adding and kind of SSD/HDD info as vendors have a TON of various configurations and also SAN storage have many categories depending on the connection (fibre channel, iSCSI etc). Let's not spend any time on this for now.
> 
> If you feel like it's too much, let me know and let's create an issue for that.

I think

lzap (commented):
Ok some shuffling around is still needed but it is almost done.

internal/clients/embed.go
lzap: Are you saying to use `aws` instead of `ec2`? Sure, if you prefer this term just use it!

internal/clients/ec2/aws_supported_types.go
adiabramovitch: I removed that function as Ondrej suggested https://github.com/RHEnVision/provisioning-backend/pull/169#discussion_r954642225

internal/clients/embed.go
adiabramovitch: I think it is double nesting, the whole struct is called supportedInstanceTypes / supportedTypes

internal/clients/embed.go
adiabramovitch: Why? Cant it be suitable for Azure as well?

internal/clients/client_models.go
adiabramovitch: I will add the storage in a different PR

internal/clients/embed.go
lzap: So it's always a good practice to have some kind of root element in XML/JSON/YAML, because later when you need to add just one more thing, you don't need to break it apart. So you can do:

```yaml
original_data:
  - xxx
  - xxx
some_additional_stuff_5_years_later:
  - xxx
  - xxx
  - xxx
```

But I am heavily biased by my years of XML programming where it was a problem, maybe YAML does not suffer from this. Then just remove the extra level I guess.

internal/clients/embed.go
lzap: True, we can convert different labels, yeah. Let's keep it, sure.

lzap (approved):
I am good, OpenAPI needs a refresh, unit test is missing tho.

ezr-ondrej (approved):
I'd be fine if you create a follow-up ticket for next iteration to cover it with tests.
WDYT @adiabramovitch ?

adiabramovitch:
> I'd be fine if you create a follow-up ticket for next iteration to cover it with tests. WDYT @adiabramovitch ?

Sure sounds good

PR #170, opened 2022-08-19
Implement thread-safe TTL cache instead LRU

lzap:
The cache package was not designed for the future, so I have rewritten it from scratch. It now uses proper interface with two implementations: memory and noop. The memory implementation was previously using LRU cache, this patch drops the dependency in favour of TTL cache with 5 minutes default cleanup interval and configurable TTL value per entry. When we talked about caching I mentioned LRU by accident, what I mean was really TTL. The implementation is also now thread-safe, the previous library and the present one are not. Long term, we would use something like Redis/Memcache anyway.

The no-operation implementation is meant for testing where we explicitly do not want to use any kind of caching which can cause false positives/negatives.

Application ID now does not even use cache because it is essentially a global variable and it can be stored as a string pointer. Proper caching is implemented for instance types, however, when working on the patch I could not figure out how to list instance types per region. This needs to be researched, although it is not M0 I would like to code and the cache to be ready.

Cache misses are reported into logs via trace level and cache name is also reported.


internal/clients/ec2/ec2_client.go
```diff
@@ -102,7 +102,7 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client,
 	return newClient, nil
 }
 
-func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
+func (c *Client) ListInstanceTypes() ([]string, error) {
```
ezr-ondrej, line 105: I don't think this is sufficient, we will need more info about the instances (memory, cpus...)
We do not do this ATM, but we should do that soon, for the list to be more user readable, so I'd like us to keep returning the full info, or at least do not write it in a way we can't return more info... so maybe custom type?

lzap: Good point, we are talking with @adiabramovitch actually about the "supported instance types" problem and it too turns out that it would be a great benefit if we had a common type across clouds so the filtering component could be generic. Let's look into that, because I see a big value there in the future when multicloud is a thing. A customer might want need to launch an instance across clouds that is reasonably similar spec. A common type could actually help with this, we should be able to make a function that would take instance type of one cloud and find the closest matches in regard to specs for all other clouds.

This should be possible at least on the technical level (architecture, memory, CPUs, instance storage speed and size, network speed). User is then presented with options and can do the final selection according a different criteria (price, support level etc).

lzap: So we had a chat with @adiabramovitch and we agreed on creating a new type:

type InstanceType struct {
 Name string,
 CPUs integer,
 MemoryBytes integer, // is perhaps better than MemoryGiB float because dealing with floats is less human friendly
}

lzap:
> I could not figure out how to list instance types per region

So it looks like you always have to connect to the region you want to work in. Therefore our AWS client must have capability to select different regions per connection. This is not there yet, I will just hardcode `us-east-1` there and won't use the variable for the actual fetch.

lzap:
So it looks like we will be implementing preloading/precaching of instance types, so I will resurrect this PR once we have a use case. There is nothing we can currently cache I think.

lzap:
This is not ready for review, I actually do not see any endpoint that is worth caching at this moment, so I won't work on this until we have something reasonable.

lzap:
I opened a new PR with caching.

PR #172, opened 2022-08-22
Add Instance Type model

adiabramovitch:


internal/clients/client_models.go
```diff
@@ -16,3 +16,15 @@ type Source struct {
...
 	Uid *string `json:"uid,omitempty"`
 }
+
+// An instance type defines a model for an instance type that corresponds to one in a cloud provider.
+type InstanceType struct {
+	// The name of the instance type
+	Name *string `json:"name,omitempty"`
+
+	// The number of cores for an instance type
+	CPU *int32 `json:"cpu,omitempty"`
```
ezr-ondrej, line 26: so should this be cores? :thinking:

ezr-ondrej (commented):
I think it's very useful to have this defined ourselves, thanks for doing this! :)
Just one thought, whether we have CPUs or cores, but I'm ok with both if there is a reason to have CPUs :)

internal/clients/client_models.go
lzap: Yeah it's a mess on AWS, I suggest `CPUCores` which can map to AWS `DefaultCores`. They have the following semantics:

* DefaultCores - CPU cores as shown by the OS
* DefaultThreadsPerCore - threads as shown by the OS, typically 2 and can be configured to 1 or 2
* vCPUs = DefaultCores * DefaultThreadsPerCore - total number of virtual threads available

So speaking from the technical point, what only matters is vCPUs that is how you optimize your programs (N+1 to saturate CPUs), however, there are also licensing questions (e.g. you only have a subscription/license for software that runs on 4 CPUs max), that is the DefaultCores. So vCPU makes sense too :)

lzap: Here is an idea - let's have both: Cores and Threads (vCPUs on AWS). Let me check how it looks on Azure.

lzap: Looks like Azure use the same terminology for "threads" on hypervisor: vCPUs (virtual CPU = can be run on hyper thread). And for instances that have dedicated cores they use Core.

So I suggest to stick what is known in the domain:

```go
// Virtual CPU (maps to hypervisor hyper-thread)
VCPUs *int32

// Core (physical or virtual core)
Cores *int32
```

I suggest to actually use pointers for these values, because some instances will not have vCPUs or Cores (some will have both tho). But for Name and Memory I suggest not to use pointers - no reason to do so.

lzap: We can thank Intel's Hyper-Threading (TM) for this, a technology that is full of security problems and many cloud providers have special "security" types that don't feature this technology at all. AMD has similar SMT technology in their latest series (Zen). But for example ARM a1 instances on AWS they do not use this at all: https://aws.amazon.com/ec2/instance-types/a1/

I wonder how InstanceType looks like for a1 types, I would expect both vCPUs and Cores are the same (1-1, 8-8, 16-16).

ezr-ondrej (approved):
Thanks @adiabramovitch ! nice addition! :+1:

PR #173, opened 2022-08-22
Remove internal only docs

ezr-ondrej:
These docs are not useful to anyone upstream as there is no access to those deployments.


lzap (approved):
Sure.

PR #174, opened 2022-08-22
Use Account from headers in NoopService

lzap:
SSIA


ezr-ondrej (approved):
my fault we've forgotten it.
Thanks @lzap for fixing it!

PR #175, opened 2022-08-22
Regenerate clients and delete unused test

lzap:
SSIA

There was an unused client test, perhaps generated by the tool.


ezr-ondrej:
I'd rather see the tests implemented, but no problem removing them for now :)

ezr-ondrej (approved):
Thanks, although I do not agree with removing tests, empty ones are no use, so I'm ok here.

PR #176, opened 2022-08-22
Remove unused Update for pubkey resource

ezr-ondrej:
Update is not expected to be utilized on Pubkey resource, we shall not 
keep and maintain methods that are not currently or soon to be used.
This is easy to add once needed.


lzap (approved):
Absolutely, it served its purpose of being a "template" for other models I guess.

PR #177, opened 2022-08-22
Drop account endpoints

ezr-ondrej:
We might want to add this at some point for better observability, but we 
need to figure out admin/service access gating first, which is not yet 
on the table.


lzap (approved):
Thanks! More of deletions, please.

PR #179, opened 2022-08-22
Make log field truncating configurable

lzap:
Also puts the panic contents (the error string) into the message as it was not presented. Finally, fixes the truncating quoting - when it's quoted close the quote, otherwise do not close it.

**Checklist**


ezr-ondrej (approved):
Thanks @lzap ! Nice addition :+1:

PR #180, opened 2022-08-22
add identity header to ListApplicationTypes

amirfefer:
The request returns 401 from sources api, adding identity header fix it

### DIFF
--- a/internal/clients/sources/sources_client.go
+++ b/internal/clients/sources/sources_client.go
@@ -164,7 +164,7 @@ func (c *SourcesClient) GetProvisioningTypeId(ctx context.Context) (string, erro
 
 func (c *SourcesClient) loadAppId(ctx context.Context) (string, error) {
 	ctxval.Logger(ctx).Info().Msg("Fetching the Application Type ID of Provisioning for Sources")
-	resp, err := c.client.ListApplicationTypes(ctx, &ListApplicationTypesParams{})
+	resp, err := c.client.ListApplicationTypes(ctx, &ListApplicationTypesParams{}, headers.AddSourcesIdentityHeader)
 	if err != nil {
 		ctxval.Logger(ctx).Warn().Err(err).Msg("Failed to fetch ApplicationTypes from sources")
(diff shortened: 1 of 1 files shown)


internal/clients/sources/sources_client.go
```diff
@@ -164,7 +164,7 @@ func (c *SourcesClient) GetProvisioningTypeId(ctx context.Context) (string, erro
 
 func (c *SourcesClient) loadAppId(ctx context.Context) (string, error) {
 	ctxval.Logger(ctx).Info().Msg("Fetching the Application Type ID of Provisioning for Sources")
-	resp, err := c.client.ListApplicationTypes(ctx, &ListApplicationTypesParams{})
+	resp, err := c.client.ListApplicationTypes(ctx, &ListApplicationTypesParams{}, headers.AddSourcesIdentityHeader)
```
ezr-ondrej, line 167: Oh this is devel use only tho, right, because ListApplicationTypes is actually not gated by authentication IMHO, or is it? It should be publicly available endpoint

ezr-ondrej (approved):
Thanks @amirfefer :

internal/clients/sources/sources_client.go
amirfefer: According to sources docs, looks like this endpoint is gated by the authorization header/identity 
https://redocly.github.io/redoc/?url=https://github.com/RedHatInsights/sources-api-go/raw/main/public/openapi-3-v3.1.json#tag/application-types/operation/listApplicationTypes

ezr-ondrej: Interesting, that it works locally :) but what do I know. Lets add it and hope it won't become an issue later on :)

ezr-ondrej (approved):
Thanks @amirfefer ! :+1:

internal/clients/sources/sources_client.go
lzap: Maybe 3scale sets it to some default value that causes it to fail on stage. Weird.

PR #183, opened 2022-08-24
Set AWS credentials from kub secret

ezr-ondrej:
We have an account in the secrets now, lets set it in the ENV.

Fixes HMSPROV-203


mshriver:
The secret is in app-interface, not seeing anything here that goes against consoledot guidance for secrets, let's go for it.

mshriver (approved):

PR #184, opened 2022-08-24
Add reservation name and poweroff flags

lzap:
In order to safely and efficiently test on our AWS development accounts, we need a reliable way of keeping our bills low.

One measurement is to automatically launch instances with userdata setting that will immediately power off the instance after it was launched. This can be achieved via the new PowerOff flag. Set it to true for development purposes and your instances will be shut down which is significantly cheaper than running them when not terminated on time.

Second measurement is to introduce instance name. Although we will not use it in the UI at this stage, it is very useful to set instance names to something meaningful. Especially when paired with a new setting called instancePrefix and cloudwash script that is able to automatically terminate or stop instances starting with a configured prefix. We can now run cloudwash against our personal or shared AWS accounts having it terminating all instances prefixed "test-". You can configure the backend to always prepend a prefix from configuration, for example "test-lzap-" which will also improve visibility on shared accounts in the future.

Finally, this patch also adds support for image_id in the "ami-XXXX" format. When provided, image builder is not called in order to call UUID and the given AMI is used instead. This is great for testing because image builder images only lasts two weeks before they are deleted. Also, in the future when we start copying images to customer accounts, this would create unnecessary costs when working on non-image related patches. Here is how you can launch an Amazon Linux 2 AMI on us-east-1 that will be automatically shutdown after launch (this snippet is available as a `.http` file in the repo):

```json
{
  "name": "amz-linux-us-east-1",
  "source_id": "1",
  "image_id": "ami-05fa00d4c63e32376",
  "amount": 1,
  "instance_type": "t1.micro",
  "pubkey_id": 1,
  "poweroff": true
}
```

Database must be remigrated when testing this PR!

https://github.com/RedHatQE/cloudwash

### DIFF
--- a/internal/db/migrations/005_reservations.sql
+++ b/internal/db/migrations/005_reservations.sql
@@ -23,9 +23,11 @@ CREATE TABLE aws_reservation_details
   provider INTEGER NOT NULL DEFAULT provider_type_aws(),
   pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
   source_id TEXT NOT NULL,
+  name TEXT NOT NULL DEFAULT '',
   instance_type TEXT NOT NULL,
   amount INTEGER NOT NULL,
   image_id TEXT NOT NULL,
+  poweroff BOOLEAN NOT NULL DEFAULT false,
   aws_reservation_id TEXT,
 
   FOREIGN KEY (reservation_id, provider) REFERENCES reservations(id, provider) ON DELETE CASCADE,
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -45,6 +45,12 @@
           "instance_type": {
             "type": "string"
           },
+          "name": {
+            "type": "string"
+          },
+          "poweroff": {
+            "type": "boolean"
+          },
           "pubkey_id": {
             "format": "int64",
             "type": "integer"
@@ -386,7 +392,7 @@
     },
     "/reservations/aws": {
       "post": {
-        "description": "A reservation is a way to activate a job, keeps all data needed for a job to start. An AWS reservation is a reservation created for an AWS job.\n",
+        "description": "A reservation is a way to activate a job, keeps all data needed for a job to start. An AWS reservation is a reservation created for an AWS job. Image Builder UUID image is required, the service will also launch any AMI image prefixed with \"ami-\".\n",
         "operationId": "createAwsReservation",
         "requestBody": {
           "content": {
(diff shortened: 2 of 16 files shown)


internal/db/migrations/005_reservations.sql
```diff
@@ -23,9 +23,11 @@ CREATE TABLE aws_reservation_details
   provider INTEGER NOT NULL DEFAULT provider_type_aws(),
   pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
   source_id TEXT NOT NULL,
+  name TEXT NOT NULL DEFAULT '',
```
ezr-ondrej, line 26: I've got a bad news, that this won't be enough, we will want to have name per instance, as you've suggested and research shared with me yesterday confirmed.

lzap: Sure, let's utilize an array, postgres can easily do that and since it is not indexed column and will never be, no problem.

I am wondering if we should think about serializing all the reservation details into JSON, I find those columns harder and harder to maintain as it grows. We really only need the info to be able to return the information back to the user when they request reservation info (something to be displayed on the future pending reservations page). What you think @ezr-ondrej @adiabramovitch ?

Serializing into JSON is easy, postgres even have a native type for that and we could easily add more info. On the other hand, it brings one big negative - compatibility. A reservation after application upgrade might be out of date so unmarhsalling of old record could fail. On the other hand, this type of table is a good candidate for that because we will regularly delete old reservations so these kinds of bugs will naturally disappear as old reservations will get deleted after few days.

akhil-jha:
Cannot test anyway, HMSPROV-194 and 200 is blocking :disappointed:

lzap:
Well HMSPROV-194 does not actually block you from now on, this patch adds ability to skip image builder. See the description, just use `ami-XXXXX` for the image id instead (e.g. in the example above that is Amazon Linux 2 on US east 1).

HMSPROV-200 is a different story, but I would ask not to block this PR since any of the reservation features were not yet fully tested so there cannot be any regression, IIUC.

ezr-ondrej (approved):
Sure lets do multiple names in follow-up for the record I'm totaly for the JSON serialization of additional data, we do not need a column for everything, let's have columns just for data we actually want to search through.

ezr-ondrej:
Let's get it in so QE can try this on other PRs... but indeed this should unblock API testing of the whole flow (by skipping Image Builder)

PR #185, opened 2022-08-25
Add test for creating and filtering new instance types

adiabramovitch:
Depends on https://github.com/RHEnVision/provisioning-backend/pull/186


internal/clients/new_instance_types_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"context"
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/clients/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+
+	"github.com/aws/smithy-go/ptr"
+	"github.com/stretchr/testify/assert"
+)
```
lzap, line 13: Unit tests must not use real clients, only stubs are allowed.

I suggest you to import `test_init.go`:

```
	_ "github.com/RHEnVision/provisioning-backend/internal/testing/initialization"
```

That will do it.

lzap, line 13: Was going to recommend importing `_ "github.com/RHEnVision/provisioning-backend/internal/testing/initialization"` but then I realized you only test these functions, nice.

internal/clients/new_instance_types_test.go
```diff
@@ -0,0 +1,74 @@
+package clients_test
```
lzap, line 1: Can you rename `embed.go` to something reasonable and this file to `something_reasonable_test.go`?

The Idiomatic Go pattern is always keep unit tests named as their source files:

* blah.go
* blah_test.go

adiabramovitch: Yes! It bothered me as well 😀

lzap (approved):
Lol the test is here.

PR #186, opened 2022-08-25
Add x86_64 mac architecture

adiabramovitch:


lzap:
Nice but "x86_64_mac" is really not a hardware architecture. Why we need this? I was trying to find this in the SDK without any results.

It can be perhaps virtualization architecture, like configuration of a VM to run Mac. Then we probably should maybe rename this to "Platform" rather than Architecture.

lzap:
So you are right, it does exist and it is named really "x86_64_mac" and there is even "arm64_mac" now for M1 chips.

I am pretty sure that none of these instance types are supported by RHEL, so we would filter them out anyway. The only offering I see is "mac1.metal" which is not allowed in our list. So it should never happen that such an instance type is ever returned by the endpoint.

But I am not against having it as a constant, tho I really hate this format. Since we can afford rename our own architecture constants how about something nicer: `apple-x86_64` and `apple-arm64`. Here is where I found the info:

* https://aws.amazon.com/ec2/instance-types/mac/
* https://instances.vantage.sh/aws/ec2/mac2.metal

lzap (approved):
Don't want to be pesky, but would you mind writing a unit test for the mapping function? :) Maybe there is a type we will miss during review...

lzap:
Just found there is another PR with this commit AND a test :-) Merged that one. Closing this one, thanks.

PR #187, opened 2022-08-26
Rework environmental variables

lzap:
Handing of environment variables with viper library is strange. It was not possible to easily implement camelCase conversion to CAMEL_CASE due to various issues with how processing of variables is implemented.

This patch extract all viper code into a separate package, this allows to use StringReplacer interface to do replacing manually. There is now table of all known configuration values (which viper automatically converts to upper case and use dots). For every configuration value (in our config struct) there must be entry in the table in known.go file.

Names that don't have multiple words are automatically converted (e.g. APP.NAME to APP_NAME) but those which cannot be converted this way (as viper already capitalized the strings) a second column can be used to override the value.

Not only we can manually set some ENV variables to reasonable values, we can also shorten some keys (e.g. IMAGEBUILDER looks quite long), although, I haven't opted for such option for now.

But the best feature of this patch is that pbapi now creates a WARNING log after all configuration is parsed and provides a list of all environmental variables which are passed in and not recognized. This should help us to identify when we add new configuration option but forget to add it to the array in known.go.

The patch also touches clowdapp.yaml and updates few typos there, adds WORKER variables. My goal was not to add all variables, we should do this PR by PR.

During testing, I realized that "make" always recompiles even when no source files were changed. This was pretty slow, so I fixed this in Makefile.


Makefile
```diff
@@ -1,22 +1,22 @@
...
 
-.PHONY: build-pbworker
-build-pbworker:
+pbworker: $(SRC_GO) $(SRC_SQL) $(SRC_YAML)
 	CGO_ENABLED=0 go build -ldflags $(LDFLAGS) -o pbworker ./cmd/pbworker
 
-.PHONY: build-pbmigrate
-build-pbmigrate:
+pbmigrate: $(SRC_GO) $(SRC_SQL) $(SRC_YAML)
 	CGO_ENABLED=0 go build -o pbmigrate ./cmd/pbmigrate
```
lzap, line 21: This change is not relevant, however, it really helped me to speed up testing this as I was calling:

    make && TEST_ENVIRONMENT_VAR=XYZ ./pbapi

So let's keep it here? :-) Please.

configs/local_example.yaml
```diff
@@ -73,7 +73,7 @@ worker:
   # amount of goroutines for processing (CPU cores or more)
   concurrency: 50
   # heartbeat update rate (seconds)
-  hearbeatSec: 30
+  heartbeatSec: 30
   # maximum amount of heartbeat updates
```
lzap, line 77: The patch actually helped me to find a typo in our config!

deploy/clowdapp.yaml
```diff
@@ -109,8 +119,23 @@ parameters:
...
     name: CLOWDER_ENABLED
     value: "true"
-  - description: Determines if Cloud Watch logging is enabled
-    name: ENABLE_CLOUDWATCH_LOGGING
+  - description: AWS CloudWatch logging integration
+    name: CLOUDWATCH_ENABLED
     value: "false"
+  - description: AWS instance prefix adds string to all instance names, leave blank for production
+    name: AWS_INSTANCE_PREFIX
+    value: ""
```
lzap, line 127: This is where @RHEnVision/qe might be interested - set this to "test-" to have all instances prefixed so they can be cleaned out with Cloudwash.

deploy/clowdapp.yaml
```diff
@@ -109,8 +119,23 @@ parameters:
...
+    name: WORKER_CONCURRENCY
+    value: "100"
+  - description: Internal queue heartbeat interval. How often to update job liveness information.
+    name: WORKER_HEARTBEAT_SEC
+    value: "30"
+  - description: Internal queue beat threshold. Now many times the beat can be updated until job is cancelled.
+    name: WORKER_MAX_BEATS
+    value: "10"
   - name: OPEN_API_FILEPATH
     value: /var/tmp/openapi.json
```
lzap, line 141: No idea what this is for.

internal/config/5_integration_defaults.go
```diff
@@ -0,0 +1,10 @@
+//go:build integration && !test
+// +build integration,!test
+
+package config
+
+import "github.com/RHEnVision/provisioning-backend/internal/config/parser"
+
+func init() {
+	parser.Viper.Set("featureFlags.environment", "test")
+}
```
lzap, line 10: I did a lot of cleanup in these files, I renamed them to `_defaults` because they really only contain default values. Also added integration Go build tag which is same as test tag at the moment.

internal/config/5_normal_defaults.go
```diff
@@ -0,0 +1,41 @@
...
+		// load from working directory for integration tests
+		parser.Viper.AddConfigPath(".")
+		parser.Viper.SetConfigName("local")
+
+		err := parser.Viper.ReadInConfig()
+		if err != nil {
+			fmt.Println("Could not read configs/local.yaml, continuing with defaults")
+		}
+	}
+}
```
lzap, line 41: On the other hand, these are not just defaults, so maybe to rename this back? Not sure.

internal/config/main_config.go
```diff
@@ -98,7 +101,7 @@ var Sources = &config.RestEndpoints.Sources
 var Worker = &config.Worker
 
 func Initialize() {
-	err := viper.Unmarshal(&config)
+	err := parser.Viper.Unmarshal(&config)
```
lzap, line 104: Using `viper` package (default singleton variable viper.v) is no longer possible due to bug in the library which needs v2 release, so I moved everything into a package where I create proper string replace configuration for it.

internal/config/parser/known.go
```diff
@@ -0,0 +1,70 @@
...
+	{"RESTENDPOINTS.SOURCES.PASSWORD", "REST_SOURCES_PASSWORD"},
+	{"RESTENDPOINTS.SOURCES.PROXY", "REST_SOURCES_PROXY"},
+	{"RESTENDPOINTS.SOURCES.PROXY.URL", "REST_SOURCES_PROXY_URL"},
+	{"RESTENDPOINTS.TRACEDATA", "REST_TRACE_DATA"},
+	{"WORKER", ""},
+	{"WORKER.CONCURRENCY", ""},
+	{"WORKER.HEARTBEATSEC", "WORKER_HEARTBEAT_SEC"},
+	{"WORKER.MAXBEATS", "WORKER_MAX_BEATS"},
+	{"WORKER.QUEUE", ""},
+}
```
lzap, line 70: This file needs to be updated when we add a new configuration value, if not, pbapi will issue a warning in logs naming exactly what is missing here.

internal/config/parser/replacer.go
```diff
@@ -0,0 +1,39 @@
...
+	return strings.ToUpper(strings.ReplaceAll(str, ".", "_"))
+}
+
+func UnknownEnvVariables() []string {
+	result := make([]string, 0, len(unknownMap))
+	for k := range unknownMap {
+		result = append(result, k)
+	}
+	return result
+}
```
lzap, line 39: Yeah, it's horrible. Not my fault, this is how viper interface makes it hard to use it. Alternatively, let's remove viper and just use `os.GetEnv` and few functions that will convert from string to our config structure with types.

internal/config/parser/viper.go
```diff
@@ -0,0 +1,18 @@
...
+package parser
+
+import "github.com/spf13/viper"
+
+var Viper *viper.Viper
+
+func init() {
+	Viper = viper.NewWithOptions(viper.EnvKeyReplacer(camelCaseReplacer{}))
+	Viper.AutomaticEnv()
+}
```
lzap, line 18: This is the reason why viper must be extracted into package (because it is used from other `init` function - ordering issue) and because `SetEnvKeyReplacer` dos not accept `StringReplacer` interface so it's literally unusable.

lzap:
Fixed unit test.

deploy/clowdapp.yaml
lzap: @amirfefer isn't this something that should have been deleted? I remember that previously you had OpenAPI loading from the filesystem, but you refactored it to use Go embedding. Can you confirm?

Makefile
ezr-ondrej: sure, its visible enough :)

internal/config/5_normal_defaults.go
ezr-ondrej: yeah, the others are defaults, but this one we use to set variables from clowder, so those are final IMHO, so others default and this one main or smthing?

internal/config/parser/replacer.go
ezr-ondrej: I am all for it, with all these additions viper kinda looses its appeal of minimal configuration code. Lets get rid of it and do it ourselves?

internal/config/parser/viper.go
ezr-ondrej: I stand convinced viper is not a great tool xD

internal/config/5_normal_defaults.go
ezr-ondrej: shall we move these into main config? :thinking:

ezr-ondrej (commented):
Only thing I have here is the naming of normal_defaults, I believe it is more like real config, but it works so I'm ok if we keep it, it just sounds weird.

internal/config/5_normal_defaults.go
lzap: I don't know, it contains both defaults and both production code.

lzap:
You know what, I renamed them back to `xxx_config.go` because I don't know the clowder makes is weird. I am okay with this for now.

...until viper bites us, then I will shoot it down.

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #188, opened 2022-08-26
Use JSONB for reservation details and improve error handling

lzap:
The list of fields in aws details table was only growing. Most of the fields, however, are informative and they are only passed into jobs as inputs. They are saved only when user wants to fetch existing reservation for more details.

It makes sense to store these fields as a native JSON (JSONB) because we will need to unlikely search or order these fields. This patch does exactly that.

On top of that change, instance name can be now an array so instances can be named separately. There is, however, one big problem with that - AWS API does not allow naming instances in batches, all get the same name.

The patch improves error handling of jobs, we simply ignored errors so far. Now we properly fail reservation with "Error: xxx" status and Success bool is set to false as it was meant.

Therefore, this initial version of the change only recognizes one case - when there is exactly one name. In that case, it will be passed in when launching instance.

The solution is to create a new job that will be only responsible for renaming all existing instances separately, after they are launched. We have the list in the reservation, so that is doable.

This creates a problem tho: launch operation now will contain 2 or 3 jobs depending on if names were provided or not. The launch job currently sets the status to "Finished" and UI expects that. I think we need to change the way we report status.

There is currently the Success flag that is either null (not finished yet) or true/false. We should only use this.

On top of that, I propose to add two new ints, one is total number of steps and the other is actual step that is currently in progress. This way, the UI can actually show a relevant progress bar depending on how many steps the launch is actually doing. Also a new field `Error` would be helpful with detailed error message.

Then the UI would have the following logic: @amirfefer @adiabramovitch @ezr-ondrej 

* Check reservation `Success`:
 * `null` means it is still in progress
 * `true` means it is finished
 * `false` means there was a failure, error message should be found in `Error`
* Use `Status` string field to display some meaningful information (e.g. "Uploading public key")
* Use `Step` and `Steps` fields will contain integers, e.g. `2/3` meaning currently executing job 2 out of three 3.
* Use `Status` and `Error` field when a job was not successful to retrieve detailed error. While `Status` would contain user-facing error like `Uploading of public key failed`, the `Error` field would contain full error message (go error).

Let me know if you like this and I will incorporate this into the patch, or create a new one.

### DIFF
--- a/api/openapi.gen.yaml
+++ b/api/openapi.gen.yaml
@@ -242,6 +242,10 @@ components:
           type: string
         instance_type:
           type: string
+        name:
+          type: string
+        poweroff:
+          type: boolean
         pubkey_id:
           format: int64
           type: integer
@@ -307,6 +311,8 @@ components:
         created_at:
           format: date-time
           type: string
+        error:
+          type: string
         finished_at:
           format: date-time
           type: string
@@ -317,6 +323,12 @@ components:
           type: integer
         status:
           type: string
+        step:
+          format: int32
+          type: integer
+        steps:
+          format: int32
+          type: integer
         success:
           type: boolean
       type: object
--- a/internal/clients/ec2/ec2_client.go
+++ b/internal/clients/ec2/ec2_client.go
@@ -118,7 +118,7 @@ func (c *Client) ListInstanceTypesWithPaginator() ([]types.InstanceTypeInfo, err
 	return res, nil
 }
 
-func (c *Client) RunInstances(ctx context.Context, name string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error) {
+func (c *Client) RunInstances(ctx context.Context, name *string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error) {
 	log.Trace().Msg("Run AWS EC2 instance")
 	encodedUserData := base64.StdEncoding.EncodeToString(userData)
 	input := &ec2.RunInstancesInput{
@@ -129,16 +129,18 @@ func (c *Client) RunInstances(ctx context.Context, name string, amount int32, in
 		KeyName:      &keyName,
 		UserData:     &encodedUserData,
 	}
-	input.TagSpecifications = []types.TagSpecification{
-		{
-			ResourceType: types.ResourceTypeInstance,
-			Tags: []types.Tag{
-				{
-					Key:   aws.String("Name"),
-					Value: aws.String(name),
+	if name != nil {
+		input.TagSpecifications = []types.TagSpecification{
+			{
+				ResourceType: types.ResourceTypeInstance,
+				Tags: []types.Tag{
+					{
+						Key:   aws.String("Name"),
+						Value: aws.String(*name),
+					},
 				},
 			},
-		},
+		}
 	}
 	resp, err := c.ec2.RunInstances(ctx, input)
 	if err != nil {
--- a/internal/ctxval/context_values.go
+++ b/internal/ctxval/context_values.go
@@ -2,6 +2,7 @@ package ctxval
 
 import (
 	"context"
+	"errors"
 
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 	"github.com/rs/zerolog"
@@ -16,6 +17,8 @@ const (
 	requestNumCtxKey commonKeyId = iota
 	accountIdCtxKey  commonKeyId = iota
 )
+
+var MissingAccountInContextError = errors.New("operation requires account_id in context")
 
 // Identity returns identity header struct or nil when not set.
 func Identity(ctx context.Context) identity.XRHID {
@@ -56,9 +59,13 @@ func WithRequestNumber(ctx context.Context, num uint64) context.Context {
 	return context.WithValue(ctx, requestNumCtxKey, num)
 }
 
-// Account returns current account model or nil when not set.
+// Account returns current account model or panics when not set
 func AccountId(ctx context.Context) int64 {
-	return ctx.Value(accountIdCtxKey).(int64)
+	value := ctx.Value(accountIdCtxKey)
+	if value == nil {
+		panic(MissingAccountInContextError)
+	}
+	return value.(int64)
 }
 
 func WithAccountId(ctx context.Context, accountId int64) context.Context {
--- a/internal/dao/dao_interfaces.go
+++ b/internal/dao/dao_interfaces.go
@@ -1,3 +1,14 @@
+// Package dao provides a Database Access Object interface to stored data.
+//
+// All functions require the following variables to be set in the context:
+//
+// * Logger: for all context-aware logging.
+// * Account ID: for multi-tenancy, unless marked with UNSCOPED word.
+//
+// Functions marked as UNSCOPED can be safely used from contexts where there is
+// exactly zero function arguments coming from an user (e.g. ID was retrieved via
+// another DAO call that was scoped).
+//
 package dao
 
 import (
@@ -8,6 +19,7 @@ import (
 
 var GetAccountDao func(ctx context.Context) (AccountDao, error)
 
+// AccountDao TODO
 type AccountDao interface {
 	Create(ctx context.Context, pk *models.Account) error
 	GetById(ctx context.Context, id int64) (*models.Account, error)
@@ -19,6 +31,7 @@ type AccountDao interface {
 
 var GetPubkeyDao func(ctx context.Context) (PubkeyDao, error)
 
+// PubkeyDao TODO
 type PubkeyDao interface {
 	Create(ctx context.Context, pk *models.Pubkey) error
 	Update(ctx context.Context, pk *models.Pubkey) error
@@ -29,6 +42,7 @@ type PubkeyDao interface {
 
 var GetPubkeyResourceDao func(ctx context.Context) (PubkeyResourceDao, error)
 
+// PubkeyResourceDao TODO
 type PubkeyResourceDao interface {
 	GetResourceByProviderType(ctx context.Context, pubkeyId int64, provider models.ProviderType) (*models.PubkeyResource, error)
 	ListByPubkeyId(ctx context.Context, pkId int64) ([]*models.PubkeyResource, error)
@@ -38,14 +52,40 @@ type PubkeyResourceDao interface {
 
 var GetReservationDao func(ctx context.Context) (ReservationDao, error)
 
+// ReservationDao represents a reservation, an abstraction of one or more background jobs with
+// associated detail information different for different cloud providers (like number of vCPUs,
+// instance IDs created etc).
 type ReservationDao interface {
+	// CreateNoop creates no operation reservation with details in a single transaction.
 	CreateNoop(ctx context.Context, reservation *models.NoopReservation) error
+
+	// CreateAWS creates AWS reservation with details in a single transaction.
 	CreateAWS(ctx context.Context, reservation *models.AWSReservation) error
+
+	// CreateInstance inserts instance associated to a reservation.
 	CreateInstance(ctx context.Context, reservation *models.ReservationInstance) error
+
+	// GetById returns reservation for a particular account.
+	GetById(ctx context.Context, id int64) (*models.Reservation, error)
+
+	// List returns reservation for a particular account.
 	List(ctx context.Context, limit, offset int64) ([]*models.Reservation, error)
+
+	// ListInstances returns instances associated to a reservation. UNSCOPED.
 	ListInstances(ctx context.Context, limit, offset int64) ([]*models.ReservationInstance, error)
-	UpdateStatus(ctx context.Context, id int64, status string) error
+
+	// UpdateStatus sets status field and increment step counter by addSteps. UNSCOPED.
+	UpdateStatus(ctx context.Context, id int64, status string, addSteps int32) error
+
+	// UpdateReservationIDForAWS updates AWS reservation id field. UNSCOPED.
 	UpdateReservationIDForAWS(ctx context.Context, id int64, awsReservationId string) error
-	Finish(ctx context.Context, id int64, success bool, status string) error
+
+	// FinishWithSuccess sets Success flag. UNSCOPED.
+	FinishWithSuccess(ctx context.Context, id int64) error
+
+	// FinishWithError sets Success flag and Error flag. UNSCOPED.
+	FinishWithError(ctx context.Context, id int64, errorString string) error
+
+	// Delete deletes a reservation. Only used in tests and background cleanup job. UNSCOPED.
 	Delete(ctx context.Context, id int64) error
 }
--- a/internal/dao/sqlx/reservation_dao.go
+++ b/internal/dao/sqlx/reservation_dao.go
@@ -2,6 +2,8 @@ package sqlx
 
 import (
 	"context"
+	"database/sql"
+	"errors"
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	"github.com/RHEnVision/provisioning-backend/internal/db"
@@ -10,13 +12,15 @@ import (
 )
 
 const (
-	createReservation         = `INSERT INTO reservations (provider, account_id, status) VALUES ($1, $2, $3) RETURNING *`
-	createAwsDetail           = `INSERT INTO aws_reservation_details (reservation_id, pubkey_id, source_id, instance_type, amount, image_id, poweroff, name) VALUES ($1, $2, $3, $4, $5, $6, $7, $8)`
-	updateReservationStatus   = `UPDATE reservations SET status = $2 WHERE id = $1 RETURNING *`
+	createReservation         = `INSERT INTO reservations (provider, account_id, steps, status) VALUES ($1, $2, $3, $4) RETURNING *`
+	createAwsDetail           = `INSERT INTO aws_reservation_details (reservation_id, pubkey_id, source_id, image_id, detail) VALUES ($1, $2, $3, $4, $5)`
+	updateReservationStatus   = `UPDATE reservations SET status = $2, step = step + $3 WHERE id = $1 RETURNING *`
 	updateReservationIDForAWS = `UPDATE aws_reservation_details SET aws_reservation_id = $2 WHERE reservation_id = $1 RETURNING *`
-	finishReservationStatus   = `UPDATE reservations SET status = $2, success = $3, finished_at = now() WHERE id = $1 RETURNING *`
+	finishReservationSuccess  = `UPDATE reservations SET success = true, finished_at = now() WHERE id = $1 RETURNING *`
+	finishReservationError    = `UPDATE reservations SET success = false, error = $2, finished_at = now() WHERE id = $1 RETURNING *`
 	deleteReservationById     = `DELETE FROM reservations WHERE id = $1`
-	listReservations          = `SELECT * FROM reservations ORDER BY id LIMIT $1 OFFSET $2`
+	getReservationById        = `SELECT * FROM reservations WHERE account_id = $1 AND id = $2 LIMIT 1`
+	listReservations          = `SELECT * FROM reservations WHERE account_id = $1 ORDER BY id LIMIT $2 OFFSET $3`
 	createInstance            = `INSERT INTO reservation_instances (reservation_id, instance_id) VALUES ($1, $2)`
 	listInstanceReservations  = `SELECT * FROM reservation_instances ORDER BY reservation_id LIMIT $1 OFFSET $2`
 )
@@ -24,9 +28,11 @@ const (
 type reservationDaoSqlx struct {
 	name                      string
 	create                    *sqlx.Stmt
+	getById                   *sqlx.Stmt
 	createAwsDetail           *sqlx.Stmt
 	updateStatus              *sqlx.Stmt
-	finishStatus              *sqlx.Stmt
+	finishSuccess             *sqlx.Stmt
+	finishError               *sqlx.Stmt
 	deleteById                *sqlx.Stmt
 	list                      *sqlx.Stmt
 	createInstance            *sqlx.Stmt
@@ -43,6 +49,10 @@ func getReservationDao(ctx context.Context) (dao.ReservationDao, error) {
 	if err != nil {
 		return nil, NewPrepareStatementError(ctx, &daoImpl, createReservation, err)
 	}
+	daoImpl.getById, err = db.DB.PreparexContext(ctx, getReservationById)
+	if err != nil {
+		return nil, NewPrepareStatementError(ctx, &daoImpl, getReservationById, err)
+	}
 	daoImpl.createAwsDetail, err = db.DB.PreparexContext(ctx, createAwsDetail)
 	if err != nil {
 		return nil, NewPrepareStatementError(ctx, &daoImpl, createAwsDetail, err)
@@ -51,9 +61,13 @@ func getReservationDao(ctx context.Context) (dao.ReservationDao, error) {
 	if err != nil {
 		return nil, NewPrepareStatementError(ctx, &daoImpl, updateReservationStatus, err)
 	}
-	daoImpl.finishStatus, err = db.DB.PreparexContext(ctx, finishReservationStatus)
+	daoImpl.finishSuccess, err = db.DB.PreparexContext(ctx, finishReservationSuccess)
 	if err != nil {
-		return nil, NewPrepareStatementError(ctx, &daoImpl, finishReservationStatus, err)
+		return nil, NewPrepareStatementError(ctx, &daoImpl, finishReservationSuccess, err)
+	}
+	daoImpl.finishError, err = db.DB.PreparexContext(ctx, finishReservationError)
+	if err != nil {
+		return nil, NewPrepareStatementError(ctx, &daoImpl, finishReservationError, err)
 	}
 	daoImpl.list, err = db.DB.PreparexContext(ctx, listReservations)
 	if err != nil {
@@ -90,7 +104,11 @@ func (di *reservationDaoSqlx) CreateNoop(ctx context.Context, reservation *model
 	query := createReservation
 	stmt := di.create
 
-	err := stmt.GetContext(ctx, reservation, reservation.Provider, reservation.AccountID, reservation.Status)
+	err := stmt.GetContext(ctx, reservation,
+		reservation.Provider,
+		reservation.AccountID,
+		reservation.Steps,
+		reservation.Status)
 	if err != nil {
 		return NewGetError(ctx, di, query, err)
 	}
@@ -101,7 +119,11 @@ func (di *reservationDaoSqlx) CreateAWS(ctx context.Context, reservation *models
 	err := dao.WithTransaction(ctx, func(tx *sqlx.Tx) error {
 		query := createReservation
 		stmt := di.create
-		err := stmt.GetContext(ctx, reservation, reservation.Provider, reservation.AccountID, reservation.Status)
+		err := stmt.GetContext(ctx, reservation,
+			reservation.Provider,
+			reservation.AccountID,
+			reservation.Steps,
+			reservation.Status)
 		if err != nil {
 			return NewGetError(ctx, di, query, err)
 		}
@@ -112,11 +134,8 @@ func (di *reservationDaoSqlx) CreateAWS(ctx context.Context, reservation *models
 			reservation.ID,
 			reservation.PubkeyID,
 			reservation.SourceID,
-			reservation.InstanceType,
-			reservation.Amount,
 			reservation.ImageID,
-			reservation.PowerOff,
-			reservation.Name)
+			reservation.Detail)
 		if err != nil {
 			return NewExecUpdateError(ctx, di, query, err)
 		}
@@ -150,12 +169,28 @@ func (di *reservationDaoSqlx) CreateInstance(ctx context.Context, reservation *m
 	return nil
 }
 
+func (di *reservationDaoSqlx) GetById(ctx context.Context, id int64) (*models.Reservation, error) {
+	query := getReservationById
+	stmt := di.getById
+	result := &models.Reservation{}
+
+	err := stmt.GetContext(ctx, result, ctxAccountId(ctx), id)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, NewNoRowsError(ctx, di, query, err)
+		} else {
+			return nil, NewGetError(ctx, di, query, err)
+		}
+	}
+	return result, nil
+}
+
 func (di *reservationDaoSqlx) List(ctx context.Context, limit, offset int64) ([]*models.Reservation, error) {
 	query := listReservations
 	stmt := di.list
 	var result []*models.Reservation
 
-	err := stmt.SelectContext(ctx, &result, limit, offset)
+	err := stmt.SelectContext(ctx, &result, ctxAccountId(ctx), limit, offset)
 	if err != nil {
 		return nil, NewSelectError(ctx, di, query, err)
 	}
@@ -174,11 +209,11 @@ func (di *reservationDaoSqlx) ListInstances(ctx context.Context, limit, offset i
 	return result, nil
 }
 
-func (di *reservationDaoSqlx) UpdateStatus(ctx context.Context, id int64, status string) error {
+func (di *reservationDaoSqlx) UpdateStatus(ctx context.Context, id int64, status string, addSteps int32) error {
 	query := updateReservationStatus
 	stmt := di.updateStatus
 
-	res, err := stmt.ExecContext(ctx, id, status)
+	res, err := stmt.ExecContext(ctx, id, status, addSteps)
 	if err != nil {
 		return NewExecUpdateError(ctx, di, query, err)
 	}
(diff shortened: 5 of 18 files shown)


internal/clients/ec2/ec2_client.go
```diff
@@ -133,16 +133,18 @@ func (c *Client) RunInstances(ctx context.Context, name string, amount int32, in
...
+				ResourceType: types.ResourceTypeInstance,
+				Tags: []types.Tag{
+					{
+						Key:   aws.String("Name"),
+						Value: aws.String(name),
+					},
 				},
 			},
-		},
+		}
```
lzap, line 147: Empty name allows to completely avoid any tagging of instances during launch.

internal/dao/sqlx/reservation_dao.go
```diff
@@ -112,11 +112,7 @@ func (di *reservationDaoSqlx) CreateAWS(ctx context.Context, reservation *models
...
 			reservation.SourceID,
-			reservation.InstanceType,
-			reservation.Amount,
-			reservation.ImageID,
-			reservation.PowerOff,
-			reservation.Name)
+			reservation.Detail)
 		if err != nil {
 			return NewExecUpdateError(ctx, di, query, err)
 		}
```
lzap, line 118: This will be much easier to maintain now.

internal/db/migrations/005_reservations.sql
```diff
@@ -23,11 +23,7 @@ CREATE TABLE aws_reservation_details
   provider INTEGER NOT NULL DEFAULT provider_type_aws(),
   pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
   source_id TEXT NOT NULL,
-  name TEXT NOT NULL DEFAULT '',
-  instance_type TEXT NOT NULL,
-  amount INTEGER NOT NULL,
-  image_id TEXT NOT NULL,
-  poweroff BOOLEAN NOT NULL DEFAULT false,
+  detail JSONB,
   aws_reservation_id TEXT,
```
lzap, line 27: JSON/JSONB are natively supported by pgx driver, no custom code needed :-)

internal/jobs/launch_instance_aws.go
```diff
@@ -43,6 +46,18 @@ func HandleLaunchInstanceAWS(ctx context.Context, job dejq.Job) error {
...
 
+	jobErr := handleLaunchInstanceAWS(ctx, &args)
+	if jobErr != nil {
+		finishWithError(ctx, args.ReservationID, jobErr)
+	}
+	return jobErr
+}
+
+// Job logic, when error is returned the job status is updated accordingly
+func handleLaunchInstanceAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) error {
```
lzap, line 57: I am splitting handlers into exported and unexported function, the former will unmarshal and handle error. This is common across all the three jobs we have today. We kind of ignored errors until now, my bad.

internal/jobs/pubkey_upload_aws_job.go
```diff
@@ -108,12 +121,12 @@ func HandlePubkeyUploadAWS(ctx context.Context, job dejq.Job) error {
...
 	}
 
-	// mark the reservation as finished
+	// update the reservation status, but keep it unfinished yet
 	rDao, err := dao.GetReservationDao(ctx)
 	if err != nil {
 		return fmt.Errorf("cannot get reservation DAO: %w", err)
 	}
-	err = rDao.UpdateStatus(ctx, args.ReservationID, "Finished")
+	err = rDao.UpdateStatus(ctx, args.ReservationID, "Key uploaded")
```
lzap, line 129: This is why I think we definitely need to improve on status recognition - only Status field is not enough. We do have Success flag but it was unused until this patch.

scripts/rest_examples/reservation-create-aws.http
```diff
@@ -4,7 +4,7 @@ Content-Type: application/json
 X-Rh-Identity: {{identity}}
 
 {
-  "name": "amz-linux-us-east-1",
+  "names": ["amz-linux-us-east-1"],
   "source_id": "1",
   "image_id": "ami-05fa00d4c63e32376",
   "amount": 1,
```
lzap, line 10: If you pass an array with exactly one name, it works as previously - the launchInstance function will actually proceed because it can tag all of them. We should probably update the check condition, it should only do it if they are ALL the same, not when there is only one tho.

adiabramovitch:
love it! think you need to update the spec as well

internal/services/aws_reservation_service.go
```diff
@@ -112,9 +126,9 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var ami string
-	if strings.HasPrefix(reservation.ImageID, "ami-") {
+	if strings.HasPrefix(reservation.Detail.ImageID, "ami-") {
 		// Direct AMI ID was provided, no need to call image builder
-		ami = reservation.ImageID
+		ami = reservation.Detail.ImageID
```
ezr-ondrej, line 131: I'd vote for ImageId staying on the table. We might need to answer questions like: "how many instances have been provisioned from Image X?"

lzap: Moved back!

internal/dao/sqlx/reservation_dao.go
```diff
@@ -174,11 +212,11 @@ func (di *reservationDaoSqlx) ListInstances(ctx context.Context, limit, offset i
 	return result, nil
 }
 
-func (di *reservationDaoSqlx) UpdateStatus(ctx context.Context, id int64, status string) error {
+func (di *reservationDaoSqlx) UpdateStatus(ctx context.Context, id int64, status string, addSteps int32) error {
```
ezr-ondrej, line 215: Do we expect adding more steps at the same time? or is that for not adding a step, but still update status?

internal/dao/sqlx/reservation_dao.go
```diff
@@ -150,11 +171,28 @@ func (di *reservationDaoSqlx) CreateInstance(ctx context.Context, reservation *m
 	return nil
 }
 
+func (di *reservationDaoSqlx) GetByIdAllAccounts(ctx context.Context, id int64) (*models.Reservation, error) {
```
ezr-ondrej, line 174: This shall not be scoped by the account? We have the account for which the job was executed, or is there another reason to not scope by account?

ezr-ondrej (commented):
The renaming is quite a lot of code that we still don't know we will need, is that worth implementing in this PR, shall we not wait for what UX will come up with?

The job tooling is quite awesome in general, I love the addional observability of it, the JSONB looks great as well, did not go into details of implementation yet tho.

lzap:
> The renaming is quite a lot of code

What renaming specifically? Status messages? They have never came from UX so we are free to rename as we like until this gets on screen and they join the effort. I have to rename them because previously they were used after jobs were executed which is bad for UX - what should be displayed for the initial job? The same applies for the last job and final status - what should be there? This solves both ends. Thanks to the first optional job (rename instances) you will never know if the job you are working on is last or not.

internal/dao/sqlx/reservation_dao.go
lzap: This function is meant to update status string and step number in a relative way (typically 0 or 1 - add nothing, add one). If you have a better idea, let me know. I do not love this design honestly, but I could not figure out anything better since you want to update status text (beginning of a job) and status and step (end of a job). Two functions look like an overkill (two queries).

internal/dao/sqlx/reservation_dao.go
lzap: Right, I was rushing this and I forgot that I actually do have the account ID at the call site! So let's definitely add this for safety measures (someone else could accidentally use the function without account). Good call.

lzap: So I did a full review of the whole Reservation DAO interface and I added scope to the two functions which take user's input. I fully documented everything:

```go
// ReservationDao represents a reservation, an abstraction of one or more background jobs with
// associated detail information different for different cloud providers (like number of vCPUs,
// instance IDs created etc).
type ReservationDao interface {
	// CreateNoop creates no operation reservation with details in a single transaction.
	CreateNoop(ctx context.Context, reservation *models.NoopReservation) error

	// CreateAWS creates AWS reservation with details in a single transaction.
	CreateAWS(ctx context.Context, reservation *models.AWSReservation) error

	// CreateInstance inserts instance associated to a reservation.
	CreateInstance(ctx context.Context, reservation *models.ReservationInstance) error

	// GetById returns reservation for a particular account.
	GetById(ctx context.Context, id int64) (*models.Reservation, error)

	// List returns reservation for a particular account.
	List(ctx context.Context, limit, offset int64) ([]*models.Reservation, error)

	// ListInstances returns instances associated to a reservation. UNSCOPED.
	ListInstances(ctx context.Context, limit, offset int64) ([]*models.ReservationInstance, error)

	// UpdateStatus sets status field and increment step counter by addSteps. UNSCOPED.
	UpdateStatus(ctx context.Context, id int64, status string, addSteps int32) error

	// UpdateReservationIDForAWS updates AWS reservation id field. UNSCOPED.
	UpdateReservationIDForAWS(ctx context.Context, id int64, awsReservationId string) error

	// FinishWithSuccess sets Success flag. UNSCOPED.
	FinishWithSuccess(ctx context.Context, id int64) error

	// FinishWithError sets Success flag and Error flag. UNSCOPED.
	FinishWithError(ctx context.Context, id int64, error string) error

	// Delete deletes a reservation. Only used in tests and background cleanup job. UNSCOPED.
	Delete(ctx context.Context, id int64) error
}
```

lzap: FYI @avitova this PR will add few new test functions into the `reservation_test.go` from DAO test suite, I will do minimum possible changes to get it working, I will copy paste from existing functions. For the record, the new functions are:

* TestUpdateStepReservation
* TestFinishWithErrorReservation

No other changes, I know you are still working on this so hopefully this should help you with git rebasing.

ezr-ondrej:
> What renaming specifically

Sorry for not being specific I meant renaming of the instances

lzap:
You mean instance names I guess, this is only in payloads and it gets stored into reservation, however, the actual job is empty (TODO). Let's create an issue for that, it is easy to implement. This is done by creating a tag named "Name" for each instance found in the instances table for the particular reservation.

Rebased, tests are passing now.

adiabramovitch:
Started going through it 
But I am wondering,  does the JSONB change need to be in the same PR as the multiple names?

lzap:
> does the JSONB change need to be in the same PR as the multiple names?

This is exactly what is happening here, read the code you will see. The title literally says: "Support for multiple names and JSONB detail". Or I am not following!? :)

internal/jobs/launch_instance_aws.go
```diff
@@ -15,38 +15,58 @@ import (
...
-
 	args := LaunchInstanceAWSTaskArgs{}
 	err := job.Decode(&args)
 	if err != nil {
-		ctxLogger.Error().Err(err).Msg("unable to decode arguments")
+		ctxval.Logger(ctx).Error().Err(err).Msg("unable to decode arguments")
 		return fmt.Errorf("unable to decode args: %w", err)
 	}
+	logger := ctxval.Logger(ctx).With().Int64("reservation_id", args.ReservationID).Logger()
+	logger.Info().Interface("args", args).Msg("Processing no operation job")
```
adiabramovitch, line 46: I think this msg should be regarding Launch Instance Job

internal/jobs/pubkey_upload_aws_job.go
```diff
@@ -21,6 +21,7 @@ type PubkeyUploadAWSTaskArgs struct {
 	ARN           string `json:"arn"`
 }
 
+// Unmarshall arguments and handle error
 func HandlePubkeyUploadAWS(ctx context.Context, job dejq.Job) error {
 	ctxLogger := ctxval.Logger(ctx)
 	ctxLogger.Debug().Msg("Started pubkey upload AWS job")
```
adiabramovitch, line 27: Saw that you removed the log msgs from Noop and Launch Instance.
SHould we remove this as well? (it is duplicated in handlePubkeyUploadAWS)

internal/jobs/noop_job.go
```diff
@@ -14,28 +14,36 @@ type NoopJobArgs struct {
...
+		finishWithError(ctx, args.ReservationID, jobErr)
+	} else {
+		finishWithSuccess(ctx, args.ReservationID)
 	}
+	return jobErr
+}
+
+// Job logic, when error is returned the job status is updated accordingly
+func handleNoop(ctx context.Context, args *NoopJobArgs) error {
+	// status updates before and after the code logic
```
adiabramovitch, line 41: Do we want to add the log msgs here as we do for the launch instance and upload pubkeys?

internal/jobs/rename_instances_aws_job.go
```diff
@@ -0,0 +1,47 @@
...
+
+type RenameInstanceJobArgs struct {
+	AccountID     int64             `json:"account_id"`
+	ReservationID int64             `json:"reservation_id"`
+	Detail        *models.AWSDetail `json:"detail"`
+}
+
+// Unmarshall arguments and handle error
+func HandleRenameInstance(ctx context.Context, job dejq.Job) error {
+	args := NoopJobArgs{}
```
adiabramovitch, line 20: Why do you use Noop args?

internal/jobs/common.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	err = rDao.FinishWithError(ctx, reservationId, jobError.Error())
+	if err != nil {
+		ctxLogger.Warn().Err(err).Msg("unable to update job status: finish")
+	}
+}
+
+func updateStatusBefore(ctx context.Context, id int64, status string) {
+	updateStatusAfter(ctx, id, status, 0)
+}
```
adiabramovitch, line 53: I dont understand why did you choose this implementation, can you explain? 😄

internal/jobs/launch_instance_aws.go
lzap: Good catch, you know what? I see a lot of copy&pasted code in these handlers, so I created a bunch of helper functions in that package which shortens this to just:

```go
// Unmarshall arguments and handle error
func HandleNoop(ctx context.Context, job dejq.Job) error {
	args := NoopJobArgs{}
	decodeJob(ctx, job, args)
	ctx = contextLogger(ctx, job.Type(), args, args.AccountID, args.ReservationID)

	jobErr := handleNoop(ctx, &args)

	finishJob(ctx, args.ReservationID, jobErr)
	return jobErr
}

// Job logic, when error is returned the job status is updated accordingly
func handleNoop(ctx context.Context, args *NoopJobArgs) error {
        // ...
}
```

internal/jobs/pubkey_upload_aws_job.go
lzap: I created a common code for all of the jobs so logging is now always guaranteed to be same.

internal/jobs/noop_job.go
lzap: Again, the common function `contextLogger` now logs the following:

```go
	logger := ctxval.Logger(newContext).With().Int64("reservation_id", reservationId).Logger()
	logger.Info().Interface("args", args).Msgf("Processing job: '%s'", jobName)
```

All jobs should call it, so they will all contain this line:

    reservation_id=12345 Processing job: 'job_name'

internal/jobs/rename_instances_aws_job.go
lzap: Copy-pasting, this is a bug actually :-) Weird it worked the last time I tried this.

lzap: Oh wait, this is the empty job, I haven't tested it yet. Yup, copy and paste problem, fixing.

internal/jobs/common.go
lzap: So one of the drawback of the old design was that Status (string) field was used both for status messages before and after ("something started" as well as "something is done" or "something failed").

Now, the DAO exposes function called `UpdateStatus` that previously only contained `status string` argument, now also contains `addSteps int32` - a relative change in steps that needs to be performed when setting new status.

In the status updates at job start, we do not want to update step counter, so this is when you pass zero. When job ends (regardless if there was error or not - that's what `defer` does), we actually want to increase step by one. So that's why I have two different function, so it is easy to understand for someone who is implementing new job by copying code.

Otherwise the code would need to look something like this, which is pretty not intuitive:

```go
rDao.UpdateStatus(ctx, reservationId, "Starting xyz", 0)

// do some work

rDao.UpdateStatus(ctx, reservationId, "Finished xyz", 1)
```

I could design it via two separate functions, but then these would be two separate SQL queries which is slow:

```go
rDao.UpdateStatus(ctx, reservationId, "Starting xyz")

// do some work

rDao.UpdateStatus(ctx, reservationId, "Finished xyz")
rDao.IncrementStep(ctx, reservationId)
```

But if you like this more, an extra SQL is reasonable. It's just weird to issue two `UPDATE` SQL statements for the very same table twice in a row.

lzap:
I rebased, the only change is extraction of common code into `jobs/common.go` in the handler functions.

scripts/rest_examples/reservation-create-aws.http
```diff
@@ -4,11 +4,11 @@ Content-Type: application/json
...
-  "instance_type": "t1.micro",
+  "instance_type": "t1.micro2",
```
lzap, line 11: This was not meant to be commited, use this to simulate an error. Note that memory jobqueue will panic, so you need to restart the app, then list reservations to see how an error really looks like. In production there is no panic.

internal/services/aws_reservation_service.go
```diff
@@ -138,17 +157,30 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
...
+		}
+		startJobs = append(startJobs, renameJob)
+	}
+
 	// Enqueue all jobs
-	err = queue.GetEnqueuer().Enqueue(r.Context(), uploadPubkeyJob, launchJob)
+	err = queue.GetEnqueuer().Enqueue(r.Context(), startJobs...)
 	if err != nil {
 		renderError(w, r, payloads.NewEnqueueTaskError(r.Context(), "AWS reservation", err))
 		return
```
lzap, line 186: This change makes sure to schedule 2 jobs in case there are no names provided, otherwise it schedules 3 jobs (rename instances as the third one).

internal/payloads/reservation_payload.go
```diff
@@ -116,13 +134,24 @@ func NewNoopReservationResponse(reservation *models.NoopReservation) render.Rend
...
 	list := make([]render.Renderer, 0, len(reservations))
 	for _, reservation := range reservations {
+		var finishedAt *time.Time
+		if reservation.FinishedAt.Valid {
+			finishedAt = &reservation.FinishedAt.Time
+		}
+		var success *bool
+		if reservation.Success.Valid {
+			success = &reservation.Success.Bool
+		}
```
lzap, line 144: This hunk fixes `null` value for finished_at/success, it was not working properly.

internal/payloads/reservation_payload.go
```diff
@@ -50,6 +59,12 @@ type AWSReservationResponsePayload struct {
 
 	// The ID of the aws reservation which was created.
 	AWSReservationID string `json:"aws_reservation_id"`
+
+	// Optional name of the instance(s).
+	Names *[]string `json:"names"`
+
+	// Immediately power off the system after initialization
+	PowerOff bool `json:"poweroff"`
 }
```
lzap, line 68: I realized that the PowerOff field was not present in AWS payload by accident, so this fixes it.

internal/models/reservation_model.go
```diff
@@ -35,6 +45,20 @@ type NoopReservation struct {
...
+	// AWS Instance type.
+	InstanceType string `json:"instance_type"`
+
+	// Amount of instances to provision of type: Instance type.
+	Amount int32 `json:"amount"`
+
+	// Immediately power off the system after initialization
+	PowerOff bool `json:"poweroff"`
+}
+
```
lzap, line 61: This is a new type that is serialized into database as JSONB. It is embedded into the reservation type under name `Detail`.

internal/models/reservation_model.go
```diff
@@ -44,23 +68,14 @@ type AWSReservation struct {
...
-	PowerOff bool `db:"poweroff" json:"poweroff"`
-
 	// The ID of the aws reservation which was created.
 	AWSReservationID string `db:"aws_reservation_id" json:"aws_reservation_id"`
+
+	// The ID of the image from which the instance is created. AMI's must be prefixed with 'ami-'.
+	ImageID string `json:"image_id"`
+
+	// Detail information is stored as JSON in DB
+	Detail *AWSDetail `db:"detail" json:"detail"`
```
lzap, line 78: Here is where it get's referenced, `pgx` driver works seamlessly for this.

I kept `ImageID` here because it is a good idea as we will have to search by image id at some point.

internal/models/reservation_model.go
```diff
@@ -21,9 +21,19 @@ type Reservation struct {
...
+	// Active job step for this reservation. See Status for more details.
+	Step int32 `db:"step" json:"step"`
+
+	// Textual status of the reservation or error when there was a failure. This is user-facing message,
+	// it should not contain error details. See Error field for more details.
 	Status string `db:"status" json:"status"`
 
+	// Error message when reservation was not successful. Only set when Success if false.
+	Error string `db:"error" json:"error,omitempty"`
+
```
lzap, line 36: These are the new fields added for improved UX.

internal/jobs/noop_job.go
```diff
@@ -14,28 +12,26 @@ type NoopJobArgs struct {
...
+	updateStatusBefore(ctx, args.ReservationID, "No operation started")
+	defer updateStatusAfter(ctx, args.ReservationID, "No operation finished", 1)
+
+	// do nothing and return no error
+	time.Sleep(5 * time.Second)
 
 	return nil
 }
```
lzap, line 37: All of these changes are improvements to error handling, I feel like a separate method with an explicit error is a good way to ensure error handling is always correct. Also using `defer` allows me to ensure that we will never forget to set status and update step even when there is an error. All the other jobs have the very same changes, just review this one and then read the others to ensure I did not do another copy&paste mistake :-)

internal/jobs/launch_instance_aws.go
```diff
@@ -15,38 +15,50 @@ import (
...
-	// Amount of instances to launch
-	Amount int32 `json:"amount"`
-	// Immediately power off the system
-	PowerOff bool `json:"poweroff"`
-	// Amazon EC2 Instance Type
-	InstanceType string `json:"instance_type"`
+
 	// The ARN fetched from Sources which is linked to a specific source
 	ARN string `json:"arn"`
 }
```
lzap, line 35: Since we now have a "reservation detail" type, let's use it in job arguments too. It is already serializable, so less copying is left on the programmer.

internal/jobs/common.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		ctxLogger.Warn().Err(err).Msg("unable to update step number: dao")
+		return
+	}
+
+	err = rDao.UpdateStatus(ctx, id, status, int32(addSteps))
+	if err != nil {
+		ctxLogger.Warn().Err(err).Msg("unable to update step number: update")
+	}
+}
```
lzap, line 99: This is a bunch of unexported functions which hopefully make jobs easier to maintain and read. We can perhaps take this further but I did not want to do too drastic changes to how we handle jobs.

internal/db/migrations/005_reservations.sql
```diff
@@ -23,12 +26,9 @@ CREATE TABLE aws_reservation_details
...
-  name TEXT NOT NULL DEFAULT '',
-  instance_type TEXT NOT NULL,
-  amount INTEGER NOT NULL,
   image_id TEXT NOT NULL,
-  poweroff BOOLEAN NOT NULL DEFAULT false,
   aws_reservation_id TEXT,
+  detail JSONB,
 
   FOREIGN KEY (reservation_id, provider) REFERENCES reservations(id, provider) ON DELETE CASCADE,
   CHECK (provider = provider_type_aws())
```
lzap, line 34: To test this, migrations must be done.

internal/dao/sqlx/reservation_dao.go
```diff
@@ -204,11 +241,26 @@ func (di *reservationDaoSqlx) UpdateReservationIDForAWS(ctx context.Context, id
...
+
+func (di *reservationDaoSqlx) FinishWithError(ctx context.Context, id int64, errorString string) error {
+	query := finishReservationError
+	stmt := di.finishError
 
-	res, err := stmt.ExecContext(ctx, id, status, success)
+	res, err := stmt.ExecContext(ctx, id, false, errorString)
 	if err != nil {
 		return NewExecUpdateError(ctx, di, query, err)
 	}
```
lzap, line 266: These two finish functions almost do the same thing, yet the latter takes `error` field not `status` field.

I really do not want to create a generic `Update(ctx, reservation)` function because that will make the handler code really not very readable.

internal/dao/sqlx/reservation_dao.go
```diff
@@ -10,23 +12,27 @@ import (
...
 	updateReservationIDForAWS = `UPDATE aws_reservation_details SET aws_reservation_id = $2 WHERE reservation_id = $1 RETURNING *`
-	finishReservationStatus   = `UPDATE reservations SET status = $2, success = $3, finished_at = now() WHERE id = $1 RETURNING *`
+	finishReservationSuccess  = `UPDATE reservations SET success = $2, finished_at = now() WHERE id = $1 RETURNING *`
+	finishReservationError    = `UPDATE reservations SET success = $2, error = $3, finished_at = now() WHERE id = $1 RETURNING *`
 	deleteReservationById     = `DELETE FROM reservations WHERE id = $1`
-	listReservations          = `SELECT * FROM reservations ORDER BY id LIMIT $1 OFFSET $2`
+	getReservationById        = `SELECT * FROM reservations WHERE account_id = $1 AND id = $2 LIMIT 1`
+	listReservations          = `SELECT * FROM reservations WHERE account_id = $1 ORDER BY id LIMIT $2 OFFSET $3`
 	createInstance            = `INSERT INTO reservation_instances (reservation_id, instance_id) VALUES ($1, $2)`
 	listInstanceReservations  = `SELECT * FROM reservation_instances ORDER BY reservation_id LIMIT $1 OFFSET $2`
```
lzap, line 25: I am adding the new fields and details to create statements.

In the update, I do `step = step + X` to do relative update instead of absolute, which cannot be done since the order of jobs is unknown (can be shuffled around later on).

Finish function was split into two variants, one sets status, the other sets error fields.

Both get and list functions are now account-aware as these are the only functions which takes arguments from the user.

internal/dao/dao_interfaces.go
```diff
@@ -38,14 +52,40 @@ type PubkeyResourceDao interface {
...
+
+	// FinishWithSuccess sets Success flag. UNSCOPED.
+	FinishWithSuccess(ctx context.Context, id int64) error
+
+	// FinishWithError sets Success flag and Error flag. UNSCOPED.
+	FinishWithError(ctx context.Context, id int64, errorString string) error
+
+	// Delete deletes a reservation. Only used in tests and background cleanup job. UNSCOPED.
 	Delete(ctx context.Context, id int64) error
 }
```
lzap, line 91: I think the documentation speaks for itself here :-)

internal/ctxval/context_values.go
```diff
@@ -56,9 +59,13 @@ func WithRequestNumber(ctx context.Context, num uint64) context.Context {
...
-// Account returns current account model or nil when not set.
+// Account returns current account model or panics when not set
 func AccountId(ctx context.Context) int64 {
-	return ctx.Value(accountIdCtxKey).(int64)
+	value := ctx.Value(accountIdCtxKey)
+	if value == nil {
+		panic(MissingAccountInContextError)
+	}
+	return value.(int64)
 }
```
lzap, line 69: When testing, I realized that if you call a DAO function that requires account in context but it is not present (e.g. worker process did not have it previously), an nil pointer panic occurs.

This improves the error message, it now says: operation requires account_id in context

api/openapi.gen.yaml
```diff
@@ -242,6 +244,12 @@ components:
           type: string
         instance_type:
           type: string
+        names:
+          items:
+            type: string
+          type: array
+        poweroff:
+          type: boolean
```
lzap, line 252: I realized that the PowerOff field was missing in AWS response, so I added it.

lzap:
Dropped names array from the patch, it is now just a single name. Was that what you wanted @ezr-ondrej ?

api/openapi.gen.yaml
ezr-ondrej: Do we want to expose it? given it's an internal flag, that we do not want to expose to customers, I'd rather hide it (on request as well as response) :thinking:

internal/dao/sqlx/reservation_dao.go
```diff
@@ -101,7 +120,12 @@ func (di *reservationDaoSqlx) CreateAWS(ctx context.Context, reservation *models
 	err := dao.WithTransaction(ctx, func(tx *sqlx.Tx) error {
 		query := createReservation
 		stmt := di.create
-		err := stmt.GetContext(ctx, reservation, reservation.Provider, reservation.AccountID, reservation.Status)
+		err := stmt.GetContext(ctx, reservation,
+			reservation.Provider,
+			reservation.AccountID,
+			reservation.Step,
```
ezr-ondrej, line 126: same as ^^

internal/dao/sqlx/reservation_dao.go
ezr-ondrej: I did not realize we might not want to bumpt step when updating status, this is good :+1: two functions would be indead overkill for one argument difference, I just missed the second use case :+1:

internal/dao/sqlx/reservation_dao.go
```diff
@@ -10,23 +12,27 @@ import (
...
 const (
-	createReservation         = `INSERT INTO reservations (provider, account_id, status) VALUES ($1, $2, $3) RETURNING *`
-	createAwsDetail           = `INSERT INTO aws_reservation_details (reservation_id, pubkey_id, source_id, instance_type, amount, image_id, poweroff, name) VALUES ($1, $2, $3, $4, $5, $6, $7, $8)`
-	updateReservationStatus   = `UPDATE reservations SET status = $2 WHERE id = $1 RETURNING *`
+	createReservation         = `INSERT INTO reservations (provider, account_id, step, steps, status) VALUES ($1, $2, $3, $4, $5) RETURNING *`
+	createAwsDetail           = `INSERT INTO aws_reservation_details (reservation_id, pubkey_id, source_id, image_id, detail) VALUES ($1, $2, $3, $4, $5)`
+	updateReservationStatus   = `UPDATE reservations SET status = $2, step = step + $3 WHERE id = $1 RETURNING *`
 	updateReservationIDForAWS = `UPDATE aws_reservation_details SET aws_reservation_id = $2 WHERE reservation_id = $1 RETURNING *`
-	finishReservationStatus   = `UPDATE reservations SET status = $2, success = $3, finished_at = now() WHERE id = $1 RETURNING *`
+	finishReservationSuccess  = `UPDATE reservations SET success = $2, finished_at = now() WHERE id = $1 RETURNING *`
```
ezr-ondrej, line 19: similar comment here, should we have this static? Here it's just about convenience, not a big deal, but seems like extra argument.

internal/db/migrations/005_reservations.sql
```diff
@@ -4,7 +4,10 @@ CREATE TABLE reservations
   provider INTEGER NOT NULL CHECK (valid_provider(provider)),
   account_id BIGINT NOT NULL REFERENCES accounts(id),
   created_at TIMESTAMP NOT NULL DEFAULT current_timestamp,
+  steps SMALLINT NOT NULL,
+  step SMALLINT NOT NULL,
```
ezr-ondrej, line 8: Shall we default to 0?
```suggestion
  step SMALLINT NOT NULL DEFAULT 0,
```

internal/jobs/launch_instance_aws.go
```diff
@@ -15,38 +15,50 @@ import (
...
 
+// Unmarshall arguments and handle error
 func HandleLaunchInstanceAWS(ctx context.Context, job dejq.Job) error {
+	args := LaunchInstanceAWSTaskArgs{}
+	decodeJob(ctx, job, args)
```
ezr-ondrej, line 40: this allows to continue the execution even if the decoding has failed, is that what we want?

scripts/rest_examples/reservation-create-aws.http
```diff
@@ -4,7 +4,7 @@ Content-Type: application/json
 X-Rh-Identity: {{identity}}
 
 {
-  "name": "amz-linux-us-east-1",
+  "names": ["amz-linux-us-east-1"],
```
ezr-ondrej, line 7: shall we revert this? :)

ezr-ondrej (commented):
I've got few nits, but overall it's an awesome patch, I'll take it for a spin as we do not have QE test for AWS deployment yet :)

internal/dao/sqlx/reservation_dao.go
```diff
@@ -10,23 +12,27 @@ import (
 )
 
 const (
-	createReservation         = `INSERT INTO reservations (provider, account_id, status) VALUES ($1, $2, $3) RETURNING *`
-	createAwsDetail           = `INSERT INTO aws_reservation_details (reservation_id, pubkey_id, source_id, instance_type, amount, image_id, poweroff, name) VALUES ($1, $2, $3, $4, $5, $6, $7, $8)`
-	updateReservationStatus   = `UPDATE reservations SET status = $2 WHERE id = $1 RETURNING *`
+	createReservation         = `INSERT INTO reservations (provider, account_id, step, steps, status) VALUES ($1, $2, $3, $4, $5) RETURNING *`
```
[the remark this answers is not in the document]
lzap: Sure, added DEFAULT statement in SQL.

internal/dao/sqlx/reservation_dao.go
```diff
@@ -90,7 +104,12 @@ func (di *reservationDaoSqlx) CreateNoop(ctx context.Context, reservation *model
 	query := createReservation
 	stmt := di.create
 
-	err := stmt.GetContext(ctx, reservation, reservation.Provider, reservation.AccountID, reservation.Status)
+	err := stmt.GetContext(ctx, reservation,
+		reservation.Provider,
+		reservation.AccountID,
+		reservation.Step,
```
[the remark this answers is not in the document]
lzap: Sure.

internal/dao/sqlx/reservation_dao.go
lzap: I am not sure I understand here.

internal/jobs/launch_instance_aws.go
lzap: No, good point. Added a return statements at caller sites.

scripts/rest_examples/reservation-create-aws.http
lzap: Forgot this one, changed noop only. Fixed.

lzap:
Rebased.

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #189, opened 2022-08-28
Remove limitation on number of instance types

adiabramovitch:


internal/clients/ec2/ec2_client.go
```diff
@@ -116,7 +116,7 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
 
 	log.Debug().Msgf("Number of AWS EC2 instance types: %d", len(resp.InstanceTypes))
 	if len(resp.InstanceTypes) == 100 {
-		return nil, MoreThan100InstanceTypes
+		log.Debug().Msg("There are more instance types to be fetched")
```
ezr-ondrej, line 119: Shouldn't it be a warn level? :) (Nit, but would be safer to have these things visible)

ezr-ondrej (approved):
Thanks, just a nit

ezr-ondrej (approved):
Thanks @adiabramovitch ! 👍

PR #190, opened 2022-08-28
Use paginator to list instance types

adiabramovitch:
After reading some documentation it seems that we have two ways in order to present a complete instance types list. 

1. Using the token given in DescribeInstanceTypesOutput: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#DescribeInstanceTypesOutput and fetch the instance types until the token is nil.
2. Use NewDescribeInstanceTypesPaginator as implemented in this patch. 

It seems like both ways are based on the same idea but we can improve the performance of the second option of DescribeInstanceTypesPaginator using go routines as used here: https://github.com/firekillice/filebeat-kafka-mongo/blob/703d589a09cfdbfd7f84c1d990b50b6b7f62ac29/x-pack/libbeat/autodiscover/providers/aws/ec2/fetch.go#L83
**( It took about 5 seconds to get all the data after filtering, there are 557 types fetched from AWS )**

In my opinion, we have a few options: 
- use this patch and cache the list - the list is updated rarely so it is a great choice IMO
- fetching the pages using go routines (https://github.com/firekillice/filebeat-kafka-mongo/blob/703d589a09cfdbfd7f84c1d990b50b6b7f62ac29/x-pack/libbeat/autodiscover/providers/aws/ec2/fetch.go#L83).  
- Add filtering in the client-side - architecture, memory, etc, and use those attributes to filter the request to aws. (I think this should be done anyway)
- Add a "load more" button in the dropdown and only then re-request was for a page 

Would like to hear your opinion as well

### DIFF
--- a/internal/clients/ec2/ec2_client.go
+++ b/internal/clients/ec2/ec2_client.go
@@ -103,23 +103,19 @@ func (c *Client) CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (*Client, 
 	return newClient, nil
 }
 
-func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
-	log.Trace().Msg("Listing AWS EC2 instance types")
-	input := &ec2.DescribeInstanceTypesInput{
-		MaxResults: aws.Int32(100),
-	}
+func (c *Client) ListInstanceTypesWithPaginator() ([]types.InstanceTypeInfo, error) {
+	input := &ec2.DescribeInstanceTypesInput{MaxResults: aws.Int32(100)}
+	pag := ec2.NewDescribeInstanceTypesPaginator(c.ec2, input)
 
-	resp, err := c.ec2.DescribeInstanceTypes(c.context, input)
-	if err != nil {
-		return nil, fmt.Errorf("cannot list instance types: %w", err)
+	res := make([]types.InstanceTypeInfo, 0, 128)
+	for pag.HasMorePages() {
+		resp, err := pag.NextPage(c.context)
+		if err != nil {
+			return nil, fmt.Errorf("cannot list instance types: %w", err)
+		}
+		res = append(res, resp.InstanceTypes...)
 	}
-
-	log.Debug().Msgf("Number of AWS EC2 instance types: %d", len(resp.InstanceTypes))
-	if len(resp.InstanceTypes) == 100 {
-		log.Warn().Msg("There are more instance types to be fetched")
-	}
-
-	return resp.InstanceTypes, nil
+	return res, nil
 }
 
 func (c *Client) RunInstances(ctx context.Context, name string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error) {
--- a/internal/services/instance_types_service.go
+++ b/internal/services/instance_types_service.go
@@ -8,6 +8,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/clients/ec2"
 	sources "github.com/RHEnVision/provisioning-backend/internal/clients/sources"
 	"github.com/RHEnVision/provisioning-backend/internal/clients/sts"
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
 	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/render"
@@ -16,6 +17,7 @@ import (
 func ListInstanceTypes(w http.ResponseWriter, r *http.Request) {
 	sourceId := chi.URLParam(r, "ID")
 	ec2Client := ec2.NewEC2Client(r.Context())
+	logger := ctxval.Logger(r.Context())
 
 	sourcesClient, err := clients.GetSourcesClient(r.Context())
 	if err != nil {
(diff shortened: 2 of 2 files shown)


internal/clients/ec2/ec2_client.go
```diff
@@ -122,6 +122,25 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
...
+
+	res := make([]types.InstanceTypeInfo, 0)
+	for pag.HasMorePages() {
+		resp, err := pag.NextPage(c.context)
+		if err != nil {
+			return nil, fmt.Errorf("cannot list instance types: %w", err)
+		}
+		res = append(res, resp.InstanceTypes...)
+	}
+	log.Debug().Msgf("Number of AWS EC2 instance types: %d", len(res))
```
lzap, line 139: Can you make this a trace call and delete the one above? I think it's okay for a small function to leave one trace message, but two is a bit too much.

internal/clients/ec2/ec2_client.go
```diff
@@ -122,6 +122,25 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
...
+	return res, nil
+
+}
```
lzap, line 142: Extra line.

internal/clients/ec2/ec2_client.go
```diff
@@ -122,6 +122,25 @@ func (c *Client) ListInstanceTypes() ([]types.InstanceTypeInfo, error) {
 	return resp.InstanceTypes, nil
 }
 
+func (c *Client) ListInstanceTypesWithPaginator() ([]types.InstanceTypeInfo, error) {
+	log.Trace().Msg("Listing AWS EC2 instance types Using Paginator")
+
+	input := &ec2.DescribeInstanceTypesInput{MaxResults: aws.Int32(100)}
+	pag := ec2.NewDescribeInstanceTypesPaginator(c.ec2, input)
+
+	res := make([]types.InstanceTypeInfo, 0)
```
lzap, line 131: I would probably set the initial capacity to 64 or 128, almost always the runtime will have to reslice this from 8 to 16 to 32 and to 64. It smells like preliminary optimalization but it looks like amount of instances will be more than 100 all the time (it is currently 230 or so).

internal/services/instance_types_service.go
```diff
@@ -55,13 +56,14 @@ func ListInstanceTypes(w http.ResponseWriter, r *http.Request) {
...
 
-	res, err := newEC2Client.ListInstanceTypes()
+	res, err := newEC2Client.ListInstanceTypesWithPaginator()
 	if err != nil {
 		renderError(w, r, payloads.NewAWSError(r.Context(), "can't list EC2 instance types", err))
 		return
 	}
 
 	instances, err := ec2.NewInstanceTypes(r.Context(), res)
+	ctxval.Logger(r.Context()).Debug().Msgf("Number of AWS EC2 instance types after filtering: %d", len(*instances))
```
lzap, line 66: You have the same message in the function as well, so pick a place and log only once please. I would suggest a trace level as it's very detailed information. Also I suggest to introduce a `logger` variable here as someone else could copy this code in this function and we would end up calling `ctxval.Logger` multiple times (which is pretty expensive operation). The compiler will probably optimize single-use variable anyway automatically so there is nothing to worry about.

lzap (commented):
Yeah if AWS SDK provides such helper function, let's use it!

Looks great.

internal/services/instance_types_service.go
adiabramovitch: it is the number of instances **after filtering** ( filter unsupported instance types and create new ones if they support more than one arch)

lzap: Right, could you then just do a single trace here, count it before, after and make a single message containing both information? Let me explain why I worry about: ending up with massive amount of log messages typically lead to decreasing verbosity and that could be harmful for the future, so I think it is worth trying our best to keep the amount of information as small as possible.

lzap:
So to your performance concern, it's a valid one. Five seconds is really bad experience since this is literally the very first screen our users will see. This is why I am already working on improving and implementing cache for this particular call.

Let me explain why I think this cannot be improved. AWS API is designed in a way that no one can retrieve more than 100 records from any resource. To fetch more, they use the token technique which works simply by generating a unique (and pretty long) token you need to pass in order to fetch another page. Then they generate a new one. This design is specifically to enforce one thing: you cannot paralelize this, there is no way you could implement fetching that spawns 20 threads/processes to speed things up - they don't want you to do that actually. I believe the way you implemented this is the only way we can do this.

The code you refer to is a different case. I briefly read their code and it looks like they are using goroutines to fetch results from multiple regions or even multiple clients (customers). It is probably some batch operation where it definitely makes sense to use some kind of job queue, because AWS API enforces "fair use" - you cannot simply overload their servers with 1000 requests at one time. For this reason, their code have a queue and scheduler of some kind that performs all this pagination in 10 workers which ensures there are always 10 requests at one point to the AWS SDK.

This alone is also a problem for us - if we have many customers doing work via our software, at some point we might hit AWS API limits. I already thought about this and I believe this should be enough for us for the initial period, then we can probably contact Amazon and negotiate increase for our particular service: https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html

So correct me if I am wrong, but we cannot speed this up and I am fine with this patch, great work! @ezr-ondrej ?

lzap (approved):

ezr-ondrej (approved):
Great addition, thanks @adiabramovitch :+1:

I think that the goroutine solution might be slight improvement (not waitig for local operation before next request), but not much, but there might be other improvements as you've said already. I'm totatly ok trying this out first and see how it works and figure out solutions once trying out in ephemeral or even stage :+1:

lzap:
FYI my latency is 1600ms before the patch (100 records returned) and 3100ms after the patch (557 records, 6 requests).

One question tho, the filtering does not appear to work here:

```
11:45AM TRC Number of AWS EC2 instance types 557, Number after filtering: 566 
```

In my case, filtering added few records rather than filtered them out :-)

@amirfefer can you figure out some technique in the UI that would pre-fetch the list of instance types as soon as user opens up the wizard? So the moment they hit the dropdown it's already loaded up? We will cache the data, but we need to cache per-account per-region and expire the data resonably (e.g. after few hours) so loading will be still quite often thing.

adiabramovitch:
We are creating more instances if they support more than one architecture so it is possible but I would expect the number to decrease

lzap:
For the record, it looks like instance types are common across accounts, so we should be able to store them in a JSON with our project and then have a background job that would just fetch and update the list on a daily basis to keep it up to date when new are added (or we can simply do this manually too and deploy a new version).

https://repost.aws/questions/QUzbojJXafR8uL_M7LxUxQYw/are-ec-2-instance-types-unique-per-account

I think for M0 we can use what we have now.

PR #191, opened 2022-08-28
Add storage info to instance type model

adiabramovitch:


internal/clients/client_models.go
```diff
@@ -31,6 +31,9 @@ type InstanceType struct {
 	// The size of the memory, in MiB.
 	MemoryMiB int64 `json:"memory,omitempty"`
 
+	// The total size of the disks, in GB or 0 if EBS only.
+	Storage int64 `json:"storage,omitempty"`
```
lzap, line 35:
```suggestion
	StorageGiB `int64 `json:"storage,omitempty"`
```

adiabramovitch: Maybe StorageGB? just to be aligned with AWS?

lzap: Oh it's storage not memory, so GB it is, sorry my bad: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceStorageInfo.html

internal/clients/ec2/aws_supported_types.go
```diff
@@ -10,19 +10,24 @@ import (
...
 			list = append(list, clients.InstanceType{
 				Name:         string(types[i].InstanceType),
 				VCPUs:        types[i].VCpuInfo.DefaultVCpus,
 				Cores:        types[i].VCpuInfo.DefaultCores,
 				MemoryMiB:    *types[i].MemoryInfo.SizeInMiB,
 				Architecture: string(arch),
+				Storage:      storage,
 				Supported:    clients.IsSupported(string(types[i].InstanceType)),
 			})
 		}
```
lzap, line 33: There is a bug in this code, I suggest you to avoid variable completely, just set the storage flag before appending to list via temporary local one for the `InstanceType` type.

lzap (changes requested):

internal/clients/ec2/aws_supported_types.go
adiabramovitch: Thanks @lzap didn't notice :)

internal/clients/client_models.go
```diff
@@ -31,6 +31,9 @@ type InstanceType struct {
 	// The size of the memory, in MiB.
 	MemoryMiB int64 `json:"memory,omitempty"`
```
lzap, line 32:
```suggestion
	MemoryMiB int64 `json:"memory_mib,omitempty"`
```

internal/clients/client_models.go
```diff
@@ -31,6 +31,9 @@ type InstanceType struct {
...
 
+	// The total size of the disks, in GB or 0 if EBS only.
+	StorageGB int64 `json:"storage,omitempty"`
```
lzap, line 35:
```suggestion
	StorageGB int64 `json:"storage_gb,omitempty"`
```

lzap:
It's great, one last suggestion: do we want to also export the names with MB/MiB/GB so actually expose this? Because our OpenAPI (generated) don't have documentation per fields, I think this could be pretty useful.

api/openapi.gen.yaml
```diff
@@ -258,11 +258,14 @@ components:
...
           format: int64
           type: integer
         name:
           type: string
+        storage_gb:
+          format: int64
+          type: integer
         supported:
           type: boolean
         vcpus:
```
lzap, line 271: @RHEnVision/qe we are renaming memory to memory_mib to indicate that memory size is actually MiB not MB. The same for the new field storage (which is actually in GB not GiB).

lzap (approved):
TY

lzap:
For the record, we are playing around with the github's auto-merging feature, looks like it does not work well with group reviews. Anyway, please @RHEnVision/qe ack this PR since there is an OpenAPI change.

akhil-jha:
Testing with a new image.

akhil-jha (approved):
Got it. :) 
```
{'name': 'm1.large', 'vcpus': 2, 'cores': 2, 'memory_mib': 7680, 'storage_gb': 840, 'supported': True, 'architecture': 'x86_64'}
```

PR #193, opened 2022-08-30
Add region to instance type endpoint

adiabramovitch:
NEXT: to add the change in all the jobs as well

### DIFF
--- a/internal/clients/ec2/ec2_client.go
+++ b/internal/clients/ec2/ec2_client.go
@@ -24,7 +24,7 @@ type Client struct {
 	log     zerolog.Logger
 }
 
-func NewEC2Client(ctx context.Context) *Client {
+func NewEC2Client(ctx context.Context, params ...string) *Client {
 	c := Client{
 		context: ctx,
 		log:     ctxval.Logger(ctx).With().Str("client", "ec2").Logger(),
@@ -34,8 +34,13 @@ func NewEC2Client(ctx context.Context) *Client {
 	cache := aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(
 		config.AWS.Key, config.AWS.Secret, config.AWS.Session))
 
+	region := params[0]
+	if region == "" {
+		region = config.AWS.Region
+	}
+
 	c.ec2 = ec2.New(ec2.Options{
-		Region:      config.AWS.Region,
+		Region:      region,
 		Credentials: cache,
 	})
 
--- a/internal/services/instance_types_service.go
+++ b/internal/services/instance_types_service.go
@@ -16,7 +16,12 @@ import (
 
 func ListInstanceTypes(w http.ResponseWriter, r *http.Request) {
 	sourceId := chi.URLParam(r, "ID")
-	ec2Client := ec2.NewEC2Client(r.Context())
+
+	region := r.URL.Query().Get("region")
+	if region == "" {
+		renderError(w, r, payloads.NewNotFoundError(r.Context(), ec2.RegionNotFoundErr))
+	}
+	ec2Client := ec2.NewEC2Client(r.Context(), region)
 	logger := ctxval.Logger(r.Context())
 
 	sourcesClient, err := clients.GetSourcesClient(r.Context())
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -465,6 +465,15 @@
               "format": "int64",
               "type": "integer"
             }
+          },
+          {
+            "description": "List instance types of the specified region",
+            "in": "query",
+            "name": "region",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
           }
         ],
         "responses": {
--- a/api/openapi.gen.yaml
+++ b/api/openapi.gen.yaml
@@ -126,6 +126,12 @@ paths:
             format: int64
           required: true
           description: 'Source ID from Sources Database'
+        - in: query
+          name: region
+          schema:
+            type: string
+          required: true
+          description: List instance types of the specified region
         responses:
           '200':
             description: Return on success.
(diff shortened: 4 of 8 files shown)


internal/clients/ec2/ec2_client.go
```diff
@@ -24,7 +24,7 @@ type Client struct {
 	log     zerolog.Logger
 }
 
-func NewEC2Client(ctx context.Context) *Client {
+func NewEC2Client(ctx context.Context, region ...string) *Client {
```
ezr-ondrej, line 27: isn't this general args array?
I believe `params ...string` would describe the param bit better.
And then ofc somewhere `region = params[0]`

In case you wanted to add only region, I believe we should do `region string` (not an array).
Or am I missing something?

internal/clients/ec2/ec2_client.go
```diff
@@ -34,10 +34,17 @@ func NewEC2Client(ctx context.Context) *Client {
...
 
-	c.ec2 = ec2.New(ec2.Options{
-		Region:      config.AWS.Region,
-		Credentials: cache,
-	})
+	if region[0] == "" {
+		c.ec2 = ec2.New(ec2.Options{
+			Region:      config.AWS.Region,
+			Credentials: cache,
+		})
```
ezr-ondrej, line 41: having this on top of the service makes it bit clearer what are the default values (especially when we want to get rid of it in the future IMHO.
```suggestion
	if region == "" {
	  region = config.AWS.Region
    }
```

internal/services/instance_types_service.go
```diff
@@ -16,7 +16,12 @@ import (
 
 func ListInstanceTypes(w http.ResponseWriter, r *http.Request) {
 	sourceId := chi.URLParam(r, "ID")
-	ec2Client := ec2.NewEC2Client(r.Context())
+
+	region := r.URL.Query().Get("region")
+	if region == "" {
+		renderError(w, r, payloads.NewNotFoundError(r.Context(), ec2.RegionNotFoundErr))
```
ezr-ondrej, line 22: For easier transition, we could allow for empty region for the transition period (until the UI part would be merged).
But given we do not have UI pipeline yet, there is no need, just thinking out loud how to do this in the future ^_^
No action necessary here, I just needed to say this xD

ezr-ondrej (commented):
Nice and easy, thanks @adiabramovitch ! :+1: 
I have few code nitpicks and one rant, but nothing major :)

internal/clients/ec2/ec2_client.go
adiabramovitch: it is a variadic function, so we won't need to change all the calls in the jobs. 
in the next step (adding the region to the jobs) we can remove it and use a regular region parameter. 
I didn't think that using the optional struct https://bojanz.github.io/optional-parameters-go/ is good practice here

ezr-ondrej (approved):
This works :) thanks @adiabramovitch ! :+1:

ezr-ondrej:
Trying auto-merge, let's get the QE ack on this and see if that worked :)

internal/clients/ec2/ec2_client.go
```diff
@@ -34,8 +34,13 @@ func NewEC2Client(ctx context.Context) *Client {
 	cache := aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(
 		config.AWS.Key, config.AWS.Secret, config.AWS.Session))
 
+	region := params[0]
+	if region == "" {
+		region = config.AWS.Region
+	}
```
lzap, line 40: The OpenAPI says it is required, but here we do have a default value. So what was the intention?

lzap: Oh this is the client code, it is required in the service function, okay! I get it.

lzap (approved):
Nice, we literally just spoke with @avitova about the fact our instance type endpoint is not region aware yet. Well, not true anymore!

lzap (approved):
Tested and it works fine.

Response code: 200 (OK); Time: 2837ms (2 s 837 ms); Content length: 65978 bytes (65,98 kB)

akhil-jha:
Well, this does work, and I need to pass a region in the request. 

But...If I do this...
```
app.hms_provisioning.rest_client.default_api.get_instance_type_list(1, " ")
OR 
app.hms_provisioning.rest_client.default_api.get_instance_type_list(1, "qwertyuiop")
```
also gives me the entire instances list :(

akhil-jha (approved):
This does works. ACKing
Just needs a region validation.

lzap:
> Just needs a region validation.

Isn't 500 error when you provide an incorrect region a good enough? The error message should actually state it's not valid region I believe.

internal/clients/ec2/ec2_client.go
```diff
@@ -34,8 +34,13 @@ func NewEC2Client(ctx context.Context) *Client {
 	cache := aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(
 		config.AWS.Key, config.AWS.Secret, config.AWS.Session))
 
+	region := params[0]
```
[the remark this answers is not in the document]
lzap: If you want to hear my opinion on this ;-) ...

... I would generally avoid variadic functions for these cases (optional arguments). Just simply make a function named MakeEC2ClientWithRegion and keep the old one, call one from the other. This way the compiler will tell you what you have wrong exactly.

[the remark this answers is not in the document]
ezr-ondrej: we wanted this as intermediate solution not to fix everything in one PR, but we did not think it through properly.
I guess we can fix this all together in follow-up, @adiabramovitch wil you have capacity to follow-up soon, or shall we fix the panic?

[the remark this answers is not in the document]
adiabramovitch: Fixed it here: https://github.com/RHEnVision/provisioning-backend/pull/197
There is a follow-up PR to dro the variadic function after lzap's PR is merged: https://github.com/RHEnVision/provisioning-backend/pull/188

PR #194, opened 2022-08-31
Transparent HTTP gzip compression

lzap:
This patch adds configurable transparent gzip compression. Since all requests are proxied by 3scale, I do not want to mess around with modern types like brotli or zstd. Although modern browsers do support them, they are causing issues to proxies to this date. At least this is what I learned on the internet.

By default, compression is turned off. We need to test it on production first (stage is proxied differently), any only then we can enable it. Or simply enable it from the day one to see how it works, and in case of any problem, let's disable.

**Checklist**


cmd/pbapi/main.go
```diff
@@ -76,7 +77,24 @@ func main() {
 	r.Use(m.RequestNum)
 	r.Use(m.MetricsMiddleware)
 	r.Use(m.LoggerMiddleware(&log.Logger))
+
+	// Set Content-Type to JSON for chi and non-chi renderers
+	r.Use(m.ContentTypeMiddleware)
 	r.Use(render.SetContentType(render.ContentTypeJSON))
```
ezr-ondrej, line 83: Is this second content type setter also useful here now?

ezr-ondrej (commented):
So this is all or nothing compression? Might it be useful to compress some responses and not others? Just a question, I've got no idea :)

cmd/pbapi/main.go
lzap: It actually is, the `SetContentType` only works for Chi renderers, while the other is for non-Chi output. We have few of those, for example OpenAPI JSON and docs.

lzap:
> So this is all or nothing compression?

Yes, for the moment. If you have an example of endpoint that must not be compressed then we can add this conditionally. It's just a middleware after all.

Rebased.

cmd/pbapi/main.go
ezr-ondrej: oh, I thought the other will set it for all, not just non-chi, or won't it?

lzap: From what I've learned, we need both. Headers can be set multiple times, last wins. Now, when you use Chi the render framework has its content type that is stored in context and later set. It overwrites what you set from the normal middleware. The middleware on 83 actually does not set any content type, it stores content type in the context for later use by Chi.

Now, if you feel like setting content type globally to JSON is not what we want, I can move this only to routes which do not use chi, that would be OpenAPI at the moment. I think this might work, unless Chi sets the content type header too late.

lzap: Yeah that is the case, Chi sets content type the moment you start rendering, it overwrites any value that was there previously: https://github.com/go-chi/render/blob/master/responder.go#L51-L59

The actual set is done here: https://github.com/go-chi/render/blob/master/responder.go#L91-L107

I we have two options:

* Set a global content type just like in my proposal so other non-Chi routes like `/ready` or `/openapi` works fine.
* Delete it from here and move the middleware into all handlers.

lzap:
To test this use this curl command (identity 13/0003):

Chi-based route:

```
curl -X GET --location "http://localhost:8000/api/provisioning/v1/sources/1/instance_types?region=us-east-1" --http1.1 \
    -H "Content-Type: application/json" \
    -H "X-Rh-Identity: eyJpZGVudGl0eSI6IHsidHlwZSI6ICJVc2VyIiwgImFjY291bnRfbnVtYmVyIjoiMTMiLCAiaW50ZXJuYWwiOnsib3JnX2lkIjoiMDAwMDEzIn19fQo=" \
    -H "Accept-Encoding: gzip, deflate" -v 1>/dev/null
```

Plain route:

```
curl -X GET --location "http://localhost:8000/api/provisioning/v1/ready" --http1.1 \
    -H "Content-Type: application/json" \
    -H "X-Rh-Identity: eyJpZGVudGl0eSI6IHsidHlwZSI6ICJVc2VyIiwgImFjY291bnRfbnVtYmVyIjoiMTMiLCAiaW50ZXJuYWwiOnsib3JnX2lkIjoiMDAwMDEzIn19fQo=" \
    -H "Accept-Encoding: gzip, deflate" -v 1>/dev/null
```

ezr-ondrej (commented):
LGTM, tho shall we add this to known environment variables?
We should be able to set it on from ENV and thus possibly in clowder as well, right?

lzap:
Fixed conflicts, I actually deleted the global ContentType middleware, it is actually NOT needed at all, I think there was some bug or I mis-tested this with OpenAPI. Try now, it should work. I tested:

* OpenAPI JSON
* OpenAPI docs (does not load for me tho)
* /status
* /sources/1/instance_types

internal/clients/filter_instances.go
```diff
@@ -37,9 +37,9 @@ func init() {
 
 func MapArchitectures(_ context.Context, arch string) (ArchitectureType, error) {
 	switch {
-	case arch == "86_64_mac":
+	case arch == "x86_64_mac":
 		return ArchitectureTypeAppleX8664, nil
-	case arch == "arm_mac":
+	case arch == "arm64_mac":
 		return ArchitectureTypeAppleArm64, nil
 	case arch == "i386":
```
lzap, line 44: These two typos sneaked in which broke instance_types endpoint. Since this PR needs this for testing, I am fixing here.

internal/services/error_renderer.go
```diff
@@ -13,7 +13,7 @@ func writeBasicError(w http.ResponseWriter, r *http.Request, err error) {
 	if logger := ctxval.Logger(r.Context()); logger != nil {
 		logger.Error().Msgf("unable to render error %v", err)
 	}
-	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
```
lzap, line 16: Typo, will fix!

ezr-ondrej: shall we not render errors in json format?
I don't fully understand this change :thinking: is it relevant to the PR?

lzap: Yeah I just noticed this and fixed it too.

ezr-ondrej (approved):
Thanks @lzap ! Nice improvement especially for instance types :+1:

PR #195, opened 2022-08-31
ETag middleware with implementation for OpenAPI

lzap:
This new middleware can be used to cache any static or nearly static content. Everything works transparently, as long as a valid etag (string, hash, datetime, timestamp, anything unique) is provided and expiration time is set.

Caching works transparently in browsers, it will not work when API is hit from a curl utility of course.

The OpenAPI caching is just an example, this will be useful for instance types endpoint once we start embedding them into the application instead of dynamically loading them every time.

It can be also useful for any content that was cached in the future.

I tested it works fine with JSON spec, however, I am trying to open up redoc documentation and it simply renders JSON instead of giving me any HTML. What is the URL again?

http://localhost:8000/docs/openapi.json

### DIFF
--- a/api/embed.go
+++ b/api/embed.go
@@ -4,13 +4,31 @@ import (
 	_ "embed"
 	"fmt"
 	"net/http"
+	"time"
+
+	"github.com/RHEnVision/provisioning-backend/internal/middleware"
 )
 
 //go:embed openapi.gen.json
 var embeddedJSONSpec []byte
 
-func ServeOpenAPISpec(w http.ResponseWriter, _ *http.Request) {
-	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+var etag *middleware.ETag
+
+func init() {
+	var err error
+	etag, err = middleware.GenerateETagFromBuffer("json-spec", 30*time.Minute, embeddedJSONSpec)
+	if err != nil {
+		panic(err)
+	}
+}
+
+// ETagValue returns etag generated from the input file content with 30 minute expiration.
+func ETagValue() *middleware.ETag {
+	return etag
+}
+
+func ServeOpenAPISpec(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusOK)
 	_, err := w.Write(embeddedJSONSpec)
 	if err != nil {
--- a/internal/middleware/etag_middleware.go
+++ b/internal/middleware/etag_middleware.go
@@ -1,0 +1,83 @@
+package middleware
+
+import (
+	"errors"
+	"fmt"
+	"hash/crc64"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+)
+
+var InvalidETagErr = errors.New("empty etag provided")
+
+type ETag struct {
+	Name       string
+	Expiration time.Duration
+	Value      string
+	HashTime   time.Duration
+}
+
+type ETagValueFunc func() *ETag
+
+var etags = make([]*ETag, 0)
+
+func (etag *ETag) Header() string {
+	return fmt.Sprintf("\"pb-%s-%s\"", etag.Name, etag.Value)
+}
+
+func (etag *ETag) CacheControlHeader() string {
+	return fmt.Sprintf("max-age=%d", int(etag.Expiration.Seconds()))
+}
+
+func ETagMiddleware(etagFunc ETagValueFunc) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			etag := etagFunc()
+			if etag.Value == "" {
+				panic(InvalidETagErr)
+			}
+			cc := etag.CacheControlHeader()
+			w.Header().Set("ETag", etag.Header())
+			w.Header().Set("Cache-Control", cc)
+			logger := ctxval.Logger(r.Context()).With().Str("etag", etag.Value).Str("etag_name", etag.Name).Logger()
+			logger.Trace().Msgf("Returned etag with Cache-Control '%s'", cc)
+
+			if match := r.Header.Get("If-None-Match"); match != "" {
+				if strings.Contains(match, etag.Value) {
+					logger.Trace().Msgf("ETag cache hit")
+					w.WriteHeader(http.StatusNotModified)
+					return
+				}
+			}
+			next.ServeHTTP(w, r)
+		}
+		return http.HandlerFunc(fn)
+	}
+}
+
+func GenerateETagFromBuffer(name string, expiration time.Duration, buffers ...[]byte) (*ETag, error) {
+	start := time.Now()
+	hash := crc64.New(crc64.MakeTable(crc64.ECMA))
+	for _, buffer := range buffers {
+		_, err := hash.Write(buffer)
+		if err != nil {
+			return nil, fmt.Errorf("unable to generate etag from buffer: %w", err)
+		}
+	}
+	etag := &ETag{
+		Name:       name,
+		Expiration: expiration,
+		Value:      fmt.Sprintf("%x", hash.Sum64()),
+		HashTime:   time.Since(start),
+	}
+	etags = append(etags, etag)
+	return etag, nil
+}
+
+// AllETags returns all etags for diagnostic purposes
+func AllETags() []*ETag {
+	return etags
+}
--- a/internal/routes/all_routes.go
+++ b/internal/routes/all_routes.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/rs/zerolog/log"
+
 	"github.com/RHEnVision/provisioning-backend/api"
 	"github.com/RHEnVision/provisioning-backend/internal/middleware"
 	s "github.com/RHEnVision/provisioning-backend/internal/services"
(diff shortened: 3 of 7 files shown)


lzap:
Rebased not to use sha1 which is crypto-weak and liner complains about it. It now uses crc64 which is okay for etag.

api/embed.go
```diff
@@ -3,12 +3,22 @@ package api
...
 
+// ETagJSONSpec contains sha sum of the SPEC json.
+var ETagJSONSpec string
+
+func init() {
+	hash := crc64.New(crc64.MakeTable(crc64.ECMA))
+	hash.Write(embeddedJSONSpec)
+	ETagJSONSpec = fmt.Sprintf("%x", hash.Sum64())
+}
+
```
lzap, line 21: Calculates CRC64 hash of the whole JSON spec file during startup and that is returned to the browser.

internal/middleware/etag_middleware.go
```diff
@@ -0,0 +1,39 @@
...
+					logger.Trace().Msgf("ETag cache hit")
+					w.WriteHeader(http.StatusNotModified)
+					return
+				}
+			}
+			next.ServeHTTP(w, r)
+		}
+		return http.HandlerFunc(fn)
+	}
+}
```
lzap, line 39: The middleware responsible for setting the ETag and Cache-Control and respecting the If-None-Match header from the client to quickly return NotModified if browser already has that specific version.

internal/routes/all_routes.go
```diff
@@ -23,15 +24,21 @@ func SetupRoutes(r *chi.Mux) {
...
 }
 
 func apiRouter() http.Handler {
 	r := chi.NewRouter()
 
-	r.Get("/openapi.json", api.ServeOpenAPISpec)
+	r.Route("/openapi.json", func(r chi.Router) {
+		r.Use(middleware.ETagMiddleware(15*time.Minute, "json-spec", api.ETagJSONSpec))
+		r.Get("/", api.ServeOpenAPISpec)
+	})
```
lzap, line 41: And this is how to use the middleware, I set the timeout to 15 minutes. I use it both for JSON spec and documentation.

It's not super useful for OpenAPI, granted, but it's a good example. A good candidate will definitely be instance-types endpoint.

internal/routes/all_routes.go
```diff
@@ -23,15 +24,21 @@ func SetupRoutes(r *chi.Mux) {
 	r.Get("/ping", s.StatusService)
 	r.Route("/docs", func(r chi.Router) {
 		r.Use(redocMiddleware)
-		r.Get("/openapi.json", api.ServeOpenAPISpec)
+		r.Route("/openapi.json", func(r chi.Router) {
+			r.Use(middleware.ETagMiddleware(15*time.Minute, "json-spec", api.ETagJSONSpec))
```
ezr-ondrej, line 28: I've left the review here last time (sorry!)
I was trying to think of a way how we could introduce the middleware globally, but enable it per endpoint, but I guess this is the best way.

The only thing I've came up with, is to include the the tag in the return statement of the handler and middleware would be watching out for that, but it sounds bit ugly as we would change the Handler func definition just for that :thinking:

internal/middleware/etag_middleware.go
```diff
@@ -0,0 +1,39 @@
...
+
+var InvalidETagErr = errors.New("empty etag provided")
+
+func ETagMiddleware(expires time.Duration, name, etag string) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			if etag == "" {
+				panic(InvalidETagErr)
+			}
+			etagValue := fmt.Sprintf("\"hms-%s-%s\"", name, etag)
```
ezr-ondrej, line 21: Why do we want to have the `hms` prefix? HMS is internal name and only RH, so I'd rather not expose it (even in minor stuff user really don't see). Can we have `provisioning`/`prov` if we want prefix it?

internal/middleware/etag_middleware.go
```diff
@@ -0,0 +1,39 @@
...
+			controlValue := fmt.Sprintf("max-age=%d", int(expires.Seconds()))
+			w.Header().Set("ETag", etagValue)
+			w.Header().Set("Cache-Control", controlValue)
+			logger := ctxval.Logger(r.Context()).With().Str("etag", etag).Str("etag_name", name).Logger()
+			logger.Trace().Msgf("Returned etag with Cache-Control '%s'", controlValue)
+
+			if match := r.Header.Get("If-None-Match"); match != "" {
```
ezr-ondrej, line 28: I've got no idea how this is supposed to be implemented, so it's hard to review this part, why don't we use some external implementation?

ezr-ondrej (commented):
few comments, but more of general thoughts, feel free to disagree, I'd be ok to merge as is :)

internal/middleware/etag_middleware.go
adiabramovitch: Yes, I am having the same issue here, can you explain in a few words?

internal/routes/all_routes.go
adiabramovitch: So for now we are  caheching only in the openapi.json? and we are supposed to use this middleware in any endpoint which holds static information?

api/embed.go
```diff
@@ -3,12 +3,22 @@ package api
...
 	"fmt"
+	"hash/crc64"
 	"net/http"
 )
 
 //go:embed openapi.gen.json
 var embeddedJSONSpec []byte
 
+// ETagJSONSpec contains sha sum of the SPEC json.
+var ETagJSONSpec string
```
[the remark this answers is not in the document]
lzap: Absolutely, now that I think about it, I refactored this code into a helper function in the middleware.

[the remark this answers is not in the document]
lzap: Also added a test :)

internal/routes/all_routes.go
lzap: Yeah this is the cleanest approach. Frankly, I don't see us using this for anything else than instance types and OpenAPI.

@adiabramovitch yes etag only works for content that does not change often (e.g. photos uploaded to a site). In our case instance types.

ETag is complementary to normal caching, it is not here to replace it at all. While it is possible to implement ETag with cache in tandem (could be timestamp of when some item was pulled from cache), I don't think it is worth the effort for our use case. If we cache something, let's just return it without ETag since it will be pretty fast anyway.

internal/middleware/etag_middleware.go
lzap: Sure, I was aiming for something short, it's not mandatory it's recommended to use a prefix to differentiate own content (we do have a name). Changed this to "pb".

internal/middleware/etag_middleware.go
lzap: While there might be libraries providing such middleware, it is so trivial that I think it is simply worth having it in our codebase.

From Google: https://devcenter.heroku.com/articles/increasing-application-performance-with-http-cache-headers

lzap:
Rebased.

internal/routes/all_routes.go
ezr-ondrej: Well it will be fast, but the trafic still needs to happen, right, so if we are caching it, it's probably good idea to etag it as well IMHO. 200KB traffic is IMHO quite some safe

internal/middleware/etag_middleware.go
```diff
@@ -0,0 +1,49 @@
...
+			}
+			etagValue := fmt.Sprintf("\"pb-%s-%s\"", name, etag)
+			controlValue := fmt.Sprintf("max-age=%d", int(expires.Seconds()))
+			w.Header().Set("ETag", etagValue)
+			w.Header().Set("Cache-Control", controlValue)
+			logger := ctxval.Logger(r.Context()).With().Str("etag", etag).Str("etag_name", name).Logger()
+			logger.Trace().Msgf("Returned etag with Cache-Control '%s'", controlValue)
+
+			if match := r.Header.Get("If-None-Match"); match != "" {
+				if strings.Contains(match, etag) {
```
ezr-ondrej, line 30: Shouldn't this be exact match? I know it's a minor thing, but I'm interested, why? Some whitespace issues?

ezr-ondrej (approved):
The helper is nice :+1: 
Do you think we should pull all the ETag constants into routes tho?
I feel kinda unease to have them scattered all around the code. Not a big deal tho :shrug: 
Apart of that and nit on the match comparison, I'm good here :+1:

internal/middleware/etag_middleware.go
lzap: _The If-None-Match HTTP request header makes the request conditional. For [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) methods, the server will return the requested resource, with a [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) status, only if it doesn't have an [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) matching the given ones. For other methods, the request will be processed only if the eventually existing resource's [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) doesn't match any of the values listed._

From: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match

lzap:
> I feel kinda unease to have them scattered all around the code.

Hmm, let me create a type with a function type to solve this. It won't be perfect since hash calculation will always be done somewhere else, but this will be vastly better.

Rebased, how about this?

api/embed.go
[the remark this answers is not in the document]
lzap: Ha forgot to unexport this variable, not needed anymore since there is a function now.

internal/middleware/etag_middleware.go
```diff
@@ -0,0 +1,67 @@
...
+
+var InvalidETagErr = errors.New("empty etag provided")
+
+type ETag struct {
+	Name       string
+	Expiration time.Duration
+	Value      string
+}
+
+type ETagValueFunc func() ETag
```
lzap, line 22: This is the function which returns the type above which contains all information for etag to be generated from.

internal/middleware/etag_middleware.go
ezr-ondrej: ILT, thanks! (this is exactly what I meant by I don't fully understand how it should work, so it's hard to review) :)

ezr-ondrej (approved):
Nice interface! :orange_heart:
Great, I'm all up for it, we just need a rebase now.

lzap:
I rebased, I also added one more thing - duration of hash calculation is now logged via debug level. I expect it to be slower for Azure and other clouds, so I want to keep an eye on that to prevent slowing down of development cycles (app restarts). If this turns to be slow, let's make a setting for etag. Will be probably good tho.

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #197, opened 2022-09-01
Add NewEC2ClientWithRegion

adiabramovitch:


internal/clients/ec2/ec2_client.go
```diff
@@ -34,11 +34,13 @@ func NewEC2Client(ctx context.Context, params ...string) *Client {
 	cache := aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(
 		config.AWS.Key, config.AWS.Secret, config.AWS.Session))
 
-	region := params[0]
-	if region == "" {
-		region = config.AWS.Region
+	region := config.AWS.Region
+	if params != nil {
```
ezr-ondrej, line 38:
```suggestion
	if params != nil && params[0] != "" {
```

ezr-ondrej (commented):
I believe we can simplify the condition.

internal/clients/ec2/ec2_client.go
ezr-ondrej: And then remove the fallback ofc...

internal/clients/ec2/ec2_client.go
```diff
@@ -34,11 +34,13 @@ func NewEC2Client(ctx context.Context, params ...string) *Client {
...
+		region = params[0]
+		if region == "" {
+			region = config.AWS.Region
+		}
 	}
```
lzap, line 40: This does not fix the bug, variable `params` is never nil, but it can be empty in two out of three cases, therefore `params[0]` panics with out of bounds.

I highly recommend to remove variadic function in favour of two functions: `NewEC2ClientWithRegion` and `NewEC2Client` which will call the former function with config setting. This is the safest option.

lzap (changes requested):
This does not fixes the problem, here is a proof:

https://go.dev/play/p/NaFGZI10qb9

internal/clients/ec2/ec2_client.go
lzap: This does not fix it either, guys, I highly recommend not to use variadic functions. See below :-)

internal/clients/ec2/ec2_client.go
adiabramovitch: I am not sure why we need two functions if it will be changed after your PR is merged.
we need to change the jobs as well, each job should hold the region.

adiabramovitch: But I don't mind to change it if you think it is a better option

lzap (approved):
Perfect, this is what I believe is Idiomatic Go, although I haven't seen an example in Effective Go book. I strongly believe we should avoid variadic functions unless absolutely necessary. Compile-time checking is always better than runtime-checking.

PR #200, opened 2022-09-01
Simplify EC2 client initialization

ezr-ondrej:
Initiate EC2 client directly with assumed credentials through sts, 
without exposing sts hastle to consumers, we never use unassumed direct 
clients for ec2, so it is useless to keep the ability of creating those.

I've tried to keep the changes minimal:

- pulled contents of sts client into `getStsAssumedCredentials` that now lives in `ec2_client`
- dropped EC2 client initialization, as that has never actualy been used
- pulled `CreateEC2ClientFromConfig` (the real initialization of EC2 client) into `newAssumedEC2ClientWithRegion`
- made the region work for assumed client (and thus actually have effect)

I believe there is room for optimalization of both `newAssumedEC2ClientWithRegion` and `getStsAssumedCredentials`, but I think it's out of scope and is much easier once this will get in.
Also there are still some direct imports of the ec2 client for getting the types and errors, which shall be avoided IMHO, but again kept for simplification. Will try to do a follow-up


internal/clients/clients_interfaces.go
```diff
@@ -30,19 +29,12 @@ type ImageBuilder interface {
 	Ready(ctx context.Context) error
 }
 
-var GetEC2Client func(ctx context.Context) (EC2, error)
-var GetEC2ClientWithRegion func(ctx context.Context, region string) (EC2, error)
+var GetAssumedEC2Client func(ctx context.Context, arn string) (EC2, error)
+var GetAssumedEC2ClientWithRegion func(ctx context.Context, arn string, region string) (EC2, error)
```
lzap, line 33: One of these functions can be simply defined in the interface package as it will always be the same, implementations need only to redefine one of them:

```go
func GetAssumedEC2Client func(ctx context.Context, arn string) (EC2, error) {
  return GetAssumedEC2ClientWithRegion(ctx, arn, config.AWS.Region)
}
```

I know it was me who sort of pushed @adiabramovitch into this, with factory functions it do look little bit weird. I am good with this (definitely better than variadic function), but one solution to this could be simply having just a single function and when region is set to empty string, then config default would be used. 

The reason why I am bringing it here is that @avitova will soon need also non-assumed ec2 client for her instance type preloading work which can be done directly on our service account. This could possibly double the amount of factory functions.

I also suggest to clean naming a bit, to be consistent across clouds I propose the following terms:

* GetServiceClient - for our service (direct) account
* GetCustomerClient - for the customer's (assumed) client

internal/clients/clients_interfaces.go
```diff
@@ -30,19 +29,12 @@ type ImageBuilder interface {
...
-	CreateEC2ClientFromConfig(crd *stsTypes.Credentials) (EC2, error)
 	ListInstanceTypesWithPaginator() ([]types.InstanceTypeInfo, error)
 	RunInstances(ctx context.Context, name *string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error)
 }
-
-var GetSTSClient func(ctx context.Context) (STS, error)
-
-type STS interface {
-	AssumeRole(arn string) (*stsTypes.Credentials, error)
-}
```
lzap, line 48: Now, we're talking! :-)

internal/clients/http/ec2/ec2_client.go
```diff
@@ -26,31 +27,62 @@ type EC2Client struct {
...
+		return nil, fmt.Errorf("cannot create a new ec2 config: %w", err)
 	}
 
-	log.Trace().Msg("Creating new EC2 client")
-	cache := aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(
-		config.AWS.Key, config.AWS.Secret, config.AWS.Session))
+	return &EC2Client{
+		ec2:     ec2.NewFromConfig(newCfg),
+		context: ctx,
+		log:     logger.With().Str("client", "ec2").Logger(),
```
lzap, line 56: We should probably leave a field in the logger, something like "client_type" with values "service" or "customer" to differentiate between the two.

lzap (changes requested):
This is a great chance to introduce both assumed and non assumed clients. But if you feel like doing it in another PR its fine.

Just move the function call into the interface file because it can be generic.

internal/clients/http/ec2/ec2_client.go
ezr-ondrej: if we have both then yes, but I'm removing the Service one here, we can bring it back, but given both were kinda incorect, I'd rather do one right in this PR and add the Service one later. I can rename the Assumed to Customer tho, thats a good idea anyway :)

lzap:
Lint failed.

ezr-ondrej:
> This is a great chance to introduce both assumed and non assumed clients. But if you feel like doing it in another PR its fine.

I'd really rather not... btw there is a food for thought: if we have service role we can assume, we do not need to have Customer/Service clients, but we would do with one. There would be unecessary assume, but less code ^_^
These and other discussions belong do different PR IMHO, this is doing quite a lot already... But I'll be happy to follow this up (by using region in all the calls and removing the double method e.g.)

lzap (approved):
Ok I haven't tested but looks good. Thanks.

PR #201, opened 2022-09-02
Region aware reservation endpoint 

ezr-ondrej:
Reservation endpoint was missing a region to deploy into.
We are using one region for now, but we need this for future use.

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -27,13 +27,18 @@ type EC2Client struct {
 }
 
 func init() {
-	clients.GetCustomerEC2ClientWithRegion = newAssumedEC2ClientWithRegion
+	clients.GetCustomerEC2Client = newAssumedEC2ClientWithRegion
 }
 
 func newAssumedEC2ClientWithRegion(ctx context.Context, arn string, region string) (clients.EC2, error) {
 	logger := ctxval.Logger(ctx)
 
-	creds, err := getStsAssumedCredentials(ctx, arn)
+	if region == "" {
+		logger.Warn().Msg("No region passed, using us-east-1")
+		region = "us-east-1"
+	}
+
+	creds, err := getStsAssumedCredentials(ctx, arn, region)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


internal/clients/http/ec2/ec2_client.go
```diff
@@ -55,7 +55,7 @@ func newAssumedEC2ClientWithRegion(ctx context.Context, arn string, region strin
 func getStsAssumedCredentials(ctx context.Context, arn string) (*stsTypes.Credentials, error) {
 	logger := ctxval.Logger(ctx)
 
-	cfg, err := awsCfg.LoadDefaultConfig(ctx, awsCfg.WithRegion(config.AWS.Region),
+	cfg, err := awsCfg.LoadDefaultConfig(ctx, awsCfg.WithRegion("us-east-1"),
```
lzap, line 58: Wait why you hardcode it here when there is the region argument in the function declaration?

internal/jobs/launch_instance_aws.go
```diff
@@ -20,6 +20,9 @@ type LaunchInstanceAWSTaskArgs struct {
 	// Associated account
 	AccountID int64 `json:"account_id"`
 
+	// Region to provision the instances into
+	Region string `json:"region""`
+
```
lzap, line 25: You have a double quote there.

internal/payloads/reservation_payload.go
```diff
@@ -48,6 +48,9 @@ type AWSReservationResponsePayload struct {
 	// Source ID.
 	SourceID string `json:"source_id"`
 
+	//AWS region.
+	Region string `json:"region"`
+
 	//AWS Instance type.
 	InstanceType string `json:"instance_type"`
```
lzap, line 55: Can you fix all spaces in this file, or I won't sleep tonight ;-)

internal/services/aws_reservation_service.go
```diff
@@ -41,7 +41,13 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if payload.Region == "" {
+		logger.Warn().Msg("No region passed to the reservation endpoint using us-east now, but this will be failure in the future.")
```
lzap, line 45: I would not do in-code commitments, 10 years later we could still find this ;-)

    logger.Warn().Msg("No region passed, using us-east-1")

Also: We don't do full stops in logs.

Finally, I would suggest to move this logic into the client function including the warning, so all possible call behavior is the same.

internal/clients/http/ec2/ec2_client.go
ezr-ondrej: Yeah, this was first take, sorry I assume, that this config can be on global region, because role assuming is region agnostic, but I'll use the region from argument for now :)

internal/jobs/launch_instance_aws.go
ezr-ondrej: oh

lzap:
Looks great. Mind also adding service client too? It can be reused from the existing (customer's) client code.

internal/clients/http/ec2/ec2_client.go
ezr-ondrej: fixed with todo, I believe there is room for improvement in the role assume.

internal/services/aws_reservation_service.go
ezr-ondrej: Done :)

ezr-ondrej:
> Looks great. Mind also adding service client too?

I'd love to have a discussion on the topic first if we even want to do that, if we don't want to just create a service role and assume role for loading the instances as well. One advantage of this approach on top of using the same code is permission separation on the service account. The role will have only permissions we need for the service and the service account will still only have permissions to assume roles.

lzap (approved):
Sure no problem, scheduled merge.

PR #203, opened 2022-09-02
Upgrade to Go 1.18

lzap:
The following is an upgrade to Go runtime 1.18 because Azure SDK requires this version of Go currently. Unfortunately, the only way to workaround this is to put all Azure libraries into ./vendor directory and use go.mod's "replace" feature to insist on these versions. Given amount of libraries the Azure SDK consist of, the better alternative is to simply upgrade to Go 1.18 even when this is not yet part of EL8 runtime stack (go-toolset).

This patch adds additional Dockerfile and a symlink which we can easily use to switch between Go upstream and Red Hat Go Toolset Runtime. It is expected that the version 1.18 will be released soon (currently 1.17 was released a month ago).

I tested building via podman.

@ezr-ondrej can you get an approval from SRE for this?


lzap:
It builds fine locally, but on our CI we are getting:

```
15:02:22 time="2022-09-02T13:02:22Z" level=warning msg="failed, retrying in 2s ... (1/3). Error: initializing source docker://golang:1.18-bullseye: pinging container registry registry-1.docker.io: Get \"[https://registry-1.docker.io/v2/\](https://registry-1.docker.io/v2//)": dial tcp 127.0.0.0:443: connect: network is unreachable"
```

Is it possible that network is blocked on purpose, @RHEnVision/qe ? Can we somehow whitelist the Go image? I am getting this feeling that my patch might not actually work :-)

lzap:
Agreed that it is very likely we will have 1.18 Go Toolset in RHEL8/UBI8 at that time.

I rebased adding additional information about the source of the golang:1.18 from quay. It looks like a reasonable base image to use, one big advantage is that this is only compile-time container.

lzap:
For the record, from the `Dockerfile`:

```
# We cannot use the official golang images from docker.io because it's blocked
# on our CI. Instead, we use image maintained by Quay team which is updated every
# few hours automatically from theird GitHub. For more info:
#
#  https://github.com/quay/claircore/actions/workflows/golang-image.yml
#  https://github.com/quay/claircore/blob/main/.github/workflows/golang-image.yml
```

ezr-ondrej (approved):
Works like a charm, thanks @lzap ! :+1:

PR #204, opened 2022-09-02
Fix integration config loading, introduced NoError in tests

lzap:
Fixes configuration of tests, add missing ENV variable and adds NoError in all DAO tests.

Three commits, see commit messages for more details.

ACTION REQUIRED: This patch also fixes configuration loading, it now properly finds the path to the config file. Please move your `cmd/dao_test/local.yaml` to `configs/local.integration.yaml` to get it loaded properly.

### DIFF
--- a/internal/config/5_normal_config.go
+++ b/internal/config/5_normal_config.go
@@ -4,38 +4,14 @@
 package config
 
 import (
-	"fmt"
-
 	"github.com/RHEnVision/provisioning-backend/internal/config/parser"
-	clowder "github.com/redhatinsights/app-common-go/pkg/api/v1"
 )
 
 func init() {
-	if clowder.IsClowderEnabled() {
-		cfg := clowder.LoadedConfig
-		parser.Viper.Set("featureFlags.environment", "production")
-		parser.Viper.Set("database.host", cfg.Database.Hostname)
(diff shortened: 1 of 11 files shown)


avitova (approved):
LGTM

internal/config/5_normal_config.go
```diff
@@ -4,38 +4,14 @@
...
-		// load from working directory for integration tests
-		parser.Viper.AddConfigPath(".")
-		parser.Viper.SetConfigName("local")
+	parser.Viper.SetDefault("featureFlags.environment", "development")
 
-		err := parser.Viper.ReadInConfig()
-		if err != nil {
-			fmt.Println("Could not read configs/local.yaml, continuing with defaults")
-		}
-	}
```
lzap, line 40: FYI this was the reason why it did not work - this file is now excluded when integration build tag is set. So I moved the logic into the Initialize() function where it probably belongs.

ezr-ondrej (approved):
Thanks @lzap ! :)

internal/config/5_normal_config.go
ezr-ondrej: I'd agree, we (read I) went overboard with the splitting the config into files.

PR #207, opened 2022-09-04
HMSPROV-239 - add reservation status endpoint

amirfefer:


internal/routes/all_routes.go
```diff
@@ -64,6 +64,9 @@ func apiRouter() http.Handler {
 			r.Route("/{type}", func(r chi.Router) {
 				r.Post("/", s.CreateReservation)
 			})
+			r.Route("/{ID}/status", func(r chi.Router) {
```
amirfefer, line 67: I had to add the `status` suffix due to `reservations/{ID}` collides with `reservations{type}`

ezr-ondrej: I don't mind the status suffix personally :)

ezr-ondrej, line 67: Can't we directly do:
```suggestion
			r.Get("/{ID}/status", s.GetReservation)
```

internal/services/reservations_service.go
```diff
@@ -52,3 +52,32 @@ func ListReservations(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 }
+
+func GetReservation(w http.ResponseWriter, r *http.Request) {
```
ezr-ondrej, line 56: What about naming it with the status? smthing like `GetReservaionStatus`

ezr-ondrej (commented):
Few minor improvements, but apart of it it looks great :+1:
Could you pls also add a test for this and if the endpoint is ready and working, should we also add it into OpenAPI spec?

internal/routes/all_routes.go
lzap: This smells like a dirty trick, let's should push for a clean REST design by refactoring the existing endpoint which is in our way.

We briefly discussed use of URLs or URL params for filtering and I think this case shows that using URL params for filtering is a good approach. I would suggest to refactor `/{type}` to be `/?type=XXX` URL param so we can use the RESTy `/{ID}` for status.

lzap:
I know what I am asking is more work, but once we release V1 API there is no easy way to fix things, so I encourage anyone for refactoring rather now, with unstable API, then never :-)

internal/routes/all_routes.go
ezr-ondrej: @amirfefer I also belive the putting this route first would fix the issue if we allow get only there, as I'm suggesting below, it should not match the POST requests ;) which won't be more work and will be tidier :)

lzap: Please remove the '/status' and simply use 'GET /{ID}'

lzap (changes requested):
I just realized that there is no collision!

* GET '/reservations' returns all reservations
* POST '/reservations/{TYPE}' creates a reservation
* GET '/reservations/{ID}' returns a single reservation

Let's not introduce any non-REST '/status' subpaths please.

I think it is worth fixing reservation creation, the `{TYPE}` should not be used at all, I think it should be in the payload as a field rather than a URL element. But that's another story.

ezr-ondrej:
@amirfefer what is the status here? can we get this done? :)

internal/services/reservations_service.go
amirfefer: Basically, this function returns a reservation by id, not just the status

amirfefer:
@ezr-ondrej, I rebased, remove the `status` suffix, added this endpoint to spec, and verified that it works with our UI :)

internal/payloads/reservation_payload.go
```diff
@@ -140,25 +144,29 @@ func NewNoopReservationResponse(reservation *models.NoopReservation) render.Rend
...
+		Provider:   int(reservation.Provider),
+		CreatedAt:  reservation.CreatedAt,
+		FinishedAt: finishedAt,
+		Status:     reservation.Status,
+		Success:    success,
+		Steps:      reservation.Steps,
+		Step:       reservation.Step,
+		Error:      reservation.Error,
+	}
+}
```
lzap, line 172: When I look on other payload files, we actually never do copying of payloads in any other function than NewXXXListResponse, so this helper function might not be reused ever. But it's ok.

internal/services/reservations_service.go
lzap: No status please, kill status, let's be REST :-)

lzap (approved):
Perfect patch!

ezr-ondrej (approved):
Nice addition :+1:

akhil-jha (approved):

PR #208, opened 2022-09-05
Refactor instance types for Azure, types for Azure

lzap:
WIP: Initial look how Azure instance types looks like. There is no client interface yet for both http client and registered/available types or even loading of the YAML data. Looking for feedback.

Take a look on the two YAML files, one is a "registry" of all possible types (even deprecated ones), the other is availability info across regions/zones. Both are pretty big files (500kB and 144kB) so github wont' show them by default.

For the record, this is the API this patch uses, it is vaguely documented so I hope I got it right: https://docs.microsoft.com/en-us/rest/api/compute/resource-skus/list?tabs=HTTP


lzap:
> A thought after skimming through the instance types: there is many versions of the same, do we know if we need (or want) to include all the versions or could we get away with using just the last version?

Do you mean _v2 _v3 etc? They still offer most of these, for example:

```
    - Standard_D16_v3
    - Standard_D16_v4
    - Standard_D16_v5
```

These are still all valid options. In this case, v3 is supported only on gen1 VM type, v4 and v5 supports both. The original and v2 are deprecated tho, so we filter it out.

lzap:
So I have split the YAML into smaller files, now it's much more manageable, specifically github reviews.

It looks like the data is correct, e.g. if you open up germany west central zeone 3, you only have these E and L powerful machines available. You will see some of them being restricted because of quotas on your account, you can increase quotas per account. Everything else seems to be not available in that zone.

lzap:
So I am adding ability to also print loaded data from YAML, so we can easily check:

```
...
Standard_E112iads_v5 | Arch: x86_64 | vCPUs: 112 | Cores: 56 | Memory: 672000 MiB | Disk: 3891 GB | Azure Gen: V1 V2 | Supported: No
Standard_L8s_v3 | Arch: x86_64 | vCPUs: 8 | Cores: 4 | Memory: 64000 MiB | Disk: 81 GB | Azure Gen: V1 V2 | Supported: No
Standard_L64as_v3 | Arch: x86_64 | vCPUs: 64 | Cores: 32 | Memory: 512000 MiB | Disk: 655 GB | Azure Gen: V1 V2 | Supported: No
Standard_NC24r | Arch: x86_64 | vCPUs: 24 | Cores: 24 | Memory: 224000 MiB | Disk: 1474 GB | Azure Gen: V1 | Supported: No
Standard_D8pls_v5 | Arch: arm64 | vCPUs: 8 | Cores: 8 | Memory: 16000 MiB | Disk: 0 GB | Azure Gen: V2 | Supported: No
Total: 713
```

The same for regions/zones.

ezr-ondrej:
> These are still all valid options. In this case, v3 is supported only on gen1 VM type, v4 and v5 supports both. The original and v2 are deprecated tho, so we filter it out.

Right, so it's valid Azure semantic and we need to (should) support it... got it.

lzap:
So I pushed, opinions? I am working on linter etc.

The interface is WIP so far, the idea is to have a common API to get the instance data:

* Give me details for instance named XY
* Give me list of instances for a region/zone

But I want to keep this separate from this PR as changes of the instance type endpoint will be needed (new provider, refactoring of AWS a bit).

internal/clients/client_models.go
```diff
@@ -16,27 +16,3 @@ type Source struct {
...
-
-	// The total size of the disks, in GB or 0 if EBS only.
-	StorageGB int64 `json:"storage_gb"`
-
-	// Does the instance type supports RHEL
-	Supported bool `json:"supported,omitempty"`
-
-	// Instance type's Architecture: i386, arm64, x86_64
-	Architecture string `json:"architecture,omitempty"`
-}
```
lzap, line 42: I moved this into a separate file since it grew substantially.

internal/clients/filter_instances.go
```diff
@@ -4,24 +4,14 @@ import (
...
-type ArchitectureType string
-
-const (
-	ArchitectureTypeI386       ArchitectureType = "i386"
-	ArchitectureTypeX8664      ArchitectureType = "x86_64"
-	ArchitectureTypeArm64      ArchitectureType = "arm64"
-	ArchitectureTypeAppleX8664 ArchitectureType = "apple-x86_64"
-	ArchitectureTypeAppleArm64 ArchitectureType = "apple-arm64"
-)
-
```
lzap, line 24: Moved to the newly created file, I think this belongs to instance type.

internal/clients/filter_instances_test.go
```diff
@@ -28,10 +28,19 @@ func TestMapArchitectures(t *testing.T) {
...
 }
 
+func TestAzureMapArchitectures(t *testing.T) {
+	result, err := clients.MapArchitectures(context.Background(), "x64")
+	assert.Nil(t, err)
+	assert.Equal(t, clients.ArchitectureTypeX8664, result)
+	result, err = clients.MapArchitectures(context.Background(), "arm64")
+	assert.Nil(t, err)
+	assert.Equal(t, clients.ArchitectureTypeArm64, result)
+}
```
lzap, line 42: Azure currently have only x86_64 and arm64 is in beta, at least I don't see any instance types reported with this architecture.

lzap:
Fixed tests.

lzap:
This is now ready for review, I suggest to filter out YAML files from the review tab otherwise it's a pain. But once we commit the files, changes will be typically small (new type added/removed) so in that regard reviews will be okayish.

@avitova this is a rough first version on what we want to do for EC2 as well. My plan is to refactor instance type service so it will use the new data for Azure, so you can then do the same for AWS. The idea is to reuse the new types `clients.RegisteredInstanceTypes` and `clients.RegionalTypeAvailability` to store the same data, but from AWS. Both `Load` and `Save` functions will work with AWS data too, I suggest to also use separate directory: `intenral/clients/http/ec2/types` and here store both types data in `types.yaml` and region/zone data in `availability/` subdirectory the same I do it in this patch.

FYI we don't have availability zones implemented yet, we only support regions. This needs to be added, the AWS API supports that: https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instance-type-offerings.html

lzap:
Improved the CLI and added filtering too:

```
% go run cmd/types_azure/main.go -help

  -all
        print everything (long output)
  -generate
        generate new type information
  -region string
        print instance type names for a region (or 'all')
  -type string
        print specific instance type detail (or 'all')
  -zone string
        print instance type names for a zone (region is needed too)
```

Filtering now works too, I added a test list which will be replaced with a real one soon. I would like to refactor EC2 the same way this is done.

adiabramovitch (approved):
looks great!
 I have no comments what about you? @ezr-ondrej

lzap:
Added one new commit since @adiabramovitch already did review - it renames the tool to `typesctl` and it breaks it via interface so @avitova can easily add a new EC2 provider to the CLI, all you need to do is to implement the three functions and that's it. The CLI will pick the new provider automatically.

It now requires a new option `-provider` which currently only takes `azure` obviously as this is the only implementation to this date.

```
% go run cmd/typesctl/main.go -help                              

  -all
        print everything (long output)
  -generate
        generate new type information
  -provider string
        provider: [azure] (required)
  -region string
        print instance type names for a region (or 'all')
  -type string
        print specific instance type detail (or 'all')
  -zone string
        print instance type names for a zone (region is needed too)

```

ezr-ondrej (approved):
Thanks a lot @lzap ! :+1:
I'm happy to have these pregenerated, I'd love to filter the instances by supported, but we'll have to wait for that, so lets just use all for now :+1:

PR #209, opened 2022-09-06
Pubkey resource unscoped by account

ezr-ondrej:
Rename pubkey reservation dao methods.

This might be bit too much to add prefix, suffix might have enough visibility, happy to change to whatever we agree on :)

### DIFF
--- a/internal/dao/dao_interfaces.go
+++ b/internal/dao/dao_interfaces.go
@@ -42,12 +42,15 @@ type PubkeyDao interface {
 
 var GetPubkeyResourceDao func(ctx context.Context) (PubkeyResourceDao, error)
 
-// PubkeyResourceDao TODO
+// PubkeyResourceDao interface for manipulating data of Pubkey deployed into cloud provider.
+// Allows to track the key in the cloud provider and manage it.
+// All methods are unscoped by Account and thus should not get user input as input directly,
+// input should first be validated by PubkeyDao.GetById
 type PubkeyResourceDao interface {
-	GetResourceByProviderType(ctx context.Context, pubkeyId int64, provider models.ProviderType) (*models.PubkeyResource, error)
-	ListByPubkeyId(ctx context.Context, pkId int64) ([]*models.PubkeyResource, error)
-	Create(ctx context.Context, pkr *models.PubkeyResource) error
-	Delete(ctx context.Context, id int64) error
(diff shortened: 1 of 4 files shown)


internal/dao/tests/pubkey_resource_test.go
```diff
@@ -33,7 +33,7 @@ func createPubkeyResourceAzure() *models.PubkeyResource {
 	}
 }
 
-func setupResource(t *testing.T) (dao.PubkeyResourceDao, context.Context) {
+func setupPubkeyResource(t *testing.T) (dao.PubkeyResourceDao, context.Context) {
```
ezr-ondrej, line 36: I've also renamed this, as when I searched for usage, this got me confused and I thought I'm renaming wrong method.
Hope it's ok :)

internal/dao/dao_interfaces.go
```diff
@@ -42,12 +42,15 @@ type PubkeyDao interface {
...
 type PubkeyResourceDao interface {
-	GetResourceByProviderType(ctx context.Context, pubkeyId int64, provider models.ProviderType) (*models.PubkeyResource, error)
-	ListByPubkeyId(ctx context.Context, pkId int64) ([]*models.PubkeyResource, error)
-	Create(ctx context.Context, pkr *models.PubkeyResource) error
-	Delete(ctx context.Context, id int64) error
+	UnscopedGetResourceByProviderType(ctx context.Context, pubkeyId int64, provider models.ProviderType) (*models.PubkeyResource, error)
+	UnscopedListByPubkeyId(ctx context.Context, pkId int64) ([]*models.PubkeyResource, error)
+	UnscopedCreate(ctx context.Context, pkr *models.PubkeyResource) error
+	UnscopedDelete(ctx context.Context, id int64) error
 }
```
lzap, line 54: Wow that's quite harsh, I like it! No one will ever miss the fact that a DAO function is not scoped.

lzap (approved):
I thought a suffix is little bit less intrusive but I LOVE THIS.

internal/dao/dao_interfaces.go
ezr-ondrej: @adiabramovitch opinions on this? I'm fine suffix or prefix, I just wanted to go with more strict to provoke a though and see what y'all think :)

PR #210, opened 2022-09-07
Unify service test assertions and error handling

ezr-ondrej:
Puts all service tests under the services_test package.
Uses require for cases where we want to stop further test execution
Unify failures handling accros the test files.


lzap (approved):
WFM

PR #211, opened 2022-09-07
Add azure instance types endpoint

lzap:
This adds a new endpoint, since we don't have Azure sources integration yet, I am adding it as a separate endpoint that is also documented as undocumented :-) Meaning, there is no OpenAPI for this endpoint, it can be used for quick lookups in the instance type database without having any source record. I propose to keep such endpoint for our own development purposes.

This can serve as an example how to implement the same for EC2 for @avitova. In that case, we already have sources integration so both `/sources/ID/instance_types` and `/instance_types` should be implementing returning the very same results. Note EC2 endpoint currently only supports region, we need to add zones as well.

### DIFF
--- a/internal/clients/http/azure/types/azure_types.go
+++ b/internal/clients/http/azure/types/azure_types.go
@@ -1,49 +1,92 @@
 package types
 
 import (
+	"bytes"
 	"embed"
+	"fmt"
+	"path/filepath"
+	"time"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/middleware"
 )
 
 //go:embed *.yaml availability/*.yaml
 var fsTypes embed.FS
 
-var registeredTypes clients.RegisteredInstanceTypes
-var regionalAvailability clients.RegionalTypeAvailability
+const availabilityPath = "availability"
+
+var etag *middleware.ETag
+
+var typeInfo clients.InstanceTypeInfo
 
 func init() {
 	// load instance types
-	buffer, err := fsTypes.ReadFile("types.yaml")
+	typesBuf, err := fsTypes.ReadFile("types.yaml")
 	if err != nil {
 		panic(err)
 	}
-	err = registeredTypes.Load(buffer)
+	err = typeInfo.RegisteredTypes.Load(typesBuf)
 	if err != nil {
 		panic(err)
 	}
 
 	// load supported types
-	buffer, err = fsTypes.ReadFile("supported.yaml")
+	supportedBuf, err := fsTypes.ReadFile("supported.yaml")
 	if err != nil {
 		panic(err)
 	}
-	err = registeredTypes.LoadSupported(buffer)
+	err = typeInfo.RegisteredTypes.LoadSupported(supportedBuf)
 	if err != nil {
 		panic(err)
 	}
 
 	// load availability information
-	err = regionalAvailability.Load(fsTypes, "availability")
+	err = typeInfo.RegionalAvailability.Load(fsTypes, availabilityPath)
+	if err != nil {
+		panic(err)
+	}
+
+	// calculate etag
+	// concatenate all yamls into one 1 MiB buffer
+	dirEntries, err := fsTypes.ReadDir(availabilityPath)
+	if err != nil {
+		panic(err)
+	}
+	availBuf := bytes.NewBuffer(make([]byte, 0, 2^20))
+	for _, dirEntry := range dirEntries {
+		if dirEntry.IsDir() {
+			continue
+		}
+		file := filepath.Join(availabilityPath, dirEntry.Name())
+		buffer, errBuf := fsTypes.ReadFile(file)
+		if errBuf != nil {
+			panic(errBuf)
+		}
+		availBuf.Write(buffer)
+	}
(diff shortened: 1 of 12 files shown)


internal/clients/http/azure/types/azure_types.go
```diff
@@ -9,16 +9,15 @@ import (
 //go:embed *.yaml availability/*.yaml
 var fsTypes embed.FS
 
-var registeredTypes clients.RegisteredInstanceTypes
-var regionalAvailability clients.RegionalTypeAvailability
+var typeInfo clients.InstanceTypeInfo
```
lzap, line 12: I moved this into a new type from the clients package so the lookup function can be shared across providers.

internal/payloads/error_payload.go
```diff
@@ -63,6 +55,19 @@ func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
...
+		logger.Error().Msg(msg)
+	}
+	return &ResponseError{
+		HTTPStatusCode: 400,
+		Message:        msg,
+		RequestId:      ctxval.RequestId(ctx),
+		Context:        ctx,
+	}
+}
+
```
lzap, line 70: The `ParamMissingError` looked weird, the purpose of this payloads package is to have payload errors, these are serializable JSON structures, not Go errors.

internal/payloads/instance_types_payload.go
```diff
@@ -19,9 +19,9 @@ func (s *InstanceTypeResponse) Render(_ http.ResponseWriter, _ *http.Request) er
...
 }
 
-func NewListInstanceTypeResponse(sl *[]clients.InstanceType) []render.Renderer {
-	list := make([]render.Renderer, 0, len(*sl))
-	for _, instanceType := range *sl {
+func NewListInstanceTypeResponse(sl []*clients.InstanceType) []render.Renderer {
+	list := make([]render.Renderer, 0, len(sl))
+	for _, instanceType := range sl {
 		list = append(list, &InstanceTypeResponse{instanceType})
 	}
```
lzap, line 26: This change makes sure we are not copying instance types over and over again, it is now slice of pointers, not pointer to slice. Slices are effectively passed as reference.

internal/services/builtin_types_service.go
```diff
@@ -0,0 +1,52 @@
...
+	if zone == "" {
+		renderError(w, r, payloads.NewMissingRequestParameterError(r.Context(), "zone"))
+		return
+	}
+
+	if provider == "azure" {
+		listBuiltinTypesAzure(w, r, region, zone, supported)
+		return
+	}
+}
```
lzap, line 52: The new instance types endpoint is now much simpler.

lzap:
Fixed the linter.

lzap:
Added ETag support, rebased.

internal/payloads/error_payload.go
```diff
@@ -63,6 +55,19 @@ func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
 	}
 }
 
+func NewMissingRequestParameterError(ctx context.Context, param string) *ResponseError {
+	msg := fmt.Sprintf("missing parameter: %s", param)
+	if logger := ctxval.Logger(ctx); logger != nil {
+		logger.Error().Msg(msg)
```
ezr-ondrej, line 61: Do we want to log it as error? It is not an application error, it is client error, so we shall log it to Warn or maybe even just info IMHO :)

internal/services/instance_types_service_test.go
```diff
@@ -55,15 +58,15 @@ func TestListInstanceTypesHandler(t *testing.T) {
...
+		assert.Nil(t, err, fmt.Sprintf("Error creating a new request: %v", err))
 
 		rr := httptest.NewRecorder()
 		handler := http.HandlerFunc(services.ListInstanceTypes)
 		handler.ServeHTTP(rr, req)
 
 		require.Equal(t, http.StatusBadRequest, rr.Code, "Handler returned wrong status code")
 
-		assert.Error(t, payloads.ParamMissingError{})
+		assert.Contains(t, rr.Body.String(), "missing parameter")
```
ezr-ondrej, line 69: Seems like that except this change, this reverts my changes from #210 is that intentional?

internal/services/instance_types_service_test.go
```diff
@@ -3,21 +3,21 @@ package services_test
...
 	"github.com/RHEnVision/provisioning-backend/internal/dao/stubs"
-	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 	"github.com/RHEnVision/provisioning-backend/internal/services"
 	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestListInstanceTypesHandler(t *testing.T) {
+func TestListInstanceTypes(t *testing.T) {
```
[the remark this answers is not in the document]
lzap: Did not mean to change it.

internal/services/instance_types_service_test.go
lzap: Nah, this was a bad conflict merge. I meant to change only this line. Fixed.

lzap:
Rebased.

ezr-ondrej (approved):
Cool thanks @lzap, it works nicely. It's missing a test, but I'll take it as an oportunity for me to be useful ^_^

PR #213, opened 2022-09-08
Split Makefile and create devel docs

lzap:
No changes in the Makefile other than split. Inspired from https://github.com/content-services/content-sources-backend/tree/main/mk and big thanks for the AWK command which generates the help page.

I haven't implemented the loading as it was designed there, I am using include glob to pick all the mk files.

There are some targets which we might adopt as well, feel free to review and we might pull something from there. Container looks interesting, they support both docker and podman. That could be nice.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ TAGS
 /deploy/bonfire.yaml
 /Containerfile
 /tern.conf
+/mk/private.mk
 /scripts/sources-*/
 /scripts/sources.local.conf
 /scripts/rest_examples/*.local.http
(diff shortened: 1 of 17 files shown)


ezr-ondrej:
QE failure related: `make: *** No rule to make target 'prep'.  Stop.`

lzap:
Oh I renamed this one, I thought it's unused anyway. Created an alias.

ezr-ondrej:
> Oh I renamed this one, I thought it's unused anyway. Created an alias.

Or we can change its usage in the Docker file, if you want to rename :)

PR #215, opened 2022-09-08
Add test for Azure instance types service

ezr-ondrej:
Azure instance types has been merged without any test.

### DIFF
--- a/internal/services/instance_types_service_test.go
+++ b/internal/services/instance_types_service_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"testing"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
@@ -17,7 +18,6 @@ import (
 )
 
 func TestListInstanceTypesHandler(t *testing.T) {
-
 	t.Run("with region", func(t *testing.T) {
 		var names []string
 		ctx := stubs.WithAccountDaoOne(context.Background())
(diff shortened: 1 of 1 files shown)


docs/configure-amazon-role.md
```diff
@@ -35,7 +35,7 @@ Create a new policy for the Service account:
                 "sts:AssumeRole",
                 "sts:GetFederationToken"
             ],
-            "Resource": "arn:aws:iam::123456789:role/*"
+            "Resource": "arn:aws:iam::*:role/*"
         }
```
lzap, line 39: Just to double check, this allows assuming to any role on the service account, so it is a safe change.

internal/services/instance_types_service_test.go
```diff
@@ -64,5 +64,24 @@ func TestListInstanceTypesHandler(t *testing.T) {
 
 		assert.Contains(t, rr.Body.String(), "missing parameter")
 	})
+}
+
+func TestListAzureBuiltinInstanceTypesHandler(t *testing.T) {
+	ctx := context.Background()
+
+	req, err := http.NewRequestWithContext(ctx, "GET", "/api/provisioning/v1/instance_types/azure", nil)
```
lzap, line 72: We should probably pull the prefix out into a constant for all the tests, good for now.

lzap (approved):
Sorry I forgot, my bad.

lzap:
Tests do fail tho.

docs/configure-amazon-role.md
ezr-ondrej: yup, sorry it should probably be separate, but I really want to add this, because we need the service acc to assume role in any account, not a specific one :)

ezr-ondrej:
I've fixed the tests

docs/configure-amazon-role.md
ezr-ondrej: I'll split it off, it really does not belong to this PR :)

ezr-ondrej: Splitted into #218

PR #216, opened 2022-09-09
Improve error handling of sources client

lzap:
SSIA

FYI this is broken into separate commits for easier review.

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -17,7 +17,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	stsTypes "github.com/aws/aws-sdk-go-v2/service/sts/types"
 	"github.com/rs/zerolog"
-	"github.com/rs/zerolog/log"
 )
 
 type EC2Client struct {
@@ -31,14 +30,15 @@ func init() {
 }
 
 func newAssumedEC2ClientWithRegion(ctx context.Context, arn string, region string) (clients.EC2, error) {
-	logger := ctxval.Logger(ctx)
+	logger := ctxval.Logger(ctx).With().Str("client", "ec2").Logger()
 
 	if region == "" {
+		// TODO: use DefaultRegion from config
 		logger.Warn().Msg("No region passed, using us-east-1")
 		region = "us-east-1"
 	}
 
-	creds, err := getStsAssumedCredentials(ctx, arn, region)
+	creds, err := getStsAssumedCredentials(ctx, logger, arn, region)
 	if err != nil {
 		return nil, err
 	}
@@ -53,13 +53,11 @@ func newAssumedEC2ClientWithRegion(ctx context.Context, arn string, region strin
 	return &EC2Client{
 		ec2:     ec2.NewFromConfig(newCfg),
 		context: ctx,
-		log:     logger.With().Str("client", "ec2").Logger(),
+		log:     logger,
 	}, nil
 }
 
-func getStsAssumedCredentials(ctx context.Context, arn string, region string) (*stsTypes.Credentials, error) {
-	logger := ctxval.Logger(ctx)
-
+func getStsAssumedCredentials(ctx context.Context, logger zerolog.Logger, arn string, region string) (*stsTypes.Credentials, error) {
 	// TODO: role assume should be region agnostic
 	cfg, err := awsCfg.LoadDefaultConfig(ctx, awsCfg.WithRegion(region),
 		awsCfg.WithCredentialsProvider(
@@ -88,7 +86,7 @@ func getStsAssumedCredentials(ctx context.Context, arn string, region string) (*
 
 // ImportPubkey imports a key and returns AWS ID
 func (c *EC2Client) ImportPubkey(key *models.Pubkey, tag string) (string, error) {
-	log.Trace().Msgf("Importing AWS key-pair named '%s' with tag '%s'", key.Name, tag)
+	c.log.Trace().Msgf("Importing AWS key-pair named '%s' with tag '%s'", key.Name, tag)
 	input := &ec2.ImportKeyPairInput{}
 	input.KeyName = aws.String(key.Name)
 	input.PublicKeyMaterial = []byte(key.Body)
@@ -118,7 +116,7 @@ func (c *EC2Client) ImportPubkey(key *models.Pubkey, tag string) (string, error)
 }
 
 func (c *EC2Client) DeleteSSHKey(handle string) error {
-	log.Trace().Msgf("Deleting AWS key-pair with handle %s", handle)
+	c.log.Trace().Msgf("Deleting AWS key-pair with handle %s", handle)
 	input := &ec2.DeleteKeyPairInput{}
 	input.KeyPairId = aws.String(handle)
 	_, err := c.ec2.DeleteKeyPair(c.context, input)
@@ -152,7 +150,7 @@ func (c *EC2Client) ListInstanceTypesWithPaginator() ([]types.InstanceTypeInfo, 
 }
 
 func (c *EC2Client) RunInstances(ctx context.Context, name *string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error) {
-	log.Trace().Msg("Run AWS EC2 instance")
+	c.log.Trace().Msg("Run AWS EC2 instance")
 	encodedUserData := base64.StdEncoding.EncodeToString(userData)
 	input := &ec2.RunInstancesInput{
 		MaxCount:     aws.Int32(amount),
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -107,7 +107,7 @@ func (c *EC2Client) ImportPubkey(key *models.Pubkey, tag string) (string, error)
 		if isAWSUnauthorizedError(err) {
 			err = clients.UnauthorizedErr
 		} else if isAWSOperationError(err, "InvalidKeyPair.Duplicate") {
-			err = clients.DuplicatePubkeyErr
+			err = DuplicatePubkeyErr
 		}
 		return "", fmt.Errorf("cannot import SSH key %s: %w", key.Name, err)
 	}
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -19,10 +19,11 @@ import (
 	"github.com/rs/zerolog"
 )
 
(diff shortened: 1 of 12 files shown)


ezr-ondrej (approved):
Nice cleanup so far :)

internal/clients/http/sources/sources_client.go
```diff
@@ -107,51 +115,55 @@ func (c *SourcesClient) ListProvisioningSources(ctx context.Context) (*[]clients
...
-		return "", NewNotFoundError(ctx, "authentications for source weren't found in sources app")
-	}
-	if !parsing.IsHTTPStatus2xx(statusCode) {
-		ctxval.Logger(ctx).Warn().Msgf("Sources replied with unexpected status while fetching Authentications per source: %v", statusCode)
-		return "", SourcesClientErr
+
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
+	if err != nil {
+		if errors.Is(err, clients.NotFoundError) {
+			return "", SourceNotFoundErr
```
ezr-ondrej, line 130: I'd really love, if all our clients would throw the same errors, at least wrapped? I liked it the way linter did not xD

lzap: Yes I wrapped all of them.

ezr-ondrej: you are creating a `clients.NotFoundError` that is very useful to the outside world, but then you drop it here completely and return `sources.SourceNotFoundErr` that is IMHO completely useless to outside world (outside of http package)

internal/clients/http/image_builder/image_client.go
```diff
@@ -44,66 +45,63 @@ func newImageBuilderClient(ctx context.Context) (clients.ImageBuilder, error) {
...
-	ctxval.Logger(ctx).Debug().Msgf("Verifying AWS type")
+
+	logger.Trace().Msgf("Verifying AWS type")
 	if imageStatus.Type != UploadTypesAws {
-		ctxval.Logger(ctx).Warn().Err(err).Msg("Image is not AWS type")
-		return "", BadImageTypeErr
+		return "", fmt.Errorf("%w: expected image type AWS", UnknownImageTypeErr)
 	}
-	awsStatus, ok := imageStatus.Options.(map[string]interface{})
+	awsStatus, ok := imageStatus.Options.(map[string]interface{})["ami"]
```
ezr-ondrej, line 77: Lets rename the var, if we want to change content of it? :)

internal/clients/http/image_builder/image_client.go
```diff
@@ -44,66 +45,63 @@ func newImageBuilderClient(ctx context.Context) (clients.ImageBuilder, error) {
...
-		return nil, ClientErr
-	}
 
-	err = verifyImage(ctx, resp.JSON200)
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
 	if err != nil {
-		ctxval.Logger(ctx).Warn().Err(err).Msg("Image status in not ready")
-		return nil, err
+		if errors.Is(err, clients.NotFoundError) {
+			return nil, ComposeNotFoundErr
```
ezr-ondrej, line 97: here again, is it wise to return internal errors from the http/xxx packages? How would I compare the error without importing the package? I want to compare with `clients` errors, because that's what I have imported, right?

internal/clients/http/image_builder/image_client.go
lzap: Renamed.

internal/clients/http/image_builder/image_client.go
lzap: The idea is that all client functions may return either common errors or specific errors. The caller does not care as it should be using `error.Is/As` to extract errors.

lzap: To elaborate, it is okay (IMHO and for the liter) to return an error that is an exported error from other package. What is not okay (and linter complains about) is when you get a generic `error` and return it unwrapped.

ezr-ondrej: but you loose the information about this being clients.NotFoundError completely, right? I'm not able to figure out that this was `clients.NotFoundError` by any use of `error.Is/As` or am I?

lzap:
I am done with this patch, please do full review. It is broken down into smaller commits.

internal/clients/http/image_builder/image_client.go
lzap: Added wrapping to all errors in the last commit.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -53,13 +54,11 @@ func newAssumedEC2ClientWithRegion(ctx context.Context, arn string, region strin
 	return &EC2Client{
 		ec2:     ec2.NewFromConfig(newCfg),
 		context: ctx,
-		log:     logger.With().Str("client", "ec2").Logger(),
+		logger:  logger,
```
ezr-ondrej, line 57: Why do you think it was wrang to enrich the logger by the client field? It might be useful to know, that the logs came from the client and not our app.

internal/clients/http/image_builder/image_client.go
```diff
@@ -40,70 +45,63 @@ func newImageBuilderClient(ctx context.Context) (clients.ImageBuilder, error) {
...
-		return nil, ClientErr
-	}
 
-	err = verifyImage(ctx, resp.JSON200)
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
 	if err != nil {
-		ctxval.Logger(ctx).Warn().Err(err).Msg("Image status in not ready")
-		return nil, err
+		if errors.Is(err, clients.NotFoundErr) {
+			return nil, fmt.Errorf("fetch image status call: %w", ComposeNotFoundErr)
```
ezr-ondrej, line 97: I don't think this is fixing the issue.
I as a caller would want to do `errors.Is(err, clients.NotFoundErr)` nor `error.As` I believe.
What is the `ComposeNotFoundErr` good for? I can not assert it from higher lvl package then http, because I don't want to import `clients/http/image_builder` just for one error type :thinking:

internal/clients/http/sources/sources_client.go
```diff
@@ -107,51 +115,54 @@ func (c *SourcesClient) ListProvisioningSources(ctx context.Context) (*[]clients
...
-	if !parsing.IsHTTPStatus2xx(statusCode) {
-		ctxval.Logger(ctx).Warn().Msgf("Sources replied with unexpected status while fetching Authentications per source: %v", statusCode)
-		return "", SourcesClientErr
+
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
+			return "", fmt.Errorf("get source ARN call: %w", SourceNotFoundErr)
+		}
+		return "", fmt.Errorf("get source ARN call: %w", err)
```
ezr-ondrej, line 131: What about just wrapping it?
```suggestion
		return "", fmt.Errorf("get source ARN call: %w", err)
```

internal/clients/http/sources/sources_client.go
```diff
@@ -67,38 +72,41 @@ type dataElement struct {
...
-	}
-	if !parsing.IsHTTPStatus2xx(statusCode) {
-		ctxval.Logger(ctx).Warn().Msgf("Sources replied with unexpected status while fetching sources of application type: %v", statusCode)
-		return nil, SourcesClientErr
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
+			return nil, fmt.Errorf("list provisioning sources call: %w", SourceNotFoundErr)
+		}
+		return nil, fmt.Errorf("list provisioning sources call: %w", err)
```
ezr-ondrej, line 107: IMHO this looks better:

```suggestion
		return nil, fmt.Errorf("list provisioning sources call: %w", err)
```

internal/clients/http/image_builder/image_client.go
```diff
@@ -40,70 +45,63 @@ func newImageBuilderClient(ctx context.Context) (clients.ImageBuilder, error) {
...
+		}
+		return nil, fmt.Errorf("fetch image status call: %w", err)
```
ezr-ondrej, line 99: Wouldn't this be good enough?
```suggestion
		return nil, fmt.Errorf("fetch image status call: %w", err)
```

internal/clients/http/sources/sources_client.go
```diff
@@ -107,51 +115,54 @@ func (c *SourcesClient) ListProvisioningSources(ctx context.Context) (*[]clients
...
-	if !parsing.IsHTTPStatus2xx(statusCode) {
-		ctxval.Logger(ctx).Warn().Msgf("Sources replied with unexpected status while fetching Applications: %v", statusCode)
-		return "", SourcesClientErr
+
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
+			return "", fmt.Errorf("get source ARN call: %w", ApplicationNotFoundErr)
+		}
+		return "", fmt.Errorf("get source ARN call: %w", err)
```
ezr-ondrej, line 154:
```suggestion
		return "", fmt.Errorf("get source ARN call: %w", err)
```

internal/clients/http/sources/sources_client.go
```diff
@@ -167,28 +178,34 @@ func (c *SourcesClient) GetProvisioningTypeId(ctx context.Context) (string, erro
...
 	}
-	if !parsing.IsHTTPStatus2xx(resp.StatusCode) {
-		ctxval.Logger(ctx).Warn().Msgf("Sources replied with unexpected status while fetching ApplicationTypes: %v", resp.Status)
-		return "", fmt.Errorf("%w, status: '%s'", ApplicationTypesFetchUnsuccessful, resp.Status)
-	}
 	defer resp.Body.Close()
+
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode)
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
```
ezr-ondrej, line 192: :disappointed: :panda_face:

ezr-ondrej (commented):
Great cleanup, but I still struggle with dropping the ClientNotFoundErr information, that is lost and not retrievable by client. Can we do something about it? I struggle with understanding how I as a caller of such function would use this.

internal/clients/http/ec2/ec2_client.go
lzap: Oh I moved this up, look it few lines up. If it's not there, then it's a bug, but I swear I did put it there.

ezr-ondrej (approved):

PR #217, opened 2022-09-09
Refactor clients package, add documentation, drop old APIClient

lzap:
In this patch, I move things around in the client package so it's more clear. Here is how I propose it:

* clients/ - contains only errors.go and interface.go as well as all shared types which are now stored in single files (singular noun): architecture.go, instance_type.go, instance_type_info.go, regional_availability.go and registered_types.go (it's slice so plural).
* clients/supported - temporary package for our current (AWS) supported list and filtering, this is supposed to be moved into the cmd/typesctl utility where we will preload and mark as supported into YAML
* clients/http - now contains common and specific errors (`{ec2,sources,image_builder}_errors.go` and common_http_errors.go) as well as `*_doer.go` interface and implementations

The common theme is:

* Package clients/ should only contain the interface, common errors and common types (we previously called them "models" although these are not database models at all).
* Package clients/http contains HTTP related common errors, helpers (doers) and service-related errors.
* Packages clients/http/XXX contain the actual implementations.
* Package clients/stubs contains test stubs.

It's not perfect as I would very much rather prefer a design where each client is in its very own package (e.g. clients/ec2 containing interface and errors with clients/ec2/http and clients/ec2/stub). In the current design, errors are "hanging" a bit in "http" package where they should be really somewhere else so they can be reused by stubs. We can move them to the main clients/ package but they all must be named differently. It's keep it this way for more time and if this turns to be a problem, we can always refactor it.

This is a pure refactoring PR, no code changes were made. See multiple commits to see what happened.

Edit: I pushed two additional commits which are not refactoring:

* Adds documentation to the client interfaces, we should really keep those very important interfaces documented.
* It looks like we have an old APIClient interface which is unused, I remember we had it in the first version when the interface was based on the generated API rather than our own custom one. I removed it.


ezr-ondrej (approved):
Looks good to me, I was trying to check if it was only moving around packages, and it looks like it.
I believe that code should have good home, and this is great improvement towards that ^_^
Thanks @lzap !

ezr-ondrej:
> * It looks like we have an old APIClient interface which is unused, I remember we had it in the first version when the interface was based on the generated API rather than our own custom one. I removed it.

I disagree with this point, it could be useful to test the wrapping client, but it does not serve this purpose now, so lets remove it, we can add it once needed :)

lzap:
> could be useful to test the wrapping client, but it does not serve this purpose now, so lets remove it, we can add it once needed :)

Sure but isn't better to test our interface which also tests the generated client API? Anyway, let's add it if we need it. Thanks, have a good one! Cheers.

ezr-ondrej:
> Sure but isn't better to test our interface which also tests the generated client API? Anyway, let's add it if we need it. Thanks, have a good one! Cheers.

Well the generated client actually makes real http calls, but I'd agree, that it would be better to mock it on the network (http) layer :thinking:

PR #219, opened 2022-09-11
Add GOOGLE_APPLICATION_CREDENTIALS env var

adiabramovitch:


internal/config/parser/known.go
```diff
@@ -23,6 +23,7 @@ var known = [][]string{
 	{"AWS.REGION", ""},
 	{"AWS.SECRET", ""},
 	{"AWS.SESSION", ""},
+	{"GCP.APPCRDS", "GOOGLE_APPLICATION_CREDENTIALS"},
```
adiabramovitch, line 26: @ezr-ondrej Not sure about that

ezr-ondrej: depends if we want to rename this variable.

ezr-ondrej: `GOOGLE_CREDENTIALS` sounds good enough to me tho :) or even `GCP_CREDENTIALS` and we could have `gcp.credentials` and
```
gcp:
  credentials
```

And we need no renaming :)

adiabramovitch:
```golang	
{"GCP.CREDENTIALS", "GOOGLE_CREDENTIALS"},
```

ezr-ondrej (commented):
This is fine as is, but I'd love to see this with also usage of it (initialization of the GCE client).
Is that a plan, or what do we think here?

adiabramovitch:
> This is fine as is, but I'd love to see this with also usage of it (initialization of the GCE client).
> 
> Is that a plan, or what do we think here?

That is the plan I'll add the changes in this pr but in several commits

configs/local_example.yaml
```diff
@@ -47,6 +47,9 @@ azure:
   # Secret value
   clientSecret:
 
+gcp:
+  credentials:
+
```
lzap, line 52: Can you add comments on what these are? Perhaps a brief explanation. I mean, is that a UUID, secret key or JSON string expected here?

internal/config/main_config.go
```diff
@@ -64,6 +64,9 @@ var config struct {
 		ClientID       string
 		ClientSecret   string
 	}
+	Gcp struct {
```
lzap, line 67: We do have `AWS` in the main config, so let's keep this for this case:

```suggestion
	GCP struct {
```

internal/config/0_common_config.go
```diff
@@ -52,4 +52,7 @@ func init() {
 	parser.Viper.SetDefault("aws.secret", "")
 	parser.Viper.SetDefault("aws.session", "")
 	parser.Viper.SetDefault("aws.instancePrefix", "")
+
+	//GCP
```
lzap, line 56:
```suggestion
	// GCP
```

lzap (changes requested):
Can you tell what is expected in the field? I've seen GCP documentation and it looks like this might be a JSON string. If that's the case, can you also add a validator (we already have few of them) which will stop the app if the provided value is not JSON (can be empty but not anything else than JSON).

If this is not a JSON, then just ignore my comment.

internal/config/main_config.go
ezr-ondrej: for some reason that didn't work for @adiabramovitch, but might be issue somewhere else? If it works, I'd also prefered to be consistent.

internal/jobs/launch_instance_gcp.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot bulk insert instances: %w", err)
+	}
+
+	// TODO: info regarding the reservation and info about the created instances (create a db record for each instance)
+	if err = op.Wait(ctx); err != nil {
+		panic(err)
+	}
+	return nil
+
```
lzap, line 121: Ok this is very important - the wait function will wait forever? Is there a timeout?

By default, our job queue will do X hearbeats every Y seconds (this is configurable), so we need to actually set a context timeout here so when the job queue will almost consider this job to be locked (dead), then we will actually time this out and report an error.

I suggest to create a function in the `config` package called `DefaultJobTimeout` which would return time.Duration of `config.Worker.HearbeatSec * config.Worker.MaxBeats - 10 sec`. This would always be guaranteed to be at least 30 seconds (or some reasonable value), if not a panic would be raised (in case someone misconfigure it).

Then use it:

```go
ctx, _ = context.WithTimeout(ctx, config.DefaultJobTimeout())
if err = op.Wait(ctx); err != nil {
// ...
}
```

internal/db/migrations/005_reservations.sql
```diff
@@ -33,3 +33,18 @@ CREATE TABLE aws_reservation_details
   FOREIGN KEY (reservation_id, provider) REFERENCES reservations(id, provider) ON DELETE CASCADE,
   CHECK (provider = provider_type_aws())
 );
+
+CREATE TABLE gcp_reservation_details
+(
+  reservation_id BIGINT NOT NULL,
+  provider INTEGER NOT NULL DEFAULT provider_type_gcp(),
+  -- pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
+  pubkey_id TEXT NOT NULL,
```
lzap, line 42: Not a huge fan of this, I would rather see pubkey reference from the day one, but it will do for now.

internal/models/reservation_model.go
```diff
@@ -80,6 +80,41 @@ type AWSReservation struct {
...
+type GCPReservation struct {
+	Reservation
+
+	// Pubkey ID.
+	PubkeyID string `db:"pubkey_id" json:"pubkey_id"`
+
+	// Source ID.
+	SourceID string `db:"source_id" json:"source_id"`
+
+	// The ID of the aws reservation which was created.
```
lzap, line 108:
```suggestion
	// The ID of the GCP reservation which was created.
```

internal/models/reservation_model.go
```diff
@@ -80,6 +80,41 @@ type AWSReservation struct {
...
+	GCPReservationID string `db:"gcp_reservation_id" json:"gcp_reservation_id"`
```
lzap, line 109: I believe this is unused, I suggest to drop or comment out unused fields for now so that we won't forget to remove them in the future if they end up not being used at all.

internal/jobs/launch_instance_gcp.go
```diff
@@ -0,0 +1,122 @@
...
+		},
+	}
+
+	logger.Trace().Msg("Executing Insert")
+	op, err := instancesClient.BulkInsert(ctx, req2)
+	if err != nil {
+		return fmt.Errorf("cannot bulk insert instances: %w", err)
+	}
+
+	// TODO: info regarding the reservation and info about the created instances (create a db record for each instance)
```
lzap, line 116: Honestly, I would rather see this feature in the PR since you also add the field into the API (payload,  model). Or you can keep it this way and comment it out from there.

internal/services/gcp_reservation_service.go
```diff
@@ -0,0 +1,121 @@
...
+
+func CreateGCPReservation(w http.ResponseWriter, r *http.Request) {
+	logger := ctxval.Logger(r.Context())
+
+	var accountId int64 = ctxval.AccountId(r.Context())
+
+	// TODO: meanwhile these values are hardcoded until we will have instance types, sources, ssh endpoints,
+	payload := &payloads.GCPReservationRequestPayload{
+		// TODO: The actual value of the pubkey change to pubkey id
+		PubkeyID: "aabramov:ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC8w6DONv1qn3IdgxSpkYOClq7oe7davWFqKVHPbLoS6+dFInru7gdEO5byhTih6+PwRhHv/b1I+Mtt5MDZ8Sv7XFYpX/3P/u5zQiy1PkMSFSz0brRRUfEQxhXLW97FJa7l+bej2HJDt7f9Gvcj+d/fNWC9Z58/GX11kWk4SIXaKotkN+kWn54xGGS7Zvtm86fP59Srt6wlklSsG8mZBF7jVUjyhAgm/V5gDFb2/6jfiwSb2HyJ9/NbhLkWNdwrvpdGZqQlYhnwTfEZdpwizW/Mj3MxP5O31HN45aE0wog0UeWY4gvTl4Ogb6kescizAM6pCff3RBslbFxLdOO7cR17 lzap+rsakey+{{$randomInt}}@redhat.com",
```
lzap, line 27: You can simply use the seeded pubkey ID=1. Feel free to seed your key too and use it as ID=2 :-)

internal/services/gcp_reservation_service.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		renderError(w, r, payloads.NewEnqueueTaskError(r.Context(), "GCP reservation", err))
+		return
+	}
+
+	// Return response payload
+	if err := render.Render(w, r, payloads.NewGCPReservationResponse(reservation)); err != nil {
+		renderError(w, r, payloads.NewRenderError(r.Context(), "reservation", err))
+	}
+
```
lzap, line 120: Empty line, I also saw one above if you could kill those.

internal/services/gcp_reservation_service.go
```diff
@@ -0,0 +1,121 @@
...
+		// TODO: The project id change to source id from sources
+		ImageID:  payload.ImageID,
+		SourceID: payload.SourceID,
+		Detail:   detail,
+	}
+
+	reservation.AccountID = accountId
+	reservation.Status = "Created"
+	reservation.Provider = models.ProviderTypeGCP
+	reservation.Steps = 2
```
lzap, line 67: Isn't this just a single step?

```suggestion
	reservation.Steps = 1
```

lzap (changes requested):
Excellent work, I do have one major concern and that is in regard to the timeout of the job. We  ignored it for now because all EC2 operations are instant, but here it looks like the Wait is actually polling, so we need to implement proper timeout before a job is actually considered as dead and resubmitted.

Luckily, it's extremely easy assuming the Go SDK GCP library supports context cancellation: https://pkg.go.dev/context#example-WithTimeout

I think that ideally we should actually split the waiting and submission into two job actions, because if waiting fails, job queue can easily resubmit it and it won't try to submit the launch request again (it will just wait more time). I hope the SDK allows postponed waiting somehow (using some kind of reservation ID/handle).

internal/jobs/launch_instance_gcp.go
```diff
@@ -0,0 +1,122 @@
...
+		Zone:    args.Zone,
+		BulkInsertInstanceResourceResource: &computepb.BulkInsertInstanceResource{
+			NamePattern: ptr.String("inst-####"),
+			Count:       ptr.Int64(args.Detail.Amount),
+			MinCount:    ptr.Int64(args.Detail.Amount),
+			InstanceProperties: &computepb.InstanceProperties{
+				Disks: []*computepb.AttachedDisk{
+					{
+						InitializeParams: &computepb.AttachedDiskInitializeParams{
+							DiskSizeGb:  ptr.Int64(30),
```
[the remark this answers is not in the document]
adiabramovitch: This is not possible, I can't assign a size smaller than the image size which is over 20

internal/models/reservation_model.go
adiabramovitch: This is something I am still trying to figure out. I don't get the instance and reservation info right away

internal/services/gcp_reservation_service.go
adiabramovitch: This was copied from aws, so maybe we need to change it there as well

internal/jobs/launch_instance_gcp.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot create instances REST client: %w", err)
+	}
+	defer instancesClient.Close()
+
+	req2 := &computepb.BulkInsertInstanceRequest{
+		Project: args.ProjectID,
+		Zone:    args.Zone,
+		BulkInsertInstanceResourceResource: &computepb.BulkInsertInstanceResource{
+			NamePattern: ptr.String("inst-####"),
```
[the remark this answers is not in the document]
ezr-ondrej: yup I definitely agree ^_^ I'd start with this in our UI and try to avoid name field per instance :)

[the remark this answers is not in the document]
ezr-ondrej: I've added a https://issues.redhat.com/browse/HMSPROV-269 to track this :)

internal/services/gcp_reservation_service.go
ezr-ondrej: We have pubkey step in AWS, we do not have that yet here :)

ezr-ondrej: @lzap this is very hard to remember to update, when adding another step.

internal/jobs/launch_instance_gcp.go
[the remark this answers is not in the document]
adiabramovitch: So args.Name is not needed. 
First I used InsertInstanceResource which needs the name manually. 
But I saw I cant define the amount of instances I want to create so I moved to BulkInsertInstanceResource which automatically creates instances names.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot create new instances REST client: %w", err)
+	}
+
+	return &gcpClient{
+		instancesClient: instancesClient,
+		context:         ctx,
+		logger:          logger,
+	}, nil
+
```
adiabramovitch, line 39: I will remove that extra line in the next update 😸 
@lzap Isn't go linter suppose to warn about this?

ezr-ondrej: Go linter does not enforce these. But it would be nice if it would :)
Maybe there is an option? :)

ezr-ondrej (commented):
:cupid: it :+1: great addition!
I just wonder where are we using the credentials, isn't this supposed to use the json to authenticate the service account?

adiabramovitch:
> :cupid: it :+1: great addition!
> 
> I just wonder where are we using the credentials, isn't this supposed to use the json to authenticate the service account?

Yes you add the json in the local.yml file

internal/clients/http/image_builder/image_client.go
```diff
@@ -81,6 +81,31 @@ func (c *ibClient) GetAWSAmi(ctx context.Context, composeID string) (string, err
...
+	}
+	imageName, ok := imageStatus.Options.(map[string]interface{})["image_name"]
+	if !ok {
+		return "", http.NameNotFoundInStatusErr
+	}
+
+	projectID, ok := imageStatus.Options.(map[string]interface{})["project_id"]
+	if !ok {
+		return "", http.NameNotFoundInStatusErr
+	}
```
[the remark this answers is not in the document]
lzap: I don't see this resolved in the code. Might be a github.com hiccup.

internal/clients/http/gcp/gcp_client.go
lzap: It's fine, it's just I can't sleep that's all ... :-)

internal/clients/http/image_builder/image_client.go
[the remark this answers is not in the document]
lzap: Oh sure, I had to refresh.

internal/jobs/launch_instance_gcp.go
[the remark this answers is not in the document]
lzap: Oh okay, weird. Well played, Google, well played...

internal/clients/stubs/gcp_stub.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"errors"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+)
+
+type gcpCtxKeyType string
+
+var gcpCtxKey gcpCtxKeyType = "gcp-interface"
+var NotImplementedErr = errors.New("stub not yet implemented")
```
lzap, line 13: Nitpick: I would move this into `stubs/errors.go` or something so it's obvious this error is for all stubs.

internal/config/main_config.go
```diff
@@ -64,6 +64,9 @@ var config struct {
 		ClientID       string
 		ClientSecret   string
 	}
+	GCP struct {
+		Credentials string
+	}
```
lzap, line 69: Nitpick: I would add a validator, I am pretty sure people will try to put some invalid inputs into this setting and not JSON. We do have a validate function already, just try to parse it with JSON and panic on errors or something.

internal/dao/stubs/reservation_dao.go
```diff
@@ -2,11 +2,14 @@ package stubs
...
 import (
 	"context"
+	"errors"
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 )
 
+var NotImplementedErr = errors.New("stub not yet implemented")
+
```
lzap, line 12: Nitpick: This could be in another `errors.go` too. A second file, it's a different package.

internal/models/reservation_model.go
lzap: It's okay to leave it here just put "TODO" here. Or maybe do not export this via OpenAPI so QA folks are not confused why it's blank.

lzap (approved):
Few nitpicks, I am good. @ezr-ondrej ?

I highly suggest to implement the PowerOff flag as soon as possible. It should not be very different from what we do for EC2 today. The template will be exactly the same, you just need to pass this via the API request.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,91 @@
...
+		Zone:    zone,
+		BulkInsertInstanceResourceResource: &computepb.BulkInsertInstanceResource{
+			NamePattern: namePattern,
+			Count:       ptr.Int64(amount),
+			MinCount:    ptr.Int64(amount),
+			InstanceProperties: &computepb.InstanceProperties{
+				Disks: []*computepb.AttachedDisk{
+					{
+						InitializeParams: &computepb.AttachedDiskInitializeParams{
+							DiskSizeGb:  ptr.Int64(DiskSizeGb),
```
lzap, line 54: Please remove the disk size, works without it.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,91 @@
...
+				NetworkInterfaces: []*computepb.NetworkInterface{
+					{
+						Name: ptr.String("global/networks/default"),
+					},
+				},
+				Metadata: &computepb.Metadata{
+					Items: []*computepb.Items{
+						{
+							Key:   ptr.String("ssh-keys"),
+							Value: ptr.String(keyName),
```
lzap, line 72: This looks weird, this should be a key body/value instead of name?

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,91 @@
...
+	c.logger.Trace().Msg("Executing Insert")
+	op, err := c.instancesClient.BulkInsert(ctx, req)
+	if err != nil {
+		return fmt.Errorf("cannot bulk insert instances: %w", err)
+	}
+
+	// TODO: info regarding the reservation and info about the created instances (create a db record for each instance)
+	if err = op.Wait(ctx); err != nil {
+		return fmt.Errorf("cannot wait for run instances operation to end: %w", err)
+	}
```
lzap, line 89: I think you can drop waiting since you don't do anything after it finishes. We need to figure out how to get request/operation ID and store it in our database and how to do waiting later on. But for EC2 we also do not wait so let's just ignore this for now.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,91 @@
+package gcp
+
+import (
+	"context"
+	"fmt"
+
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+
+	compute "cloud.google.com/go/compute/apiv1"
```
lzap, line 9: One question tho - aren't you using too low-level API? That looks like some protobuf layer. I need to read some examples about this.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func newGCPClient(ctx context.Context) (clients.GCP, error) {
+	logger := ctxval.Logger(ctx).With().Str("client", "gcp").Logger()
+
+	instancesClient, err := compute.NewInstancesRESTClient(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create new instances REST client: %w", err)
+	}
+
```
lzap, line 33: This is missing `defer instancesClient.Close()`

lzap: Ok after a small chat, I update: Close() must be part of the interface, make sure to document that all callers of GetGCPClient are responsible to defer the close call too. This is not HTTP, it's gRPC.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,92 @@
...
+	return &gcpClient{
+		instancesClient: instancesClient,
+		context:         ctx,
+		logger:          logger,
+	}, nil
+
+}
+
+func (c *gcpClient) Close() {
+	defer c.instancesClient.Close()
```
ezr-ondrej, line 43: defer makes the statement run at the end of the function, so in this case it is really not needed, given we have this statement at the end of the function ^_^

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,92 @@
...
+		},
+	}
+
+	c.logger.Trace().Msg("Executing Insert")
+	_, err := c.instancesClient.BulkInsert(ctx, req)
+	if err != nil {
+		return fmt.Errorf("cannot bulk insert instances: %w", err)
+	}
+
+	c.Close()
```
ezr-ondrej, line 89: this sounds dangerous, the caller can still use the client, I'd instead make this call in the caller function.

internal/clients/http/gcp/gcp_client.go
adiabramovitch: I don't have it at the end of the function

internal/jobs/launch_instance_gcp.go
```diff
@@ -0,0 +1,90 @@
...
+
+	pk, err := pkD.GetById(ctx, args.PubkeyID)
+	if err != nil {
+		return fmt.Errorf("cannot get pubkey by id: %w", err)
+	}
+
+	gcpClient, err := clients.GetGCPClient(ctx)
+	if err != nil {
+		return fmt.Errorf("cannot get gcp client: %w", err)
+	}
```
ezr-ondrej, line 82:
```suggestion
	gcpClient, err := clients.GetGCPClient(ctx)
	if err != nil {
		return fmt.Errorf("cannot get gcp client: %w", err)
	}
	defer gcpClient.Close()
```

internal/clients/http/gcp/gcp_client.go
ezr-ondrej: it's the only statement in the function, so you do have it at the end :)

internal/services/gcp_reservation_service.go
lzap: Understood, it is informative only luckily. But we should find a better solution, agreed.

internal/clients/http/gcp/gcp_client.go
lzap: After reading some examples, it really looks like this is it. Their API is heavily based around gRPC.

internal/clients/http/gcp/gcp_client.go
lzap: Yeah, you simply call `Close()` in this function and then you use `defer` at the call-site of this function. In other words, move `defer` from here to where you call this function.

internal/clients/http/gcp/gcp_client.go
lzap: If you could remove it, I'd appreciate it :-)

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,92 @@
...
+		logger:          logger,
+	}, nil
+
+}
+
+func (c *gcpClient) Close() {
+	defer c.instancesClient.Close()
+}
+
+func (c *gcpClient) RunInstances(ctx context.Context, projectID string, namePattern *string, imageName *string, amount int64, machineType string, zone string, keyBody string) error {
```
lzap, line 46: When I finish with Azure, we should create some types to bring the amount of arguments down.

internal/clients/interface.go
```diff
@@ -57,3 +60,13 @@ type EC2 interface {
 	// RunInstances launches one or more instances
 	RunInstances(ctx context.Context, name *string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error)
 }
+
+var GetGCPClient func(ctx context.Context) (GCP, error)
+
+type GCP interface {
+	// Caller is responsible for closing the client using Close() call
+	Close()
```
lzap, line 68: I suggest to move the comment from here to the variable `GetGCPClient` above. That is what you see in your editor/IDE when you actually use the function. Here you can just say something like "Close performs close on the gRPC client" or something like that.

internal/clients/stubs/gcp_stub.go
```diff
@@ -0,0 +1,38 @@
...
+	if si, ok = ctx.Value(gcpCtxKey).(*GCPClientStub); !ok {
+		err = &contextReadError{}
+	}
+	return si, err
+}
+
+func (mock *GCPClientStub) Close() {
+
+}
+func (mock *GCPClientStub) RunInstances(ctx context.Context, projectID string, namePattern *string, imageName *string, amount int64, machineType string, zone string, keyBody string) error {
```
lzap, line 36: Empty line should be between functions.

internal/clients/stubs/image_builder_stub.go
```diff
@@ -41,3 +41,6 @@ func (*ImageBuilderClientStub) Ready(ctx context.Context) error {
 func (mock *ImageBuilderClientStub) GetAWSAmi(ctx context.Context, composeID string) (string, error) {
 	return "ami-0c830793775595d4b-test", nil
 }
+func (mock *ImageBuilderClientStub) GetGCPImageName(ctx context.Context, composeID string) (string, error) {
```
lzap, line 44: Empty line would be appreciated :-)

internal/models/reservation_model.go
```diff
@@ -80,6 +80,41 @@ type AWSReservation struct {
...
+	// Pubkey ID.
+	PubkeyID int64 `db:"pubkey_id" json:"pubkey_id"`
+
+	// Source ID.
+	SourceID string `db:"source_id" json:"source_id"`
+
+	// The ID of the gcp reservation which was created.
+	GCPReservationID string `db:"gcp_reservation_id"`
+
+	// The ID of the image from which the instance is created. AMI's must be prefixed with 'ami-'.
```
lzap, line 111:
```suggestion
	// The ID of the image from which the instance is created.
```

internal/models/reservation_model.go
```diff
@@ -80,6 +80,41 @@ type AWSReservation struct {
...
+
+	// GCP Machine type.
+	MachineType string `json:"machine_type"`
+
+	// Amount of instances to provision of type: Instance type.
+	Amount int64 `json:"amount"`
+
+	// Immediately power off the system after initialization
+	PowerOff bool `json:"poweroff"`
+}
```
lzap, line 97: Important observation: You are missing Region, but you have a Zone. Zones live within a Region, I think we should support both from the very beginning.

internal/payloads/reservation_payload.go
```diff
@@ -112,10 +163,17 @@ func (p *AWSReservationResponsePayload) Render(_ http.ResponseWriter, _ *http.Re
 	return nil
 }
 
+func (p *GCPReservationResponsePayload) Render(_ http.ResponseWriter, _ *http.Request) error {
+	return nil
+}
 func (p *NoopReservationResponsePayload) Render(_ http.ResponseWriter, _ *http.Request) error {
```
lzap, line 169: Empty line is missing.

lzap:
I left some additional comments, most of them minor. I suggest to add Region and Zone to the API, we can hardcode the Region either in the UI or at the payload level, but our app should be ready from the day one. That is my plan for Azure, we already pushed code for instance types in Azure with both Region and Zone.

internal/models/reservation_model.go
adiabramovitch: it does not seem like there is a region specification when creating an instance https://github.com/googleapis/go-genproto/blob/18f8339a66a5/googleapis/cloud/compute/v1/compute.pb.go#L27032

lzap (approved):
I am good, nice!

lzap:
@ezr-ondrej please do re-read this is an important patch!

internal/clients/http/gcp/gcp_client.go
```diff
@@ -48,30 +48,30 @@ func (c *gcpClient) RunInstances(ctx context.Context, projectID string, namePatt
 		Zone:    zone,
 		BulkInsertInstanceResourceResource: &computepb.BulkInsertInstanceResource{
 			NamePattern: namePattern,
-			Count:       ptr.Int64(amount),
-			MinCount:    ptr.Int64(amount),
+			Count:       ptr.ToInt64(amount),
+			MinCount:    ptr.ToInt64(amount),
```
lzap, line 52: Just a sidenote: You can just use `ptr.To` and generics will figure out the type for you :-)

You only need to use `ToInt64` when the type is `int` and not `int64`, like when you use type inference.

Let's just leave this, no big deal.

ezr-ondrej: or you can just `&amount` :P which would also do it :shrug:

lzap: Lol that's right. Busted, for the second time today.

lzap:
We are now on stage which means we must start doing proper migrations. Please move the SQL changes into a new migration file, including renaming of the helper function.

internal/db/migrations/007_create_gcp_reservation_details_table.sql
```diff
@@ -0,0 +1,14 @@
...
+  provider INTEGER NOT NULL DEFAULT provider_type_gce(),
+  pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
+  source_id TEXT NOT NULL,
+  image_id TEXT NOT NULL,
+  gcp_reservation_id TEXT,
+  detail JSONB,
+
+  FOREIGN KEY (reservation_id, provider) REFERENCES reservations(id, provider) ON DELETE CASCADE,
+  CHECK (provider = provider_type_gce())
+);
```
lzap, line 14: Since it is not as easy as before to do SQL DDL, let's think twice about the `gcp_reservation_id`, that is probably not the name we end up using. Can you figure out how the "job system" of GCP work and what identifier are we getting? I slightly looked into this and could not figure it out quickly, so I gave up :-)

internal/db/migrations/002_utility_functions.sql
```diff
@@ -41,13 +41,13 @@ BEGIN
...
-CREATE OR REPLACE FUNCTION provider_type_gcp()
+CREATE OR REPLACE FUNCTION provider_type_gce()
   RETURNS INTEGER AS
-$provider_type_gcp$
+$provider_type_gce$
 BEGIN
   RETURN(SELECT 4);
 END;
-$provider_type_gcp$ LANGUAGE 'plpgsql' IMMUTABLE PARALLEL SAFE;
+$provider_type_gce$ LANGUAGE 'plpgsql' IMMUTABLE PARALLEL SAFE;
```
lzap, line 50: You cannot do this, modifications of existing SQL files are no longer allowed. You have to do ALTER FUNCTION RENAME TO kind of DDL statement.

lzap (changes requested):
With full migrations in place, let's make sure the table is final to avoid multiple migrations.

Can you rename the file to remove `_table` suffix? We don't use it in that directory.

Also move the function rename into the same migration, I think it's okay to do it in a single file.

lzap:
So can you rename `gcp_reservation_id` to `gcp_operation_name` (TEXT) and set the value, I am not sure how you can extract this from the Operation value but there must be way:

* https://cloud.google.com/compute/docs/api/how-tos/api-requests-responses#handling_api_responses
* https://cloud.google.com/compute/docs/reference/rest/v1/regionOperations/get

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func init() {
+	clients.GetGCPClient = newGCPClient
+}
+
+func newGCPClient(ctx context.Context) (clients.GCP, error) {
+	logger := ctxval.Logger(ctx).With().Str("client", "gcp").Logger()
+
+	instancesClient, err := compute.NewInstancesRESTClient(ctx)
```
lzap, line 29: For the record, there is instances endpoint: https://cloud.google.com/compute/docs/reference/rest/v1/instances/bulkInsert

And regionInstances endpoint which also supports bulk insert, but across regions: https://cloud.google.com/compute/docs/reference/rest/v1/regionInstances

internal/clients/http/gcp/gcp_client.go
```diff
@@ -0,0 +1,89 @@
...
+		},
+	}
+
+	c.logger.Trace().Msg("Executing Insert")
+	op, err := c.instancesClient.BulkInsert(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("cannot bulk insert instances: %w", err)
+	}
+
+	return ptr.To(op.Name()), nil
```
ezr-ondrej, line 88: 1. Do we need to return pointer?
2. if we do:
```suggestion
	return &op.Name(), nil
```

ezr-ondrej: ok, it was a nit ;)

ezr-ondrej (approved):
I'm happy here :+1:

internal/db/migrations/007_gcp_details.sql
```diff
@@ -0,0 +1 @@
+ALTER FUNCTION provider_type_gce () RENAME TO provider_type_gcp;
```
lzap, line 1: I would name this `007_rename_gce_type_function` or something, not like this.

internal/services/gcp_reservation_service.go
```diff
@@ -0,0 +1,139 @@
...
+func CreateGCPReservation(w http.ResponseWriter, r *http.Request) {
+	logger := ctxval.Logger(r.Context())
+
+	var accountId int64 = ctxval.AccountId(r.Context())
+
+	// TODO: meanwhile these values are hardcoded until we will have instance types, sources, ssh endpoints,
+	payload := &payloads.GCPReservationRequestPayload{
+		PubkeyID: 2,
+		// TODO: The project id change to source id from sources
+		SourceID:    "citric-expanse-361512",
```
lzap, line 28: I don't understand why we name a string which is obviously GCP Project ID as "SourceID", isn' this a typo? I would expect this to be named ProjectID instead. Even if sources name it badly, let's if this on our end.

scripts/rest_examples/reservation_create_gcp.http
```diff
@@ -0,0 +1,4 @@
+// @no-log
+POST http://{{hostname}}:{{port}}/{{prefix}}/reservations/gcp HTTP/1.1
+Content-Type: application/json
+X-Rh-Identity: {{identity}}
```
lzap, line 4: Nitpick: We use dashes for all files here, can you rename this?

internal/services/gcp_reservation_service.go
```diff
@@ -0,0 +1,139 @@
...
+	if ibErr != nil {
+		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "image builder client", ibErr))
+		return
+	}
+
+	// Get Image Name
+	name, ibErr := IBClient.GetGCPImageName(r.Context(), reservation.ImageID)
+	if ibErr != nil {
+		renderError(w, r, payloads.ClientError(r.Context(), "Image Builder", "can't get name from image builder", ibErr, 500))
+	}
```
lzap, line 108: This is missing return statement, it returns invalid reply in case of error:

```
{
  "msg": "Image Builder error: can't get name from image builder: fetch image status call: image compose not found: backend service returned not found (404) or no data",
  "request_id": "cchit927qo0jv3l0r19g"
}
{
  "reservation_id": 2,
  "pubkey_id": 1,
  "source_id": "redhat-360508",
  "zone": "us-central1-a",
  "machine_type": "n1-standard-1",
  "amount": 2,
  "image_id": "bc97177c-9d07-4db9-ad59-8b2c0bf4174e",
  "poweroff": true
}
```

internal/services/gcp_reservation_service.go
adiabramovitch: Project Id will be kept as a source, we will fetch that from sources.

adiabramovitch: so we will keep the source id and will fetch the project id like we did in was and arn.

lzap: Ah right! TY

lzap (approved):

lzap:
Merged thank, reminder, these are unchecked items from the description:

* Add dao tests for GCP reservation
* Add test for GCP reservation service

PR #220, opened 2022-09-13
Avoid appending when copying slices

lzap:
I have realized that we do unnecessary `append` calls when copying slices from one to the other when length is fixed (known before). In these cases, it's actually more efficient and more readable to allocate array buffer to the exact size and copy without append call.

@ezr-ondrej @avitova @adiabramovitch @amirfefer

### DIFF
--- a/internal/clients/http/sources/sources_client.go
+++ b/internal/clients/http/sources/sources_client.go
@@ -49,15 +49,6 @@ func newSourcesClient(ctx context.Context) (clients.Sources, error) {
 	return &sourcesClient{client: c, logger: logger}, nil
 }
 
-func copySource(src Source) clients.Source {
-	return clients.Source{
-		Id:           src.Id,
-		Name:         src.Name,
-		SourceTypeId: src.SourceTypeId,
-		Uid:          src.Uid,
-	}
-}
-
 type appType struct {
 	Id          string `json:"id"`
 	Name        string `json:"name"`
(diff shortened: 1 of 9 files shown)


internal/clients/http/sources/sources_client.go
```diff
@@ -83,7 +74,7 @@ func (c *sourcesClient) Ready(ctx context.Context) error {
 	return nil
 }
 
-func (c *sourcesClient) ListProvisioningSources(ctx context.Context) (*[]clients.Source, error) {
+func (c *sourcesClient) ListProvisioningSources(ctx context.Context) ([]*clients.Source, error) {
```
lzap, line 77: This was an overlook, it should be slice of pointers, not pointer to slice. It is actually required to be able to refactor copy loop in this function effectively.

internal/clients/http/sources/sources_client.go
```diff
@@ -49,15 +49,6 @@ func newSourcesClient(ctx context.Context) (clients.Sources, error) {
 	return &sourcesClient{client: c, logger: logger}, nil
 }
 
-func copySource(src Source) clients.Source {
```
ezr-ondrej, line 52: it looked bit more clean to me to have this separate, why did you inlined it? any specific reason?

lzap: Because it's not used anywhere else, we do a ton of copying in other places and we do it without helper functions.

ezr-ondrej (approved):
I :cupid: performance improvements, thanks @lzap :+1: 
This one even makes the code look bit cleaner :tada:

PR #221, opened 2022-09-13
Package ptr with generics

lzap:
Up until now we used the `ptr` package from smithy library. With Go 1.18, we can now take advantage of generics and have a single function `ptr.To` that will work with any type. This patch creates a new `internal/ptr` package with functions `To` and `From` for easy pointer copying. On top of that, there are also `ToInt64` (and 32) variants so you don't have to type `ptr.To(int64(42))` for constants because the default type for number is `int`.

Feel free to add new functions for other types (unsigned ints) or floats if needed: https://go.dev/tour/basics/11

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,11 +11,15 @@ linters:
     - sql
     - module
     - unused
+    - format
+    - import
   enable:
     - predeclared
   disable:
     - maligned # deprecated by fieldalignment
     - scopelint # deprecated by exportloopref
+    - gci # too strict
+    - gofumpt # too strict
 linters-settings:
   govet:
     check-shadowing: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,7 +10,6 @@ linters:
     - performance
     - sql
     - module
-    - unused
     - format
     - import
   enable:
@@ -20,9 +19,13 @@ linters:
     - scopelint # deprecated by exportloopref
     - gci # too strict
     - gofumpt # too strict
+    - contextcheck # too restrictive
 linters-settings:
   govet:
     check-shadowing: true
+  gosec:
+    excludes:
+      - G112 # slowloris attack - TODO (easyfix)
 issues:
   exclude-rules:
     # If a Stmt is prepared on a DB, it will remain usable for the lifetime of the DB. When the Stmt needs to execute
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -10,7 +10,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
-	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 	awsCfg "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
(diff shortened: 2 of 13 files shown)


ezr-ondrej:
our `fmt` still formats for go 1.16 I guess :)
TBH I do not see a big value in this usage of generic, not much less typing here, I mean it's great example of how to use generics, but the usage seems mostly the same as we use only String and int, where we still endup using different functions per type xD

lzap:
I understand that it does not look obvious when you see the diff, but it feels different. You simply type `ptr.To` and don't think about a type. Well, unless it's a integer constant, but that does not make it worse than it was before.

So I could not get the current GHA action working, but I realized that actually the golangci-lint does support both fmt and imports, so I enabled it and let's just use a single tool for everything.

lzap:
Since I renamed GHA job names, the status won't perhaps be reported at all. Or I am not sure why it's stuck.

.golangci.yml
```diff
@@ -10,15 +10,22 @@ linters:
...
     - scopelint # deprecated by exportloopref
+    - gci # too strict
+    - gofumpt # too strict
+    - contextcheck # too restrictive
 linters-settings:
   govet:
     check-shadowing: true
+  gosec:
+    excludes:
+      - G112 # slowloris attack - TODO (easyfix)
```
lzap, line 28: This is easy fix, will do in a different PR. Just two occurrences but new setting is needed.

.golangci.yml
```diff
@@ -10,15 +10,22 @@ linters:
     - performance
     - sql
     - module
-    - unused
+    - format
+    - import
```
lzap, line 14: These two (format, import) actually enables "go fmt" and "goimports" chekcing

The unused was accidentally added, I thought it is some kind of "unused code" check while it is actually a deprecated group.

ezr-ondrej:
> You simply type `ptr.To` and don't think about a type. Well, unless it's a integer constant, but that does not make it worse than it was before.

I understand this, but we are using only string and int at the moment and I see us only using those, because for other types, you usually have a variable and use dereference, so you'll always need to think about type, because string is `To` and integer is `ToIntXX` so the point of this code seems really to showcase usage of generics, which I love and hate from `C++` where after its introduction it looked cool as hell, but it made everything more confusing (especially debugging).

I'm not against this, I'm just saying we are now maintaining a code, that feels like it should be a shared go library (similar to `smithy-go/ptr`)

ezr-ondrej (approved):
I love the GH action improvements!
And I'm all for getting rid of unecessary dependencies and the code looks quite clean :+1:

lzap:
>  you'll always need to think about type

This statement is not correct. You only need to think about it, if you are making a pointer to local variable which you created via type inference which defaults to `int` or `float64`. It works fine for functions with variables passed in as arguments:

```go
func Function(x int64) {
 _ = ptr.To(x)
}
```

Granted we do have a bunch of them but mostly in tests/stubs, if you look closer to `RunInstances` function you will see that it actually gets easier and cleaner. Fun fact: I actually skipped these because I searched only for `ptr.` but this code actually uses `aws.Int32` which is a different package from AWS SDK to do the same thing. Let me fix that as well.

lzap:
I just pushed a commit that also changes the real EC2 client, not stubs/tests. This is where I think the new `To` and `From` functions really shine. :-)

internal/clients/http/ec2/ec2_client.go
```diff
@@ -167,8 +167,8 @@ func (c *ec2Client) RunInstances(ctx context.Context, name *string, amount int32
 				ResourceType: types.ResourceTypeInstance,
 				Tags: []types.Tag{
 					{
-						Key:   aws.String("Name"),
-						Value: aws.String(*name),
+						Key:   ptr.To("Name"),
+						Value: ptr.To(*name),
```
ezr-ondrej, line 171: Interesting code actually :thinking: :smile:

ezr-ondrej (approved):
Thanks @lzap ! I like the showcase of generics, lets get some hands-on with it :tada:

ezr-ondrej:
Thanks @lzap ! :+1:

PR #225, opened 2022-09-14
Fix Viper Default Values

avitova:
Updating a few lines.

### DIFF
--- a/internal/config/0_common_config.go
+++ b/internal/config/0_common_config.go
@@ -12,9 +12,9 @@ func init() {
 	// Database
 	parser.Viper.SetDefault("database.host", "localhost")
 	parser.Viper.SetDefault("database.port", 5432)
-	parser.Viper.SetDefault("database.name", "provisioning")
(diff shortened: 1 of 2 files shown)


internal/config/0_common_config.go
```diff
@@ -12,9 +12,9 @@ func init() {
 	// Database
 	parser.Viper.SetDefault("database.host", "localhost")
 	parser.Viper.SetDefault("database.port", 5432)
-	parser.Viper.SetDefault("database.name", "provisioning")
```
ezr-ondrej, line 15: I thought we are using this default in GH accion, but apparently not, so lets get rid of it ^_^

ezr-ondrej (approved):

lzap:
Perfect, thanks.

I wonder if you don't provide any config what happens. I would appreciate a validator that would nicely say "Please configure database in config/local.yaml or environment variables" instead of some random panic. But I haven't checked.

PR #226, opened 2022-09-14
Rename database to match stage

ezr-ondrej:


adiabramovitch (approved):

ezr-ondrej:
Ok, we need to revert, this works less then it worked before xD

PR #230, opened 2022-09-14
Enable gofumpt

lzap:
Looks like @adiabramovitch is leaving too many blank lines behind which makes me nervous.

While I was looking on updating golang-ci linter this morning, I have found out it also supports https://github.com/mvdan/gofumpt - a slightly stricter "go fmt". This patch enables the check for the linter and adds a makefile step to easily run both in case linter complains.

It looks like it's a great fit for our codebase because I only see the following changes (which would be enforced from now on):

* Variable grouping for package-level variables
* Blank lines between functions
* No blank lines at the start/end of a function
* Prefer variable inference (go fmt does the same but we had it disabled: `-s` = simplify code)


ezr-ondrej (approved):
Great addition :+1: test failures are almost for sure unrelated, so I'm ok to merge, but will wait for @adiabramovitch to check it, if she's something against it :)

lzap:
Rebased to retrigger tests.

adiabramovitch (approved):
Great! thanks @lzap

PR #232, opened 2022-09-14
Add Region To PubkeyResource Model

avitova:
We need a region for a pubkey. I felt this should belong to PubkeyResource. Correct me if I am wrong. I was not sure if the region should be required, but I felt like that should not be a problem. The value is prefilled for now.

### DIFF
--- a/internal/dao/sqlx/pubkey_resource_dao.go
+++ b/internal/dao/sqlx/pubkey_resource_dao.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	createPubkeyResource            = `INSERT INTO pubkey_resources (pubkey_id, provider, source_id, handle, tag) VALUES ($1, $2, $3, $4, $5) RETURNING id, tag`
+	createPubkeyResource            = `INSERT INTO pubkey_resources (pubkey_id, provider, source_id, handle, tag, region) VALUES ($1, $2, $3, $4, $5, $6) RETURNING id, tag`
 	deletePubkeyResourceById        = `DELETE FROM pubkey_resources WHERE id = $1`
 	listByPubkeyId                  = `SELECT * FROM pubkey_resources WHERE pubkey_id = $1`
 	getPubkeyResourceByProviderType = `SELECT * FROM pubkey_resources WHERE pubkey_id = $1 AND provider = $2`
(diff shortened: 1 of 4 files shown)


internal/db/migrations/004_pubkeys.sql
```diff
@@ -17,7 +17,8 @@ CREATE TABLE pubkey_resources
   pubkey_id BIGINT NOT NULL REFERENCES pubkeys(id),
   provider INTEGER NOT NULL CHECK (valid_provider(provider)),
   source_id TEXT NOT NULL,
-  handle TEXT NOT NULL
+  handle TEXT NOT NULL,
+  region TEXT NOT NULL
```
ezr-ondrej, line 21: I'm so sorry you've had to be the first, but we'll need to do a proper migration now that we are in stage :(

lzap: To elaborate, create a new SQL file with a new sequence numer and then do ALTER TABLE command to add a new column.

internal/dao/sqlx/pubkey_resource_dao.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	createPubkeyResource            = `INSERT INTO pubkey_resources (pubkey_id, provider, source_id, handle, tag) VALUES ($1, $2, $3, $4, $5) RETURNING id, tag`
+	createPubkeyResource            = `INSERT INTO pubkey_resources (pubkey_id, provider, source_id, handle, tag, region) VALUES ($1, $2, $3, $4, $5, 'us-east-1') RETURNING id, tag`
 	deletePubkeyResourceById        = `DELETE FROM pubkey_resources WHERE id = $1`
```
lzap, line 16: You also need to modify the create function, it need to set it.

ezr-ondrej (approved):
LGTM, thanks @avitova :+1:

@lzap? :)

internal/db/migrations/007_add_region.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE pubkey_resources
+ADD region text;
```
lzap, line 2: We are just about to merge a huge PR that adds 007 and 008 migration files, can you rebase and rename this to 009?

internal/models/pubkey_resource.go
```diff
@@ -26,6 +26,9 @@ type PubkeyResource struct {
 
 	// Resource handle (id). Format is provider-dependant. Required.
 	Handle string `db:"handle" json:"handle"`
+
+	// Required.
```
lzap, line 30:
```suggestion
	// Region name. This is provider-dependant. Required for providers which don't have global public keys.
```

lzap (changes requested):
You need to rebase because of migration conflict.

It would be nice to create custom types for both `Handle` and `Region` that would have `Provider` and `Value` fields and would support custom serialization to DB as a concatenated string (e.g. `provider-xxx`). But it's just a suggestion for future improvements, it's fine for now.

avitova:
I forgot to rebase, now this should be ready.

PR #233, opened 2022-09-15
Add a make/CI target to check migration changes

lzap:
This makefile/CI job will enforce that no migration files are modified, only added/deleted files are allowed. The makefile target has a nice explanation of what happened and guides you through the process of either skipping the check (by creating an extra/empty commit), or generating a new migration file using the tern utility.

It is executed on GH after integration tests, it's better to run it after because if someone skips the check via the extra commit, tests must be passing.

I also moved few things around in makefile, updated the makefile document markdown and fixed migration target which was still using gomigrate instead of the new tern utility.

### DIFF
--- a/.github/workflows/gotest.yml
+++ b/.github/workflows/gotest.yml
@@ -58,7 +58,7 @@ jobs:
           DATABASE_USER: postgres
           DATABASE_PASSWORD: foobar
           DATABASE_NAME: provisioning_test
-        run: make integration-test
+        run: make integration-test migrations
 
   openapi:
     name: "🪆 Generated code diff check"
--- a/.github/workflows/gotest.yml
+++ b/.github/workflows/gotest.yml
@@ -58,7 +58,7 @@ jobs:
           DATABASE_USER: postgres
           DATABASE_PASSWORD: foobar
           DATABASE_NAME: provisioning_test
-        run: make integration-test migrations
+        run: make integration-test check-migrations
 
   openapi:
     name: "🪆 Generated code diff check"
--- a/mk/z_common.mk
+++ b/mk/z_common.mk
@@ -4,13 +4,6 @@
 
 MAKE_DOC=docs/make.md
 
-.PHONY: validate-make
-validate-make:
-	echo '# Make documentation' > $(MAKE_DOC)
-	echo '```' >> $(MAKE_DOC)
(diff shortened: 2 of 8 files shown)


mk/z_common.mk
```diff
@@ -4,13 +4,6 @@
 
 MAKE_DOC=docs/make.md
 
-.PHONY: validate-make
-validate-make:
-	echo '# Make documentation' > $(MAKE_DOC)
-	echo '```' >> $(MAKE_DOC)
-	make help | sed -r "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g" >> $(MAKE_DOC)
-	echo '```' >> $(MAKE_DOC)
-
```
lzap, line 13: I moved this into the more appropriate `help.mk`.

scripts/check_migrations.sh
```diff
@@ -0,0 +1,20 @@
...
+  echo
+  echo "If you still want to pass CI tests, create an additional commit to get this"
+  echo "check passing as it only checks the latest one. Use with care!"
+  echo
+  echo "The following files were modified:"
+  echo
+  echo $CHANGED
+  echo $CHANGED_COMMITED
+  exit 1
+fi
```
lzap, line 20: To test this, modify one of the existing migrations and run: `make fmt` :-)

How cool is that? ;-)

.github/workflows/gotest.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           DATABASE_USER: postgres
           DATABASE_PASSWORD: foobar
           DATABASE_NAME: provisioning_test
-        run: make integration-test
+        run: make integration-test migrations
```
ezr-ondrej, line 61:
```suggestion
        run: make integration-test check-migrations
```

ezr-ondrej (approved):
Cool! love it! :)
Just one typo, and we're good to merge I believe :)

.github/workflows/gotest.yml
lzap: Damn, good eyes. I renamed this just as the last thing I thought the prefix is more clear.

lzap:
Rebased.

lzap:
All green, thanks. Let's get this in ASAP.

PR #234, opened 2022-09-16
Refactor Sources and Cloud provider APIs

lzap:
_Warning: This is a monster patch, it's not my fault my cat jumped on enter and squashed all commits..._

Jokes aside, it's big because nobody was working, PTOs and sick and hackathon. So as the PR got bigger and bigger, I kept adding more and more changes. In general, the theme of this patch is **refactoring of our Sources integration and cloud clients interface** and worry do not, I will write an extensive explanation of what is going on in the PR. Before you start any code changes around sources or ec2/azure/gcp beware of possible conflicts with this PR. @adiabramovitch @avitova @ezr-ondrej @amirfefer 

**Generalizing sources ARN**

Previously, our Sources API was designed around AWS and its ARN token (string). Since Sources are a generic service and can hold authentication artifacts of any kind, I created a new generic interface `clients.Authentication`. You can think of it as basically a wrapper for ARN/SubscriptionID/ProjectID. It carries a payload (string) and a provider type. There are some helper functions to make decisions easier.

The `GetArn` function therefore changes to `GetAuthentication` and callers need to decide if they only want the payload which can be fetched via `String()` function, or to do some decisions based on type. It is worth nothing that I found a bug in Sources OpenAPI (missing type) which doesn't allow to retrieve authentication type from their API via generated clients, so I fixed it and this PR contains the updated OpenAPI spec. I am assuming it gets merged soon: https://github.com/RedHatInsights/sources-api-go/pull/566 and we can simply merge this as-is because the next time we sync OpenAPI specs it will get pulled too.

**Improving our cloud interface**

When working on the new Azure interface, I realized that we don't pass context in all functions and we keep logger and context in the client struct itself. I don't think this is a good practice, tho, I think it was me who created this. This patch makes sure that:

* All functions in all cloud interfaces (ec2, azure, gcp) accepts Context as the first parameter.
* All implementations has a `logger` function that extracts and configures logger for local use.
* Client struct only has credentials information, clients themselves are always initialized from functions. This is important for GCP since that is not a REST API, it's gRPC HTTP/2 protocol which requires explicit closing, so this is much safer.
* All implementation-clients are created via `newXXXClient` functions.
* All errors are wrapped properly and root cause is always returned to the caller.

**Status call for each source**

To properly test this patch with all three providers, I had to have some call available for Azure. Actually, this feature was actually the initial thought thus the branch name. What this does it it adds a new endpoint `/sources/ID/status` which is very simple. It returns 200 if the source is active and correctly configured (have valid credentials), 500 otherwise. It performs a quick "ping" call, but since none of the three cloud APIs have a ping/status call, I use something that is fast: "get subscription detail" for Azure and "list all regions" for others.

To be actually able to implement those status requests, new functions are available for each implementation: `ListAllRegions`, `ListAllZones` and `ListAllRegionsAndZones` for GCP (it's just a single call in that API). These functions are exported but unused at the moment, the plan is to use them in the `typesctl` command line utility to actually extract the data into YAML. New types were created in the `clients` package: `Region` and `Zone`, both are strings for the moment but we might want to expand this (e.g. there are multiple Zone types).

Now, to test this you simply hit `/sources/1/status` or 2 or 3 depending on your source number (I have three, 1 for EC2, 2 for Azure and 3 for GCP). I created `source-X-status.http` file that can be used to do the call, it's just a simple GET with no payload. Before that, you need to update sources backend app to the latest version, edit your `sources.local.conf` (see sources.conf for more details about ENV variables) and run the `sources.seed.sh` script. This will create you sources 1, 2 and 3 that can be then used nicely with these `.http` files directly from your IDE.

You will find a new small interface called `Statuser` which has just a single function `Status(ctx)` and this is embedded into all three interfaces. This allows a nice and generic code in the status service function. This is idiomatic Go, these small interfaces are very often called Somethinger (Reader, Stringer etc in the standard library).

**Configuration changes**

The patch adds new GCP configuration tree with `json` field which must be base64-encoded credentials GCP JSON. I propose this solution because I think it's the cleanest way of providing a string full of quotes via environment variables. On Linux, there is typically very big limit on env var length (32kB or something), my own GCP JSON is around 3k characters encoded. So as long as clowder and k8s supports this, it's the best option. There is also a validator that immediately stops the app if the value was not decoded correctly.

All cloud provider configuration now has default region setting (zone for GCP). This setting can be used for requests which are non-regional (e.g. global, typically fetching instance types or other generic info). It should be as close as possible to the datacenter that production runs in. My wild guess is useast1, therefore this is the default value.

I am moving the instance prefix from AWS tree to the APP tree, therefore AWS_INSTANCE_PREFIX will be now APP_INSTANCE_PREFIX. Reason being we want the same setting to have for all clouds typically (for development and testing), therefore there's no sense to configure this separately as this will almost always be `test-` anyways. @RHEnVision/qe please keep this in mind.

### DIFF
--- a/configs/local_example.yaml
+++ b/configs/local_example.yaml
@@ -36,14 +36,16 @@ app:
     # Individual application cache features
     typeAppId: true
     account: true
+  # Instance name prefix for all providers (useful only for testing)
+  instancePrefix:
 
 # AWS service account credentials
 aws:
   key:
   secret:
   session:
-  # instance name prefix for all VMs created via the app
-  #instancePrefix:
+  # default region for non-regional requests
+  defaultRegion: us-east-1
 
 # Azure service account credentials
 azure:
@@ -55,6 +57,15 @@ azure:
   clientId: 00000000-0000-0000-0000-000000000000
   # Secret value
   clientSecret:
+  # default region for non-regional requests
+  defaultRegion: eastus
+
+# GCP service account credentials
+gcp:
+  # Base64-encoded credentials JSON file (defaults to empty JSON)
+  json:
+  # default region for non-regional requests
+  defaultZone: us-east1-b
 
 # AWS CloudWatch logger
 cloudwatch:
--- a/internal/clients/authentication.go
+++ b/internal/clients/authentication.go
@@ -1,0 +1,62 @@
+package clients
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+)
+
+type Authentication struct {
+	ProviderType models.ProviderType `json:"type"`
+	Payload      string              `json:"payload"`
+}
+
+func NewAuthentication(str string, provType models.ProviderType) *Authentication {
+	a := Authentication{
+		Payload:      str,
+		ProviderType: provType,
+	}
+	return &a
+}
+
+func NewAuthenticationFromSourceAuthType(ctx context.Context, str, authType string) *Authentication {
+	a := Authentication{Payload: str}
+	switch authType {
+	case "provisioning-arn":
+		a.ProviderType = models.ProviderTypeAWS
+	case "provisioning_lighthouse_subscription_id":
+		a.ProviderType = models.ProviderTypeAzure
+	case "provisioning_project_id":
+		a.ProviderType = models.ProviderTypeGCP
+	default:
+		ctxval.Logger(ctx).Warn().Msgf("Unknown auth type returned from sources: %s", authType)
+		a.ProviderType = models.ProviderTypeUnknown
+	}
+	return &a
+}
+
+// Type returns authentication provider type
+func (auth *Authentication) Type() models.ProviderType {
+	return auth.ProviderType
+}
+
+// Is checks if Authentication is of a given provider type
+func (auth *Authentication) Is(providerType models.ProviderType) bool {
+	return auth.ProviderType == providerType
+}
+
+// MustBe returns nil, if authentication is of given type. Otherwise, returns an error.
+func (auth *Authentication) MustBe(providerType models.ProviderType) error {
+	if !auth.Is(providerType) {
+		return fmt.Errorf("%w: %s", UnknownAuthenticationTypeErr, auth.ProviderType.String())
+	}
+
+	return nil
+}
+
+// String returns authentication payload string (ARN, Subscription UUID, Project-ID...)
+func (auth *Authentication) String() string {
+	return auth.Payload
+}
--- a/internal/clients/errors.go
+++ b/internal/clients/errors.go
@@ -3,8 +3,12 @@ package clients
 import "errors"
 
 var (
+	// Common errors
 	NotFoundErr       = errors.New("backend service returned not found (404) or no data")
 	UnauthorizedErr   = errors.New("backend service returned unauthorized (401)")
 	ForbiddenErr      = errors.New("backend service returned forbidden (403)")
 	Non2xxResponseErr = errors.New("backend service did not return 2xx")
+
+	// Sources errors (some others are defined in http package too)
+	UnknownAuthenticationTypeErr = errors.New("unknown authentication type")
 )
--- a/internal/clients/http/azure/azure_client.go
+++ b/internal/clients/http/azure/azure_client.go
@@ -1,0 +1,77 @@
+package azure
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/rs/zerolog"
+)
+
+type azureClient struct {
+	subscriptionID string
+	credential     *azidentity.ClientSecretCredential
+}
+
+func init() {
+	clients.GetAzureClient = newAzureClient
+}
+
+func logger(ctx context.Context) zerolog.Logger {
+	return ctxval.Logger(ctx).With().Str("client", "azure").Logger()
+}
+
+func newAzureClient(ctx context.Context, auth *clients.Authentication) (clients.Azure, error) {
+	opts := azidentity.ClientSecretCredentialOptions{}
+	identityClient, err := azidentity.NewClientSecretCredential(config.Azure.TenantID, config.Azure.ClientID, config.Azure.ClientSecret, &opts)
+	if err != nil {
+		return nil, fmt.Errorf("unable to init Azure credentials: %w", err)
+	}
+
+	return &azureClient{
+		subscriptionID: auth.Payload,
+		credential:     identityClient,
+	}, nil
+}
+
+func (c *azureClient) newResourcesClient(ctx context.Context) (*armresources.Client, error) {
+	client, err := armresources.NewClient(c.subscriptionID, c.credential, nil)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create resources Azure client: %w", err)
+	}
+	return client, nil
+}
+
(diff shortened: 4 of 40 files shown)


internal/routes/all_routes.go
```diff
@@ -57,6 +57,9 @@ func apiRouter() http.Handler {
 		r.Route("/sources", func(r chi.Router) {
 			r.Get("/", s.ListSources)
 			r.Route("/{ID}", func(r chi.Router) {
+				r.Get("/status", s.SourcesStatus)
+
+				// TODO move this to outside of /sources (see below)
```
adiabramovitch, line 62: Why do you want to remove it outside?

adiabramovitch, line 62: Why do you want to remove it outside?

internal/clients/authentication.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+)
+
+type Authentication interface {
+	// Is checks if Authentication is of a given provider type
+	Is(providerType models.ProviderType) bool
+
+	// MustBe returns nil, if authentication is of given time. Otherwise, returns an error.
```
adiabramovitch, line 15: What do you mean by "of given time"?

internal/clients/authentication.go
```diff
@@ -0,0 +1,65 @@
...
+	MustBe(providerType models.ProviderType) error
+
+	// String returns authentication string (ARN, Subscription UUID, Project-ID...)
+	String() string
+}
+
+type auth struct {
+	ProviderType models.ProviderType `json:"type"`
+	Payload      string              `json:"payload"`
```
adiabramovitch, line 24: I  am not sure payload is the best name here

internal/clients/authentication.go
```diff
@@ -0,0 +1,65 @@
...
+	return &auth
+}
+
+func (auth *auth) Is(providerType models.ProviderType) bool {
+	return auth.ProviderType == providerType
+}
+
+func (auth *auth) MustBe(providerType models.ProviderType) error {
+	if !auth.Is(providerType) {
+		return fmt.Errorf("%w: %s", UnknownAuthenticationTypeErr, auth.ProviderType)
```
adiabramovitch, line 57: I guess you need to change it to %v

adiabramovitch (commented):
First round :)

internal/routes/all_routes.go
lzap: The plan for instance types is to preload them, see below, this is already implemented for Azure (`/instance_types/azure`) and I hope @avitova can work on refactoring AWS in the same way. According to testing, loading up instances from embedded YAML is < 1ms compared to 6 seconds of live loading.

internal/clients/authentication.go
lzap: I am not sure either, but recommend a better alternative. It's either ARN or subscription id or project id.

internal/clients/authentication.go
lzap: Right, weird that lint does not warn about this. Also provider type did not implement `String()` so it was also an issue.

lzap:
That's all? I was really hoping for more :-)

I am going to fix tests and rebase.

internal/clients/authentication.go
lzap: Oh actually no, the linter was only complaining about the 2nd parameter, the first one needs to be wrapped error.

adiabramovitch:
> That's all? I was really hoping for more :-)
> 
> 
> 
> I am going to fix tests and rebase.

Just the first round 😊

internal/clients/errors.go
```diff
@@ -3,8 +3,12 @@ package clients
...
 var (
+	// Common errors
 	NotFoundErr       = errors.New("backend service returned not found (404) or no data")
 	UnauthorizedErr   = errors.New("backend service returned unauthorized (401)")
 	ForbiddenErr      = errors.New("backend service returned forbidden (403)")
 	Non2xxResponseErr = errors.New("backend service did not return 2xx")
+
+	// Sources errors (some others are defined in http package too)
+	UnknownAuthenticationTypeErr = errors.New("unknown authentication type")
 )
```
lzap, line 14: I do not want to make more refactorings here, but I believe we should move all errors from the `http` package to the `clients` package at some point.

lzap:
This is now ready for full review, I tested this on all three providers, the status call returns 200 for all of them.

configs/local_example.yaml
```diff
@@ -46,6 +48,15 @@ azure:
   clientId: 00000000-0000-0000-0000-000000000000
   # Secret value
   clientSecret:
+  # default region for non-regional requests
+  defaultRegion: useast
```
lzap, line 52: This needs to be `eastus` this was a typo.

mk/compile.mk
```diff
@@ -33,3 +33,7 @@ run: pbapi ## Build and run backend API
 .PHONY: clean
 clean: ## Clean build artifacts
 	rm pbapi pbmigrate pbworker
+
+.PHONY: generate
+generate: ## Run go generate tool (stringer etc)
+	go generate ./...
```
adiabramovitch, line 39: What is the goal of the target?

adiabramovitch:
tried /retest but didn't trigger any test so maybe I don't have the permissions? @mshriver @lzap

internal/clients/http/azure/azure_client.go
```diff
@@ -0,0 +1,81 @@
...
+	client, err := c.newSubscriptionsClient(ctx)
+	if err != nil {
+		return fmt.Errorf("unable to initialize status request: %w", err)
+	}
+	_, err = client.Get(ctx, c.subscriptionID, nil)
+	if err != nil {
+		return fmt.Errorf("unable to perform status request: %w", err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
lzap: To "ping" Azure, they don't provide any "ping" or "status" call, so what I do is I do a simple GET my subscription call which appears to be quite fast.

internal/clients/http/azure/azure_client.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/rs/zerolog"
+)
+
+type azureClient struct {
+	logger         zerolog.Logger
```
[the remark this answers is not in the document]
lzap: Oh an overlook, my intention was to remove these from all clients. Fixing.

internal/config/0_common_config.go
```diff
@@ -52,4 +52,12 @@ func init() {
 	parser.Viper.SetDefault("aws.secret", "")
 	parser.Viper.SetDefault("aws.session", "")
 	parser.Viper.SetDefault("aws.instancePrefix", "")
+	parser.Viper.SetDefault("aws.defaultRegion", "us-east-1")
+
+	// Azure
+	parser.Viper.SetDefault("azure.defaultRegion", "useast")
```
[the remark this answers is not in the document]
lzap: Good catch, yes!

internal/clients/http/gcp/gcp_client.go
```diff
@@ -14,37 +18,101 @@ import (
 )
 
 type gcpClient struct {
-	instancesClient *compute.InstancesClient
-	context         context.Context
-	logger          zerolog.Logger
```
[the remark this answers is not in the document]
lzap: Here I removed logger (as well as other fields) and replaced it only with auth and options. This is intentional.

internal/config/main_config.go
```diff
@@ -167,6 +188,10 @@ func IsProduction() bool {
 	return Features.Environment == "production"
 }
 
+func GCPCredentials() []byte {
+	return gcpCredentialsJSON
+}
+
```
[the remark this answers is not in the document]
lzap: It's a different type now, it was string, now this is byte slice. But you know what, let's convert it back to string and then convert when it's needed, that's a good suggestion.

mk/compile.mk
lzap: A leftover from my experiments, removing.

lzap:
Pushed @adiabramovitch 's remarks.

adiabramovitch:
@lzap  please rebase and hopefully the ci will be green 😄

lzap:
@RHEnVision/qe any idea why this fails?

lzap:
AWS EC2 reservation does not work, it's a bug! Let me fix that.

lzap:
All green now, thanks all!

PR #235, opened 2022-09-20
HMSPROV-285: Reservation detail for AWS

lzap:
Although I thought for all filtering we would use URL parameters and provider type can be seen as a filter, because of OpenAPI, we must use the provider in the URL because payloads can be different only for different URLs in OpenAPI.

This adds a new endpoint `/reservations/TYPE/ID` that returns reservation with all the details loaded, currently it only works for AWS but the code is ready for extension.

I am leaving the old `/reservations/ID` still active, it returns only generic information and it can be useful for UI to get just basic info for the progress bar. Should be a tad faster as it does not need to join the two tables, so it's a better fit for polling.

There is still the open question for possible conflict between `TYPE` and `ID`, but I don't see this as a problem until someone complains about this, it's `GET` vs `POST` so it should be fine.

There are new `.http` files to test this feature.

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -391,7 +391,7 @@
     },
     "/reservations": {
       "get": {
-        "description": "A reservation is a way to activate a job, keeps all data needed for a job to start. This operation returns list of all reservations for particular account.\n",
+        "description": "A reservation is a way to activate a job, keeps all data needed for a job to start. This operation returns list of all reservations for particular account. To get a reservation with common fields, use /reservations/ID. To get a detailed reservation with all fields which are different per provider, use /reservations/aws/ID.\n",
         "operationId": "getReservationsList",
         "responses": {
           "200": {
@@ -414,6 +414,40 @@
       }
     },
     "/reservations/aws": {
+      "get": {
+        "description": "Return an AWS reservation with details by id",
+        "operationId": "getAWSReservationByID",
+        "parameters": [
+          {
+            "description": "Reservation ID, must be an AWS reservation otherwise 404 is returned",
+            "in": "path",
+            "name": "ID",
+            "required": true,
+            "schema": {
+              "format": "int64",
+              "type": "integer"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/v1.AWSReservationResponse"
+                }
+              }
+            },
+            "description": "Returns detailed reservation information for an AWS reservation."
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound"
+          },
+          "500": {
+            "$ref": "#/components/responses/InternalError"
+          }
+        }
+      },
       "post": {
         "description": "A reservation is a way to activate a job, keeps all data needed for a job to start. An AWS reservation is a reservation created for an AWS job. Image Builder UUID image is required, the service will also launch any AMI image prefixed with \"ami-\".\n",
         "operationId": "createAwsReservation",
@@ -466,7 +500,7 @@
     },
     "/reservations/{ID}": {
       "get": {
-        "description": "Return a reservation by id",
+        "description": "Return a generic reservation by id",
         "operationId": "getReservationByID",
         "parameters": [
           {
@@ -489,7 +523,7 @@
                 }
               }
             },
-            "description": "Returned on success."
+            "description": "Returns generic reservation information like status or creation time."
           },
           "404": {
             "$ref": "#/components/responses/NotFound"
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -414,6 +414,36 @@
       }
     },
     "/reservations/aws": {
+      "post": {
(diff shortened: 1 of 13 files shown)


internal/models/reservation_model.go
```diff
@@ -74,10 +92,10 @@ type AWSReservation struct {
 	AWSReservationID string `db:"aws_reservation_id" json:"aws_reservation_id"`
 
 	// The ID of the image from which the instance is created. AMI's must be prefixed with 'ami-'.
-	ImageID string `json:"image_id"`
+	ImageID string `db:"image_id" json:"image_id"`
 
 	// Detail information is stored as JSON in DB
-	Detail *AWSDetail `db:"detail" json:"detail"`
+	Detail *AWSDetail `json:"detail"`
```
adiabramovitch, line 98: Why did you remove the 'db'?

api/openapi.gen.json
```diff
@@ -414,6 +414,40 @@
       }
     },
     "/reservations/aws": {
```
akhil-jha, line 416: Missing {ID} I think.
`/reservations/aws/{ID}`

lzap: Right, good catch. Fixing.

internal/models/reservation_model.go
lzap: Oh this was not intentional, fixing. I was playing around with this, it did not work well with pgx/sqlx and I left a TODO comment on two functions which I had to implement. If we used pgx/scanny natively it would work out of box as pgx driver supports all of the postgres native types (this is JSONB).

internal/dao/dao_interfaces.go
```diff
@@ -73,6 +73,9 @@ type ReservationDao interface {
 	// GetById returns reservation for a particular account.
 	GetById(ctx context.Context, id int64) (*models.Reservation, error)
 
+	// GetAWSById returns reservation for a particular account.
+	GetAWSById(ctx context.Context, id int64) (*models.AWSReservation, error)
+
```
adiabramovitch, line 78: Maybe we should also add tickets for creating those methods for GCP and Azure?

internal/models/reservation_model.go
```diff
@@ -61,6 +65,31 @@ type AWSDetail struct {
 	PowerOff bool `json:"poweroff"`
 }
 
+// TODO: Use pgx native driver with scany library instead of sqlx which does not
```
adiabramovitch, line 68: Can you elaborate?

api/openapi.gen.json
akhil-jha: It did get fixed, and it's working,  but with the latest commit, I am getting the following for `GET /reservations/{ID}`
```
*** iqe_provisioning_api.exceptions.ApiTypeError: Invalid type for variable 'received_data'. Required value type is V1ReservationResponse and passed type was str at ['received_data']
```

lzap: Ok this is a bug, working on this!

internal/dao/dao_interfaces.go
lzap: Yeah, well, we will eventually figure this out when implementing these features :-)

internal/services/reservations_service.go
```diff
@@ -55,31 +55,56 @@ func ListReservations(w http.ResponseWriter, r *http.Request) {
...
-func GetReservation(w http.ResponseWriter, r *http.Request) {
+func renderNotFoundOrDAOError(w http.ResponseWriter, r *http.Request, err error, daoError string) {
+	var e dao.NoRowsError
+	if errors.As(err, &e) {
+		renderError(w, r, payloads.NewNotFoundError(r.Context(), err))
+	} else {
+		renderError(w, r, payloads.NewDAOError(r.Context(), daoError, err))
+	}
+}
+
```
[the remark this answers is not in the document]
lzap: Absolutely, we already have `error_renderer.go` in that package, let's move it there.

lzap:
Fixed the problem with the old endpoint, moved the code as @adiabramovitch suggested.

adiabramovitch (approved):
@lzap Thanks 😄 we need @RHEnVision/qe ack

lzap:
@akhil-jha please merge once you confirm it works, thanks!

akhil-jha (approved):

internal/db/migrations/010_rename_provider_detail.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE aws_reservation_details RENAME provider TO provider_fk;
```
ezr-ondrej, line 1: Why are we renaming to `_fk`? Sounds like a weird col name, I'd expect constraint having this name

lzap: There was a conflicts for the joined tables, otherwise we would need to drop "*" from the SQL replacing it with 13 columns. I thought I would rename instead of this.

ezr-ondrej: ok, make sense, I'd still love less confusing name, but it's fine, not a big deal at all, just curious :) Thanks for the explanation :+1:

PR #240, opened 2022-09-26
Implement EC2 types precaching

lzap:
This adds a new endpoint `/instance_types/aws` that returns pre-cached types generated via the `typesctl` utility. It looks like EC2 have the very same instance types across zones and also across regions, so all the YAML files are the same. But the code is generic, so it's not big deal to have few more files in memory.

I had to add some more permissions to the service account, everything is documented see docs/ directory. Also, I found that some regions are locked behind an explicit confirmation because of backward compatibility of STS v2 token. This is also documented, otherwise the generation will fail with authorization error. Finally, the app now must explicitly use endpoint that is hard-coded in the configuration - use the closest datacenter available. Without this, it does not work.


lzap:
As part of this PR, I also regenerated types for Azure because the code now sorts the list, this will be more user-friendly when doing diffs and it looks like AWS always return them in random order. This is why you see a lot of changes for Azure types.

internal/clients/architecture.go
```diff
@@ -10,7 +10,8 @@ import (
 type ArchitectureType string
 
 const (
-	ArchitectureTypeI386       ArchitectureType = "i386"
+	ArchitectureTypeI386 ArchitectureType = "i386"
+	// TODO rename to X86_64
 	ArchitectureTypeX8664      ArchitectureType = "x86_64"
```
lzap, line 15: I can do this after review.

docs/configure-amazon-role.md
```diff
@@ -38,13 +38,23 @@ The policy JSON to use in step above:
...
+        },
+        {
+            "Sid": "DescribeGenericData",
+            "Effect": "Allow",
+            "Action": [
+                "ec2:DescribeInstanceTypeOfferings",
+                "ec2:DescribeRegions",
+                "ec2:DescribeInstanceTypes"
+            ],
+            "Resource": "*"
```
ezr-ondrej, line 57: shall we split this off to a second policy? I believe this is only necessary for the development accounts that do want to refetch the instance types, right?

ezr-ondrej (approved):
I've got one documentation nit, but apart of it, it's great, I'd go ahead and merge this and we can even start using it in the UI I guess :)

internal/clients/http/azure/types/azure_types.go
```diff
@@ -31,41 +29,14 @@ func init() {
...
-		file := filepath.Join(availabilityPath, dirEntry.Name())
-		buffer, errBuf := fsTypes.ReadFile(file)
-		if errBuf != nil {
-			panic(errBuf)
-		}
-		availBuf.Write(buffer)
-	}
-	etag, err = middleware.GenerateETagFromBuffer("azure-types", 30*time.Minute, typesBuf, supportedBuf, availBuf.Bytes())
+	availBuf := clients.ConcatBuffers(fsTypes, availabilityPath)
+	etag, err = middleware.GenerateETagFromBuffer("azure-types", 30*time.Minute, typesBuf, availBuf)
```
avitova, line 39: I know this was there before, but why is there '30 minutes'? This could get less confusing using a constant.

avitova (commented):
This works and looks great.

docs/configure-amazon-role.md
lzap: Perhaps it's a good practice overall, but here there is no risk. Let me split this sure.

lzap: You know what, after I edited the file, it feels a lot more complex, you cannot simply copy and paste a single JSON anymore. Let's keep that in mind and if there are more dangerous permissions added, let's split this. What you think?

internal/clients/http/azure/types/azure_types.go
lzap: Absolutely valid point, it is the time all user browsers will consider the reply to cached. So when we roll out a new version of the application and any data changes within a list, the customers will see the change after 30 minutes in the worst case.

This was my wild guess, I was thinking when someone calls our support saying "my type is missing" so we would add it. Now that I think about it, I think we can do something like several hours. Let me extract this into a constant.

lzap:
Amended @avitova 's point by introducing new two constants. I don't think we need to have a configurable value for this, it is unlikely to ever change. I am going for 4 hours because that is a decent timeframe.

lzap:
By the way, I forgot to mention this but this patch also filters out unsupported instances (< 1.5 GiB RAM). So it actually implements what we want.

lzap:
An error on the CI: `request send failed, Post \"[https://STS..amazonaws.com/\":](https://sts..amazonaws.com//) dial tcp: lookup STS..amazonaws.com: no such host`

Hmmm looks like some configuration issue, looking into this.

lzap:
Yeah so the `signingRegion` had no default value, I meant it to default to `us-east-1` this was a bug. Rebased.

lzap:
All good!

lzap:
Since this has been reviewed, tests are passing and I need to start working on some changes on top of this PR, I am merging. Thanks!

PR #243, opened 2022-10-03
Unify multiple versions of same libraries

lzap:
Also I looked into the SPEC generator, it uses a non-standard YAML library but I have found that it leverages a feature not possible via the standard library, so I am keeping the code as-is, tho I improved variable naming it was a bit messy.

Finally, I performed an overall deps bump in the third commit.


ezr-ondrej (approved):
Thanks @lzap, nice cleanup :+1:

PR #245, opened 2022-10-04
Add the new instance_types to OpenAPI

lzap:
SSIA

Note: only EC2 and Azure are implemented. Also keep in mind that region and zone are required only for some clouds:

* EC2: region required, zone not (must not be passed in)
* Azure: region required, zone required

### DIFF
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -118,7 +118,7 @@ paths:
           $ref: "#/components/responses/InternalError"
   /sources/{ID}/instance_types:
       get:
-        description: 'Return a list of instance types'
+        description: 'Return a list of instance types (DEPRECATED: use /instance_types)'
         operationId: getInstanceTypeList
         parameters:
         - in: path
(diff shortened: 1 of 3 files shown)


lzap:
Providing zone for EC2 will likely error out, we should work on a nicer error when that happens :-)

cmd/spec/path.yaml
```diff
@@ -147,6 +147,48 @@ paths:
...
             $ref: "#/components/responses/InternalError"
+  /instance_types/{TYPE}:
+    get:
+      description: >
+        Return a list of instance types for particular provider. A region must be provided. A zone must be provided
+        for Azure.
+      operationId: getInstanceTypeListAll
+      parameters:
+        - in: path
+          name: TYPE
```
ezr-ondrej, line 158: Wouldn't `PROVIDER` be more descriptive?

lzap:
Oh yeah that's what I meant, way a typo, PROVIDER it is.

ezr-ondrej (approved):
Works for me, I guess we want to wait for the tests to be switched to this endpoint tho, right?
@akhil-jha ?

lzap:
Ignore the above comment, it was meant for a different PR :-) Carry on...

akhil-jha (approved):

PR #246, opened 2022-10-04
Regenerate types

lzap:
Big bang generation of new types.

For Azure, due to recent change the list is sorted, therefore lots of changes.

For EC2, most of the regions were missing so new files.

I think after this is merged, regeneration patches will be much smaller.


ezr-ondrej (approved):
I refuse to review, my browser is crashing if I try. But I believe it was auto generated and there is no other change, so I'm all for it :)) I agree that next time it will be likely easier to review :+1:

PR #247, opened 2022-10-05
Signing region fix

lzap:
Oh this was tough, I was doing it wrong. After reading https://aws.amazon.com/premiumsupport/knowledge-center/iam-validate-access-credentials/ I thought that I have to use regional STS service instead of global one even after I enabled v2 tokens for all regions on my account. This is why I implemented the endpoint callback function.

In reality, that was not the case. After I added logging for the aws client, I could clearly see that STS token is NOT acquired at all during instance type generation. It was simply not allowing me access to some regions! And I have found that indeed some regions must be explicitly enabled: https://docs.aws.amazon.com/general/latest/gr/rande-manage.html

I also have found that SDK v2 by default use regional STS anyway, not the global one. Therefore in this patch I am dropping the SigninRegion, it is not needed. New configuration option Logging was added, it is turned off by default and beware that enabling it will cause a lot of trace log messages, it is perhaps only useful in development only.

The aws client creation code was simplified, there is now a helper function and since regional STS is used, no endpoint functions are necessary. The root cause of why it did not work was that I had some regions disabled. After I enabled them, it generated a lot more types:

```
12:25PM TRC Number of instance types returned: 137, after filtering: 137 hostname=mone.home.lan
12:25PM TRC Number of instance types returned: 230, after filtering: 230 hostname=mone.home.lan
12:25PM TRC Number of instance types returned: 370, after filtering: 374 hostname=mone.home.lan
12:25PM TRC Number of instance types returned: 278, after filtering: 282 hostname=mone.home.lan
12:25PM TRC Number of instance types returned: 345, after filtering: 349 hostname=mone.home.lan
12:25PM TRC Number of instance types returned: 240, after filtering: 240 hostname=mone.home.lan
12:25PM TRC Number of instance types returned: 549, after filtering: 558 hostname=mone.home.lan
12:25PM TRC Number of instance types returned: 136, after filtering: 140 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 340, after filtering: 344 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 156, after filtering: 156 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 484, after filtering: 493 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 118, after filtering: 118 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 335, after filtering: 344 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 340, after filtering: 344 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 194, after filtering: 194 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 424, after filtering: 433 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 413, after filtering: 422 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 123, after filtering: 123 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 470, after filtering: 475 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 557, after filtering: 566 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 500, after filtering: 504 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 332, after filtering: 341 hostname=mone.home.lan
12:26PM TRC Number of instance types returned: 557, after filtering: 566 hostname=mone.home.lan
```

For easier review there are three commits.


ezr-ondrej (approved):
Thanks, code make sense and it gives back my t1.micro ! :tada: xD

PR #249, opened 2022-10-05
Add otel to EC2 client

lzap:
Only to the most important functions.


lzap:
Actually stop, I forgot one thing. When there is an error, you need to manually set error into the telemetry span, otherwise it would be reported as a success.

ezr-ondrej (approved):
:orange_heart:

PR #253, opened 2022-10-07
Rename architectures X86_64

avitova:
I have renamed x86_64 architectures. We are using types.ArchitectureTypeX8664 from ec2 package. I hope that it will not confuse us in the future.


lzap (approved):
TY

PR #254, opened 2022-10-07
Improve HTTP client transport creation

lzap:
It was a bit messy, now we properly create transports using `Clone` function as it should be.

Note that transports can be cloned before any HTTP communication happens, therefore this is done in an `init` function. This rules out use of configuration values, so these are constants. I think it's a good tradeoff, it could be doable to do configuration but I don't think we will ever need to change these.


lzap:
Ha it fails on CI because the " HTTP client proxy can be only used in development" error is actually triggered. Is it correct that we have HTTP stage proxy for ephemeral environment? @ezr-ondrej @akhil-jha

lzap:
Sorry I have a bug in my code! Jeeez.

lzap:
Two bugs, TWO bugs, would you believe this... Rebased.

internal/telemetry/http_client.go
```diff
@@ -13,43 +13,72 @@ import (
...
+
+func init() {
+	// HTTP transports must be initialized before use
+	dialer := &net.Dialer{
+		Timeout:   TransportTimeout,
+		KeepAlive: TransportKeepAlive,
+	}
+
+	sourcesTransport = http.DefaultTransport.(*http.Transport).Clone()
+	sourcesTransport.Proxy = proxyFromConfig(func() string {
```
ezr-ondrej, line 36: This very much feels like first-class citizen already and it was very hard to parse for me what actually is happening here and I still don't get why we need to pass a function, not the string directly?
The string is not defined at this point?

Can't we just add the proxy when building the client (in `HTTP{Sources,Imagebuilder}Client`)?
It would feel much more readable and less like Proxy is first-class citizen.

(I believe we should avoid devel code, Proxy in this case, to get too rooted in the code, that's why I'd like to simlify it as much as possible even if it's slightly less performant)

ezr-ondrej (commented):
The core of the change is great, I only have an issue with making the Proxy code (which is devel only) more complex and more rooted in the code.

internal/telemetry/http_client.go
lzap: It's not nice, but we have to use functions because this is `init()` where order is unknown and configuration might not yet be available. Luckily the HTTP libraries support setting key options via functions.

The documentation/discussion in Go library indicates that HTTP transports should be cloned before any connection is made, which is guaranteed in `init()`, we do not do any HTTP requests there and should absolutely not do that.

I can move this to clients, it's just I feel like we have already quite a lot of code there and it is going to be copy-pasted around when we introduce new clients. I am aiming for the opposite - extract as much code from there to somewhere else.

Now, I understand the concern and it does not definitely look pretty. So the alternative is basically to keep the previous code, unfortunately, there is no way to get the default values (timeouts, all the numbers) from the Go library so we just need to set constants and hope for the best.

lzap:
Let me take another look into this.

lzap:
You are right, it was utterly complex and most importantly, this should have never been in `telemetry` context, I moved this into our `http` context, renamed the function to simply `NewPlatformClient` and made it more simple - no cloning, it just use a new transport with default values, it looks like this works too.

Please test, I am currently recreating my config which I accidentally deleted too.

lzap:
Tested it works.

lzap:
Updated commit message and rebased to follow the new commit message formatting guideline: https://github.com/RHEnVision/provisioning-backend/pull/261

ezr-ondrej (approved):
Works well, nice cleanup here :)

PR #256, opened 2022-10-07
Remove unnecessary error check

avitova:
I do not think we need these.


lzap:
Linter has errors. You can run all lint tools just by running `make fmt`.

lzap (approved):
TY 🎧

lzap:
I mean ❤️

PR #257, opened 2022-10-07
Rename clowder service to 'api'

lzap:
This will fail until https://gitlab.cee.redhat.com/insights-qe/iqe-provisioning-plugin/-/blob/master/iqe_provisioning/conf/provisioning.default.yaml#L56 is updated too. Not sure how to approach this without breaking the pipeline tho.

@RHEnVision/qe

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -68,7 +68,7 @@ objects:
               requests:
                 cpu: ${CPU_REQUESTS}
                 memory: ${MEMORY_REQUESTS}
-        - name: service
+        - name: api
           minReplicas: ${{MIN_REPLICAS}}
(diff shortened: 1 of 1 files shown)


mshriver:
Checked app-interface config, I don't think there's anything there linked to that.  Making update to iqe-provisioning-plugin env_parser now, will retest this once that's ready.

ezr-ondrej:
Thanks! :orange_heart:

mshriver (approved):

ezr-ondrej:
Right this broke stage routing, so our app is not routed properly now xD

lzap:
Just a follow-up: @ezr-ondrej has requested a configuration change so we can keep the new name.

PR #260, opened 2022-10-10
Some types were added twice

lzap:
This fixes it, added unit test, regenerated all types.


ezr-ondrej (approved):
Ah, right :)

internal/clients/regional_availability.go
```diff
@@ -62,12 +62,16 @@ func (rit *RegionalTypeAvailability) Add(region, zone string, it InstanceType) {
 	if _, ok := rit.types[raz]; !ok {
 		rit.types[raz] = make([]InstanceTypeName, 0)
 	}
-	rit.types[raz] = append(rit.types[raz], it.Name)
+	// keep lists of types sorted for faster lookups
+	if _, found := slices.BinarySearch(rit.types[raz], it.Name); !found {
+		rit.types[raz] = append(rit.types[raz], it.Name)
+		slices.Sort(rit.types[raz])
+	}
```
lzap, line 69: This is sub-ideal and map of map would be probably better with O(1), however, this is only called when creating a new type list via `typesctl` so not a big deal. Slices serialize better in YAML without any custom code.

PR #261, opened 2022-10-11
Add CHANGELOG and GHA commit action

lzap:
This is a proposal to enforce more user-friendly commit messages which allows nicer CHANGELOG generation and history tracking of features and bugs. It introduces https://github.com/git-chglog/git-chglog tool, a CHANGELOG file generator. It also adds GHA that enforces few simple rules on commits:

* Must start with: chore demo docs feat fix refactor revert style test
* Optionally can contain subject of commit in the form of `type(subject): xxx`
* Maximum length of 70 characters (for easier text terminal use)

Additional recommendations (not enforced):

* Use lower-case everywhere, except function names or acronyms etc.
* Commit messages should not be sentences, avoid verbs, skip articles, be short.
* Don't hesitate to continue with explanation in the commit body.

Example commits:

* `feat: new feature`
* `fix(test): fix in validator unit test`
* `docs(azure): regional check`
* `test(dao): tests for GetNewStub`

### DIFF
--- a/.chglog/CHANGELOG.tpl.md
+++ b/.chglog/CHANGELOG.tpl.md
@@ -1,0 +1,56 @@
+{{ if .Versions -}}
+<a name="unreleased"></a>
+## [Unreleased]
+
+{{ if .Unreleased.CommitGroups -}}
+{{ range .Unreleased.CommitGroups -}}
+### {{ .Title }}
+{{ range .Commits -}}
+- {{ if .Scope }}**{{ .Scope }}:** {{ end }}{{ .Subject }}
+{{ end }}
+{{ end -}}
+{{ end -}}
+{{ end -}}
+
+{{ range .Versions }}
+<a name="{{ .Tag.Name }}"></a>
+## {{ if .Tag.Previous }}[{{ .Tag.Name }}]{{ else }}{{ .Tag.Name }}{{ end }} - {{ datetime "2006-01-02" .Tag.Date }}
+{{ range .CommitGroups -}}
+### {{ .Title }}
+{{ range .Commits -}}
+- {{ if .Scope }}**{{ .Scope }}:** {{ end }}{{ .Subject }}
+{{ end }}
+{{ end -}}
+
+{{/* tags before 0.8.0 have no grouping */}}
+{{ if regexMatch "^0.[12345678].0$" .Tag.Name }}
+{{ range .Commits -}}
+- {{ .Header }}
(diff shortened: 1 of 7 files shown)


.chglog/CHANGELOG.tpl.md
```diff
@@ -0,0 +1,56 @@
...
+- {{ if .Scope }}**{{ .Scope }}:** {{ end }}{{ .Subject }}
+{{ end }}
+{{ end -}}
+
+{{/* tags before 0.8.0 have no grouping */}}
+{{ if regexMatch "^0.[12345678].0$" .Tag.Name }}
+{{ range .Commits -}}
+- {{ .Header }}
+{{ end }}
+{{ end }}
```
lzap, line 30: This template was unchanged, except this block. By default, when there is no type (type is the `fix` or `chore` or `feat` prefix), the tool skips commits. Since we don't have any types in commit messages prior this PR, we would have pretty much empty CHANGELOG up until 0.8.0 (this commit). This fixes it.

lzap:
I have tags 0.1.0 to 0.8.0 locally, let me know if you agree with the CHANGELOG proposal and I will push those tags into the main branch.

lzap:
One more note - the tool supports Jira to some degree, but I haven't bothered implementing it for now. It only supports reading it from the subject, the regexp is already pretty complicated and I am not sure where we would put this. It's something we can take a look later.

lzap:
Looks like GHA performs merge by default, this should fix it: https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

ezr-ondrej (approved):
Nice, thanks! :+1:

lzap:
Pushed the tags:

```
lzap@mone provisioning-backend % g push --tags upstream main
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
To https://github.com/RHEnVision/provisioning-backend
 * [new tag]         0.1.0 -> 0.1.0
 * [new tag]         0.2.0 -> 0.2.0
 * [new tag]         0.3.0 -> 0.3.0
 * [new tag]         0.5.0 -> 0.5.0
 * [new tag]         0.6.0 -> 0.6.0
 * [new tag]         0.7.0 -> 0.7.0
 * [new tag]         0.8.0 -> 0.8.0
```

Here is also a small followup: https://github.com/RHEnVision/provisioning-backend/pull/262

PR #262, opened 2022-10-11
changelog: update types and add docs

lzap:
The CI check did not allow commit messages without component: `feat: no component`.

This fixes it, also sorts all possible types and sync them with the check script (and configuration of the changelog tool).

Finally, adds some documentation.


ezr-ondrej (approved):
Thanks

PR #263, opened 2022-10-12
chore(oapi): update clients, missing make doc

lzap:
After the `configs` directory rename, the timestamps for `make` were not correct. This fixes it (just by running `make` once).

Also I updated clients, it now contains the change I submitted to sources (missing type, enums).

Finally, I realized that `make generate-clients` target was not documented at all.


ezr-ondrej (approved):
Thanks @lzap !

PR #264, opened 2022-10-12
chore(scripts): update sources seed script

lzap:
An unique index was created recently in sources which prevents from bulk creations of multiple sources. This breaks the seed operation into three distinctive requests.

https://github.com/RedHatInsights/sources-api-go/pull/326


ezr-ondrej (approved):
Nice, that fixes it for me :tada:

PR #265, opened 2022-10-13
feat: welcome HTML page

lzap:
This adds a nice welcome page, this is pure HTML 3.2 how I remember it. Here is a screenshot:

Snímek obrazovky 2022-10-13 v 16 18 51

The reason was I was hitting `/docs/` instead of the correct `/docs`, I don't need to remember this anymore with this patch. Also shows current version and sha commit on the page (when compiled in).

Also cleans up routes a bit, it was a bit confusing to me.

Finally, adds an `index.html` file which should be available on through github to always show us latest and greatest OpenAPI documentation. Untested but it should work as this is exactly how Go redoc does it (I actually copy and pasted the HTML from there).


ezr-ondrej (approved):
I don't know why the refactoring of routers, but seems to work ^_^
Thanks @lzap ! :+1:

lzap:
I have found it little bit less readable before the change.

PR #266, opened 2022-10-14
Use not found or dao error

avitova:
These are changes made to the error_payload that do not include touching the function header. More changes are coming but let's split this. We will be adding message parameters everywhere, so I prepared the renderNotFoundOrDAOError to deal with that.


internal/services/error_renderer.go
```diff
@@ -28,7 +28,7 @@ func renderError(w http.ResponseWriter, r *http.Request, renderer render.Rendere
 	}
 }
 
-func renderNotFoundOrDAOError(w http.ResponseWriter, r *http.Request, err error, daoError string) {
+func renderNotFoundOrDAOError(w http.ResponseWriter, r *http.Request, err error, daoError string, notFoundError string) {
```
lzap, line 31: This parameter does not appear to be used. I think you can simply add `resource string` parameter and this can be passed in a way that both errors look like:

* NotFoundError: Name="not found", Message="[resource] not found", Error="[resource] not found: %w"
* DAOError: Name="dao error", Message="[resource] DAO error", Error="[resource] DAO error: %w"

internal/payloads/error_payload.go
```diff
@@ -49,10 +41,10 @@ func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
...
 	return &ResponseError{
 		HTTPStatusCode: 400,
+		Name:           "invalid request error",
 		Message:        msg,
-		RequestId:      ctxval.TraceId(ctx),
-		Err:            err,
-		Context:        ctx,
+		TraceId:        ctxval.TraceId(ctx),
+		Error:          err,
 	}
```
lzap, line 48: Can you include the actual "Name" in error message too for all relevant errors?

```go
func NewInvalidRequestError(ctx context.Context, err error) *ResponseError {
    name := "invalid request error"
	err = fmt.Errorf("%s: %w", name, err)
	if logger := ctxval.Logger(ctx); logger != nil {
		logger.Warn().Err(err).Msg(msg)
	}
	return &ResponseError{
		HTTPStatusCode: 400,
		Name:           name,
		Message:        name, // you can do this for most errors when name is appropriate
		Error:            err,
		TraceId:        ctxval.TraceId(ctx),
	}
```

lzap (changes requested):
Looks good, I have some minor comments.

FYI the issue is https://issues.redhat.com/browse/HMSPROV-299

internal/payloads/error_payload.go
avitova: I thought I might add a 'message' parameter later and pass it from the calling function with a more specific message than just an "invalid request error". Possibly something like "invalid request: parameter missing" or something like that. That might be more useful for the user. Adding parameters would force changing more files, so I planned to make these changes in an upcoming PR.
Does this make sense?

avitova:
I feel like this could be enough before implementing upcoming changes.

internal/payloads/error_payload.go
```diff
@@ -15,44 +15,36 @@ type ResponseError struct {
 	// HTTP status code
 	HTTPStatusCode int `json:"-"`
 
-	// error message including message from wrapped error
-	Message string `json:"msg"`
+	// error type name
+	Name string `json:"name"`
```
adiabramovitch, line 19: I am wondering if name is a good term here. 
Maybe type is better? what is the purpose of the field?

avitova: In the future, each error will get its value passed from the place where the error was caught on services. We will have the error itself in the `.Error`, and the `.Message` is supposed to be shown to the user, which will include more details. So I guess the `'Name'` could be handy when we need to use a simple string.
For example, dao error messages tend to be a bit messy. I am sure that Lukas will be able to correct me and clarify this, as we were discussing this before I started working on this. I am unsure if we should use different naming for this. I will probably leave this decision to @lzap.

lzap: Oh no don't leave those decisions on me, we just brainstormed "name" the other day, I did not put much thought about it.

So the background around this is that we wanted to have all REST error payloads to have:

* Something we can show to the user, a meaningful message like "pubkey upload failed". This can be translated in the UI into other languages via i18n so it should be a short and simple English statement (perhaps not a full sentence).
* Full error as Go returns it (the whole error message chain like `pubkey upload failed: AWS client error: 401 fingerprint already exists`). This would be shown when user clicks on some kind of "Detail".

During a discussion, we were talking about "error codes" and how useful there were in Foreman, we briefly thought about implementing it but then we thought that "error type name" should be better. Now that I think about it, I am not so sure about importance of this information to the user and even for us. Because we also discussed that all our error handling in the application should be done via string errors and wrapping, the de-facto Go standard. At some places we need to check for the type via `errors.As`, but result of this is typically an error value.

So all and all, I think we might actually drop this field completely to be honest, I don't see much value in it.

avitova:
Okay, let's leave out Name for now.

lzap (approved):

lzap:
Linter complaints tho.

internal/services/gcp_reservation_service.go
```diff
@@ -2,10 +2,9 @@ package services
 
 import (
 	"errors"
-	"net/http"
-
 	"github.com/go-chi/render"
 	"github.com/lzap/dejq"
+	"net/http"
```
ezr-ondrej, line 7: linter dislikes this move, net/http needs to be above the github (external) imports.
This is difference between go fmt and go imports style.
In case you're using goland, you can change the style at [`Settings/Preferences | Editor | Code Style | <language> | Imports`](https://www.jetbrains.com/help/go/creating-and-optimizing-imports.html#optimize-imports) we are using the `goimports` sorting style :)

lzap: @avitova what I do is simply running `make fmt` and it will run all formatters we require.

lzap (approved):
Thanks, try to avoid "chore" as much as possible. This could easily be `refactor(errors)` or whatever, but this is fine. I`ve already done that multiple times :-) Merging!

PR #267, opened 2022-10-17
feat(cache): add Redis cache and queue

lzap:
The patch adds both Redis app cache (configurable) and dejq (job queue) which is an easy in-memory only implementation. We will use this until we implement proper SQS or Postgres queues.

This patch will likely fail IQE tests, until "Add provisioning-elasticache-stage resource" app-interface MR is merged. That will add a new Redis cluster to stage with secrets published via clowder.

To check this works, a noop reservation should pass and in logs there should be "queue_type" attribute set to "redis".


internal/clients/errors.go
```diff
@@ -1,10 +1,16 @@
...
 var (
+	ClientErr = errors.New("HTTP client")
+
 	// Common errors
-	NotFoundErr       = errors.New("backend service returned not found (404) or no data")
+	NotFoundErr = fmt.Errorf("%w: backend service returned not found (404) or no data", ClientErr)
+
 	UnauthorizedErr   = errors.New("backend service returned unauthorized (401)")
 	ForbiddenErr      = errors.New("backend service returned forbidden (403)")
 	Non2xxResponseErr = errors.New("backend service did not return 2xx")
```
lzap, line 16: This was not intended, removing.

internal/cache/memcache/apptype_cache.go
```diff
@@ -11,13 +13,21 @@ var (
...
 
 	return appTypeId, appTypeId != ""
 }
 
 func SetAppTypeId(_ context.Context, newAppTypeId string) {
+	if !config.CacheEnabled() {
+		return
+	}
+
 	appTypeIdMutex.Lock()
```
lzap, line 31: These checks can be removed, functions are dynamically assigned via `init()`.

config/api.env.example
```diff
@@ -118,6 +112,20 @@
...
+#     	redis password (default "")
+#   APP_CACHE_REDIS_DB int
+#     	redis database number (default "0")
+#   APP_CACHE_MEM_EXPIRATION int64
+#     	in-memory cache expiration (time interval syntax) (default "1h")
+#   APP_CACHE_MEM_CLEANUP_INTERVAL int64
+#     	in-memory expiration interval (time interval syntax) (default "5m")
 #   REST_ENDPOINTS_IMAGE_BUILDER_PROXY_URL string
 #     	proxy URL (dev only) (default "")
 #   REST_ENDPOINTS_SOURCES_PROXY_URL string
```
lzap, line 131: This is auto-generated.

mk/help.mk
```diff
@@ -16,12 +17,12 @@ help: ## Print out the help content
...
 	echo '# Make documentation' > $(MAKE_DOC)
 	echo '```' >> $(MAKE_DOC)
 	make help | sed -r "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g" >> $(MAKE_DOC)
 	echo '```' >> $(MAKE_DOC)
 
 .PHONY: generate-example-config
-generate-example-config:
-	go run cmd/confgen/main.go > configs/api.env.example
+generate-example-config: ## Generate example configuration
+	go run cmd/confgen/main.go > config/api.env.example
```
lzap, line 28: Found this when I was re-generating example config.

lzap:
I am gonna rebase soon.

deploy/clowdapp.yaml
```diff
@@ -180,16 +186,19 @@ parameters:
   - description: Instance prefix adds string to all instance names, leave blank for production
     name: APP_INSTANCE_PREFIX
     value: ""
+  - description: Application cache type (none, memory or redis)
+    name: APP_CACHE_TYPE
+    value: "redis"
   - description: Internal queue type (memory/sqs/postgres).
     name: WORKER_QUEUE
-    value: "memory"
```
ezr-ondrej, line 185: memory's gone? :)

lzap: What do you mean, the value is set on line `194` below to "redis".

lzap:
So it looks like syntax of `clowdapp.yml` was not correct, inMemoryDb and inMemoryDBConfig appear to be two different elements: https://redhatinsights.github.io/clowder/clowder/dev/api_reference.html

Let me just enable inMemoryDb (set to true) and hope that Redis is the default, which I think it is since sources do it this way.

lzap:
Looks like it works! The end to end fails, but I guess that's something you are looking into @RHEnVision/qe ?

lzap:
For the record, backend api no longer dequeues work, but a worker is: https://ci.int.devshift.net/job/RHEnVision-provisioning-backend-pr-check/536/artifact/artifacts/k8s_artifacts/ephemeral-vrvnvo/logs/provisioning-backend-worker-868894c849-tx48g_provisioning-backend-worker.log

We currently run a single worker, that's definitely enough for non-production deployment.

akhil-jha:
> Looks like it works! The end to end fails, but I guess that's something you are looking into @RHEnVision/qe ?

Well, the tests passed. It failed on removing the VM as part of cleanup. Looking into it.

lzap:
Thanks I figured, it also fails in other PRs. :-)

lzap:
It is green, Redis is working both for cache and queue (worker is picking up jobs).

internal/clients/errors.go
avitova: I think this is still in the commit, right?

ezr-ondrej:
@lzap there is a conflict and @avitova's comments sound like she had a really good mentor and are probably worth addressing or explaining why you've decided to go against your own recomendations in those cases? :)

internal/cache/redis/redis_cache.go
```diff
@@ -0,0 +1,64 @@
...
+		Username: config.Application.Cache.Redis.User,
+		Password: config.Application.Cache.Redis.Password,
+		DB:       config.Application.Cache.Redis.DB,
+	})
+	return &cache{
+		client: client,
+	}
+}
+
+func (c *cache) FindAccountId(ctx context.Context, OrgID, AccountNumber string) (*models.Account, bool) {
```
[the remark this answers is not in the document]
lzap: Good point, initially the only implementation was a memory implementation where there could be no errors, thus the boolean. I actually considered introducing errors, but then the Redis API seemed to also work with boolean and error was like a 2nd citizen to the API. However, you are right that this is not good - errors should be _always_ returned, here I just log it and that's bad.

I will refactor the whole interface to use errors.

internal/cache/memcache/mem_cache.go
```diff
@@ -0,0 +1,71 @@
...
+		accountId: newMemoryCache[accountKey, *models.Account](config.Application.Cache.Memory.CleanupInterval),
+	}
+
+	_ = promauto.NewGaugeFunc(prometheus.GaugeOpts{
+		Name:        "app_cache_account_items",
+		Help:        "The total number of cache items for account ID",
+		ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName},
+	}, func() float64 {
+		return float64(c.accountId.Count())
+	})
```
adiabramovitch, line 36: Just to make sure after reading the documentation
the purpose here is to create an “info” style metric with a constant value of 1 for The total number of cache items for account ID?

internal/cache/noop/noop_cache.go
```diff
@@ -0,0 +1,25 @@
...
+	// noop
+}
+
+func (_ *Cache) FindAppTypeId(_ context.Context) (string, bool) {
+	return "", false
+}
+
+func (_ *Cache) SetAppTypeId(_ context.Context, _ string) {
+	// noop
+}
```
adiabramovitch, line 25: I am missing something.
 why do we have the same functions name here and internal/cache/memcache/mem_cache.go ? 
is it supposed to be an interface? I would imagine to see the functions implemented under Redis or memcache

internal/cache/redis/redis_cache.go
```diff
@@ -0,0 +1,64 @@
...
+	result := c.client.Get(ctx, key)
+	if result.Err() != nil {
+		ctxval.Logger(ctx).Warn().Err(result.Err()).Msgf("Redis returned an error for account %s", OrgID)
+		return nil, false
+	}
+
+	buf, err := result.Bytes()
+	if err != nil {
+		ctxval.Logger(ctx).Warn().Err(err).Msgf("Redis returned an error for account %s", OrgID)
+		return nil, false
```
adiabramovitch, line 42:
```suggestion
	ctxval.Logger(ctx).Warn().Msgf("Redis returned an error for account with orgID %s", account.OrgID)
```

internal/cache/redis/redis_cache.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (c *cache) FindAccountId(ctx context.Context, OrgID, AccountNumber string) (*models.Account, bool) {
+	key := OrgID + AccountNumber
+	account := models.Account{}
+
+	result := c.client.Get(ctx, key)
+	if result.Err() != nil {
+		ctxval.Logger(ctx).Warn().Err(result.Err()).Msgf("Redis returned an error for account %s", OrgID)
+		return nil, false
```
adiabramovitch, line 36:
```suggestion
		ctxval.Logger(ctx).Warn().Err(result.Err()).Msgf("Redis returned an error for account with orgID %s", OrgID)
		return nil, false
```

internal/clients/errors.go
```diff
@@ -1,10 +1,13 @@
...
-import "errors"
+import (
+	"errors"
+)
 
 var (
 	// Common errors
-	NotFoundErr       = errors.New("backend service returned not found (404) or no data")
+	NotFoundErr = errors.New("backend service returned not found (404) or no data")
+
```
adiabramovitch, line 10: extra line

internal/config/config.go
```diff
@@ -169,6 +177,19 @@ func Initialize(configFiles ...string) {
 		config.Prometheus.Port = cfg.MetricsPort
 		config.Prometheus.Path = cfg.MetricsPath
 
+		// in-memory cache
+		if cfg.InMemoryDb == nil {
+			panic("ERROR: Redis required in clowder environment")
```
adiabramovitch, line 182:
```suggestion
			panic("ERROR: Redis is required in clowder environment")
```

internal/cache/redis/redis_cache.go
```diff
@@ -0,0 +1,64 @@
...
+
+	return &account, true
+}
+
+func (c *cache) SetAccountId(ctx context.Context, OrgID, AccountNumber string, account *models.Account) {
+	key := OrgID + AccountNumber
+
+	buf, err := json.Marshal(account)
+	if err != nil {
+		ctxval.Logger(ctx).Warn().Msgf("Unable to marshal account %s", account.OrgID)
```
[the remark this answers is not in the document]
lzap: This is subject of proper error handling too, therefore I am dropping this log message and return a proper error now.

internal/cache/memcache/mem_cache.go
```diff
@@ -0,0 +1,71 @@
...
+	return &c
+}
+
+func (c *cache) FindAccountId(_ context.Context, OrgID, AccountNumber string) (*models.Account, bool) {
+	key := accountKey{OrgID: OrgID, AccountNumber: AccountNumber}
+
+	if value, ok := c.accountId.Get(key); ok {
+		return value, ok
+	}
+	return nil, false
```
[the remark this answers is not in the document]
lzap: Absolutely, great point!

internal/cache/memcache/mem_cache.go
lzap: Constant value? No, this gauge is supposed to contain total amount of accounts currently held in memory. So that could be 0, 1, thousands.

internal/cache/noop/noop_cache.go
lzap: There is an interface, yes. See the `cache_interface.go` file, there are three implementations and a decorator that is only meant to deliver hit metrics (only for account):

https://en.wikipedia.org/wiki/Decorator_pattern

internal/clients/errors.go
lzap: Oh right, thanks.

internal/cache/redis/redis_cache.go
lzap: I am dropping those log lines because the new version I am about to rebase to handles errors properly.

lzap:
There you go, great reviews indeed. Well done! This was a hell lot of refactoring but I feel like proper error handling is important, regardless that booleans are easier to write. Rebased.

internal/cache/memcache/mem_cache.go
adiabramovitch: yes, paste it from documentation but it means the a number not just 0

lzap: Oh I see, no, this is a normal "gauge" measurement that counts all items in the account cache. It has nothing to do with that constant metric, the documentation just mention that this is also one of other uses of a gauge metric (to return constant only to be able to use labels to return some informative values like application version etc). We don't do that, the only label we return here is the application service name (which is "provisioning").

lzap:
Fixed the unit test.

deploy/clowdapp.yaml
ezr-ondrej: ah, I've missread it :(

adiabramovitch (approved):

PR #268, opened 2022-10-20
refactor(changelog): different tool for changelog

lzap:
The tool I initially implemented for CHANGELOG generation was way complicated and it actually striked the moment I tried to regenerate it - it was not working correctly and I spent an hour trying to figure it out. Then I simply swapped the tool for a very simply Python-based solution that is the 3rd most used changelog generator according to Google.

Works out of box without any changes of the template, nicely keeps our old-style commit messages in a clean form (links to github). I think this would do it.


scripts/check_commits.sh
```diff
@@ -6,7 +6,7 @@ if [[ "$MESSAGE" =~ ^WIP ]]; then
 fi
 
 MAX_LENGTH=70
-TYPES="feat fix refactor perf revert chore demo docs style test"
+TYPES="build ci docs feat fix perf refactor test chore"
```
lzap, line 9: This is the list that the changelog generation tool supports, for simplicity I want to simply stick with it :-) It is not configurable. Basically it's all minus demo/style plus build and ci. Makes sense.

ezr-ondrej (approved):
This is unlikely to have caused the regression, but any idea what is happening @akhil-jha ?

PR #271, opened 2022-10-20
Add gcp request payload

adiabramovitch:


scripts/rest_examples/reservation-create-gcp.http
```diff
@@ -2,3 +2,14 @@
...
+{
+  "name": "gcp-linux-us-central1-a",
+  "zone": "us-central1-a",
+  "source_id": "3",
+  "image_id": "80967e7f-efef-4eee-85b0-bd4cef4c455d",
+  "amount": 2,
+  "machine_type": "n1-standard-1",
+  "pubkey_id": 1,
+  "poweroff": true
+}
```
lzap, line 15: I think we will need to parametrize source_ids in those .http files, I use 1, 2, 3, you appear to have something different. A simply `{{gcp_source_id}}` would do the trick.

lzap (approved):

PR #272, opened 2022-10-20
fix: throw an error when pubkey is duplicated (HMSPROV-309)

adiabramovitch:


ezr-ondrej (approved):
Thanks, lets merge before someone messes up the DB in our fresh stage environment ^_^

lzap:
Thanks, the code still needs more love as we need to actually check all keys on AWS prior any action, looking it up in our database is unfortunately not enough as users might imported ssh keys manually.

adiabramovitch:
@lzap This is a temporary fix until we will discuss a needed refactoring for the pubkeys

ezr-ondrej:
Yeah, I totaly agree we need to take it further, but for now I prefer failures then inconsistent DB :)

PR #273, opened 2022-10-21
feat: AWS file credentials support

lzap:
Since we still don't have AWS team account for development and my free credits has expired already, I need to use our SAML-enabled account via the `saml-aws.py` script. It is possible only if we add a special configuration for this.

This patch adds exactly that, however, I would very much rather prefer getting a proper service/customer account for the whole team instead of this.


lzap:
Ok looks like the team account has been set already.

PR #274, opened 2022-10-22
build(deploy): Allow passing image builder url

ezr-ondrej:
Allows to passimage builder URL in non-clowder environments.

Other apps are passing parameter IMAGEBUILDER_URL, so we've decided to keep the same convention.

But internaly we prefer IMAGE_BUILDER_URL format internaly, thus the discrepancy.

We pass the env variable to both `api` and `worker`.


ezr-ondrej:
Merging this, as build passed and it's quite safe

PR #276, opened 2022-10-24
Split Azure interface into service and customer ones

lzap:
It's cleaner to have service operations within a type that can be reused. It introduces another interface and a second Get function, but it is only for service account operations so it will likely be small.

We should do the same for EC2 and GCP.


internal/clients/http/azure/azure_client.go
```diff
@@ -10,61 +10,55 @@ import (
...
-
-func newAzureClient(ctx context.Context, auth *clients.Authentication) (clients.Azure, error) {
+func newCustomerClient(ctx context.Context, auth *clients.Authentication) (clients.Azure, error) {
 	opts := azidentity.ClientSecretCredentialOptions{}
 	identityClient, err := azidentity.NewClientSecretCredential(config.Azure.TenantID, config.Azure.ClientID, config.Azure.ClientSecret, &opts)
 	if err != nil {
 		return nil, fmt.Errorf("unable to init Azure credentials: %w", err)
 	}
 
-	return &azureClient{
```
adiabramovitch, line 37: I would keep azureClient for consistency or change in ec2 and GCP to client as well

cmd/typesctl/providers/azure_types.go
```diff
@@ -151,13 +50,5 @@ func generateTypesAzure() error {
 		return fmt.Errorf("unable to generate types: %w", err)
 	}
```
adiabramovitch, line 52: Why we don't want it as part of the interface as well?

internal/clients/interface.go
```diff
@@ -75,13 +75,20 @@ type EC2 interface {
...
+var GetCustomerAzureClient func(ctx context.Context, auth *Authentication) (Azure, error)
+
+// GetServiceAzureClient returns an Azure client for the service account itself.
+var GetServiceAzureClient func(ctx context.Context) (AzureService, error)
 
 type Azure interface {
 	ClientStatuser
 }
 
+type AzureService interface {
```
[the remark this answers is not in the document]
lzap: Oh yeah it is AzureCustomer vs AzureService. We might need to find a better name, granted, it reads weird.

internal/clients/http/azure/azure_client.go
```diff
@@ -10,61 +10,55 @@ import (
...
 
-type azureClient struct {
+type client struct {
 	subscriptionID string
 	credential     *azidentity.ClientSecretCredential
 }
 
 func init() {
-	clients.GetAzureClient = newAzureClient
+	clients.GetCustomerAzureClient = newCustomerClient
```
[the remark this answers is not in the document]
lzap: Because of the two different (clients) types. For Azure it is literally a different type, for EC2 it can be the same client (with just different configuration), is it is technically possible to have a single interface for EC2, but for consistency, I would recommend to do the same approach for all the three.

internal/clients/http/azure/azure_client.go
lzap: Because each implementation lives its in own package, you can just use `client` and there will be no conflicts. Whatever you like.

cmd/typesctl/providers/azure_types.go
lzap: The `typesctl` is a quick and dirty code, it uses three function pointers, feel free to refactor this if you want.

internal/clients/http/azure/azure_service_client.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to init Azure credentials: %w", err)
+	}
+
+	return &serviceClient{
+		credential: identityClient,
+	}, nil
+}
+
+func (c *serviceClient) RegisterInstanceTypes(ctx context.Context, instanceTypes *clients.RegisteredInstanceTypes, regionalTypes *clients.RegionalTypeAvailability) error {
```
avitova, line 35: This function is 90 lines long. I suggest splitting this, there are some logic parts that can be separated. Especially the huge for loop. Moreover, on the line 118, the for loop is nested on level four.
We could have a helper function to return `instanceType.Cores` value. *c.name comparisons seem like a good example for decomposition as well.

internal/clients/http/azure/azure_service_client.go
```diff
@@ -0,0 +1,147 @@
...
+					}
+					instanceType.MemoryMiB = int64(memoryGB * 1000)
+				}
+				if *c.Name == "HyperVGenerations" {
+					if strings.Contains(*c.Value, "V1") {
+						instanceType.AzureDetail.GenV1 = true
+					}
+					if strings.Contains(*c.Value, "V2") {
+						instanceType.AzureDetail.GenV2 = true
+					}
```
avitova, line 99:
```suggestion
					instanceType.AzureDetail.GenV1 = strings.Contains(*c.Value, "V1")
					instanceType.AzureDetail.GenV2 = strings.Contains(*c.Value, "V2") 
```

internal/clients/http/azure/azure_service_client.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+			if vcpusPerCore != 0 {
+				instanceType.Cores = instanceType.VCPUs / vcpusPerCore
+			} else {
+				// Some types have no HT (HPC AMD CPUs) thus vcpus per core is zero:
+				// Standard_B16ms, Standard_B20ms, Standard_HB60rs, Standard_HC44rs,
+				// Standard_M208ms_v2, Standard_M208s_v2, Standard_M416ms_v2,
+				// Standard_M416s_v2 and Standard_PB6s
+				instanceType.Cores = instanceType.VCPUs
+			}
```
avitova, line 110:
```suggestion
			instanceType.Cores = instanceType.VCPUs
			if vcpusPerCore != 0 {
				instanceType.Cores /= vcpusPerCore
			}
```

This is also possible, but if we need the comments, a helper function might also come in handy.

internal/clients/http/azure/azure_service_client.go
```diff
@@ -0,0 +1,147 @@
...
+			vcpusPerCore := int32(0)
+			for _, c := range v.Capabilities {
+				if *c.Name == "CpuArchitectureType" {
+					instanceType.Architecture, pageErr = clients.MapArchitectures(ctx, *c.Value)
+					if pageErr != nil {
+						return fmt.Errorf("unable to generate types: %w", pageErr)
+					}
+				}
+				if *c.Name == "vCPUs" {
```
avitova, line 64: I expect the *c.Name to have only one of these values, should we maybe use `else if` just for the better readability?

internal/clients/interface.go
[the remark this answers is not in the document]
adiabramovitch: reminder: we decided to change it to ServiceAzure and keep Azure

internal/clients/http/azure/azure_service_client.go
adiabramovitch: Maybe a switch case of c.Name?

internal/clients/http/azure/azure_service_client.go
lzap: That is a fair point, I extracted a new function called `typeFromSKU` that moves probably half of the function into a separate one. That cut's it at half, a good starting point I guess.

Unfortunately, I accidentally merged it into the same commit, this should have been done in a separate commit, or separate PR ideally. Sorry for that, it is gonna be hard to follow the change now since the code moved from a different file, but it was essentially just extraction using GoLand "extract method" feature plus I renamed variable `v` to `resourceSKU` which is much more appropriate.

internal/clients/http/azure/azure_service_client.go
lzap: Damn, you really hate these, do you? Right, nice point.

internal/clients/http/azure/azure_service_client.go
lzap: Sure, I shortened the comment.

internal/clients/http/azure/azure_service_client.go
lzap: Switch it is.

lzap:
Done, thanks for comments. Again, the type refresh is in a separate commit so make sure to just look on the code change one.

internal/clients/http/azure/azure_client.go
```diff
@@ -10,61 +10,55 @@ import (
...
 
-type azureClient struct {
+type client struct {
 	subscriptionID string
 	credential     *azidentity.ClientSecretCredential
 }
 
 func init() {
-	clients.GetAzureClient = newAzureClient
+	clients.GetAzureClient = newCustomerClient
```
adiabramovitch, line 21: In GCP(and aws) the customer function is named newGCPClient
(Without a change) can we be consistent?

adiabramovitch (approved):
I am good here, just a nit. 
@avitova what about you?

internal/clients/http/azure/azure_client.go
ezr-ondrej: to add my 2c I like `newAzureClient` more, as Customer might be missleading, if this service were to be used outside of it's the RH context

adiabramovitch: @lzap Can we keep newAzureClient?

lzap: Absolutely, adding an extra commit.

lzap:
New commit is up.

adiabramovitch:
@lzap  Thanks :)

PR #277, opened 2022-10-24
Typo in etag prefix and gitignore update

lzap:
Two small things I ran into today, the ec2 prefix is obvious and was spotted by @adiabramovitch 

The other commit is just a small change - I now have a symlink and it looks like git does not recognize symlinks with `something/` gitignore pattern as the slash requires it to be a directory. This fixes it.


lzap:
Thanks!

PR #278, opened 2022-10-25
feat(flags): feature flags endpoint

lzap:
https://issues.redhat.com/browse/HMSPROV-314

Implements feature flag endpoint that our/ib UI will use to show/hide the launch button. The backend service we use is named Unleash

* https://www.getunleash.io/
* https://github.com/Unleash/unleash
* https://github.com/Unleash/unleash-client-go

Therefore all the configuration variables are also named `UNLEASH_` and it is possible to configure the service locally and enable it for debugging purposes. When Unleash is disabled through config, all queries for feature flags return true. When running in Clowder, Unleash gets enabled by default and configured from Clowder.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -116,6 +116,13 @@ var config struct {
 		Heartbeat   time.Duration `env:"HEARTBEAT" env-default:"30s" env-description:"heartbeat interval (time interval syntax)"`
 		MaxBeats    int           `env:"MAX_BEATS" env-default:"10" env-description:"maximum amount of heartbeats allowed"`
 	} `env-prefix:"WORKER_"`
+	Unleash struct {
+		Enabled     bool   `env:"ENABLED" env-default:"false" env-description:"unleash service (feature flags)"`
+		Environment string `env:"ENVIRONMENT" env-default:"" env-description:"unleash environment"`
+		Prefix      string `env:"PREFIX" env-default:"app.provisioning" env-description:"unleash flag prefix"`
+		URL         string `env:"URL" env-default:"http://localhost:4242" env-description:"unleash service URL"`
+		Token       string `env:"TOKEN" env-default:"" env-description:"unleash service client access token"`
+	} `env-prefix:"UNLEASH_"`
 }
 
 // Config shortcuts
@@ -133,6 +140,7 @@ var (
 	ImageBuilder  = &config.RestEndpoints.ImageBuilder
 	Sources       = &config.RestEndpoints.Sources
 	Worker        = &config.Worker
+	Unleash       = &config.Unleash
 )
 
 // Errors
@@ -189,6 +197,14 @@ func Initialize(configFiles ...string) {
 		}
 		if cfg.InMemoryDb.Password != nil {
 			config.App.Cache.Redis.Password = *cfg.InMemoryDb.Password
+		}
+
+		// feature flags
+		config.Unleash.Enabled = true
+		url := fmt.Sprintf("%s://%s:%d", cfg.FeatureFlags.Scheme, cfg.FeatureFlags.Hostname, cfg.FeatureFlags.Port)
+		config.Unleash.URL = url
+		if cfg.FeatureFlags.ClientAccessToken != nil {
+			config.Unleash.Token = *cfg.FeatureFlags.ClientAccessToken
 		}
 
 		// HTTP proxies are not allowed in clowder environment
--- a/internal/config/flags.go
+++ b/internal/config/flags.go
@@ -1,0 +1,70 @@
+package config
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/version"
+	"github.com/Unleash/unleash-client-go/v3"
+	"github.com/rs/zerolog"
+)
+
+// FeatureEnabled returns a user-specific or global feature flag. When such flag is not found,
+// true is returned.
+func FeatureEnabled(ctx context.Context, name string) bool {
+	// When unleash is not configured, all features are enabled by default.
+	if !config.Unleash.Enabled {
+		return true
+	}
+
+	uctx := ctxval.UnleashContext(ctx)
+	return unleash.IsEnabled(name, unleash.WithContext(uctx), unleash.WithFallback(true))
+}
+
+// LaunchEnabled: launch button in UI initiates the application workflow
+func LaunchEnabled(ctx context.Context) bool {
+	return FeatureEnabled(ctx, fmt.Sprintf("%s.launch", config.Unleash.Prefix))
+}
+
+func unleashLogger(ctx context.Context) *zerolog.Logger {
+	logger := ctxval.Logger(ctx).With().Bool("unleash", true).Logger()
+	return &logger
+}
+
+// InitializeFeatureFlags configures unleash client and starts poller routine. Callers
+// must close poller by calling StopFeatureFlags function.
+func InitializeFeatureFlags(ctx context.Context) error {
+	if !config.Unleash.Enabled {
+		return nil
+	}
+
+	listener := logListener{
+		logger: unleashLogger(ctx),
+	}
+	err := unleash.Initialize(
+		unleash.WithListener(&listener),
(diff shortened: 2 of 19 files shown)


ezr-ondrej (approved):
Awesome work :heart_eyes:

ezr-ondrej:
I'll keep it up in case there are others who wants to take a look, but I'm all for merging right away :)

adiabramovitch (commented):
LGTM, just a small question:
will ib will use this endpoint in order to know if the launch instances button should be visible or not?

lzap:
> will ib will use this endpoint in order to know if the launch instances button should be visible or not?

That was the plan and the reason why we actually expose this via an endpoint.

@ezr-ondrej before merging, I think we should put this into OpenAPI. I am not sure.

adiabramovitch:
I think it needs to be in the OpenAPI

internal/routes/all_routes.go
```diff
@@ -114,6 +114,10 @@ func apiRouter() http.Handler {
 				r.Get("/", s.ListBuiltinInstanceTypes(ec2_types.InstanceTypesForZone))
 			})
 		})
+
+		r.Route("/feature_flag/{FLAG}", func(r chi.Router) {
```
ezr-ondrej, line 118: I'd probably not expose it for now, it seems that we can consume other endpoints in the UI, that are already exposed for us, backend should consume feature flags to make it's own decisions only if I understand it correctly.

internal/config/flags.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/rs/zerolog"
+)
+
+func FeatureEnabled(name string) bool {
+	// When unleash is not configured, all features are enabled by default.
+	if !config.Unleash.Enabled {
+		return true
+	}
+
+	return unleash.IsEnabled(name)
```
ezr-ondrej, line 20: Also, we shall probably pass a context with some user info, so we can use the rollout strategies (like A-B testing).

https://github.com/Unleash/unleash-client-go#unleash-context

ezr-ondrej (commented):
few thoughts on improvements, but no blockers ^_^
I'd say we do not need to expose it at all, but if we do we don't need to expose it in the OpenAPI, as this is not customer facing endpoint and we don't want to have QE tests for it IMO :)

adiabramovitch:
> few thoughts on improvements, but no blockers ^_^
> 
> I'd say we do not need to expose it at all, but if we do we don't need to expose it in the OpenAPI, as this is not customer facing endpoint and we don't want to have QE tests for it IMO :)

I am not sure why , I would assume we want we to test when the bottom appears and under which circumstances

internal/config/flags.go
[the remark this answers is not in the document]
lzap: I do prefer more readable style at cost of LOC but you know what let's do that so I look more clever :-)

lzap: Yeah I did not know what strategies are used on our instances, so I will simply use orgID for the "userId" in the context for the initial version.

ezr-ondrej:
> I am not sure why , I would assume we want we to test when the bottom appears and under which circumstances

I'm not sure there, @RHEnVision/qe what would you say? :)

lzap:
Well you gave an answer above, @ezr-ondrej - you said that we should not expose anything at all. Therefore no OpenAPI is needed.

In general, think about OpenAPI as a huge commitment. We should only add things that we are 100% sure to have there for years and when they are stable. I am not so sure about this one.

internal/routes/all_routes.go
lzap: Well I thought so but you wrote this into the ticket. IMHO both UI and API should have its own unleash code checking against the server, no endpoint would ever be needed.

ezr-ondrej: yeah, you're right, I was wrong. Let me update the ticket.

internal/config/flags.go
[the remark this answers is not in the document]
avitova: If it feels less readable, let's not force it to be there.:)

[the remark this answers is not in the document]
lzap: Aah after I implemented contexts, it is too messy so I had to fallback to the if statement at the beginning.

lzap:
Rebased, implemented user context. I am going for org_id for userId mapping for now, if we have our own instance that is exactly what we need. If we are going to use a shared one, then I can change this if needed.

I slightly refactored the endpoint, it is still undocumented via OpenAPI, it is more simple now tho - it exposes any flag that starts with "app.provisioning". I am not adding this into OpenAPI because chances are that we might be sharing a single unleash instance with image builder, and in that case no such endpoint would be necessary.

Please test this before merging :-)

internal/config/config.go
```diff
@@ -107,6 +107,13 @@ var config struct {
 		Heartbeat   time.Duration `env:"HEARTBEAT" env-default:"30s" env-description:"heartbeat interval (time interval syntax)"`
 		MaxBeats    int           `env:"MAX_BEATS" env-default:"10" env-description:"maximum amount of heartbeats allowed"`
 	} `env-prefix:"WORKER_"`
+	Unleash struct {
+		Enabled     bool   `env:"ENABLED" env-default:"false" env-description:"unleash service (feature flags)"`
+		Environment string `env:"ENVIRONMENT" env-default:"" env-description:"unleash environment"`
+		Prefix      string `env:"PREFIX" env-default:"app.provisioning" env-description:"unleash flag prefix"`
```
lzap, line 113: Also environment is configurable option now, this will get used if we will have a single instance across environments, which is not the case I think. But I added it, just do not set it and an empty string is the default value that should not hurt anything in the client.

internal/services/reservations_service.go
```diff
@@ -20,6 +21,10 @@ var (
 
 // CreateReservation dispatches requests to type provider specific handlers
 func CreateReservation(w http.ResponseWriter, r *http.Request) {
+	if !config.LaunchEnabled(r.Context()) {
+		writeUnauthorized(w, r)
+	}
+
 	pType := models.ProviderTypeFromString(chi.URLParam(r, "TYPE"))
 	switch pType {
 	case models.ProviderTypeNoop:
```
lzap, line 30: This is the place - when launching is disabled for an account or globally, then it returns unauthorized. Listing reservations is possible (through API) tho.

internal/config/config.go
```diff
@@ -107,6 +107,13 @@ var config struct {
...
+		URL         string `env:"URL" env-default:"http://localhost:4242" env-description:"unleash service URL"`
+		Token       string `env:"TOKEN" env-default:"" env-description:"unleash service client access token"`
+	} `env-prefix:"UNLEASH_"`
```
lzap, line 116: I also added environment, by default it is unused and we will likely not need it (only if we share a single instance across environments which is not the case I think).

internal/services/error_renderer.go
```diff
@@ -35,3 +46,33 @@ func renderNotFoundOrDAOError(w http.ResponseWriter, r *http.Request, err error,
...
+	writeEmptyResponse(w, r, http.StatusServiceUnavailable)
+}
+
+func writeNoContent(w http.ResponseWriter, r *http.Request) {
+	writeEmptyResponse(w, r, http.StatusNoContent)
+}
+
+func writeUnauthorized(w http.ResponseWriter, r *http.Request) {
+	writeEmptyResponse(w, r, http.StatusUnauthorized)
+}
```
lzap, line 78: I moved these function from various places into a single place and renamed them so it`s more readable.

lzap:
Ok so now there's an error UNAUTHORIZED during AWS EC2 launch, that's because Unleash on ephemeral has no data, thus the flag is FALSE. How do we solve that? :-) @ezr-ondrej

ezr-ondrej:
Can we use `WithFallback` to return true on missing toggle?

https://github.com/Unleash/unleash-client-go/blob/459ff5b251921855426f98ad1e7091df21a37e23/config.go#L167

lzap:
Good idea, rebased.

lzap:
Rebased.

lzap:
All good, please re-review since I had bunch of merge conflicts, thanks!

ezr-ondrej (approved):
Nice, I'm not sure if the stage environment will pass the token correctly through clowder, but there is one way to find out I guess ;)

PR #281, opened 2022-10-27
fix(queue): improve errors and logging for dejq init

lzap:
SSIA

We have some weird errors on ephemeral, lines do not match what I see in my branch locally. Let's try again and do some extra logging this time.


avitova (approved):
LGTM. :+1:

ezr-ondrej (approved):
Thanks @lzap

lzap:
There is actually a missing `else` statement so if you provide a nonsense value, it will actually not initialize anything and jobqueue variable will be `nil`. Let me fix that in a followup PR.

PR #282, opened 2022-10-27
fix(queue): recognize unknown worker config values

lzap:
This fixes the problem we saw, we have modified a configuration via app-interface which broke all deployments which were not rebased on the Redis PR.

This should at least give a nicer error message to these.

All: Rebase all PRs on top of `main` to get Redis support which has been configured in app-interface already.


ezr-ondrej (approved):
Thanks @lzap :+1:

PR #283, opened 2022-10-31
chore(deps): update dependencies

lzap:
Both pgxpoolprometheus and scany libraries are now stable. The only remaining library which is in beta (because of the pgx/v5 major update) is the "tern" library.

Also updates other dependencies.


lzap:
Pushed a small gitignore change which allows creating of `internal/db/seeds/devel.local.sql` files. I use this for local seeding (my stage account number).

lzap:
Ok correction only scany left beta, pgxpoolprometheus is still in beta.

lzap:
Hitting https://github.com/jackc/puddle/issues/26 with this one, closing for now.

PR #285, opened 2022-11-02
Update all deps except pgxprometheus

lzap:
This updates dependencies, I started one by one to identify which one is causing a problem with pgx subdependency, it looks it is the pgxprometheus telemetry. So I am skipping update of that for now:

More info here: https://github.com/jackc/puddle/issues/26

Also contains a tiny fix for `make clean` (it was erroring out).


avitova (approved):
LGTM

PR #286, opened 2022-11-02
feat(kafka): setup, configuration, availability check

lzap:
This adds sources availability checking via kafka backend worker.

To test this:

* Use scripts/kafka.setup.sh and kafka.start.sh to download, unpack and start up a small Kafka cluster (single process mode). You will need Java 8+ installed.
* Enable and configure Kafka. For dev setup all you need to do is `KAFKA_ENABLED=true`, when running kafka remotely then also `KAFKA_BROKERS=nuc.home.lan:9092`.
* Build and start `pbstatuser` process, that is "statusER". I could not figure out a better name, we already have a "worker" which is a generic job worker.
* Hit the `/availability_status` endpoint one or multiple times, there is a `.http` file for that.
* The API backend will enqueue and send to Kafka after few seconds. The statuser will pick them up (it can be delayed up to 10 seconds for performance reasons).
* It will currently only print a "trace" log, it does nothing. The actual implementation is outside of scope for this PR. @adiabramovitch was interested in helping with this, so here is where you can pick this up.

What was NOT tested:

* TLS
* SASL

I only tested with unencrypted plain connections (that is the default behavior). I will take a look into this in a separate PR where I will enable Kafka from the cluster. But "it should work".

More info on the design:

https://docs.google.com/document/d/1i18N1EwU3ccRWBAkqZFaJcGWTlY1rw6W-0gRt05f_Jc/edit


internal/bus/kafka.go
```diff
@@ -0,0 +1,186 @@
...
+	logger := ctxval.Logger(ctx)
+	r := NewReader(ctx, topic)
+	defer r.Close()
+
+	for {
+		kmsg, err := r.ReadMessage(ctx)
+		if err != nil && errors.Is(err, io.EOF) {
+			break
+		} else {
+			logger.Warn().Err(err).Msg("error when reading message")
```
avitova, line 151: This logs an error even if `err == nil`, right?

lzap: Not sure what you mean, but the intention was:

* If there was an error, but it was end of file, then break the loop (and exit the goroutine).
* If there was any other error, just print a warning in log and continue reading.
* Otherwise log a trace log message and deliver the message into the callback handler function.

Do you see there a problem? I have to be missing something.

lzap:
Rebased, added tests, tested the code, added stub, interface and documentation. This is now ready for full reviews.

internal/bus/kafka.go
avitova: Yes, you fixed this with the [6e5197c](https://github.com/RHEnVision/provisioning-backend/commit/6e5197c3c82f38595811089a2127a01a0a84ca1b) push. :+1: Not relevant anymore.

cmd/pbapi/main.go
```diff
@@ -83,7 +85,20 @@ func main() {
...
+		if err != nil {
+			logger.Fatal().Err(err).Msg("Unable to initialize the platform bus")
+		}
+	}
+
+	// initialize background goroutines
+	bgCtx, bgCancel := context.WithCancel(ctx)
+	bg.Initialize(bgCtx)
+	defer bgCancel()
+
```
lzap, line 100: There are two new packages, `bus` represents our platform Kafka bus, I wanted something short for the package name. Also it is a generic interface with three implementations:

* noop broker (does nothing, messages are thrown away and never delivered, this is the default if you don't enable kafka)
* stub broker (in memory message delivery, used in unit tests)
* kafka (well, kafka broker)

The second new package is `bg` because we need REST API to be fast, so all availability checks create messages and send them over a Go channel to a background routine which picks it up and send it in 1k batches (not later than in 5 seconds). Kafka works best with multiple messages, not one by one. This allows the REST API to be super-fast.

cmd/pbstatuser/main.go
```diff
@@ -0,0 +1,118 @@
...
+	// 2) Reads from AWS channel, performs availability check, has configurable throttling.
+	// 3) Reads from Azure channel, performs availability check, has configurable throttling.
+	// 4) Reads from GCP channel, performs availability check, has configurable throttling.
+
+	// Beware there is no database connection, no job queue, most of what api/worker processes have
+	// is not available because it should not be needed. The worker will simply use SDKs to check
+	// statuses and send out results via bus.Send function to Sources.
+	// Also make sure that all goroutines are closed correctly on context cancel. A WaitGroup
+	// would be a really good here: https://gist.github.com/fracasula/b579d52daf15426e58aa133d0340ccb0
+}
```
lzap, line 37: This is the missing bit - the actual checking - which needs to be implemented later on.

cmd/pbstatuser/main.go
```diff
@@ -0,0 +1,118 @@
...
+		Str("hostname", hostname).
+		Bool("statuser", true).
+		Logger()
+
+	// initialize telemetry
+	tel := telemetry.Initialize(&log.Logger)
+	defer tel.Close(ctx)
+
+	// initialize platform bus
+	log.Info().Msg("Initializing platform bus")
```
ezr-ondrej, line 72: Why are we using log and not looger here?

cmd/pbstatuser/main.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+		}
+	}()
+
+	log.Info().Msg("Worker started")
+	<-waitForSignal
+	log.Info().Msg("Consumer shutdown initiated")
+	consumerCancelFunc()
+	log.Info().Msg("Shutdown finished, exiting")
+}
```
lzap, line 118: The statuser process is really simple - no database, no job queue, just kafka connection and prometheus.

Which reminds me, we should add some telemetry, but in another PR. Kafka exposes statistics via `Stats` function.

cmd/pbstatuser/main.go
```diff
@@ -0,0 +1,118 @@
...
+	defer clsFunc()
+	log.Logger = logger
+	logging.DumpConfigForDevelopment()
+
+	// setup structured logging
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown-hostname"
+	}
+	logger = logger.With().
```
ezr-ondrej, line 61: We initialize it here, but not passing it to `log.Logger` so then the logging seems inconsistent in the rest of this file, is there a reason for it?

Also could we use context to set the logger so it's available throughout the code?

internal/bg/availability_queue_test.go
```diff
@@ -0,0 +1,80 @@
...
+		asm, _ := bus.NewAvailabilityStatusMessage(msg)
+		require.EqualValues(t, "1", asm.SourceID)
+		wg.Done()
+	})
+
+	msg, _ := bus.AvailabilityStatusMessage{SourceID: "1"}.GenericMessage()
+	EnqueueAvailabilityStatusRequest(&msg)
+	senderCancel()
+	wg.Wait()
+}
```
lzap, line 80: This tests all the three cases described above.

internal/bus/avail_status_msg.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return GenericMessage{}, fmt.Errorf("unable to marshal message: %w", err)
+	}
+
+	return GenericMessage{
+		Topic: AvailabilityStatusRequestTopic,
+		Key:   []byte(m.SourceID),
+		Value: payload,
+	}, nil
+}
```
lzap, line 33: This is the payload for the check.

internal/bus/bus.go
```diff
@@ -0,0 +1,24 @@
...
+var broker Broker = &noopBroker{}
+
+//nolint:wrapcheck
+func Send(ctx context.Context, messages ...*GenericMessage) error {
+	return broker.Send(ctx, messages...)
+}
+
+func Consume(ctx context.Context, topic string, handler func(ctx context.Context, message *GenericMessage)) {
+	broker.Consume(ctx, topic, handler)
+}
```
lzap, line 24: This is the main bus interface.

internal/bg/initialize.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+)
+
+// Maximum batch size for each batch send, also an incoming buffered channel size to prevent
+// incoming requests to overload the sender.
+const availabilityStatusBatchSize = 1024
+
+// Initialize starts background goroutines. Use context cancellation to stop them.
+func Initialize(ctx context.Context) {
+	logger := ctxval.Logger(ctx).With().Bool("bg", true).Logger()
```
ezr-ondrej, line 18: could we use `background` or something even more descriptive? `bg` is quite vague and could be missinterpreted by someone not familiar with the code itself.

internal/bus/kafka.go
```diff
@@ -0,0 +1,204 @@
...
+func (b *kafkaBroker) NewReader(ctx context.Context, topic string) *kafka.Reader {
+	return kafka.NewReader(kafka.ReaderConfig{
+		Brokers:     config.Kafka.Brokers,
+		Dialer:      b.dialer,
+		Topic:       topic,
+		StartOffset: kafka.LastOffset,
+		Logger:      kafka.LoggerFunc(newContextLogger(ctx)),
+		ErrorLogger: kafka.LoggerFunc(newContextErrLogger(ctx)),
+	})
+}
```
lzap, line 146: One thing tho - when statuser is started, it should start reading from the end (last offset), we do not want to read the whole history of the topic. However, during my testing it looks like when I start a new reader, it reads the whole history regardless of the `StartOffset` setting. Do you see the same behavior? I am not sure why I see that.

ezr-ondrej (commented):
Just first read, I still don't fully understand the code, so I'll dig deeper.

internal/bus/message.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Key is used for topic partitioning. Can be nil.
+	Key []byte
+
+	// Value is the payload. Typically, a JSON marshaled data.
+	Value []byte
+
+	// List of key-value pairs for each message.
+	Headers map[string]string
+}
```
lzap, line 20: This is the main message struct that is used in our own bus interface. It can be converted (copyied) to/from Kafka native message, which is exactly the same (plus few more fields).

internal/config/config.go
```diff
@@ -207,6 +220,32 @@ func Initialize(configFiles ...string) {
...
+				config.Kafka.SASL.SaslMechanism = *b.Sasl.SaslMechanism
+			}
+			if b.Sasl.Username != nil && *b.Sasl.Username != "" {
+				config.Kafka.SASL.Username = *b.Sasl.Username
+			}
+			if b.Sasl.Password != nil && *b.Sasl.Password != "" {
+				config.Kafka.SASL.Password = *b.Sasl.Password
+			}
+		}
+
```
lzap, line 248: Clowder passes in multiple brokers, each one of them with a unique TLS CA and SASL credentials. I am assuming that they are all the same, all other services I looked up simply use the first server and ignore the rest. That is weird, nobody was able to answer me questions in this regard. So let's try this in stage to see what happens. Likely this can fail, then we would also use only the first server.

scripts/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Populate database with some data, with the sources app running do:
+
+    ./sources.seed.sh
+
+### Clean up
+
+When you want to start over, to delete the database and user and git checkouts:
+
+    ./sources.clean.sh
```
lzap, line 66: Moved both Sources and Kafka under a single umbrella - a single README in markdown.

internal/bg/availability_queue.go
```diff
@@ -0,0 +1,64 @@
...
+				messageBuffer = make([]*bus.GenericMessage, 0, batchSize)
+			}
+		case <-ctx.Done():
+			ticker.Stop()
+			length := len(messageBuffer)
+
+			if length > 0 {
+				logger.Trace().Int("messages", length).Msgf("Sent %d availability request messages (cancel)", length)
+				send(ctx, messageBuffer...)
+				messageBuffer = make([]*bus.GenericMessage, 0, batchSize)
```
lzap, line 60: Ha these three reallocations need to be reslicing instead, this creates unnecessary GC pressure. Will fix this in the next rebase.

				messageBuffer = messageBuffer[:0]

cmd/pbstatuser/main.go
ezr-ondrej: @lzap ? any reason for the inconsistent `log`/`logger` here?

internal/bg/availability_queue.go
```diff
@@ -0,0 +1,64 @@
...
+// buffered channel for incoming requests
+var asRequestChan = make(chan *bus.GenericMessage, availabilityStatusBatchSize)
+
+// EnqueueAvailabilityStatusRequest prepares a status request check to be sent in the next
+// batch to the platform bus. Messages can be delayed up to several seconds until sent.
+// The function can block if the enqueueing channel is full.
+func EnqueueAvailabilityStatusRequest(msg *bus.GenericMessage) {
+	asRequestChan <- msg
+}
+
```
adiabramovitch, line 20: This is the msg being sent to Kafka through the asRequestChan? 
I would change the name of the channel.

adiabramovitch (commented):
Haven't checked that yet, just some clarifications :)

internal/bus/kafka.go
lzap: Ah rebasing is bad, I should stop doing that and push separate commits. Its just I am worried about us accidentally merging all of them without squash :-)

internal/bus/bus.go
lzap: Sure, just a small detail: suggest a name :)

cmd/pbstatuser/main.go
lzap: That was not intentional, should lead to the same result tho. Renamed.

cmd/pbstatuser/main.go
lzap: Oh no that's a bug. Should definitely use logger, yeah.

internal/bg/initialize.go
lzap: Sure, I renamed this and the package since you mentioned it. It is a bit too long to my taste, but I am okay.

internal/bus/kafka.go
lzap: When testing this, keep this in mind - then there are existing messages in a topic and a worker starts, it should NOT process them, only new ones. It does not work this way currently, I don't know why yet.

internal/bg/availability_queue.go
lzap: After a short chat, I understand. Renaming to `kafkaAvailabilityRequest`. :-D

internal/bg/availability_queue.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+)
+
+// buffered channel for incoming requests
+var asRequestChan = make(chan *bus.GenericMessage, availabilityStatusBatchSize)
+
+// EnqueueAvailabilityStatusRequest prepares a status request check to be sent in the next
+// batch to the platform bus. Messages can be delayed up to several seconds until sent.
+// The function can block if the enqueueing channel is full.
+func EnqueueAvailabilityStatusRequest(msg *bus.GenericMessage) {
```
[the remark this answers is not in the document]
lzap: You are right, it does not block (unless channel buffer is empty) and it enqueues messages for a Kafka batch send.
Changed the channel name.

internal/bg/availability_queue.go
```diff
@@ -0,0 +1,64 @@
...
+	asRequestChan <- msg
+}
+
+func send(ctx context.Context, messages ...*bus.GenericMessage) {
+	err := bus.Send(ctx, messages...)
+	if err != nil {
+		ctxval.Logger(ctx).Warn().Err(err).Msg("Unable to send availability check messages")
+	}
+}
```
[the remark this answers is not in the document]
lzap: Yep, that`s exactly what it does: `// send a message to the background worker for availability check`

internal/bus/kafka.go
```diff
@@ -0,0 +1,204 @@
...
+
+var _ Broker = &kafkaBroker{}
+
+var (
+	DifferentTopicErr       = errors.New("messages in batch have different topics")
+	UnknownSaslMechanismErr = errors.New("unknown SASL mechanism")
+)
+
+func createSaslMechanism(saslMechanismName string, username string, password string) (sasl.Mechanism, error) {
```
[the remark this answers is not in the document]
lzap: Renamed.

internal/bus/kafka.go
```diff
@@ -0,0 +1,204 @@
...
+// therefore topic must be set on the message-level. Make sure to close it with Close() function.
+func (b *kafkaBroker) NewWriter(ctx context.Context) *kafka.Writer {
+	return &kafka.Writer{
+		Addr:        kafka.TCP(config.Kafka.Brokers...),
+		Transport:   b.transport,
+		Logger:      kafka.LoggerFunc(newContextLogger(ctx)),
+		ErrorLogger: kafka.LoggerFunc(newContextErrLogger(ctx)),
+	}
+}
+
```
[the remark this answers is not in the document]
lzap: This function transform from "our" message (GenericMessage) to native Kafka message from the library implementation.

lzap:
Rebased.

scripts/kafka.setup.sh
```diff
@@ -0,0 +1,21 @@
...
+
+    echo "Extracting Kafka..."
+    tar -xzf kafka.tgz --strip 1
+  fi
+
+  echo "Formatting journal..."
+  bin/kafka-storage.sh format -t $UUID -c config/kraft/server.properties
+popd
+
+echo "All done, run ./kafka.start.sh"
```
lzap, line 21: This is I think missing `test -d kafka || mkdir kafka` at the very beginning.

scripts/kafka.start.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+BASEDIR=$(dirname $0)
+source $BASEDIR/kafka.conf
+[[ -f $BASEDIR/kafka.local.conf ]] && source $BASEDIR/kafka.local.conf
+
```
adiabramovitch, line 5: After checking the scripts: creating the Kafka directory is missing. 
other than that, it works :) 
image

scripts/kafka.setup.sh
ezr-ondrej: We might even want to do test -d $BASEDIR/kafka || mkdir $BASEDIR/kafka ?

lzap: Yep let me do that.

scripts/kafka.start.sh
lzap: I am not sure what exactly do you mean, but after you reboot all files in `/tmp` will be gone. Also systemd can clean up this directory, feel free to edit the config/kraft/server.properties file and use /var/lib/kafka or some other directory you want to keep.

You know what, let me move formatting into the start script, this way the directory will always be formatted.

lzap:
Rebased.

lzap:
Panic during migration, taking a look:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x787902]

goroutine 1 [running]:
github.com/RHEnVision/provisioning-backend/internal/config.Initialize({0xc0003dfe80?, 0x2?, 0xb97220?})
	/build/internal/config/config.go:235 +0x942
main.main()
	/build/cmd/pbmigrate/main.go:21 +0xa7
```

scripts/kafka.start.sh
adiabramovitch: Not sure what are you commenting about. 
I tried to run the script but the Kafka directory was missing.
I created it manually, then ran the script and it worked.
just wanted to remind you to add the creation of the Kafka directory in the script :)

lzap:
All good, although this is far from perfect, I think we can merge if you don't have any comments and improve from here. Next steps:

* Creating topic on app-interface
* Clowder enablement
* Fixing SASL / config (if needed)
* Adding RHN headers to the GenericMessage
* Implementing actual checks and sending responses

ezr-ondrej (approved):
I take it as @adiabramovitch has tested it, I've as well and it was sending some messages to Kafka, so lets do it so we can continue :)

PR #288, opened 2022-11-03
refactor: Error payload with messages

avitova:
Probably the last part of error payload refactoring. Feel free to tear this apart; there are several 'problematic' parts, for which I saved this one as the last part covering the HMSPROV-250.
This should flag the error with the error status correctly. For that purpose, I have added a ClientErr and SourcesErr, these should be in each error in the clients and http packages. Client and Sources Errors are probably worth the attention.


internal/clients/http/sources_errors.go
```diff
@@ -8,10 +8,12 @@ import (
 )
 
 var (
-	MoreThanOneAuthenticationForSourceErr = errors.New("more than one authentication")
-	AuthenticationForSourcesNotFoundErr   = fmt.Errorf("authentications for source weren't found in sources app: %w", clients.NotFoundErr)
-	ApplicationNotFoundErr                = fmt.Errorf("application not found is sources app: %w", clients.NotFoundErr)
-	ApplicationTypeNotFoundErr            = fmt.Errorf("application type 'provisioning' not found: %w", clients.NotFoundErr)
-	SourceNotFoundErr                     = fmt.Errorf("source not found: %w", clients.NotFoundErr)
-	AuthenticationSourceAssociationErr    = errors.New("authentication associated to source id not found")
+	SourcesErr                            = errors.New("HTTP client")
```
lzap, line 11:
```suggestion
	SourcesErr                            = errors.New("sources HTTP client")
```

internal/clients/http/sources_errors.go
```diff
@@ -8,10 +8,12 @@ import (
...
+	MoreThanOneAuthenticationForSourceErr = fmt.Errorf("%w: more than one authentication", SourcesErr)
+	ApplicationNotFoundErr                = errors.New(SourcesErr.Error() + fmt.Errorf("application not found is sources app: %w", clients.NotFoundErr).Error())
+	ApplicationTypeNotFoundErr            = errors.New(SourcesErr.Error() + fmt.Errorf("application type 'provisioning' not found: %w", clients.NotFoundErr).Error())
+	SourceNotFoundErr                     = errors.New(SourcesErr.Error() + fmt.Errorf("source not found: %w", clients.NotFoundErr).Error())
+	AuthenticationSourceAssociationErr    = fmt.Errorf("%w: authentication associated to source id not found", SourcesErr)
+	AuthenticationForSourcesNotFoundErr   = errors.New(SourcesErr.Error() + fmt.Errorf("authentications for source weren't "+
+		"found in sources app: %w", clients.NotFoundErr).Error())
```
lzap, line 18: Wait what? :-)

Why don't you use `fmt.Errorf` just like for the client errors above?

Wrapping errors like this via `errors.New` will not work as intended.

internal/dao/dao_errors.go
```diff
@@ -8,10 +8,12 @@ import (
 
 var (
 	// ErrNoRows is returned when there are no rows in the result
+	// Typically, REST requests should end up with 404 error
 	ErrNoRows = pgx.ErrNoRows
 
 	// ErrAffectedMismatch is returned when unexpected number of affected rows
 	// was returned for INSERT, UPDATE and DELETE queries.
+	// Typically, REST requests should end up with 404 error
 	ErrAffectedMismatch = errors.New("unexpected affected rows")
```
lzap, line 17: Hmm when you update or delete something and it fails, it should not be 404 but perhaps something else. I suggest [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) perhaps.

internal/clients/http/sources_errors.go
avitova: Unfortunately, fmt.Errorf can not combine two "%w". We would have to wrap it in a different way in that case.

internal/clients/http/sources_errors.go
lzap, line 11:
```suggestion
	SourcesErr                            = errors.New("sources HTTP client")
```

internal/payloads/error_payload.go
```diff
@@ -67,40 +69,117 @@ func PubkeyDuplicateError(ctx context.Context, message string, err error) *Respo
...
+		logger.Error().Msg(message)
 	}
 	return &ResponseError{
 		HTTPStatusCode: 500,
-		Message:        msg,
+		Message:        message,
 		TraceId:        ctxval.TraceId(ctx),
 		Error:          err,
 	}
 }
```
lzap, line 173: I think you could remove this special error for client initialization and replace it with sources/image builder error defined the same way as the other errors (wrapping a common error).

internal/payloads/error_payload.go
```diff
@@ -67,40 +69,117 @@ func PubkeyDuplicateError(ctx context.Context, message string, err error) *Respo
...
+	if logger := ctxval.Logger(ctx); logger != nil {
+		logger.Error().Msg(message)
+	}
+	return &ResponseError{
+		HTTPStatusCode: status,
+		Message:        message,
+		TraceId:        ctxval.TraceId(ctx),
+		Error:          err,
+	}
+}
```
lzap, line 148: I think you can take advantage of wrapping and have just a single payload function for all client errors:

* common client errors
* sources client errors
* image builder client errors (to be done)

One switch statement could find which one is it out of the three, the other one can then decide the message and code.

internal/payloads/error_payload.go
lzap, line 173: I think you could remove this special error for client initialization and replace it with sources/image builder error defined the same way as the other errors (wrapping a common error).

lzap:
It's a good start, major concerns:

I very much like how you use error wrapping for client/sources errors. This way any caller can easily find out which kind is it and also which type of error. Can you do the very same but for the image builder client? We currently don't have any errors defined there yet.

This approach gives you powerful thing to do: you can only define one payload function `NewClientError` that can be used for all client errors (I would also include the initialization error which for some reason is handled differently but it should not be). This function would investigate the error and assign appropriate messages and HTTP codes. It would work for any client error:

* generic http client
* sources
* image builder

internal/clients/http/sources_errors.go
adiabramovitch: Why do we need errors.New?
Why not incorporate SourcesErr.Error() inside the string of fmt.Errorf?

internal/payloads/error_payload.go
```diff
@@ -67,40 +69,117 @@ func PubkeyDuplicateError(ctx context.Context, message string, err error) *Respo
 	}
 }
 
-func ClientError(ctx context.Context, client string, message string, err error, status int) *ResponseError {
-	msg := fmt.Sprintf("%v error: %s: %v", client, message, err)
+func ClientError(ctx context.Context, err error) *ResponseError {
+	var status int
+	var message string
+	switch err {
```
adiabramovitch, line 75: fix needed :)

internal/services/aws_reservation_service.go
```diff
@@ -116,14 +113,15 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
...
+			renderError(w, r, payloads.NewClientInitializationError(r.Context(), "image builder", ibErr))
 			return
 		}
 
 		// Get AMI
 		ami, ibErr = IBClient.GetAWSAmi(r.Context(), reservation.ImageID)
 		if ibErr != nil {
-			renderError(w, r, payloads.ClientError(r.Context(), "Image Builder", "can't get ami from image builder", ibErr, 500))
+			renderNewErrorFromClientErr(w, r, err)
+			return
```
adiabramovitch, line 124: Why are we using renderError and the other time we use renderNewErrorFromClientErr? what is the guideline?

internal/services/error_renderer.go
```diff
@@ -41,12 +44,22 @@ func renderError(w http.ResponseWriter, r *http.Request, renderer render.Rendere
...
+		renderError(w, r, payloads.NewNotFoundError(r.Context(), resource, err))
 	} else {
 		renderError(w, r, payloads.NewDAOError(r.Context(), resource, err))
 	}
 }
 
+func renderNewErrorFromClientErr(w http.ResponseWriter, r *http.Request, err error) {
+	if errors.Is(err, clients.ClientErr) {
+		renderError(w, r, payloads.ClientError(r.Context(), err))
+	} else if errors.Is(err, clientshttp.SourcesErr) {
```
adiabramovitch, line 56: What about IB error?

internal/services/gcp_reservation_service.go
```diff
@@ -93,14 +90,14 @@ func CreateGCPReservation(w http.ResponseWriter, r *http.Request) {
 	IBClient, ibErr := clients.GetImageBuilderClient(r.Context())
 	logger.Trace().Msg("Creating IB client")
 	if ibErr != nil {
-		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "image builder client", ibErr))
+		renderError(w, r, payloads.NewClientInitializationError(r.Context(), "image builder", ibErr))
```
adiabramovitch, line 93: ditto

avitova: Okay, great! I was not aware of this. I can see your point now. I should change this. :)

internal/clients/http/sources_errors.go
lzap: When do you need to wrap two errors?  All client/sources/ib errors should only wrap one of the three "root" errors:

```
RootErr = errors.New("a basic error which we can use for errors.Is to find which kind of error it is")
SomeSpecificErr    = fmt.Errorf("%w: xxx", RootErr)
```

And you can of course wrap this again in the code:

    return fmt.Errorf("something is wrong: %w", SomeSpecificErr)

I am not sure what you are trying to achieve here, but concatenating `Error()` calls is not what we want to do, that is giving up on error handling completely.

internal/services/error_renderer.go
lzap: Yeah let's treat all the three clients the same way:

* generic HTTP
* sources
* image builder

lzap:
Just to give you little more context, we discussed this about a month ago and the outcome was that we should simplify error handling by strictly using the standard Go string errors (errors.New) with error wrapping (fmt.Errorf) for all the error handing in the up.

Up until the point when we process these errors in the service package and create one special error payload that will render to JSON. We want to define a clean error JSON output that will give some good information to the user (a string which can be translated in the UI later) as well as detailed info for us to debug issues.

lzap:
Okay we had a chat together and we agreed on the following solution:

* All NewXXXError payloads either must have "message" as a parameter, or will figure a message on itself via `errors.Is/errors.As` calls.
* For all errors that needs to be 404 we should wrap a value error (e.g. clients.NotFound or dao.NotFoundErr) so the payload function can easily determine this via `errors.Is`.
* For all errors that needs to be 405 (not allowed) we should wrap a value error (e.g. dao.MismatchErr) so the payload function can easily determine this via `errors.Is`.
* Since we also want to tell the user in which component there was an error, a new wrapping types `SourceError` and `ImageBuilderError` (see below) will be created. They will be used for all errors, e.g. `SourceNotFoundErr = fmt.Errorf("source not found: %w", WrapSourcesErr(clients.NotFoundErr))`. This way the payload code can use `errors.As` to find which type of error it was and construct the user message accordingly.
* Use the both techniques described above for any other case when we want to provide a good user message or a good HTTP return code. Typically only 404 and 405 are viable options for REST.
* In all our error handling, we should be only using `fmt.Errorf` to keep Go errors as simple as possible.
* All client errors should ideally be in a single `client/errors.go` file (this should be done separately after this is merged) 
* Let's just drop my idea of list of constant string for better i18n, we can add i18n later. If we stick to this plan, all user messages will be either in NewXXXError argument or in the function code itself.

```go
type SourceError struct {
	err error
}

func WrapSourceError(err error) error {
	return &SourceError{err: error}
}

func (e *wrapError) Error() string {
	return e.err.Error()
}

func (e *wrapError) Unwrap() error {
	return e.err
}
```

lzap:
With a recent bug we had in stage, we should add more:

* Every error response must also log into the context logger with either Error() level or Warn() level (for 4xx)
* Every log line must have also `Err(err)` call which creates "err" key with the full error message

I think this is pretty useful. I created a PR with this https://github.com/RHEnVision/provisioning-backend/pull/297 although I haven't implemented `Warn().Err(err)` which I think can be still useful to have (in case we accidentally error something as 4xx but in fact it is some error we need to investigate further)

internal/clients/http/sources_errors.go
adiabramovitch: I think this chang is required @avitova

PR #289, opened 2022-11-06
ci: Add rhsm-api-proxy for rhel images

akhil-jha:
Image-builder depends upon rhsm-api-proxy for creating rhel images.


akhil-jha:
It's breaking the deploy.  :/

PR #291, opened 2022-11-07
chore: update all deps

lzap:
SSIA

This one was a hard one until I figured out how to downgrade a package.


mk/deps.mk
```diff
@@ -11,6 +11,8 @@ download-deps: ## Download Go modules
 .PHONY: update-deps
 update-deps: ## Update Go modules to latest versions
 	go get -u all
+	go get github.com/jackc/puddle/v2@v2.0.0 # https://github.com/jackc/puddle/issues/26
+	go get github.com/jackc/tern/v2@v2.0.0-beta.3 # not released yet
```
ezr-ondrej, line 15: so these two just won't be in `go.mod` nor `go.sum`?
it seems bit weird, now we won't be able to use go tools to install deps, which might be contraintuitive for many :thinking:

lzap: They are in go.mod, however, the moment you do "go get -u" the puddle/v2 will be upgraded to newer version that requires Go 1.19. The line 14 is "downgrade", this is what I was told on the Gophers Slack as the best solution.

The other one is opposite, just making sure that beta is used and go get does not downgrade it to stable, which it can actually do in some cases.

ezr-ondrej: I was looking and I did not find any mention of tern in go.sum, I'm not sure about puddle.
Tho for me `go get -u all` fails, so this wouldn't fix that, or would it? :thinking:

ezr-ondrej: Ok, I'm just blind, I missed it, it just did not change :)

ezr-ondrej (approved):

PR #292, opened 2022-11-08
fix(logging): Disable middlewares for status routes

ezr-ondrej:
Mainly to disable logging for ping route, but other middlewares shall be enabled for application routes only too.

Fixes HMSPROV-333


lzap (approved):
Sure.

PR #295, opened 2022-11-08
fix(config): unleash token as bearer header

ezr-ondrej:
From others I'm guessing that this should fix unleash connection when token is required.
Also uses unleash project.


ezr-ondrej:
I'm sorry if this causes trouble, but I have no other way to try this :)

PR #296, opened 2022-11-11
fix(config): guard for non-exixtend kafka config

ezr-ondrej:


lzap (approved):
Yeah sorry this code is definitely clunky, something I meant to further improve along the way.

PR #297, opened 2022-11-11
fix: print full errors in logs

lzap:
We see an error in stage, this should show the root cause.


ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #298, opened 2022-11-11
fix: null for aws_reservation_id when pending

lzap:
When reservation is created and it is still not fulfilled, the AWS reservation id was NULL in the database, but the struct field was string and not pointer to string. This caused scany library not to be able to scan the `nil` value. This fixes it.

OpenAPI clients must be aware that the field can be `null` in case the reservation is not yet finished.

### DIFF
--- a/internal/payloads/reservation_payload.go
+++ b/internal/payloads/reservation_payload.go
@@ -60,8 +60,8 @@ type AWSReservationResponsePayload struct {
 	// The ID of the image from which the instance is created.
 	ImageID string `json:"image_id"`
 
-	// The ID of the aws reservation which was created.
-	AWSReservationID string `json:"aws_reservation_id"`
+	// The ID of the aws reservation which was created, or missing if not created yet.
+	AWSReservationID string `json:"aws_reservation_id,omitempty"`
 
 	// Optional name of the instance(s).
 	Name *string `json:"name"`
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
I thought the error is on https://github.com/RHEnVision/provisioning-backend/blob/a5b3d9c552953f3ddb7824c67712719b7a83bd27/internal/models/reservation_model.go#L123 but I might have been wrong, lets try, we should handle reservation not being made yet as well :+1:

lzap:
Nah that one cannot be nil: `instance_id TEXT NOT NULL UNIQUE`

The CI is not loading for me and I am off for a weekend in a few. Feel free to take it from here, but I believe this was the issue.

lzap:
Akhil commented:

```
19:15:54 E               iqe_provisioning_api.exceptions.ApiTypeError: Invalid type for variable 'aws_reservation_id'. Required value type is str and passed type was NoneType at ['received_data']['aws_reservation_id']

ipdb> response_kwargs
{'reservation_id': 6, 'pubkey_id': 5, 'source_id': '5', 'region': 'us-east-1', 'instance_type': 't1.micro', 'amount': 1, 'image_id': 'ami-0ba22849118c7101a', 'aws_reservation_id': None, 'name': 'test_WSqna2jmsv', 'poweroff': True}
```

```
ipdb> type(response_kwargs['aws_reservation_id'])
<class 'NoneType'>
And from apispec:
      "v1.AWSReservationResponse": {
        "properties": {
          "amount": {
            "format": "int32",
            "type": "integer"
          },
          "aws_reservation_id": {
            "type": "string"
          },
```

ezr-ondrej:
Yeah, we will need to change the OpenAPI to expect null for instances ( I was thinking of simple fix by empty array `[]`, but that would be missleading IMHO).

lzap:
Let me rebase on top of your PR from today to see if this fixes it.

ezr-ondrej:
I've rebased for you, let :eyes:

internal/payloads/reservation_payload.go
```diff
@@ -60,8 +60,8 @@ type AWSReservationResponsePayload struct {
 	// The ID of the image from which the instance is created.
 	ImageID string `json:"image_id"`
 
-	// The ID of the aws reservation which was created.
-	AWSReservationID string `json:"aws_reservation_id"`
+	// The ID of the aws reservation which was created, or missing if not created yet.
+	AWSReservationID string `json:"aws_reservation_id,omitempty"`
```
ezr-ondrej, line 64: Trying to omit it on empty. Not sure if that's more clear, tho. I'd love to be able to pass null

lzap:
It might be as simple as the IQE not expecting any JSON nulls, but omitting is okay I guess.

ezr-ondrej:
Let's try ommiting

PR #299, opened 2022-11-14
fix(reservation): generic reservation by id

ezr-ondrej:
Correctly render generic reservation on generic reservation endpoint.

Fixes HMSPROV-349

### DIFF
--- a/internal/services/reservations_service.go
+++ b/internal/services/reservations_service.go
@@ -14,6 +14,7 @@ import (
 
 var (
 	UnknownProviderTypeError         = errors.New("unknown provider type parameter")
+	ProviderTypeMismatchError        = errors.New("reservation type does not match requested provider type")
 	ProviderTypeNotImplementedError  = errors.New("provider type not yet implemented")
 	InvalidRequestPubkeyNewError     = errors.New("provide either existing (via NewName/NewBody) or new pubkey (ExistingID)")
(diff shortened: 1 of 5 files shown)


internal/services/reservations_service.go
```diff
@@ -58,6 +58,7 @@ func ListReservations(w http.ResponseWriter, r *http.Request) {
 }
 
 func GetReservationDetail(w http.ResponseWriter, r *http.Request) {
+	provider := chi.URLParam(r, "TYPE")
```
lzap, line 61: Hmm okay I remember deleting this :-)

Can we check if the provider passed in as URL parameter matches provider in the database? If not, we should definitely print out an error. Tho I think we won't be able to map fields form the database correctly before we actually load up the type from SQL column. So this might only be post check, hmm.

internal/services/reservations_service.go
```diff
@@ -72,7 +73,12 @@ func GetReservationDetail(w http.ResponseWriter, r *http.Request) {
 		return
 	}
```
lzap, line 74: At this point we should be able to validate the type I think.

akhil-jha (approved):
These changes work for me :) 
I see lzap has some comments. If the code changes, let me know, I'll test it out again :)

ezr-ondrej:
Added the validation in new commit for easy review, happy to squash once code checked :)

ezr-ondrej:
Merging as it has many approvals.

PR #300, opened 2022-11-15
feat: availability check kafka topic

lzap:
This adds the internal topic for the check.


lzap:
Let me add a log into statuser log output to see if a message is actually being received.

lzap:
According to:

https://github.com/RedHatInsights/platform-mq/blob/master/topics/topics.json

the common prefix is "platform" so rebased to: `platform.provisioning.internal.availability-check`

lzap:
https://github.com/RedHatInsights/platform-mq/pull/250

ezr-ondrej (approved):
We are ready in stage, so lets do this :tada:

PR #301, opened 2022-11-16
refactor: Error payload with messages

avitova:
The previous PR #288 got closed due to some changes in the repository I had to make. So hopefully you will be able to tell the differences. :)


lzap (changes requested):
Small changes here and there, but I think this is ready.

Please rebase and ping me so we can get this in ASAP, you suffered enough rebasing. Next time, I think we should split this into smaller PRs.

lzap:
So there is one big problem with wrapping - linter wants us to not use dynamic errors and that is a very good point, we would need to export both unwrapped and wrapped errors which would create confusion.

Therefore, I suggest to simplify this and drop the wrapper types and simply define three new types (example for sources):

```go
type SourceError string

func (e SourceError) Error() string {
	return string(e)
}
```

And then this should be easy to use:

```go
var MoreThanOneAuthenticationForSourceErr = SourceError("more than one authentication")
```

Then finding out if an error is of source type, you can simply use `errors.As` and to find a particular error just use `errors.Is`. This also looks way cleaner when I see this, I am not sure why I was pushing so hard on wrapping.

Tho we should still use normal `fmt.Errorf` for normal wrapping of these errors, that's not changing.

avitova:
Okay, as we have agreed, this is now simplified; each Client, Sources, and ImageBuilder errors are now recognizable. Let me know what you think about this.

internal/clients/http/sources_errors.go
```diff
@@ -8,10 +8,11 @@ import (
 )
 
 var (
-	AuthenticationForSourcesNotFoundErr = fmt.Errorf("authentications for source weren't found in sources app: %w", clients.NotFoundErr)
-	ApplicationNotFoundErr              = fmt.Errorf("application not found in sources app: %w", clients.NotFoundErr)
-	ApplicationReadErr                  = fmt.Errorf("application read returned no application type in sources: %w", clients.NotFoundErr)
-	ApplicationTypeNotFoundErr          = fmt.Errorf("application type 'provisioning' not found: %w", clients.NotFoundErr)
-	SourceNotFoundErr                   = fmt.Errorf("source not found: %w", clients.NotFoundErr)
-	AuthenticationSourceAssociationErr  = errors.New("authentication associated to source id not found")
+	MoreThanOneAuthenticationForSourceErr = errors.New("more than one authentication")
```
lzap, line 11: Have you rebased on top of "main"? Because this error is no longer in use. But it's fine we can merge and remove later.

internal/payloads/error_payload.go
```diff
@@ -66,40 +69,89 @@ func PubkeyDuplicateError(ctx context.Context, message string, err error) *Respo
...
+		return 500, "unknown image type"
+	} else if errors.Is(err, httpClients.AmiNotFoundInStatusErr) {
+		return 404, "image builder did not find AMI in status"
+	} else if errors.Is(err, httpClients.NameNotFoundInStatusErr) {
+		return 404, "image builder did not find image name in image status"
+	} else if errors.Is(err, httpClients.IDNotFoundInStatusErr) {
+		return 404, "image builder did not find project id in image status"
+	}
+	return 0, ""
+}
```
lzap, line 117: Do not export these functions outside of the package, unless they are used elsewhere, which I think they should not be.

Reviewers: We sort of landed on this solution we would like to test if it successful. The idea is to have a single place where we set all HTTP result codes and user-facing messages. If this turns out to be a success, we can refactor this into some types (maps perhaps) and make it more generic. For now this is only for HTTP clients.

internal/payloads/error_payload.go
```diff
@@ -66,40 +69,89 @@ func PubkeyDuplicateError(ctx context.Context, message string, err error) *Respo
...
+		return Response(ctx, status, message, err)
 	}
+	if status, message := SourcesErrorHelper(err); status != 0 {
+		return Response(ctx, status, message, err)
+	}
+	if status, message := ClientErrorHelper(err); status != 0 {
+		return Response(ctx, status, message, err)
+	}
+	return Response(ctx, 500, "HTTP service returned unknown client error", err)
 }
```
lzap, line 145: I see one major issue: this function is called `NewError` thus most of us will shortly start it using for ALL errors while what it actually does is setting up details for HTTP client errors only. I suggest to rename it to `NewClientError` for now and work with it for some weeks and if this approach proves to be valid, we can even extend this to all other components (DAO, generic errors).

lzap (changes requested):
Few nitpick, tho I will ask others to chip in reviews.

We spent a lot of time with Anna on this trying to find a good balance of easy to use error handling for HTTP clients and clean code. We landed on common set of errors defined in the client package which directly map to HTTP codes: `clients/errors.go`. These are wrapped into individual errors where appropriate.

Then there is set of helper functions (for now) which do set those HTTP codes and provide more details. We could put the HTTP code into the error itself by creating a new type, but we dropped the idea of creating complex error types and the message will be always different.

The main idea is to have a single place for all user-facing information errors going forward. Let me know how you like this, I suggest to try to work with this for some time and then decide if we want to expand this approach elsewhere (DAO, other errors).

lzap:
Requested reviews from others because Go error handling is a hot topic :-) See my review or motivation.

internal/payloads/error_payload.go
ezr-ondrej: I like the approach :+1:

internal/payloads/error_payload.go
```diff
@@ -66,40 +69,89 @@ func PubkeyDuplicateError(ctx context.Context, message string, err error) *Respo
...
+	if errors.Is(err, clients.NotFoundErr) {
+		return 404, "service returned not found or no data"
+	} else if errors.Is(err, clients.UnauthorizedErr) {
+		return 401, "service returned unauthorized"
+	} else if errors.Is(err, clients.ForbiddenErr) {
+		return 403, "service returned forbidden"
+	} else if errors.Is(err, clients.Non2xxResponseErr) {
+		return 500, "service did not return 2xx"
+	}
+	return 0, ""
```
ezr-ondrej, line 82: I'd love to see this being `500, "Unknown client error"` and logging it as warning into logs:
`logger.Warn().Err(err).Msg("Unrecognized client error")`

This would account for situation we add a new error in clients but forget to add it here.

ezr-ondrej (commented):
Yup, this is basically what I had imagined, I'd only cover the corner case of forgotten error, not to return status code 0 with empty message, but apart of it I'm happy to merge and maybe refactor to some mapping Lukas talks about later :+1:

internal/payloads/error_payload.go
lzap: Actually you cannot do that, look at the call site of these three functions - it tests for zero and in that case there is a final `else` 500 statement.

It is clunky, I agree. I would probably keep having three separate `NewSourceError`, `NewImageBuilderError` and `NewHttpClientError` function doing this separately. That is sort of cleaner and we keep the context. But I am also fine with this "one function rules them all" solution which has big advantage if this is successful - we can simply start using just a single error payload function if we apply the same technique everywhere.

ezr-ondrej (approved):
Ok, got it ignore my comment ^_^
LGTM :+1:

internal/payloads/error_payload.go
ezr-ondrej: Oh, right, I missed that final case, that make sense :+1:
Lets do it then, I'm good :+1:

lzap (approved):

lzap:
It is Friday, I do not want to let Anna wait another week as it is a relatively large PR. Thanks, let's merge and improve.

Possible improvements from here:

* Bring errors in `clients/` package into a single file as they are scattered around.
* Add new error payload fields: version, build time, stacktrace: https://issues.redhat.com/browse/HMSPROV-368
* Do similar technique to DAO and other packages (in separate error functions like NewDaoError etc).

PR #302, opened 2022-11-16
feat(HMSPROV-345): Add source check availability per each provider

adiabramovitch:
This draft PR is meant to align the usage of channels and go routines. 
It is missing:
- testing

### DIFF
--- a/cmd/pbstatuser/main.go
+++ b/cmd/pbstatuser/main.go
@@ -7,11 +7,22 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
+	"time"
 
-	"github.com/RHEnVision/provisioning-backend/internal/clients/http/cloudwatchlogs"
-	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/redhatinsights/platform-go-middlewares/identity"
+
+	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/azure"
+	"github.com/RHEnVision/provisioning-backend/internal/clients/http/cloudwatchlogs"
+	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/ec2"
+	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/gcp"
+	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/image_builder"
+	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/sources"
+	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/kafka"
 	"github.com/RHEnVision/provisioning-backend/internal/logging"
 	"github.com/RHEnVision/provisioning-backend/internal/telemetry"
@@ -20,25 +31,193 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const ChannelBuffer = 32
+
+type SourceInfo struct {
+	Authentication clients.Authentication
+
+	SourceID string
+
+	Identity identity.XRHID
+}
+
+var (
+	chAws        = make(chan SourceInfo, ChannelBuffer)
+	chAzure      = make(chan SourceInfo, ChannelBuffer)
+	chGcp        = make(chan SourceInfo, ChannelBuffer)
+	chSend       = make(chan kafka.SourceResult, ChannelBuffer)
+	receiverWG   = sync.WaitGroup{}
+	processingWG = sync.WaitGroup{}
+	senderWG     = sync.WaitGroup{}
+)
+
 func processMessage(ctx context.Context, message *kafka.GenericMessage) {
 	logger := ctxval.Logger(ctx)
-	// This is just a test to see if Kafka is configured properly in ephemeral
-	logger.Warn().Msg("RECEIVED KAFKA MESSAGE")
 
-	// TODO: implement the source checking for AWS, Azure, GCP
+	// Get source id
+	asm, err := kafka.NewAvailabilityStatusMessage(message)
+	if err != nil {
+		logger.Warn().Msgf("could not get availability status message %s", err)
+		return
+	}
 
-	// This needs to be done in 4 goroutines.
-	// 1) Reads messages, calls sources to get auth type and ARN and then enqueue the work
-	// in one of three channels (AWS, Azure, GCP)
-	// 2) Reads from AWS channel, performs availability check, has configurable throttling.
-	// 3) Reads from Azure channel, performs availability check, has configurable throttling.
-	// 4) Reads from GCP channel, performs availability check, has configurable throttling.
+	sourceId := asm.SourceID
 
-	// Beware there is no database connection, no job queue, most of what api/worker processes have
-	// is not available because it should not be needed. The worker will simply use SDKs to check
-	// statuses and send out results via kafka.Send function to Sources.
-	// Also make sure that all goroutines are closed correctly on context cancel. A WaitGroup
-	// would be a really good here: https://gist.github.com/fracasula/b579d52daf15426e58aa133d0340ccb0
+	// Get sources client
+	sourcesClient, err := clients.GetSourcesClient(ctx)
+	if err != nil {
+		logger.Warn().Msgf("Could not get sources client %s", err)
+		return
+	}
+
+	// Fetch authentication from Sources
+	authentication, err := sourcesClient.GetAuthentication(ctx, sourceId)
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
+			logger.Warn().Msgf("Not found error: %s", err)
+			return
+		}
+		logger.Warn().Msgf("Could not get authentication: %s", err)
+		return
+	}
+
+	s := SourceInfo{
+		Authentication: *authentication,
+		SourceID:       sourceId,
+		Identity:       ctxval.Identity(ctx),
+	}
+
+	switch authentication.ProviderType {
+	case models.ProviderTypeAWS:
+		chAws <- s
+	case models.ProviderTypeAzure:
+		chAzure <- s
+	case models.ProviderTypeGCP:
+		chGcp <- s
+	case models.ProviderTypeNoop:
+	case models.ProviderTypeUnknown:
+		logger.Warn().Msg("Authentication provider type is unknown")
+	}
+}
+
+func checkSourceAvailabilityAzure(ctx context.Context) {
+	defer processingWG.Done()
+
+	for s := range chAzure {
+		sr := kafka.SourceResult{
+			SourceID: s.SourceID,
+			Identity: s.Identity,
+		}
+		// TODO: check if source is avavliable - WIP
+		sr.Status = kafka.StatusAvaliable
+		chSend <- sr
+	}
+}
+
+func checkSourceAvailabilityAWS(ctx context.Context) {
+	logger := ctxval.Logger(ctx)
+	defer processingWG.Done()
+
+	for s := range chAws {
+		sr := kafka.SourceResult{
+			SourceID: s.SourceID,
+			Identity: s.Identity,
+		}
+		_, err := clients.GetEC2Client(ctx, &s.Authentication, "")
+		if err != nil {
+			sr.Status = kafka.StatusUnavailable
+			sr.Err = err
+			logger.Warn().Msgf("Could not get aws assumed client %s", err)
+			chSend <- sr
+		} else {
+			sr.Status = kafka.StatusAvaliable
+			chSend <- sr
+		}
+	}
+}
+
+func checkSourceAvailabilityGCP(ctx context.Context) {
+	logger := ctxval.Logger(ctx)
+	defer processingWG.Done()
+
+	for s := range chGcp {
+		sr := kafka.SourceResult{
+			SourceID: s.SourceID,
+			Identity: s.Identity,
+		}
+		gcpClient, err := clients.GetGCPClient(ctx, &s.Authentication)
+		if err != nil {
+			sr.Status = kafka.StatusUnavailable
+			sr.Err = err
+			logger.Warn().Msgf("Could not get gcp client %s", err)
+			chSend <- sr
+			continue
+		}
+		_, err = gcpClient.ListAllRegions(ctx)
+		if err != nil {
+			sr.Status = kafka.StatusUnavailable
+			sr.Err = err
+			logger.Warn().Msgf("Could not list gcp regions %s", err)
+			chSend <- sr
+		} else {
+			sr.Status = kafka.StatusAvaliable
+			chSend <- sr
+		}
+	}
+}
+
+func sendResults(ctx context.Context, batchSize int, tickDuration time.Duration) {
+	logger := ctxval.Logger(ctx)
+	messages := make([]*kafka.GenericMessage, 0, batchSize)
+	ticker := time.NewTicker(tickDuration)
+	defer senderWG.Done()
+
+	for {
+		select {
+
+		case sr := <-chSend:
+			ctx = ctxval.WithIdentity(ctx, sr.Identity)
+			msg, err := sr.GenericMessage(ctx)
+			if err != nil {
+				logger.Warn().Msgf("Could not generate generic message %s", err)
+				continue
+			}
+			messages = append(messages, &msg)
+			length := len(messages)
+
+			if length >= batchSize {
+				logger.Trace().Int("messages", length).Msgf("Sending %d source availability status messages (full buffer)", length)
+				err := kafka.Send(ctx, messages...)
+				if err != nil {
+					logger.Warn().Msgf("Could not send source availability status messages (full buffer) %s", err)
+				}
+				messages = messages[:0]
+			}
+		case <-ticker.C:
+			length := len(messages)
+			if length > 0 {
+				logger.Trace().Int("messages", length).Msgf("Sending %d source availability status messages (tick)", length)
+				err := kafka.Send(ctx, messages...)
+				if err != nil {
+					logger.Warn().Msgf("Could not send source availability status messages (tick) %s", err)
+				}
+				messages = messages[:0]
+			}
+		case <-ctx.Done():
+			ticker.Stop()
+			length := len(messages)
+
+			if length > 0 {
+				logger.Trace().Int("messages", length).Msgf("Sending %d source availability status messages (cancel)", length)
+				err := kafka.Send(ctx, messages...)
+				if err != nil {
+					logger.Warn().Msgf("Could not send source availability status messages (cancel) %s", err)
+				}
+			}
+
+			return
+		}
+	}
 }
 
 func main() {
@@ -81,8 +260,22 @@ func main() {
 	}
 
 	// start the consumer
+	receiverWG.Add(1)
 	cancelCtx, consumerCancelFunc := context.WithCancel(ctx)
-	go kafka.Consume(cancelCtx, kafka.AvailabilityStatusRequestTopic, processMessage)
+	go func() {
+		defer receiverWG.Done()
+		kafka.Consume(cancelCtx, kafka.AvailabilityStatusRequestTopic, processMessage)
+	}()
+
+	// start processing goroutines
+	processingWG.Add(3)
+
+	go checkSourceAvailabilityAWS(cancelCtx)
+	go checkSourceAvailabilityGCP(cancelCtx)
+	go checkSourceAvailabilityAzure(cancelCtx)
+
+	senderWG.Add(1)
+	go sendResults(cancelCtx, 1024, 5*time.Second)
 
 	// metrics
 	logger.Info().Msgf("Starting new instance on port %d with prometheus on %d", config.Application.Port, config.Prometheus.Port)
@@ -96,7 +289,7 @@ func main() {
 	waitForSignal := make(chan struct{})
 	go func() {
 		sigint := make(chan os.Signal, 1)
-		signal.Notify(sigint, os.Interrupt, syscall.SIGTERM)
+		signal.Notify(sigint, syscall.SIGINT, syscall.SIGTERM)
 		<-sigint
 		if err := metricsServer.Shutdown(context.Background()); err != nil {
 			logger.Fatal().Err(err).Msg("Metrics service shutdown error")
@@ -117,6 +310,21 @@ func main() {
 
 	logger.Info().Msg("Worker started")
 	<-waitForSignal
+
+	// stop kafka receiver (can take up to 10 seconds) and wait until it returns
+	consumerCancelFunc()
+	receiverWG.Wait()
+
+	// close all processors and wait until it exits the range loop
+	close(chAws)
+	close(chAzure)
+	close(chGcp)
+	processingWG.Wait()
+
+	// close the sending channel and wait until it exits the range loop
+	close(chSend)
+	senderWG.Wait()
+
 	logger.Info().Msg("Consumer shutdown initiated")
 	consumerCancelFunc()
 	logger.Info().Msg("Shutdown finished, exiting")
--- a/internal/background/availability_queue_test.go
+++ b/internal/background/availability_queue_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/RHEnVision/provisioning-backend/internal/testing/identity"
 	_ "github.com/RHEnVision/provisioning-backend/internal/testing/initialization"
 
 	"github.com/RHEnVision/provisioning-backend/internal/kafka"
@@ -15,6 +16,8 @@ import (
 
 func TestQueueNormalSend(t *testing.T) {
 	ctx := context.Background()
+	ctx = identity.WithIdentity(t, ctx)
+
 	_ = kafka.InitializeStubBroker(16)
 
 	wg := sync.WaitGroup{}
@@ -28,7 +31,7 @@ func TestQueueNormalSend(t *testing.T) {
 		wg.Done()
 	})
 
-	msg, _ := kafka.AvailabilityStatusMessage{SourceID: "1"}.GenericMessage()
+	msg, _ := kafka.AvailabilityStatusMessage{SourceID: "1"}.GenericMessage(ctx)
 	EnqueueAvailabilityStatusRequest(&msg)
 	EnqueueAvailabilityStatusRequest(&msg)
 	wg.Wait()
@@ -36,6 +39,7 @@ func TestQueueNormalSend(t *testing.T) {
 
 func TestFullQueueSend(t *testing.T) {
 	ctx := context.Background()
+	ctx = identity.WithIdentity(t, ctx)
 	_ = kafka.InitializeStubBroker(16)
 
 	wg := sync.WaitGroup{}
@@ -50,7 +54,7 @@ func TestFullQueueSend(t *testing.T) {
 		wg.Done()
 	})
 
-	msg, _ := kafka.AvailabilityStatusMessage{SourceID: "1"}.GenericMessage()
+	msg, _ := kafka.AvailabilityStatusMessage{SourceID: "1"}.GenericMessage(ctx)
 	EnqueueAvailabilityStatusRequest(&msg)
 	EnqueueAvailabilityStatusRequest(&msg)
 	time.Sleep(100 * time.Millisecond)
@@ -60,12 +64,15 @@ func TestFullQueueSend(t *testing.T) {
 
 func TestQueueCancelSend(t *testing.T) {
 	ctx := context.Background()
+	ctx = identity.WithIdentity(t, ctx)
 	_ = kafka.InitializeStubBroker(16)
 
 	// enqueue message to be sent first
-	msg, _ := kafka.AvailabilityStatusMessage{SourceID: "1"}.GenericMessage()
+	msg, _ := kafka.AvailabilityStatusMessage{SourceID: "1"}.GenericMessage(ctx)
 	EnqueueAvailabilityStatusRequest(&msg)
-
+	// set the receiving message function up
+	wg := sync.WaitGroup{}
+	wg.Add(2)
 	// start sending messages
 	senderCtx, senderCancel := context.WithCancel(ctx)
 	go sendAvailabilityRequestMessages(senderCtx, 2, 5*time.Second)
@@ -73,9 +80,6 @@ func TestQueueCancelSend(t *testing.T) {
 	// allow the other goroutine to put the message into the buffer
 	runtime.Gosched()
 
-	// set the receiving message function up
-	wg := sync.WaitGroup{}
-	wg.Add(1)
 	consumeCtx, consumeCancel := context.WithCancel(ctx)
 	defer consumeCancel()
 	go kafka.Consume(consumeCtx, kafka.AvailabilityStatusRequestTopic, func(ctx context.Context, msg *kafka.GenericMessage) {
@@ -88,6 +92,8 @@ func TestQueueCancelSend(t *testing.T) {
 	time.Sleep(20 * time.Millisecond)
 
 	// cancel the sender before the 5 seconds timeout (so cancel branch is triggered)
+	msg, _ = kafka.AvailabilityStatusMessage{SourceID: "1"}.GenericMessage(ctx)
+	EnqueueAvailabilityStatusRequest(&msg)
 	senderCancel()
 
 	// wait until the message is consumed
--- a/internal/ctxval/context_values.go
+++ b/internal/ctxval/context_values.go
@@ -2,7 +2,10 @@ package ctxval
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"errors"
+	"fmt"
 
 	ucontext "github.com/Unleash/unleash-client-go/v3/context"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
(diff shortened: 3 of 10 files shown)


cmd/pbstatuser/main.go
```diff
@@ -19,21 +29,101 @@ import (
...
+
+// Beware there is no database connection, no job queue, most of what api/worker processes have
+// is not available because it should not be needed. The worker will simply use SDKs to check
+// statuses and send out results via kafka.Send function to Sources.
+// Also make sure that all goroutines are closed correctly on context cancel. A WaitGroup
+// would be a really good here: https://gist.github.com/fracasula/b579d52daf15426e58aa133d0340ccb0
+
+var (
+	chProvider = make(chan clients.StatusResponse)
+	chBatch    = make(chan clients.SourcesKafkaPayload)
```
lzap, line 47: These structures are only used in this file, do not export them.

cmd/pbstatuser/main.go
```diff
@@ -19,21 +29,101 @@ import (
...
+
+	// Get source id
+	asm, _ := kafka.NewAvailabilityStatusMessage(message)
+	sourceId := asm.SourceID
+
+	// Get sources client
+	sourcesClient, err := clients.GetSourcesClient(ctx)
+	if err != nil {
+		return
+	}
```
lzap, line 62: This will create a new client for every single message received, but this is the correct approach. I actually found out that we can reuse transports for sources/ib so I will make a PR. You can keep it here, that is correct.

cmd/pbstatuser/main.go
```diff
@@ -19,21 +29,101 @@ import (
...
+
+	// Fetch authentication from Sources
+	authentication, err := sourcesClient.GetAuthentication(ctx, sourceId)
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
+			return
+		}
+		return
+	}
```
lzap, line 71: Never swallow errors like that, please these both should be warning logs. You can use ctxval to get logger from context, it will already contain some interesting info.

cmd/pbstatuser/main.go
```diff
@@ -19,21 +29,101 @@ import (
...
+		go azure.CheckSourceAvailability(ctx, authentication, chProvider)
+	// TODO: Complete
+	case models.ProviderTypeGCP:
+		go gcp.CheckSourceAvailability(ctx, authentication, chProvider)
+	case models.ProviderTypeNoop:
+	case models.ProviderTypeUnknown:
+		// TODO: Complete
+	}
+	go createBatchKafkaSourcesPayload(ctx)
+}
```
lzap, line 87: This is not how a typical Go program should look like when dealing with multiple background workers (goroutines). This function is called for every received message, you do not want to spawn goroutines. You want to start them only once from the `main()` function.

Then use package-level channels to communicate between them - pass structs from this function (Kafka receiver and authorization finder) to one of the three channels (aws, azure, gcp) where three functions (running a `range` loop as goroutines) will pick them up, work on them and pass results to final channel where another function will pick the results and send them.

cmd/pbstatuser/main.go
```diff
@@ -19,21 +29,101 @@ import (
...
+			ResourceType: "Application",
+			ResourceID:   *msg.SourceID,
+			Status:       msg.Status,
+			// Err:         ,
+		}
+		chBatch <- payload
+	}
+
+	go sendPayloadsToSources(ctx)
+}
```
lzap, line 103: This is a weird code where functions spawns other functions as goroutines which creates a hardly readable flow.

internal/clients/http/azure/azure_client.go
```diff
@@ -69,3 +69,10 @@ func (c *client) Status(ctx context.Context) error {
 	}
 	return nil
 }
+
+func CheckSourceAvailability(ctx context.Context, auth *clients.Authentication, c chan clients.StatusResponse) {
+	c <- clients.StatusResponse{
+		Status:   clients.StatusUnavailable,
+		SourceID: nil,
+	}
+}
```
lzap, line 78: Those channels you want to use to communicate goroutines should be only in the `main` package, you do not want the client API to be aware that they are called like this (from other goroutines).

lzap:
Here is a pseudo code which represents the diagram I added to the design document (untested but to get the idea):

```go
package scratches

import (
	"context"
	"sync"
)

type Pipeline struct {
	sourceId string
	auth string
	response string
	accountID string
	orgID string
}

const ChannelBuffer = 1024

var chAws = make(chan Pipeline, ChannelBuffer)
var chAzure = make(chan Pipeline, ChannelBuffer)
var chGcp = make(chan Pipeline, ChannelBuffer)
var chSend = make(chan Pipeline, ChannelBuffer)

var receiverWG = sync.WaitGroup{}
var processingWG = sync.WaitGroup{}
var senderWG = sync.WaitGroup{}

func receiveFromKafka(ctx context.Context, message string) {
	// suppose this is called from the Kafka API for each message

	// get authorization
	p := Pipeline{
		sourceId: "xxx",
		auth: "aws",
		accountID: "xxx",
		orgID: "xxx",
	}
	// if aws then
	chAws <- p
}

func processAws() {
	defer processingWG.Done()

	for p := range chAws {
		// check aws
		p.response = "available"
		chSend <- p
	}
}

func processAzure() {
	defer processingWG.Done()

	for p := range chAzure {
		// check Azure
		p.response = "available"
		chSend <- p
	}
}

func processGcp() {
	defer processingWG.Done()

	for p := range chGcp {
		// check gcp
		p.response = "available"
		chSend <- p
	}
}

func sendResults() {
	for p := range chSend {
		// send the data from p

		_ = p
		// construct message, add headers
		// kafka.Send(p)
	}
}

func main() {
	ctx := context.Background()
	ctx, cancel := context.WithCancel(ctx)

	// start receiver goroutine
	receiverWG.Add(1)
	go func() {
		defer receiverWG.Done()
		// suppose this is kafka.Consume(ctx, ...)
		receiveFromKafka(ctx, "a message")
	}()

	// start processing goroutines
	processingWG.Add(3)
	go processAws()
	go processAzure()
	go processGcp()

	// start sending goroutine
	senderWG.Add(1)
	go sendResults()

	// wait until term signal and then

	// stop kafka receiver (can take up to 10 seconds) and wait until it returns
	cancel()
	receiverWG.Wait()

	// close all processors and wait until it exits the range loop
	close(chAws)
	close(chAzure)
	close(chGcp)
	processingWG.Wait()

	// close the sending channel and wait until it exits the range loop
	close(chSend)
	senderWG.Wait()

	// the program can now exit without killing any goroutines still doing some work
}
```

adiabramovitch:
> Here is a pseudo code which represents the diagram I added to the design document (untested but to get the idea):
> 
> ```go
> package scratches
> 
> import (
> 	"context"
> 	"sync"
> )
> 
> type Pipeline struct {
> 	sourceId string
> 	auth string
> 	response string
> 	accountID string
> 	orgID string
> }
> 
> const ChannelBuffer = 1024
> 
> var chAws = make(chan Pipeline, ChannelBuffer)
> var chAzure = make(chan Pipeline, ChannelBuffer)
> var chGcp = make(chan Pipeline, ChannelBuffer)
> var chSend = make(chan Pipeline, ChannelBuffer)
> 
> var receiverWG = sync.WaitGroup{}
> var processingWG = sync.WaitGroup{}
> var senderWG = sync.WaitGroup{}
> 
> func receiveFromKafka(ctx context.Context, message string) {
> 	// suppose this is called from the Kafka API for each message
> 
> 	// get authorization
> 	p := Pipeline{
> 		sourceId: "xxx",
> 		auth: "aws",
> 		accountID: "xxx",
> 		orgID: "xxx",
> 	}
> 	// if aws then
> 	chAws <- p
> }
> 
> func processAws() {
> 	defer processingWG.Done()
> 
> 	for p := range chAws {
> 		// check aws
> 		p.response = "available"
> 		chSend <- p
> 	}
> }
> 
> func processAzure() {
> 	defer processingWG.Done()
> 
> 	for p := range chAzure {
> 		// check Azure
> 		p.response = "available"
> 		chSend <- p
> 	}
> }
> 
> func processGcp() {
> 	defer processingWG.Done()
> 
> 	for p := range chGcp {
> 		// check gcp
> 		p.response = "available"
> 		chSend <- p
> 	}
> }
> 
> func sendResults() {
> 	defer senderWG.Done()
> 	for p := range chSend {
> 		// send the data from p
> 
> 		_ = p
> 		// construct message, add headers
> 		// kafka.Send(p)
> 	}
> }
> 
> func main() {
> 	ctx := context.Background()
> 	ctx, cancel := context.WithCancel(ctx)
> 
> 	// start receiver goroutine
> 	receiverWG.Add(1)
> 	go func() {
> 		defer receiverWG.Done()
> 		// suppose this is kafka.Consume(ctx, ...)
> 		receiveFromKafka(ctx, "a message")
> 	}()
> 
> 	// start processing goroutines
> 	processingWG.Add(3)
> 	go processAws()
> 	go processAzure()
> 	go processGcp()
> 
> 	// start sending goroutine
> 	senderWG.Add(1)
> 	go sendResults()
> 
> 	// wait until term signal and then
> 
> 	// stop kafka receiver (can take up to 10 seconds) and wait until it returns
> 	cancel()
> 	receiverWG.Wait()
> 
> 	// close all processors and wait until it exits the range loop
> 	close(chAws)
> 	close(chAzure)
> 	close(chGcp)
> 	processingWG.Wait()
> 
> 	// close the sending channel and wait until it exits the range loop
> 	close(chSend)
> 	senderWG.Wait()
> 
> 	// the program can now exit without killing any goroutines still doing some work
> }
> ```

I understand what I missed. update some fixes.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
+// 3) Reads from Azure channel, performs availability check, has configurable throttling.
+// 4) Reads from GCP channel, performs availability check, has configurable throttling.
+
+// Beware there is no database connection, no job queue, most of what api/worker processes have
+// is not available because it should not be needed. The worker will simply use SDKs to check
+// statuses and send out results via kafka.Send function to Sources.
+// Also make sure that all goroutines are closed correctly on context cancel. A WaitGroup
+// would be a really good here: https://gist.github.com/fracasula/b579d52daf15426e58aa133d0340ccb0
+
+const ChannelBuffer = 1024
```
lzap, line 42: I suggest to either bring this really down to like 32 or 16 because if we slow things down, any larger queue could actually make shutting down operation to take a lot of time. Imagine if we slow down to operations per second, then to finish up 1k buffer it would take 200 seconds while 32 would be just 6 seconds. I think k8s will not allow too much time for shutdown.

Even unbuffered channel is a good option here.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
+type StatusType string
+
+const (
+	StatusUnavailable StatusType = "unavailable"
+	StatusAvaliable   StatusType = "available"
+)
+
+type Pipeline struct {
+	// Resource type of the source
+	ResourceType string `json:"resource_type"`
```
lzap, line 53: I don't see the need of passing ResourceType along goroutines, you can simply make a local variable in the send function I guess. Or I am not sure why you have it in the struct.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
+		chAzure <- Pipeline{
+			Authentication: authentication,
+			SourceID:       sourceId,
+		}
+	case models.ProviderTypeGCP:
+		chGcp <- Pipeline{
+			Authentication: authentication,
+			SourceID:       sourceId,
+		}
+	case models.ProviderTypeNoop:
```
lzap, line 122: Create the pipeline variable just once :-)

Also I just thrown away this term during our brainstorming, you can use anything you like, maybe a "job". Not sure if pipeline is the best term :-)

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
 
-	// Beware there is no database connection, no job queue, most of what api/worker processes have
-	// is not available because it should not be needed. The worker will simply use SDKs to check
-	// statuses and send out results via kafka.Send function to Sources.
-	// Also make sure that all goroutines are closed correctly on context cancel. A WaitGroup
-	// would be a really good here: https://gist.github.com/fracasula/b579d52daf15426e58aa133d0340ccb0
+	case models.ProviderTypeUnknown:
+		// TODO: Complete
+	}
```
lzap, line 126: Yeah probably just some warning in logs for now.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
+}
+
+func checkSourceAvailabilityAzure(ctx context.Context) {
+	defer processingWG.Done()
+
+	for p := range chAzure {
+		// TODO: check if source is avavliable
+		p.Status = StatusUnavailable
```
lzap, line 134: I think when error occurs we should always return Available not to block customers in case this code breaks.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
+		chSend <- p
+	}
+}
+
+func checkSourceAvailabilityAWS(ctx context.Context) {
+	defer processingWG.Done()
+
+	for p := range chAws {
+		_, err := clients.GetEC2Client(ctx, p.Authentication, "us-east-1")
```
lzap, line 143: I think if you pass empty region it will use the DefaultRegion configuration.

```suggestion
		_, err := clients.GetEC2Client(ctx, p.Authentication, "")
```

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
+		if err != nil {
+			if err != nil {
+				p.Status = StatusUnavailable
+				p.Err = err
+				chSend <- p
+			}
+		}
+		p.Status = StatusAvaliable
+		chSend <- p
+	}
```
lzap, line 174: Ditto: else and catch only the relevant error.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
+		msg := &kafka.GenericMessage{
+			Topic: kafka.AvailabilityStatusRequestTopic,
+			Value: []byte(values),
+			Headers: map[string]string{
+				"Content-Type":                "application/json",
+				"x-rh-identity":               identity.GetIdentityHeader(ctx),
+				"x-rh-sources-org-id":         id.Identity.OrgID,
+				"x-rh-sources-account-number": id.Identity.AccountNumber,
+			},
+		}
```
lzap, line 195: The headers creation needs to be moved into the `kafka` package. Now we need a "constructor", something like `NewGenericMessage` that will take context, topic and value and return a generic message with all the platform headers set correctly.

Then you want to use this in the sending code (in the api process) to send the headers as well. Right now they will be empty as nothing is really sent into our internal topic.

I have an extra ask, can you please use "X-RH-Sources-Account-Number" case? Headers (HTTP/Kafka) are not case sensitive, but it looks much nicer and it is the "defacto" standard convention for HTTP headers.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +26,179 @@ import (
...
+	for p := range chAws {
+		_, err := clients.GetEC2Client(ctx, p.Authentication, "us-east-1")
+		if err != nil {
+			p.Status = StatusUnavailable
+			p.Err = err
+			chSend <- p
+		}
+		p.Status = StatusAvaliable
+		chSend <- p
+	}
```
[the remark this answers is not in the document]
adiabramovitch: Maybe I should add return at the end? 
```suggestion
		for p := range chGcp {
		gcpClient, err := clients.GetGCPClient(ctx, p.Authentication)
		if err != nil {
			p.Status = StatusUnavailable
			p.Err = err
			chSend <- p
			return
		}
		_, err = gcpClient.ListAllRegions(ctx)
		if err != nil {
				p.Status = StatusUnavailable
				p.Err = err
				chSend <- p
				return
		}
		p.Status = StatusAvaliable
		chSend <- p
	}
		
```

cmd/pbstatuser/main.go
adiabramovitch: At first, I wanted to create two structs: 
one for passing information between the go routines and the second as the payload to send over to sources according to their needs: https://consoledot.pages.redhat.com/docs/dev/services/sources/integration.html#_prepare_an_endpoint_for_the_availability_checking
```
{
  "resource_type": "Application",
  "resource_id": "12345",
  "status": "unavailable",
  "error": "something went wrong with..."
}
```
Do you think I should separate the pipeline struct into two?

cmd/pbstatuser/main.go
adiabramovitch: I don't understand why do you think an unbuffered channel is better

go.mod
```diff
@@ -143,9 +143,9 @@ require (
...
 	go.opentelemetry.io/otel/metric v0.33.0 // indirect
-	golang.org/x/net v0.1.0 // indirect
-	golang.org/x/oauth2 v0.1.0 // indirect
-	golang.org/x/sys v0.1.0 // indirect
+	golang.org/x/net v0.2.0 // indirect
+	golang.org/x/oauth2 v0.2.0 // indirect
+	golang.org/x/sys v0.2.0 // indirect
 	golang.org/x/text v0.4.0 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/grpc v1.50.1 // indirect
```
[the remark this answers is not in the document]
adiabramovitch: Should I remove go.sum as well?

cmd/pbstatuser/main.go
lzap: Because the processing will likely be delayed artificially therefore the receiver/sources requester goroutine will likely be always faster, thus there will never be lack of messages in the channel. The checker goroutines will not have to wait too much. As I said, small buffer will do, but 32 is probably the highest number I would go for.

cmd/pbstatuser/main.go
lzap: Yeah I think these are two definitely different structures, at the end of the pipeline you just need to copy few values from one to other, but that is a small tax to pay for better readability. In this version by looking at the type, I am unable to tell what its main purpose.

cmd/pbstatuser/main.go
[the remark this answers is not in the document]
lzap: Well `return` will close the goroutine, mark the work group as done and end up all processing. I don't think that is what you want to do.

go.mod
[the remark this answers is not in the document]
lzap: Yeah just do `git reset -- file1 file2` that is the easiest thing to do.

cmd/pbstatuser/main.go
adiabramovitch: Do you mean to add the headers here as well https://github.com/RHEnVision/provisioning-backend/blob/75bc8471cff0ffbb30908ad0bfec125fba3216e6/internal/kafka/avail_status_msg.go#L31? or to use the new function I created here internal/kafka/kafka.go

lzap: Yes, headers need to be handled properly:

* sending it from the `services.AvailabilityStatus` (currently this is completely ignored and no headers are set)
* passing headers from/to GenericMessage to KafkaMessage
* sending it from the statuser send function

The type we use for ASM is currently this structure:

```
type AvailabilityStatusMessage struct {
	SourceID string `json:"source_id"`
}
```

It has no headers obviously, I suggest to refactor `GenericMessage(identity identity.XRHID)` function to get a context and construct headers from our own context (we should carry the whole identity there - see `ctxval.Identity()`).

The other direction is more tricky, one idea I have is to change the `NewAvailabilityStatusMessage` function to return three results: ASM, identity.XRHID and error, so you can then use the identity when constructing the result message back to sources (with the function above `GenericMessage(identity)`.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +30,173 @@ import (
...
+		return
+	}
+
+	s := SourceInfo{
+		Authentication: authentication,
+		SourceID:       sourceId,
+	}
+
+	// Send the identity to the sendResult channel
+	chHeader <- ctxval.Identity(ctx)
```
ezr-ondrej, line 103: I fail to see how do we synchronize this channel with the messages from the three goroutines.
From what I'm reading there is no certainty that the messages won't get missaligned and Identity for message1 will get assigned to response for message2 and vice versa.
Why can't we pass the identity in SourceInfo and then to SourceResult? It would greatly simplify the synchronization of that info with the given message.

I might be wrong tho, do we have some guidance on what this auth is supposed to achieve or on how it behaves for sources?

ezr-ondrej (commented):
I think the synchronization of one flow through mutliple channels might bite us in the butt.
I'd love to avoid it if possible.
If impossible pls make sure to update the arch of this once we have final version.

cmd/pbstatuser/main.go
adiabramovitch: 1. It is possible to pass the identity through sourceInfo and SourceResult but it will change the clean implementation of the generic message. 
Because in one case we are reading the identity from the context and in the latter, we will read the identity from the sourceResult struct. I think it is a bit repetitive to have two functions with the same logic but doable. 
2. I think and @lzap correct me if I am wrong, we can achieve this synchronization using wait groups, waiting until the consume go routine will end, but I am not sure it is a good idea (I believe we are losing parallelism). 
3. What do you mean by auth? the identity header? if so it is needed to authenticate to the sources service. 
I am not sure if there is another usage of it (I'll check that)

@lzap What do you think?

adiabramovitch: 1. It is possible to pass the identity through sourceInfo and SourceResult but it will change the clean implementation of the generic message.
Because in one case we are reading the identity from the context and in the latter, we will read the identity from the sourceResult struct. I think it is a bit repetitive to have two functions with the same logic but doable.

2. I think and @lzap correct me if I am wrong, we can achieve this synchronization using wait groups, waiting until the consume go routine will end, but I am not sure it is a good idea (I believe we are losing parallelism).

3. What do you mean by auth? the identity header? if so it is needed to authenticate to the sources service.
I am not sure if there is another usage of it (I'll check that)

@lzap What do you think?

adiabramovitch: Solved, thanks!

internal/kafka/kafka.go
```diff
@@ -171,11 +172,24 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, handler func(ct
...
+			ctx, err = ctxval.WithIdentityFrom64(ctx, returnHeaderValue(ctx, msg.Headers))
+			if err != nil {
+				logger.Trace().Msgf("Could not add identity to context: %s", err)
+			}
 			handler(ctx, NewMessageFromKafka(&msg))
 		}
 	}
 }
 
+func returnHeaderValue(ctx context.Context, headers []protocol.Header) string {
```
ezr-ondrej, line 184: naming is weird here, I'd love to name it `getIdentityHeaderValue` or something along those lines :)

ezr-ondrej (approved):
Nice, really great progress here. I think it might be even ready, I've got one naming nit and then few improvements and refactorings:
1. I'd love to get the code from `main.go` in a package and just call some `InitStatuser` that would initialize this.
2. I'm missing tests - which is hard to do for code that is in main file, so 1. would need to be addressed first I guess.
3. `WithIdentity*` could be in future contributed to https://github.com/RedHatInsights/platform-go-middlewares/blob/b89d6d455a93828359654fe68b3dddb83f03536e/identity/identity.go so others use the same code to deal with kafka auth headers (if there is consensus of using kafka headers for auth).
4. The authentication handling would be nice to document in the ADR or in the architecture documentation.

None of these are blockers for me, 1) and 2) would need a ticket so we don't completely forget about them. But with a ticket open and follow-up planned I'm happy to merge this.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +30,162 @@ import (
...
 
+const ChannelBuffer = 32
+
+type SourceInfo struct {
+	Authentication clients.Authentication `json:"-"`
+
+	SourceID string `json:"resource_id"`
+
+	Identity identity.XRHID `json:"-"`
+}
```
lzap, line 41: I believe this type never gets serialized to JSON, can we drop the tags from here?

cmd/pbstatuser/main.go
```diff
@@ -20,25 +30,162 @@ import (
...
+	receiverWG   = sync.WaitGroup{}
+	processingWG = sync.WaitGroup{}
+	senderWG     = sync.WaitGroup{}
+)
+
 func processMessage(ctx context.Context, message *kafka.GenericMessage) {
 	logger := ctxval.Logger(ctx)
+
 	// This is just a test to see if Kafka is configured properly in ephemeral
 	logger.Warn().Msg("RECEIVED KAFKA MESSAGE")
```
lzap, line 57: Please drop this, also there was some big comment block by me which can go as well. I meant this as a temporary thing.

cmd/pbstatuser/main.go
```diff
@@ -20,25 +30,162 @@ import (
...
 
-	// TODO: implement the source checking for AWS, Azure, GCP
+	// Get source id
+	asm, err := kafka.NewAvailabilityStatusMessage(message)
+	if err != nil {
+		logger.Warn().Msgf("could not get availability status message %s", err)
```
lzap, line 62: We do log messages with capitals, errors with lower case.

```suggestion
		logger.Warn().Msgf("Could not get availability status message %s", err)
```

There's a lot more down below. It's like 90% of the logs in this format, let's try to keep that number high

cmd/pbstatuser/main.go
```diff
@@ -20,25 +30,162 @@ import (
...
+			logger.Warn().Msgf("could not generate generic message %s", err)
+			continue
+		}
+		messages = append(messages, &msg)
+	}
+	err := kafka.Send(ctx, messages...)
+	if err != nil {
+		logger.Warn().Msgf("could not send message through kafka %s", err)
+	}
 }
```
lzap, line 189: This function is not right. This will panic when you receive 101th message from the channel. There are two options how to implement this:

1. Create and send messages on the fly as they are received from the channel, this is less performant solution, however, good for the first version.
2. Do whatever we do in the sending endpoint - a buffer with sending messages either when a buffer is full or every 2 seconds whatever comes first.

internal/background/availability_queue_test.go
```diff
@@ -60,22 +64,22 @@ func TestFullQueueSend(t *testing.T) {
...
 	go sendAvailabilityRequestMessages(senderCtx, 2, 5*time.Second)
 
 	// allow the other goroutine to put the message into the buffer
 	runtime.Gosched()
 
-	// set the receiving message function up
-	wg := sync.WaitGroup{}
-	wg.Add(1)
 	consumeCtx, consumeCancel := context.WithCancel(ctx)
 	defer consumeCancel()
```
lzap, line 84: I do not understand this hunk, can you comment what was the intention? The way I read this is that this test should get stuck waiting forever, you add 2 to the waitgroup but Done is called just once so Wait will block forever.

internal/ctxval/context_values.go
```diff
@@ -35,6 +43,25 @@ func Logger(ctx context.Context) *zerolog.Logger {
...
 }
 
+func WithIdentity(ctx context.Context, id identity.XRHID) context.Context {
+	return context.WithValue(ctx, identity.Key, id)
+}
+
+func WithIdentityFrom64(ctx context.Context, id string) (context.Context, error) {
+	idRaw, err := base64.StdEncoding.DecodeString(id)
+	if err != nil {
+		return nil, DecodeErr
```
lzap, line 53: Root cause is swallowed, please use `fmt.Errorf`.

internal/ctxval/context_values.go
```diff
@@ -35,6 +43,25 @@ func Logger(ctx context.Context) *zerolog.Logger {
...
+	}
+
+	var jsonData identity.XRHID
+	err = json.Unmarshal(idRaw, &jsonData)
+	if err != nil {
+		return nil, MissingValidJSONErr
```
lzap, line 59: Root cause is swallowed, please use `fmt.Errorf`.

internal/ctxval/context_values.go
```diff
@@ -2,14 +2,22 @@ package ctxval
...
 
 	ucontext "github.com/Unleash/unleash-client-go/v3/context"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 )
 
+var (
+	DecodeErr           = fmt.Errorf("unable to b64 decode x-rh-identity header")
+	MissingValidJSONErr = fmt.Errorf("x-rh-identity header does not contain valid JSON")
```
lzap, line 18: Use `errors.New` here this is a dynamic error assignment, our linter should complain.

internal/kafka/avail_status_msg.go
```diff
@@ -19,15 +23,27 @@ func NewAvailabilityStatusMessage(msg *GenericMessage) (*AvailabilityStatusMessa
...
-		Key:   []byte(m.SourceID),
+		Topic: topic,
+		Key:   []byte(key),
 		Value: payload,
+		Headers: map[string]string{
+			"Content-Type":                "application/json",
+			"x-rh-identity":               identity.GetIdentityHeader(ctx),
+			"x-rh-sources-org-id":         id.Identity.OrgID,
+			"X-RH-Sources-Account-Number": id.Identity.AccountNumber,
+		},
```
lzap, line 47: Okay  this is good, but God I hate mixed style. Can you sync up on either all lowercase or "Camel-Case-Style"? :-D

internal/kafka/kafka.go
```diff
@@ -171,11 +172,24 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, handler func(ct
 			logger.Warn().Err(err).Msgf("Error when reading message: %s", err.Error())
 		} else {
 			logger.Trace().Bytes("payload", msg.Value).Msgf("Received message with key: %s", msg.Key)
+			ctx, err = ctxval.WithIdentityFrom64(ctx, returnHeaderValue(ctx, msg.Headers))
+			if err != nil {
+				logger.Trace().Msgf("Could not add identity to context: %s", err)
```
lzap, line 177: This message does not tell you much, imagine you are reading logs because investigating an issue and read the message. Identity where? Is this HTTP context? Kafka context? I suggest:

```suggestion
				logger.Trace().Msgf("Could extract identity from context to Kafka message: %s", err)
```

internal/kafka/kafka.go
```diff
@@ -171,11 +172,24 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, handler func(ct
...
+			}
 			handler(ctx, NewMessageFromKafka(&msg))
 		}
 	}
 }
 
+func returnHeaderValue(ctx context.Context, headers []protocol.Header) string {
+	for _, h := range headers {
+		if h.Key == "x-rh-identity" {
```
lzap, line 186: Headers should be case insensitive, please do:

```suggestion
		if strings.EqualFold(h.Key, "X-RH-Identity") {
```

Or whatever style for HTTP/Kafka headers you prefer.

lzap: One comment - there is an inconsistency between Kafka native message headers (slice of struct) and our Generic message headers (map). I suggest to change our API to reflect Kafka's because very likely for smaller amount of headers, a list (slice, linked list) is more performant than hash table.

Created: https://issues.redhat.com/browse/HMSPROV-376

internal/kafka/kafka.go
lzap: Absolutely, this should probably be an exported function so others can use this. Looks like this is a pattern, I suggest something like `LookupKafkaHeader`. We will likely need something similar for our generic message that would be `LookupGenericHeader`.

lzap (changes requested):
Nice, there is a problem in the sending function tho.

internal/kafka/kafka.go
adiabramovitch: Should we do that as a follow-up?

internal/background/availability_queue_test.go
adiabramovitch: I got an error of negative counter for the wait group. 
After looking at other tests I saw that we are missing one because we have a goroutine for the **Consume** function. 
when I changed the wait group to two the error was solved.

internal/kafka/kafka.go
adiabramovitch: @ezr-ondrej Regarding 1, I don't have a problem doing it but is it a bit inconsistent, because we do not have a function initAPI for the pbapi. 
Maybe we should add a ticket to unify that?

internal/kafka/kafka.go
ezr-ondrej: Yeah, the second comment is meant as a follow-up I believe :)

internal/kafka/kafka.go
ezr-ondrej: My main point on 1 was: lets move as much code as we can into a package.
Maybe you're right, there is some leftover code in pbapi that could be moved, but if we get to the level of pbapi, and extract functions from main into package I'm good.
Tho pbapi initializes pretty standard stack, that if doesn't work, we will notice right away and it would mean some major library (like `chi`) is broken, here we are initializing custom stuff, so might be a good idea to test that and thus extract the init code to a package so it's testable as well.
But as I said, if we extract only the functions, and have tests for them I'm good ^_^

ezr-ondrej (changes requested):
Oh commit message does not compy with the new commit message style.
Mainly pls link to the Jira issue for such a big piece of code ^_^

internal/kafka/kafka.go
lzap: Absolutely, let's not make it more complex, I actually created a ticket for that. This is flaw in my design of GenericMessage, I can do it as well if you don't want to. Let me know.

ezr-ondrej (approved):
I've rebased on top of main to fix the commit message check.
I've not checked the sendResult, but looks sane to me, but I'd love @lzap to take a look as well :)

lzap (approved):
I am good, this looks great.

akhil-jha:
> The CI error is some kind of connection issue for AWS test, not related to this PR which does not touch existing code and we do not have test coverage for sources availability yet. Merging, thank you!

Is there any steps I could follow(or endpoint?) so I could add test in the iqe plugin?

lzap:
Here you have it:

https://github.com/RHEnVision/provisioning-backend/pull/333

PR #303, opened 2022-11-16
Multiple Kafka concurrency fixes

lzap:
When I was working with Adi on Kafka producer, I noticed few bugs in my code.

FIrst, the consumer did not break on context cancel correctly which has no effect any other than if you use some method to wait until goroutine is finished e.g.:

```go
	ctx := context.Background()
	ctx, cancel := context.WithCancel(ctx)

	receiverWG.Add(1)
	go func() {
		defer receiverWG.Done()
		// suppose this is kafka.Consume(ctx, ...)
		receiveFromKafka(ctx, "a message")
	}()

	// wait until term signal and then

	// stop kafka receiver (can take up to 10 seconds) and wait until it returns
	cancel()
	receiverWG.Wait()
```

The `Wait` call would never end. The first commit fixes it.

Second, we use a custom HTTP client doer for Sources and IB, however, it creates a new transport for each and every client. This is wasting of resources since http.Transport can and should be shared according to Go documentation. This allows reusing of connections or connection caching/pooling which is built in. The second patch fixes this, this should lead to better performance for sources/ib clients.

Finally, similar issue to the first one we have in the availability queuer where look is not broken on the context cancel leading to active goroutine when shutting down gracefully. Again, no practical consequences other than program exiting with a goroutine still looping (waiting) in the background, but it's good to fix it too. That's the third commit.

### DIFF
--- a/internal/background/availability_queue.go
+++ b/internal/background/availability_queue.go
@@ -58,8 +58,9 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
 			if length > 0 {
 				logger.Trace().Int("messages", length).Msgf("Sent %d availability request messages (cancel)", length)
 				send(ctx, messageBuffer...)
-				messageBuffer = messageBuffer[:0]
 			}
+
+			return
 		}
 	}
 }
--- a/internal/clients/http/platform_client.go
+++ b/internal/clients/http/platform_client.go
@@ -9,13 +9,14 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
+// Shared HTTP transport for all platform clients to utilize connection caching
+var transport = &http.Transport{}
+
 // NewPlatformClient returns new HTTP client (doer) with W3C Trace Context, logging tracing
 // and/or HTTP proxy (non-clowder environment only) according to application configuration.
 // Use this function to create HTTP clients for communication with all platform services.
 func NewPlatformClient(ctx context.Context, proxy string) HttpRequestDoer {
-	var rt http.RoundTripper
-	transport := &http.Transport{}
-	rt = transport
+	var rt http.RoundTripper = transport
 
 	if proxy != "" {
 		if config.InClowder() {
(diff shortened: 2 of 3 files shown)


internal/background/availability_queue.go
```diff
@@ -60,6 +60,8 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
 				send(ctx, messageBuffer...)
 				messageBuffer = messageBuffer[:0]
 			}
+
+			break
```
ezr-ondrej, line 64: sorry it took long, but I struggled to find time to verify a hunch that the `break` might be breaking the select and it appears it does: https://stackoverflow.com/questions/25469682/break-out-of-select-loop

So I guess we want to `return` here?
```suggestion
			return
```

adiabramovitch: Don't we want a `default` option as well? 
Correct me if I am wrong but if we get a signal to terminate won't we want to break the select block as well?

ezr-ondrej: return should break both `select` and `for` :)
default would cause the select not wait for the signals and would go to the default right away I belive :)

lzap: Oh nice find, I haven't actually tested this. Shame on me.

lzap:
Pushed.

internal/background/availability_queue.go
```diff
@@ -60,6 +60,8 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
 				send(ctx, messageBuffer...)
 				messageBuffer = messageBuffer[:0]
```
ezr-ondrej, line 61: Ah, right .. no point of cleaning the buffer when you return right after :) well that's what linter says

ezr-ondrej: @lzap can you pls remove this line so we can merge this? :)

lzap: Hmm weird check, but okay. I can live with that although my guts tell me the buffer should really be cleared... :-)

lzap:
Rebased.

ezr-ondrej (approved):
Thanks @lzap !

PR #309, opened 2022-11-21
test: add unit test for Authentication failure

ezr-ondrej:
Adds test that uses [`httptest.Server`](https://pkg.go.dev/net/http/httptest#Server) to mock the outbound connection and stub its response.
I believe this is the best way to test the clients as it tests the client implementation together with the generated client and stubs on the HTTP layer.

### DIFF
--- a/internal/clients/http/sources/sources_client.go
+++ b/internal/clients/http/sources/sources_client.go
@@ -31,7 +31,13 @@ func logger(ctx context.Context) zerolog.Logger {
 }
 
 func newSourcesClient(ctx context.Context) (clients.Sources, error) {
-	c, err := NewClientWithResponses(config.Sources.URL, func(c *Client) error {
+	return NewSourcesClientWithUrl(ctx, config.Sources.URL)
+}
+
+// NewSourcesClientWithUrl allows customization of the URL for the underlying client.
+// It is meant for testing only, for production please use clients.GetSourcesClient.
+func NewSourcesClientWithUrl(ctx context.Context, url string) (clients.Sources, error) {
+	c, err := NewClientWithResponses(url, func(c *Client) error {
 		c.Client = http.NewPlatformClient(ctx, config.Sources.Proxy.URL)
 		return nil
 	})
--- a/internal/clients/http/sources/sources_client_test.go
+++ b/internal/clients/http/sources/sources_client_test.go
@@ -1,0 +1,32 @@
+package sources_test
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients/http/sources"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSourcesClient_GetAuthentication(t *testing.T) {
+	t.Run("source with missing Provisioning auth", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


internal/clients/interface.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // GetSourcesClient returns Sources interface implementation. There are currently
 // two implementations available: HTTP and stub
-var GetSourcesClient func(ctx context.Context) (Sources, error)
+var GetSourcesClient func(ctx context.Context, url string) (Sources, error)
```
lzap, line 12: This looks like an undocumented change. What is the purpose of this? I don't believe this belongs in here, you are forcing users of this interface to keep passing URLs around just because of a test.

I think it is worth looking into a way how to configure this via app configuration, problem is this is a global thing and tests can possibly run in parallel in the future.

lzap (changes requested):
This new test is IMHO not a unit test, a unit test would only test the function that was subject for the bug. This is what I call an integration test because it tests the whole HTTP stack, router, service, HTTP client. I think the best approach would be to mark all these tests with "integration" Go build flag and setup the app in a special way for integration tests (that all HTTP clients are actually pointing to localhost). We already sort of do this for DAO, I think it make sense to add HTTP stack tests into the mix, what you think?

lzap:
We already have this, see:

* https://github.com/RHEnVision/provisioning-backend/blob/main/internal/config/override_integration.go
* https://github.com/RHEnVision/provisioning-backend/blob/main/internal/config/override_test.go

I believe the test you wrote actually is integration, therefore I would add the required configuration changes to `override_integration.go`.

ezr-ondrej:
So I've come up with a different approach and that is context overriding for config values.
I believe it checks all the boxes - it allows us to keep the global value intact, but override per test without impacting any other part of the code. The question is whether we want to allow this for only some values, because it has a potential to blow out of proportions and I'd not like to keep a copy of config in the context.

ezr-ondrej:
To the point of "not being unit test" I disagree tho, could you pls explain again why you think this is not a unit test?
I know that in strict sense we should stub even the real sources client, but apart of that nothing else is involved in this test.

This starts completely mocked httptest "server" that intercepts all communication to url it generates so there is not even a real http connection. I believe this is as isolated as it can. We test the function and the generated client that supports it, but not anything more.

lzap:
Let's not start this again, just add this to unit test environment file, I really don't care. What matters to me is to avoid adding unnecessary parameters and we are good.

lzap:
_Updated the above comment to give more context._

ezr-ondrej:
Well the issue is that this is idiomatic way to test http requests in go and I'd not like to divert into some custom stubbing just because we want to be special. You've thought me that! My approach would be exactly what you're saying - using special library that'd stub special urls and I'd set this url in config. But that's not the way of go  apparently.

I like the last idea quite a lot, it's not great, but fixes the issue of passing the url around. I thought of setting this in test to the config, but it's messing with concurrency issues.

lzap:
A getter with URL that will be not the only getter is fine for me. You can even export it, just document it's purpose and make a link to the non-URL version in case an engineer is confused.

ezr-ondrej:
Added new getter, I tried to play with some stubbing, but I did not find any satisfactory solution. This is simple enough imho and with the documentation it is clear what the purpose is.

lzap (approved):
Perfect, not breaking the contract.

PR #310, opened 2022-11-22
fix(kafka): CA config from string

adiabramovitch:


ezr-ondrej (approved):
Looks good to me :+1:

lzap:
Can you start giving descriptions for PRs which are not that obvious? :-)

It looks okay, but I would like to know more. Where did you find it should be string? Because clowder is capable of deploying files before it starts the application, in fact, that is exactly what it does for Kafka - it deploys a JSON configuration file, so technically this could be file.

Also, I copy-pasted this from an existing platform application too.

lzap:
For the record, it was an error: `can't open file ---Begin cert` 

Nice find indeed!

PR #123, opened 2022-07-28
Add GH action run for DAO tests

ezr-ondrej:
@avitova does it look ok to you? :)

PR #312, opened 2022-11-22
build(changelog): Add support for Jira issues

ezr-ondrej:
Add support for Jira issues in changelog.

This will also enforce commits with `feat` or `fix` type to be aligned to a Jira issue.
The format would be one of:
```
feat(HMSPROV-XXX): short description

longer description
```
or
```
fix: short description

longer description

Fixes: HMSPROV-XXX
```
More details is shared in the patch to CONTRIBUTE.md

This is hardcoded in custom python script, which uses git-changelog python library,
but given the simplicity it seems fine for now.

:warning: this disables parsing of github issues, but it seems fine as we are not using them. We might want to disable the feature and point our potential reporting users to the Jira.


lzap:
Generally I agree, could you incorporate everything into a single PR: https://github.com/RHEnVision/provisioning-backend/pull/313

So we can discuss this all within the team what others think.

My only comment is that HMSPROV makes it too long, I vote for something shorter like `#123` or even just `J123` because `#` sometimes confuses github as it thinks its a GH issue.

ezr-ondrej:
I see the issue of this being mouthful, but there is an effort underway of unifying all HMS projects to one Jira and thus in the future it might be `HMS-XXX`.
I'd like to keep the prefix, because user feedback might come to a different project and thus we might need to link a different project later on. I'm usually copying the issue identifier so it is not problem for me, but if others see it as a big deal, we can make JXXX mean main jira project (HMSPROV now and HMS in the future) and other jira projects would just use full Jira identifier.

.github/workflows/gotest.yml
```diff
@@ -13,6 +13,7 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 2
+      - run: pip install git-changelog
       - run: |
           make check-commits
```
lzap, line 18: Nitpick: to speed this up you could utilize native GHA action which will cache the Python modules for our job to make it super fast.

CONTRIBUTING.md
```diff
@@ -5,7 +5,7 @@
 Make sure your commit message follows this subject style:
 
         type: brief summary up to 70 characters or
-        type(scope): brief summary up to 70 characters
+        type(jiraIssue): brief summary up to 70 characters
```
lzap, line 8:
```suggestion
        type(HMSPROV-1234): brief summary up to 70 characters
```

looks better for an example.

CONTRIBUTING.md
```diff
@@ -20,17 +20,11 @@ Type must be one of the following:
...
-* clients
-* oapi
-* readme
-* changelog
-* scripts
-
-For WIP branches, start commit message with `WIP` and it will pass on CI.
+For **feat** and **fix** types a Jira issue link is required.
+Please use the scope like **feat(HMSPROV-XXX): subject**
+or put the issue reference in commit body as `Fixes: HMSPROV-XXX` or `Refs: HMSPROV-XXX`
```
lzap, line 27: So you want to use full HMSPROV- prefix? That's way to long, but I won't block this if others like it.

Remove the info about commit body, I believe this is wrong, the github log Python library is not cabable scanning anything other than the first line, is it?

lzap:
I am okay, want to hear opinion from others. I don't like the very long `HMSPROV-1234` and I see the point that for consistency, it's good, but if we ever change ticketing system, it will be done at some point in time and we can easily modify the changelog generator to use old instance up to some date in git history. And this looks so much better:

`fix(1234): very nice commit message`

lzap:
FYI I know for fact that this is possible because I was digging in its codebase:

https://github.com/pawamoy/git-changelog/blob/master/src/git_changelog/commit.py#L22

ezr-ondrej:
> There could be a transition period, but nothing we could not see from the numbers (high vs low number). But I do not insist.

Sure, I agree it can be done and I'm not strictly against it, it would just add more complexity to a support script and I'd like to keep it as simple as possible.
But I agree that the shorter format looks better (just can't be copy-pasted from Jira) so happy to implement that if we agree :)

Happy to hear what others think :)

CONTRIBUTING.md
ezr-ondrej: It is and it even works with the current template we are using it works for GH issues the same way and I kinda like it, you want to link a ticket, but in the subject it doesn't tell you much usually, so having it in subject is not best IMHO.

See the changes in CHANGELOG.md in this patch and the relevant commits, most of them have the Jira issue linked in body, not subject.

scripts/changelog.py
```diff
@@ -0,0 +1,77 @@
...
+  if len(commit.subject) > length_limit:
+    sys.stderr.write("ERROR: Commit message length too long (limit is " + 70 + "): " + commit.subject)
+    return 1
+
+  if commit.style["type"] == "":
+    sys.stderr.write("ERROR: Commit message must have a type 'type: subject': " + commit.subject)
+    return 2
+
+  if commit.style["type"] in ["Features", "Bug Fixes"]:
+    if commit.style["scope"] == "" or JIRA_ISSUE_SYMBOL not in commit.style["scope"]:
```
adiabramovitch, line 61: I saw that you had changed the example in the CONTRIBUTION.MD :
https://github.com/RHEnVision/provisioning-backend/pull/312/files#r1031449366
Should "scope" be changed here as well?

adiabramovitch:
LGTM
regarding the name, I don't mind keeping it long but lzap's suggestion work for me as well :)

scripts/changelog.py
ezr-ondrej: `scope` here is comming from the underlying library responsible for parsing the commit into metadata, so we can't easily change it here :)

adiabramovitch (approved):

lzap (approved):
I am good, thanks.

PR #314, opened 2022-11-22
fix: scope existing pubkey search by source id

ezr-ondrej:
For AWS the key needs to be available in a region we Launch in.
This fixes the search so we verify pubkey for given account and region.

Fixes: HMSPROV-366

### DIFF
--- a/internal/dao/dao_interfaces.go
+++ b/internal/dao/dao_interfaces.go
@@ -40,7 +40,7 @@ type PubkeyDao interface {
 	Delete(ctx context.Context, id int64) error
 
 	UnscopedCreateResource(ctx context.Context, pkr *models.PubkeyResource) error
-	UnscopedGetResourceByProviderType(ctx context.Context, pubkeyId int64, provider models.ProviderType) (*models.PubkeyResource, error)
+	UnscopedGetResourceBySourceAndRegion(ctx context.Context, pubkeyId int64, sourceId string, region string) (*models.PubkeyResource, error)
 	UnscopedListResourcesByPubkeyId(ctx context.Context, pkId int64) ([]*models.PubkeyResource, error)
 	UnscopedDeleteResource(ctx context.Context, id int64) error
 }
--- a/internal/dao/pgx/pubkey_pgx.go
+++ b/internal/dao/pgx/pubkey_pgx.go
@@ -129,11 +129,11 @@ func (x *pubkeyDao) UnscopedCreateResource(ctx context.Context, pkr *models.Pubk
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


ezr-ondrej:
Maybe we need to scope this by region as well @adiabramovitch ?

adiabramovitch:
> Maybe we need to scope this by region as well @adiabramovitch ?

Why did you remove the provider type? Why do you think we should add region as well?

ezr-ondrej:
> Why did you remove the provider type?

I believe Source is already scoped by provider so source_id is scoping by provider as well.

>  Why do you think we should add region as well?

I think that the keys are scoped to region so you need to import it to every region manually, so it needs to exist in given region if I'm not mistaken.

lzap:
> I think that the keys are scoped to region so you need to import it to every region manually, so it needs to exist in given region if I'm not mistaken.

On AWS, yes. That is the case. I think on GCP it's global.

lzap:
The change makes sense yeah.

adiabramovitch:
I am not sure about the pubkeys, maybe for AWS but for GCP it is not the case

lzap:
Let my dump my thinking here:

This is the current DDL of the table:

```sql
create table if not exists pubkey_resources
(
    id        bigint generated by default as identity
        primary key,
    tag       text default ''::text not null,
    pubkey_id bigint                not null
        references pubkeys
            on delete cascade,
    provider  integer               not null
        constraint pubkey_resources_provider_check
            check (valid_provider(provider)),
    source_id text                  not null,
    handle    text                  not null,
    region    text
);
```

* A user creates a AWS source, it is region idependent.
* The user launches new instance and creates a new key on us-east-1.
* The key is uploaded and resource_key is created with type/source id/region.
* The user launches on us-west-1 with the same key.
* The key resource does not exist, thus it is uploaded.

So I think this patch does not solve the problem, AWS credentials are global. I think that's the same case for all clouds. 
In all cases, we want to search the resource by:

* pubkey id (the associated key body)
* source_id or type (they are essentially the same thing)
* region

I think we can use either of the two with the same result. Or am I missing something?

ezr-ondrej:
> Search by region

Well for AWS yes, but I'm not sure if Azure would have the same. GCP will probably skip this step altogether.
But lets add it and figure out GCP/Azure once we are implementing those.

> I think we can use either of the two with the same result. Or am I missing something?

I believe we need to use source id as one customer can have two sources for one provider type.

lzap (approved):
Yeah, we can add additional DAO functions as we need them. This is good for now, looks green.

lzap:
Oh commit message, let's wait until we agree on final solution.

ezr-ondrej:
Heh, I've removed the scope so we don't hang PRs around for no reason.
This will still link the ticket, just not in subject.

PR #315, opened 2022-11-22
fix: ensure pubkey is always present on AWS

ezr-ondrej:
Ensure pubkey is on AWS and save its name to ASWDetails.

Fixes: HMSPROV-339 HMSPROV-332

Refactors the pubkey job to Ensure the pubkey exists on AWS.
First it tries to look for the key on AWS, if it exits, it gets its name.
If it doesn't exist, it imports the key with the name we have in our DB.

This name gets saved to AWS details so next job can fetch it and pass the correct name to AWS `RunInstances`.
Adds tests for the refactored job.

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -23,7 +23,7 @@ import (
 	"go.opentelemetry.io/otel/codes"
 )
 
-const TraceName = "github.com/EnVision/provisioning/internal/clients/http/ec2"
+const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/http/ec2"
 
 type ec2Client struct {
 	ec2     *ec2.Client
@@ -133,7 +133,8 @@ func getStsAssumedCredentials(ctx context.Context, arn string, region string) (*
 	return output.Credentials, nil
 }
 
-// ImportPubkey imports a key and returns AWS ID
+// ImportPubkey imports a key and returns AWS KeyPair name.
+// The AWS name will be set to value of models.Pubkey Name.
 func (c *ec2Client) ImportPubkey(ctx context.Context, key *models.Pubkey, tag string) (string, error) {
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "ImportPubkey")
 	defer span.End()
@@ -170,6 +171,33 @@ func (c *ec2Client) ImportPubkey(ctx context.Context, key *models.Pubkey, tag st
 	}
 
 	return *output.KeyPairId, nil
+}
+
+func (c *ec2Client) GetPubkeyName(ctx context.Context, fingerprint string) (string, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "fetchPubkeyName")
+	defer span.End()
+
+	if !c.assumed {
+		return "", http.ServiceAccountUnsupportedOperationErr
+	}
+	logger := logger(ctx)
+	logger.Trace().Msgf("Fetching AWS key with fingerprint '%s' to get its name", fingerprint)
+	input := &ec2.DescribeKeyPairsInput{}
+	input.Filters = []types.Filter{{Name: ptr.To("fingerprint"), Values: []string{fingerprint}}}
+	output, err := c.ec2.DescribeKeyPairs(ctx, input)
+	if err != nil {
+		if isAWSUnauthorizedError(err) {
+			err = clients.UnauthorizedErr
+		}
+		span.SetStatus(codes.Error, err.Error())
+		return "", fmt.Errorf("cannot fetch SSH key to update its tag %s: %w", fingerprint, err)
+	}
+
+	if len(output.KeyPairs) == 0 {
+		span.SetStatus(codes.Error, fmt.Sprintf("no KeyPair with fingerpring (%s) found", fingerprint))
+		return "", fmt.Errorf("SSH key not found by its fingerprint (%s): %w", fingerprint, http.PubkeyNotFoundErr)
+	}
+	return *output.KeyPairs[0].KeyName, nil
 }
 
 func (c *ec2Client) DeleteSSHKey(ctx context.Context, handle string) error {
--- a/internal/clients/stubs/ec2_stub.go
+++ b/internal/clients/stubs/ec2_stub.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/clients/http"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
 )
@@ -12,7 +13,9 @@ type ec2CtxKeyType string
 
 var ec2CtxKey ec2CtxKeyType = "ec2-interface"
 
-type EC2ClientStub struct{}
+type EC2ClientStub struct {
+	Imported []*models.Pubkey
+}
 
 func init() {
 	clients.GetEC2Client = newEC2CustomerClientStubWithRegion
@@ -41,7 +44,15 @@ func (mock *EC2ClientStub) Status(ctx context.Context) error {
 }
 
 func (mock *EC2ClientStub) ImportPubkey(ctx context.Context, key *models.Pubkey, tag string) (string, error) {
-	return "", nil
+	mock.Imported = append(mock.Imported, key)
+	return "key-12345678912345678", nil
+}
+
+func (mock *EC2ClientStub) GetPubkeyName(ctx context.Context, fingerprint string) (string, error) {
+	if fingerprint == "existing" {
+		return "awsKeyPairName", nil
+	}
+	return "", http.PubkeyNotFoundErr
 }
 
 func (mock *EC2ClientStub) DeleteSSHKey(ctx context.Context, handle string) error {
--- a/internal/clients/stubs/ec2_stub.go
+++ b/internal/clients/stubs/ec2_stub.go
@@ -2,10 +2,12 @@ package stubs
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/RHEnVision/provisioning-backend/internal/clients/http"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
 )
 
@@ -14,7 +16,7 @@ type ec2CtxKeyType string
 var ec2CtxKey ec2CtxKeyType = "ec2-interface"
 
 type EC2ClientStub struct {
-	Imported []*models.Pubkey
+	Imported []*types.KeyPairInfo
 }
 
 func init() {
@@ -44,13 +46,27 @@ func (mock *EC2ClientStub) Status(ctx context.Context) error {
 }
 
 func (mock *EC2ClientStub) ImportPubkey(ctx context.Context, key *models.Pubkey, tag string) (string, error) {
-	mock.Imported = append(mock.Imported, key)
-	return "key-12345678912345678", nil
+	ec2KeyID := fmt.Sprintf("key-%d", len(mock.Imported))
+	keyName := key.Name // copy the name
+	ec2Key := &types.KeyPairInfo{
+		KeyName:        &keyName,
+		KeyFingerprint: &key.Fingerprint,
+		KeyPairId:      &ec2KeyID,
+		PublicKey:      &key.Body,
+		Tags: []types.Tag{{
+			Key:   ptr.To("rhhc:id"),
+			Value: &tag,
+		}},
+	}
+	mock.Imported = append(mock.Imported, ec2Key)
+	return *ec2Key.KeyPairId, nil
 }
 
 func (mock *EC2ClientStub) GetPubkeyName(ctx context.Context, fingerprint string) (string, error) {
-	if fingerprint == "existing" {
-		return "awsKeyPairName", nil
+	for _, key := range mock.Imported {
+		if *key.KeyFingerprint == fingerprint {
+			return *key.KeyName, nil
+		}
 	}
 	return "", http.PubkeyNotFoundErr
 }
--- a/internal/dao/pgx/reservation_pgx.go
+++ b/internal/dao/pgx/reservation_pgx.go
@@ -210,6 +210,19 @@ func (x *reservationDao) UpdateStatus(ctx context.Context, id int64, status stri
 	return nil
 }
 
+func (x *reservationDao) UnscopedUpdateAWSDetail(ctx context.Context, id int64, awsDetail *models.AWSDetail) error {
+	query := `UPDATE aws_reservation_details SET detail = $2 WHERE reservation_id = $1`
+
+	tag, err := db.Pool.Exec(ctx, query, id, awsDetail)
+	if err != nil {
+		return fmt.Errorf("pgx error: %w", err)
+	}
+	if tag.RowsAffected() != 1 {
+		return fmt.Errorf("expected 1 row: %w", dao.ErrAffectedMismatch)
+	}
+	return nil
+}
+
 func (x *reservationDao) UpdateReservationIDForAWS(ctx context.Context, id int64, awsReservationId string) error {
 	query := `UPDATE aws_reservation_details SET aws_reservation_id = $2 WHERE reservation_id = $1`
 
--- a/internal/dao/stubs/reservation_dao.go
+++ b/internal/dao/stubs/reservation_dao.go
@@ -2,14 +2,15 @@ package stubs
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 )
 
 type reservationDaoStub struct {
-	lastId int64
-	store  []*models.AWSReservation
+	lastId   int64
+	storeAWS []*models.AWSReservation
 }
 
 func init() {
@@ -18,7 +19,7 @@ func init() {
 
 func ReservationStubCount(ctx context.Context) int {
 	pkdao := getReservationDaoStub(ctx)
-	return len(pkdao.store)
+	return len(pkdao.storeAWS)
 }
 
 func getReservationDao(ctx context.Context) dao.ReservationDao {
@@ -27,7 +28,7 @@ func getReservationDao(ctx context.Context) dao.ReservationDao {
 
 func (stub *reservationDaoStub) CreateAWS(ctx context.Context, reservation *models.AWSReservation) error {
 	reservation.ID = stub.lastId + 1
-	stub.store = append(stub.store, reservation)
+	stub.storeAWS = append(stub.storeAWS, reservation)
 	stub.lastId++
 	return nil
 }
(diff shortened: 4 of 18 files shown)


lzap (changes requested):
Few nitpick.

Tho I am not sure if we should tag existing resources, in the future we might be able to delete them and this might be uploaded extenrally.

But for keypair it is probably okay.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -163,7 +163,7 @@ func (c *ec2Client) ImportPubkey(ctx context.Context, key *models.Pubkey, tag st
 		if isAWSUnauthorizedError(err) {
 			err = clients.UnauthorizedErr
 		} else if isAWSOperationError(err, "InvalidKeyPair.Duplicate") {
-			err = http.DuplicatePubkeyErr
+			c.updatePubkeyTag(ctx, key, tag)
 		}
```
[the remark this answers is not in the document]
lzap: Ok now you have shadowing, I know this one is pesky but shadowing can introduce nasty bugs.

ezr-ondrej:
Well you might be actually right :thinking: maybe we should just blank the tag in the database and that would mean this is not a resource we've created :thinking:

lzap:
Commit message check complains, feel free to update it if you want to use tickets like this.

lzap:
Tho I would rather prefer dropping them directly in the commit message, we can easily parse them all since they have a common prefix after changelog generation via `perl` or `awk`.

ezr-ondrej:
I'm updating it in https://github.com/RHEnVision/provisioning-backend/pull/313/files

lzap:
Please rebase and reword the commit message depending on the outcome the team has on the subject.

Or you can just reword this in the old way for now so this can be merged.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -172,6 +176,45 @@ func (c *ec2Client) ImportPubkey(ctx context.Context, key *models.Pubkey, tag st
...
+		}
+		span.SetStatus(codes.Error, err.Error())
+		return "", fmt.Errorf("cannot fetch SSH key to update its tag %s: %w", key.Fingerprint, err)
+	}
+
+	if len(output.KeyPairs) == 0 {
+		return "", fmt.Errorf("SSH key not found by its fingerprint (%s): %w", key.Fingerprint, http.PubkeyNotFoundErr)
+	}
+	tagInput := &ec2.CreateTagsInput{
+		Resources: []string{*output.KeyPairs[0].KeyPairId},
```
adiabramovitch, line 203: I think it will be more readable to save *output.KeyPairs[0] as a variable

internal/jobs/pubkey_upload_aws_job.go
```diff
@@ -91,7 +91,11 @@ func handlePubkeyUploadAWS(ctx context.Context, args *PubkeyUploadAWSTaskArgs) e
 	pkr.Handle, err = ec2Client.ImportPubkey(ctx, pubkey, pkr.FormattedTag())
```
ezr-ondrej, line 91: :warning: Handle would be a name of the key now. Can we handle it? We will need to do something about our DB values on stage (cleanup of the Pubkey resources comes to mind as simple solution).

Reasoning: we are passing a name to the `RunInstances`, so it should be the handle, what for we keep the Handle if not to be used to be passed to `RunInstances`.
@lzap @adiabramovitch

internal/clients/http/ec2/ec2_client.go
ezr-ondrej: Good suggestion, but I've refactored the code completely, so now it doesn't make sense as it's used once ^_^

internal/clients/http/ec2/ec2_client.go
```diff
@@ -169,7 +170,34 @@ func (c *ec2Client) ImportPubkey(ctx context.Context, key *models.Pubkey, tag st
 		return "", fmt.Errorf("cannot import SSH key %s: %w", key.Name, err)
 	}
 
-	return *output.KeyPairId, nil
+	return *output.KeyName, nil
+}
+
+func (c *ec2Client) GetPubkeyName(ctx context.Context, key *models.Pubkey) (string, error) {
```
lzap, line 176:
```suggestion
func (c *ec2Client) FindNameByFingerprint(ctx context.Context, key *models.Pubkey) (string, error) {
```

Perhaps? We should probably only pass fingerprint, not the whole pubkey as well.

internal/jobs/launch_instance_aws.go
```diff
@@ -70,6 +73,10 @@ func handleLaunchInstanceAWS(ctx context.Context, args *LaunchInstanceAWSTaskArg
 	if err != nil {
 		return fmt.Errorf("cannot get pubkey by id: %w", err)
 	}
+	pkr, err := pkD.UnscopedGetResourceBySourceAndRegion(ctx, pk.ID, args.SourceID, args.Region)
+	if err != nil {
+		return fmt.Errorf("unable to fetch pubkey resource: %w", err)
+	}
```
lzap, line 79: Isn't more straightforward to simply fetch this before the job and pass both pubkey and its resources? I mean, DB lookups are fine to be done outside of background job, they are fast. It's AWS API calls what should be performed in a job.

internal/jobs/pubkey_upload_aws_job.go
lzap: When I designed this, I was more focused on the job scheduler part than on the AWS part, so I sort of mindlessly created this "upload pubkey" job. I think it is the time to revisit this from the ground up and rethink this. Few points:

* The "aws handle" field does not have to be the "identifier" we use for keys, it can serve as an informative field too. I very much like the idea of searching pubkeys by fingerprint because it looks like on AWS this is what the API supports and it also enforces uniqueness therefore it is guaranteed to find 0 or 1 entry.
* The job is poorly named, I think we should think about our job scheduler as a small "ansible", it should be idempotent. Therefore I think we should rename it to EnsurePubkeyAWS or something like that - it should do whatever must be done to have that pubkey on the AWS account.
* It is absolutely no problem to perform more AWS API requests in a background job, we have all the time in the world, we can slow down jobs if we hit any API limits. So whoever want's to redesign this, they can even simply just query AWS first, if a key is found, then we can compare our handle and update it (or maybe it can be completely dropped we don't know if this is possible on other clouds). If it's not found, upload it and update our records.
* I think it is a good time to think about documenting each job, I suggest to have something close to the code, so code comment describing what the job should do would be a really nice idea. We can maybe start writing the new "job story" on the next tech meeting and once we agree we have a good approach that encapsulates all the edge cases, we can go ahead and implement it.

internal/jobs/launch_instance_aws.go
ezr-ondrej: I guess, but what do you think by passed to the job? the whole model in args? thinking that feels unsafe to me to marshal and unmarshal it for no reason - it feels better to just fetch it from the DB in the job shrug

internal/jobs/launch_instance_aws.go
```diff
@@ -87,13 +90,11 @@ func handleLaunchInstanceAWS(ctx context.Context, args *LaunchInstanceAWSTaskArg
 	}
 
 	logger.Trace().Msg("Executing RunInstances")
-	instances, awsReservationId, err := ec2Client.RunInstances(ctx, args.Detail.Name, args.Detail.Amount, types.InstanceType(args.Detail.InstanceType), args.AMI, pk.Name, userData)
+	instances, awsReservationId, err := ec2Client.RunInstances(ctx, args.Detail.Name, args.Detail.Amount, types.InstanceType(args.Detail.InstanceType), args.AMI, reservation.Detail.PubkeyNameAWS, userData)
```
ezr-ondrej, line 93: so the Name should be now saved to PubkeyNameAWS and passed to this method from there.
It feels like workaround and I belive it would be great if we have job outputs, but Happy to discuss later and use workaround for now.

internal/dao/tests/main_test.go
```diff
@@ -27,7 +27,7 @@ func teardown() {
 }
 
 func initEnvironment() {
-	config.Initialize("config/test.env")
+	config.Initialize("../../../config/test.env")
```
ezr-ondrej, line 30: I don't know if ppl had their config in this folder, but this makes bit more sense to me :thinking:

internal/jobs/common.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 func decodeJob(ctx context.Context, job dejq.Job, argInterface interface{}) error {
-	err := job.Decode(&argInterface)
+	err := job.Decode(argInterface)
```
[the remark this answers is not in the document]
lzap: Nice find, I am not sure how generics could be helpful here, but go ahead and make a draft PR if you have something on your mind.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -172,6 +173,33 @@ func (c *ec2Client) ImportPubkey(ctx context.Context, key *models.Pubkey, tag st
...
+		span.SetStatus(codes.Error, err.Error())
+		return "", fmt.Errorf("cannot fetch SSH key to update its tag %s: %w", fingerprint, err)
+	}
+
+	if len(output.KeyPairs) == 0 {
+		span.SetStatus(codes.Error, fmt.Sprintf("no KeyPair with fingerpring (%s) found", fingerprint))
+		return "", fmt.Errorf("SSH key not found by its fingerprint (%s): %w", fingerprint, http.PubkeyNotFoundErr)
+	}
+	return *output.KeyPairs[0].KeyName, nil
+}
```
lzap, line 201: I am not sure if we should treat the 404 like a hard error, at most this should be warning, if ever.

Other than that, looks good.

internal/clients/stubs/ec2_stub.go
```diff
@@ -41,7 +44,15 @@ func (mock *EC2ClientStub) Status(ctx context.Context) error {
 }
 
 func (mock *EC2ClientStub) ImportPubkey(ctx context.Context, key *models.Pubkey, tag string) (string, error) {
-	return "", nil
+	mock.Imported = append(mock.Imported, key)
+	return "key-12345678912345678", nil
+}
+
+func (mock *EC2ClientStub) GetPubkeyName(ctx context.Context, fingerprint string) (string, error) {
+	if fingerprint == "existing" {
```
lzap, line 52: Highly encourage to use proper data than fake one :-) Like proper fingerprint of a real key... :-)

internal/dao/pgx/reservation_pgx.go
```diff
@@ -210,6 +210,19 @@ func (x *reservationDao) UpdateStatus(ctx context.Context, id int64, status stri
...
+
+	tag, err := db.Pool.Exec(ctx, query, id, awsDetail)
+	if err != nil {
+		return fmt.Errorf("pgx error: %w", err)
+	}
+	if tag.RowsAffected() != 1 {
+		return fmt.Errorf("expected 1 row: %w", dao.ErrAffectedMismatch)
+	}
+	return nil
+}
```
lzap, line 224: We should be generally trying to remove all "unscoped" DAO functions in the future, it's a bit more work to join tables to get an account, but it is dramatically safer if we misuse this in the future. I am okay to merge this in, unless you want to do an extra mile.

internal/jobs/queue/dejq/dejq_queue.go
```diff
@@ -30,7 +30,7 @@ func init() {
 func RegisterJobs(logger *zerolog.Logger) {
 	logger.Debug().Msg("Registering job queue handlers")
 	dejqQueue.RegisterHandler(queue.TypeNoop, jobs.HandleNoop)
-	dejqQueue.RegisterHandler(queue.TypePubkeyUploadAws, jobs.HandlePubkeyUploadAWS)
+	dejqQueue.RegisterHandler(queue.TypePubkeyUploadAws, jobs.HandleEnsurePubkeyOnAWS)
```
lzap, line 33: Can you also change the queue name? We do not persist queues yet so you can do this cheaply.

internal/models/reservation_model.go
```diff
@@ -59,6 +59,9 @@ type AWSDetail struct {
 
 	// Immediately power off the system after initialization
 	PowerOff bool `json:"poweroff"`
+
+	// PubkeyName on AWS in given region. Found by the EnsurePubkey job.
+	PubkeyNameAWS string `json:"pubkey_name"`
```
lzap, line 64: I suggest to be consistent:

```suggestion
	PubkeyName string `json:"pubkey_name"`
```

internal/jobs/launch_instance_aws.go
lzap: Not sure what you mean, this looks okay to me, like a stable and maintainable code. I don't think we are getting job pipeline anytime soon, even dbjobqueue does not support it (it is stored in DB yes but dependant jobs cannot access it I think).

internal/dao/tests/main_test.go
ezr-ondrej: I do not want to start from weird directory, I want to run `make integration-tests` from the root directory of the project. I think it's sensible. Without this lookup it doesn't work. It's not `CWD` in this case, it's the main fail location - which is this file, so it is relative path to this file.

My aim was to make it work with config in the main `config/` directory, but I might have done something wrong :)

internal/clients/http/ec2/ec2_client.go
ezr-ondrej: well the error gets caught in the only caller, but if you think we should just return empty string, I'm ok with it, but feels bit weird, so I thought error is bit cleaner there.

internal/clients/stubs/ec2_stub.go
ezr-ondrej: Well it is a special value at this moment, so I wanted to make it extra clear it is a special value. Real value might confuse ppl where it is defined. With real data, I'd want the key to actually be in the mock (added).

But that might be cleanest approach, I was just lazy to define fixtures for ex2, I needed to write a lot of setup code for tests already :) but if we want that, I'm happy to work on it.

internal/dao/pgx/reservation_pgx.go
ezr-ondrej: Well I don't want to. At least not without discussion, as complicating the queries might have some performance impact, I'd love to discuss if we don't want to do it in two queries. Which makes it a problem on its own IMHO and would slow down this PR which is already quite big and will be hard to get in :)

internal/jobs/launch_instance_aws.go
ezr-ondrej: right :) well workaround because we fetch record from database just to get output from previsous job (and the "column" exists just to be passed from one job to the other ;))

internal/dao/tests/main_test.go
ezr-ondrej: I thought it will be no brainer, but if it needs to be discussed, lets take it to another PR: #337.

internal/clients/http/ec2/ec2_client.go
ezr-ondrej: Happy to indicate not found in other means, but would you have a suggestion?

internal/jobs/queue/dejq/dejq_queue.go
ezr-ondrej: Done

internal/models/reservation_model.go
ezr-ondrej: Done

internal/clients/http/ec2/ec2_client.go
lzap: Nah you are right that the caller should be responsible for skipping errors. This is fine.

internal/clients/stubs/ec2_stub.go
lzap: I do not understand why you want to use this "special" value. Can't you simply update the stub to actually accept real Pubkey entries (with real data) so you can then get pubkey by name using it's real fingerprint? We already have several real pubkeys in the test suite (mine, Anna's...) and there is nothing wrong with creating stubs that actually do stuff. I actually thing although it will take more time, it will pay off as soon as it gets more complex later on.

internal/dao/pgx/reservation_pgx.go
lzap: You want to discuss SQL JOIN? Okay, but prepare good arguments for "joins are slow" in this case. My arguments will be "it's security, let's be extra careful". We already do this in several functions, I remember writing this myself.

lzap: Given the function prerix, I do not want to block this PR on this, but generally I would like us not to have any Unscoped functions, I do not see big reason not to join accounts table on a query that returns a _single row_.

lzap:
Other than that, looks good.

internal/clients/stubs/ec2_stub.go
```diff
@@ -14,7 +15,8 @@ type ec2CtxKeyType string
 var ec2CtxKey ec2CtxKeyType = "ec2-interface"
 
 type EC2ClientStub struct {
-	Imported []*models.Pubkey
+	Imported []*types.KeyPairInfo
+	lastId   int
```
lzap, line 19: Heh, what this is used for? Why don't you simply use `len(Imported)`?

internal/clients/stubs/ec2_stub.go
```diff
@@ -44,13 +46,23 @@ func (mock *EC2ClientStub) Status(ctx context.Context) error {
 }
 
 func (mock *EC2ClientStub) ImportPubkey(ctx context.Context, key *models.Pubkey, tag string) (string, error) {
-	mock.Imported = append(mock.Imported, key)
-	return "key-12345678912345678", nil
+	ec2KeyID := fmt.Sprintf("key-%d", mock.lastId+1)
```
lzap, line 49: This should really be `len(mock.Imported)+1`

internal/clients/stubs/ec2_stub.go
```diff
@@ -44,13 +46,23 @@ func (mock *EC2ClientStub) Status(ctx context.Context) error {
...
 func (mock *EC2ClientStub) GetPubkeyName(ctx context.Context, fingerprint string) (string, error) {
-	if fingerprint == "existing" {
-		return "awsKeyPairName", nil
+	for _, key := range mock.Imported {
+		if *key.KeyFingerprint == fingerprint {
+			return *key.KeyName, nil
+		}
 	}
 	return "", http.PubkeyNotFoundErr
 }
```
lzap, line 68: Thank you this is WAAAY better than the old proposal.

lzap:
Looks great, please explain why you don't use `len` and we can merge (linter issue).

internal/clients/stubs/ec2_stub.go
ezr-ondrej: because for some reason it didn't occure to me xD Good idea :+1:

internal/dao/stubs/reservation_dao.go
```diff
@@ -2,14 +2,15 @@ package stubs
...
+	"fmt"
 
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 )
 
 type reservationDaoStub struct {
-	lastId int64
-	store  []*models.AWSReservation
+	lastId   int64
```
lzap, line 12: Oh my god we have it everywhere, it was a copy paste :-)

Ok not here, will fix this.

internal/jobs/launch_instance_aws.go
lzap: Right.

lzap (changes requested):
The code looks good, but there is a problem in the workflow.

* PKR is looked up, let's assume it already exists in the db.
* Then presence of key is checked on AWS, assume it was not found (user deleted it).
* Therefore tag (of the existing pkr) is randomized and it is imported.
* Ssh key name is stored in the reservation.
* Finally, and here is the problem, since pkr is an existing record, ID != 0, therefore the newly generated tag will not be stored.

I think the solution is to remove `pkr.Tag = ""` and just call `RandomizeTag()` which will only randomize when pkr is not existing record.

When on it, please update the "Uploading/uploaded public key" status on line 52/53 to something like "Ensuring/Ensured" maybe. Not sure about it, we can perhaps keep it.

lzap:
Also if both pkr and ssh-key exist but the tag/handle do not match, the pkr record is not updated. I think we should do this, it's one extra if statement (compare those two values and do update if they don't match AWS being the source of "truth"). I know we will eventually drop pkr but what if not for some reason - we want to have accurate data.

PR #317, opened 2022-11-23
fix: kafka port is a pointer

lzap:
Clowder types are a mess, some values are pointers (port) others are not (hostname).

People sometimes use pointers to be able to use `nil` and it's bad practice, zero values are also valid options, specifically for port it is definitely a not valid port number.


ezr-ondrej (approved):
Nice find :+1:

PR #319, opened 2022-11-23
feat: better Kafka logging

lzap:
Just to see little bit more from Kafka:

* list of hosts
* when a message is really sent
* increases to trace for ephemeral

The trace is a bit dirty but I am not sure how can I set options for clowder. It has some CLI arguments but it's cryptic without examples. If I knew how to use its `-p component/parameter=value` then we could ask QE folks to increase verbosity for PR checks.

### DIFF
--- a/cmd/pbapi/main.go
+++ b/cmd/pbapi/main.go
@@ -87,7 +87,7 @@ func main() {
 
 	// initialize platform kafka
 	if config.Kafka.Enabled {
-		err = kafka.InitializeKafkaBroker()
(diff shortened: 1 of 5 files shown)


lzap:
Wait we already have this at https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/provisioning/deploy.yml#L69-70 does this work? I am not seeing trace logs.

Dropping the hardcoded increase.

ezr-ondrej (approved):
Sure :+1:

PR #322, opened 2022-11-24
fix: intermittent failures on CI for ASM queue test

lzap:
I don't see this on my laptop, but Go scheduler can decide not to run the enqueue goroutine which puts the message into the buffer, therefore the test sometimes fail on CI.

I do not want to create some complex synchronization test-only variables in order to wait until data goes through the channel, therefore I am using the `Sched()` function to allow for context switch and that itself should already help a lot.

But in case it would not, a small delay of 20 milliseconds should definitely do the work. Unless there is a bug in the code which I don't see.

### DIFF
--- a/internal/background/availability_queue_test.go
+++ b/internal/background/availability_queue_test.go
@@ -2,6 +2,7 @@ package background
 
 import (
 	"context"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/background/availability_queue_test.go
```diff
@@ -61,20 +62,34 @@ func TestQueueCancelSend(t *testing.T) {
...
-	EnqueueAvailabilityStatusRequest(&msg)
+	// at this point hopefully the message was already buffered but just in case :-)
+	time.Sleep(20 * time.Millisecond)
+
+	// cancel the sender before the 5 seconds timeout (so cancel branch is triggered)
 	senderCancel()
+
+	// wait until the message is consumed
 	wg.Wait()
 }
```
lzap, line 95: Not ideal, but the purpose of this test is to put a message into the buffer with a high (infinite = 5 seconds in tests) tick interval and then initiate context cancel which should cause all buffered messages to be sent before it returns.

The error we are seeing is because the runtime does not give enough CPU time in order to send the data over channel, receive it and put into the buffer. In other terms, function `sendAvailabilityRequestMessages` is CPU-starving.

ezr-ondrej (approved):
Looks sensible, lets try it out :)

PR #323, opened 2022-11-29
fix: payload name not nullable

lzap:
https://issues.redhat.com/browse/HMSPROV-373

### DIFF
--- a/internal/services/aws_reservation_service.go
+++ b/internal/services/aws_reservation_service.go
@@ -48,10 +48,8 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 	reservation.Status = "Created"
 	reservation.Provider = models.ProviderTypeAWS
 	reservation.Steps = 2
-	if payload.Name != nil {
-		newName := config.Application.InstancePrefix + *payload.Name
-		reservation.Detail.Name = &newName
-	}
+	newName := config.Application.InstancePrefix + payload.Name
+	reservation.Detail.Name = &newName
 
 	// validate pubkey
 	logger.Debug().Msgf("Validating existence of pubkey %d for this account", reservation.PubkeyID)
--- a/internal/payloads/null_convert.go
+++ b/internal/payloads/null_convert.go
@@ -9,3 +9,10 @@ func SqlNullToStringPtr(s sql.NullString) *string {
 	str := s.String
 	return &str
 }
+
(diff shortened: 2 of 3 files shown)


internal/services/aws_reservation_service.go
```diff
@@ -48,8 +48,8 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 	reservation.Status = "Created"
 	reservation.Provider = models.ProviderTypeAWS
 	reservation.Steps = 2
-	if payload.Name != nil {
-		newName := config.Application.InstancePrefix + *payload.Name
+	if config.Application.InstancePrefix != "" {
```
ezr-ondrej, line 51: I belive this is introducing a regression, that the CI is finding.
Now you can't name the instance if the prefix is not configured.

ezr-ondrej (changes requested):
The commit message failure is a bug in the script: #324
But the IQE is correct about there being a regression I believe

internal/services/aws_reservation_service.go
lzap: Yeah I rushed that in, fixed.

internal/services/aws_reservation_service.go
```diff
@@ -48,10 +48,8 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 	reservation.Status = "Created"
 	reservation.Provider = models.ProviderTypeAWS
 	reservation.Steps = 2
-	if payload.Name != nil {
-		newName := config.Application.InstancePrefix + *payload.Name
-		reservation.Detail.Name = &newName
-	}
+	newName := config.Application.InstancePrefix + payload.Name
+	reservation.Detail.Name = &newName
```
ezr-ondrej, line 52: Will this behave correctly when there is an empty name?
Before we had nil for name not set, now we have empty string.
I don't know if that matters for AWS, but I'd rather be sure (our IQE pipeline does not test this as it is setting the name).

akhil-jha (approved):

internal/services/aws_reservation_service.go
lzap: What you mean? I believe that `"" + "" = ""`.

ezr-ondrej: oki make sense :) I was talking about what we send down the road :)

PR #327, opened 2022-12-02
build: Use external changelog module

ezr-ondrej:
Uses a new python package for checking the commit message style.

The package lives at https://github.com/RHEnVision/changelog


ezr-ondrej:
self merging as this is needed to fix the changelog issues :)

PR #328, opened 2022-12-05
ci: Update iqe markers for backend

akhil-jha:
When UI test runs, it errors out because `selenium` is not passed along with it, resulting in session error.
This should fix those error and run only api tests.


ezr-ondrej:
Ok, this seems green.
So let's merge to fix builds, @mshriver if you'd disagree, please let us know! :)

mshriver:
Now that I'm awake ..... yep this is right!

PR #330, opened 2022-12-06
fix(HMSPROV-368): change version to BuildCommit

avitova:
This should change the Version to BuildCommit

### DIFF
--- a/internal/payloads/error_payload.go
+++ b/internal/payloads/error_payload.go
@@ -27,7 +27,7 @@ type ResponseError struct {
 	// full root cause
 	Error error `json:"error"`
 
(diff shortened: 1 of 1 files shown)


lzap (approved):
Thank you.

If you want to dig into Makefile hole, we could actually store version AND commit as two separate strings. Previously we had no git tags, but now we do tag releases. You could create new variable BuildVersion and do a Makefile change to get latest git tag and store it in this new variable.

Then we could actually have both Version (named Version) and Commit (named Commit) in the payload. If you ask me, commit is good enough but you might be interested to play around GNU Makefile :)

PR #332, opened 2022-12-07
fix: utilize clowder topic mapping

lzap:
https://issues.redhat.com/browse/HMSPROV-343

We cannot use topics directly, there is some sort of "prefix" on the platform that is provided by the clowder. This should fix the stage, I am unable to test this until we deploy it on stage.

Also adds some debugging statements when adding the CA cert. It is a public key cert so it is safe to log into debug.

### DIFF
--- a/internal/config/helpers.go
+++ b/internal/config/helpers.go
@@ -34,6 +34,15 @@ func InProdClowder() bool {
 	return clowder.IsClowderEnabled() && strings.HasPrefix(*clowder.LoadedConfig.Metadata.EnvName, "env-stage")
 }
 
+// TopicName returns mapped topic from Clowder. When not running in Clowder mode, it returns the input topic name.
+func TopicName(topic string) string {
+	if t, ok := clowder.KafkaTopics[topic]; ok {
+		return t.Name
+	}
+
+	return topic
+}
+
 func StringToURL(urlStr string) *url.URL {
 	if urlStr == "" {
 		return nil
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -223,6 +223,7 @@ func Initialize(configFiles ...string) {
 		// kafka
 		if cfg.Kafka != nil {
 			config.Kafka.Enabled = true
+
 			config.Kafka.Brokers = make([]string, len(cfg.Kafka.Brokers))
 			for i, b := range cfg.Kafka.Brokers {
 				config.Kafka.Brokers[i] = fmt.Sprintf("%s:%d", b.Hostname, *b.Port)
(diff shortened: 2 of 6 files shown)


internal/config/helpers.go
```diff
@@ -34,6 +34,15 @@ func InProdClowder() bool {
 	return clowder.IsClowderEnabled() && strings.HasPrefix(*clowder.LoadedConfig.Metadata.EnvName, "env-stage")
 }
 
+// RequestedTopic returns mapped topic from Clowder. When not running in Clowder mode, it returns the input topic name.
+func RequestedTopic(topic string) string {
```
ezr-ondrej, line 38: `KafkaTopic` or `RealKafkaTopicName` might be better names here, but no biggie :)

internal/config/helpers.go
```diff
@@ -34,6 +34,15 @@ func InProdClowder() bool {
...
+	if t, ok := topicRequest[topic]; ok {
+		return t
```
ezr-ondrej, line 40: do we need our mapping here? I think this would work:
```suggestion
    if InClowder() {
	    if t, ok := cfg.Kafka.Topics[topic]; ok {
	        return t.Name
	    }
```

internal/config/helpers.go
lzap: Yeah I try to stick with the clowder terminology here, not that I like it but should be less confusing. The function does map requested topics to (real) topics, which I would just call "topics" I guess. Since in Go we should avoid getters, "GetTopic" would be bad, so it would end up as "Topic" or "RealTopic" which sounds non-descriptive to me.

ezr-ondrej (commented):
I think we can simplify the mapping - use the one from Clowder config helper lib and not keep our own.

internal/config/helpers.go
lzap: Are you sure? `cfg.Kafka.Topics` is a slice, not map. We could search the slice everytime but we will be sending many messages so I wanted a map so it will scale.

lzap: Turns out you mean https://github.com/RedHatInsights/app-common-go/blob/811f8bb0662990c714e87d62f633940f05390de0/pkg/api/v1/config.go#L13 which is indeed a map, Clowder does exactly what I do in our codebase, let's use it.

ezr-ondrej: You're rigth the correct thing to use would be `cfg.KafkaTopics`

ezr-ondrej (approved):
Thanks @lzap ! Lets take it for a ride ^-^

PR #333, opened 2022-12-07
feat: introduce availability status endpoint

lzap:
We can start testing in ephemeral where Kafka does work already. For stage, we are still working on a clowder mapping of topics, there is an open PR for that. This one will be a challenge to test:

* Hit the endpoint with an existing source ID that has a AWS account associated.
* Listen for the `platform.sources.event-stream` Kafka event that contains the source ID and result.
* Results can be cached in stage/prod, however, in ephemeral we should not be doing any caching so the event should be sent always (with a small delay of approx. 10 seconds).

Let's try if our checker likes URLs too:

https://issues.redhat.com/browse/HMSPROV-337

### DIFF
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -289,3 +289,23 @@ paths:
                 $ref: '#/components/schemas/v1.NoopReservationResponse'
         "500":
           $ref: '#/components/responses/InternalError'
+  /availability_status/sources:
+    post:
+      operationId: availabilityStatus
+      description: >
+        Schedules a background operation of Sources availability check. These checks are
+        are performed in separate process at it's own pace. Results are sent via Kafka
+        to Sources. There is no output from this REST operation available, no tracking
+        of jobs is possible.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/v1.AvailabilityStatusRequest'
+        description: availability status request with source id
+        required: true
+      responses:
+        '200':
+          description: 'Returned on success, empty response.'
+        "500":
+          $ref: '#/components/responses/InternalError'
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -108,6 +108,14 @@
         },
         "type": "object"
       },
+      "v1.AvailabilityStatusRequest": {
+        "properties": {
+          "source_id": {
+            "type": "string"
+          }
+        },
+        "type": "object"
+      },
       "v1.InstanceTypeResponse": {
         "properties": {
           "architecture": {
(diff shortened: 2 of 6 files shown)


lzap:
Correction: listen on `platform.sources.status` the topic is different.

lzap:
It is green, if you want we can merge or if you prefer to write test first, we can delay this. Since this is new code, I lean towards merging it as-is.

adiabramovitch:
Shouldn't we have some sort of a payload for the response? 
it is even possible? because in contrast to the reservation job, here we do not use any job or save info in the DB.

akhil-jha:
So I am trying this: I have created a source 
```
Out[12]: 
[{'id': '10',
  'name': 'test-provisioning::bulk:rhel_aws_manual_configuration-KFPDv5zl3m',
  'source_type_id': '10',
  'uid': '5771848a-9b33-40d1-9255-c2b2bbbd5183'}]
```

From cURL:
```
curl -s -L -X POST 'https://<hostname>/availability_status' -H 'Content-Type: application/json' -u user:pass -F 'source_id="10"'

{
    "msg": "invalid request: availability status",
    "trace_id": "dfb1a1994e0fb1c33dfe1cad6ce10fb9",
    "error": {
        "Offset": 2
    },
    "version": "67fc811",
    "build_time": "2022-12-07_09:58:54"
}
```


In IQE:
```
In [10]: app.provisioning.rest_client.default_api.availability_status({'source_id':'10'})
# Empty resposne. 

In [11]: app.provisioning.rest_client.default_api.availability_status({'source_id':'-1'})
# Empty again?
```

I don't know what's going on. Am I doing wrong steps?

adiabramovitch:
So it is supposed to have an empty response if the operation succeeded but I agree it is lacing information.

lzap:
We cannot return anything from the endpoint, messages are sent into a channel to be sent later on, if Kafka transport fails there will only be log in our logs. There is nothing we can do about this, that is by design - the sending Sources background process does not check responses, at best it logs errors into their logs I think.

lzap:
Idea: Please test this with 1 message and also if possible with 2096 messages quickly sent one after the other (ideally concurrently) You can use the same source, just to stress test the batching code (it should send after 2 seconds or every 1024 requests).

lzap:
Hey, you are seeing this error:

https://github.com/RHEnVision/provisioning-backend/blob/main/internal/services/availability_status_service.go#L15

This is because the payload could not be unmarshalled, it expects this according to the OpenAPI:

```json
{ "source_id": 1234 }
```

In the body (payload), no URL parameters are used in this endpoint.

lzap:
_(Should have been string not integer in the example above sorry.)_

cmd/spec/path.yaml
```diff
@@ -289,3 +289,23 @@ paths:
                 $ref: '#/components/schemas/v1.NoopReservationResponse'
         "500":
           $ref: '#/components/responses/InternalError'
+  /availability_status:
+    post:
+      operationId: availabilityStatus
```
ezr-ondrej, line 294: Could we include the `Sources` name in the enpoint name?
At this point it is clear what we mean, but later on we can have more availability checks :)

lzap: Added.

cmd/spec/path.yaml
```diff
@@ -289,3 +289,23 @@ paths:
...
+        of jobs is possible.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/v1.AvailabilityStatusRequest'
+        description: availability status request with source id
+        required: true
+      responses:
+        '200':
```
adiabramovitch, line 308: I think @akhil-jha prefers it to be 204 because of the empty response

adiabramovitch (approved):
LGTM, just a small thing

akhil-jha (approved):
Yay!

cmd/spec/path.yaml
ezr-ondrej: @lzap @adiabramovitch did we adress this, or it was not issue in IQE this time @akhil-jha ?

PR #334, opened 2022-12-07
build: add sources kafka topics in clowdapp

ezr-ondrej:


ezr-ondrej:
Right, gonna self merge :) it's green

PR #337, opened 2022-12-07
test: use path to the main config dir

ezr-ondrej:
This was the only way for me to run `make integration-tests` with the config in the main `config/` folder.
Happy to discuss better options, but if I run the make from root folder, it needs to respect configs from the main config folder, otherwise it is very weird for devs.

Taken out from other PR after a bit of discussion: https://github.com/RHEnVision/provisioning-backend/pull/315/files#r1038972780

### DIFF
--- a/internal/dao/tests/main_test.go
+++ b/internal/dao/tests/main_test.go
@@ -27,7 +27,7 @@ func teardown() {
 }
 
 func initEnvironment() {
-	config.Initialize("config/test.env")
+	config.Initialize("../../../config/test.env")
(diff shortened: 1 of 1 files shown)


lzap (approved):
Ah now I know what you mean, sorry. Looks like `go test` changes to the directory that is provided on the path.

Go 1.20 is getting new `-C` arguments for "go subcommands" so this should help:

    go test -C directory/ ...

I am fine with the workaround tho. Thanks.

PR #338, opened 2022-12-12
Fix: Add event_type header to kafka message and resource type 

adiabramovitch:


internal/kafka/avail_status_msg.go
```diff
@@ -44,6 +44,7 @@ func genericMessage(ctx context.Context, m any, key string, topic string) (Gener
 			"x-RH-Identity":               identity.GetIdentityHeader(ctx),
 			"X-RH-Sources-Org-Id":         id.Identity.OrgID,
 			"X-RH-Sources-Account-Number": id.Identity.AccountNumber,
+			"Event_Type":                  "availability_status",
```
lzap, line 47: Can you convert ALL headers to lowercase and add a comment to keep all headers in lowercase, it really looks like lower case is what other teams use.

lzap:
For the record, this is from the sources codebase:

```
  const (
          PSK               = "x-rh-sources-psk"
          AccountNumber     = "x-rh-sources-account-number"
          OrgID             = "x-rh-sources-org-id"
          PSKUserID         = "x-rh-sources-user-id"
          XRHID             = "x-rh-identity"
          InsightsRequestID = "x-rh-insights-request-id"
          ParsedIdentity    = "identity" 
          TenantID          = "tenantID"
          UserID            = "userID"
  ) 
```

It looks like it is almost all lower case :-)

lzap (approved):
Matches what is in sources, I double checked. Don't know if that's all for now, but at least what we have is expected.

ezr-ondrej (changes requested):
Can we drop the endpoint commit pls? then I'd be happy to merge, but endpoind should be gated by IQE test :)

ezr-ondrej (approved):

lzap:
Thanks, enabled auto merge.

PR #339, opened 2022-12-12
fix(HMSPROV-345): remove default tag

adiabramovitch:


lzap (approved):
Nice, it should have been capitalized but sources will do this.

PR #340, opened 2022-12-12
fix: Kafka headers are slice now

lzap:
Kafka headers are slices instead of map, this is how Kafka does it, we should follow what the library does.

https://issues.redhat.com/browse/HMSPROV-337

### DIFF
--- a/internal/kafka/kafka.go
+++ b/internal/kafka/kafka.go
@@ -177,7 +177,7 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, handler func(ct
 			logger.Warn().Err(err).Msgf("Error when reading message: %s", err.Error())
 		} else {
 			logger.Trace().Bytes("payload", msg.Value).Msgf("Received message with key: %s", msg.Key)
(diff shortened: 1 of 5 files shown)


internal/kafka/kafka.go
```diff
@@ -186,9 +186,9 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, handler func(ct
 	}
 }
 
-func LookupKafkaHeader(ctx context.Context, headers []protocol.Header) string {
+func Header(name string, headers []protocol.Header) string {
```
lzap, line 189: Made it generic, also renamed to something shorter if you don't mind.

adiabramovitch (approved):
LGTM, linter is failing can we fix that?

lzap:
Unexported the function, fixed the check. There is actually a check that detect the variadic function expects even number of arguments and it warns about this. How cool! Disabled for the test which checks for that as well.

ezr-ondrej:
I'm going to be bit pedantic here. The last commit isn't fixing a bug, so it should be `ci:` right? And thus it won't need to be linked to irrelevant jira issue? :)

ezr-ondrej:
I've taken the liberty of fixing the commit message myself so I don't block on it for no reason :)

PR #342, opened 2022-12-13
fix: Fetch tag from AWS 

ezr-ondrej:
Refs: HMSPROV-339

To resolve
- [ ] updating Handle will result in the key from AWS when deleted on our side, do we want that?

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -180,12 +180,12 @@ func (c *ec2Client) ImportPubkey(ctx context.Context, key *models.Pubkey, tag st
 	return *output.KeyPairId, nil
 }
 
-func (c *ec2Client) GetPubkeyName(ctx context.Context, fingerprint string) (string, error) {
+func (c *ec2Client) GetKeyPairByFingerprint(ctx context.Context, fingerprint string) (*clients.EC2KeyPairInfo, error) {
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "fetchPubkeyName")
 	defer span.End()
 
 	if !c.assumed {
-		return "", http.ServiceAccountUnsupportedOperationErr
+		return nil, http.ServiceAccountUnsupportedOperationErr
 	}
 	logger := logger(ctx)
 	logger.Trace().Msgf("Fetching AWS key with fingerprint '%s' to get its name", fingerprint)
@@ -197,14 +197,23 @@ func (c *ec2Client) GetPubkeyName(ctx context.Context, fingerprint string) (stri
 			err = clients.UnauthorizedErr
 		}
 		span.SetStatus(codes.Error, err.Error())
-		return "", fmt.Errorf("cannot fetch SSH key to update its tag %s: %w", fingerprint, err)
+		return nil, fmt.Errorf("cannot fetch SSH key to update its tag %s: %w", fingerprint, err)
 	}
 
 	if len(output.KeyPairs) == 0 {
 		span.SetStatus(codes.Error, fmt.Sprintf("no KeyPair with fingerpring (%s) found", fingerprint))
-		return "", fmt.Errorf("SSH key not found by its fingerprint (%s): %w", fingerprint, http.PubkeyNotFoundErr)
+		return nil, fmt.Errorf("SSH key not found by its fingerprint (%s): %w", fingerprint, http.PubkeyNotFoundErr)
 	}
-	return *output.KeyPairs[0].KeyName, nil
+
+	k := &clients.EC2KeyPairInfo{
+		KeyPairId:      output.KeyPairs[0].KeyPairId,
+		KeyName:        output.KeyPairs[0].KeyName,
+		KeyFingerprint: output.KeyPairs[0].KeyFingerprint,
+	}
+	for _, t := range output.KeyPairs[0].Tags {
+		k.Tags = append(k.Tags, clients.EC2Tag{Key: t.Key, Value: t.Value})
+	}
+	return k, nil
 }
 
 func (c *ec2Client) DeleteSSHKey(ctx context.Context, handle string) error {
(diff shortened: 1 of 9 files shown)


ezr-ondrej:
@lzap this is what we need to do and the Tag will still not be updated, this will fix just the `Handle` updates.
I'd be ok with this going in (after review), but keeping out the tag updates, as it's adding a migration for table we want to drop...

Another option would be to drop the Resource completely.

lzap (changes requested):
I have one concern in exposing AWS SDK internals within the application. I think in this case it is worth creating a new type.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -190,14 +190,14 @@ func (c *ec2Client) GetPubkeyName(ctx context.Context, fingerprint string) (stri
...
 	}
 
 	if len(output.KeyPairs) == 0 {
 		span.SetStatus(codes.Error, fmt.Sprintf("no KeyPair with fingerpring (%s) found", fingerprint))
-		return "", fmt.Errorf("SSH key not found by its fingerprint (%s): %w", fingerprint, http.PubkeyNotFoundErr)
+		return nil, fmt.Errorf("SSH key not found by its fingerprint (%s): %w", fingerprint, http.PubkeyNotFoundErr)
 	}
-	return *output.KeyPairs[0].KeyName, nil
+	return &output.KeyPairs[0], nil
 }
```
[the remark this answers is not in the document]
ezr-ondrej: Yeah, I agree, happy to do so if we decide so, but first I'd like to be sure we want to keep the Pubkey Resource, so it's worth the effort. And for now I got the feeling it is actually not worth it ;)

ezr-ondrej:
> I have no problem dropping the resource, in the light of new findings I think it is overengineered.

Agreed, tho there is the single looming problem: what to do about the keys we've created?
My idea would be: add static tag instead of the generated one marking the pubkey as created from our app.
That would allow us query AWS on deletion of pubkey by "fingerprint AND is tagged with our special tag" and delete those (if required at some point).

lzap:
Sure, let's do a static tag, maybe a dynamic (with storing it only in logs) and let's remove PKR completely to simplify things.

lzap:
Before we remove PKR, let's fix fingerprints because currently it is a bit messy - we should decide if we want to pre-calculate or generate them on the fly. Both approaches will require some refactoring.

internal/clients/stubs/ec2_stub.go
```diff
@@ -46,29 +46,31 @@ func (mock *EC2ClientStub) Status(ctx context.Context) error {
 }
 
 func (mock *EC2ClientStub) ImportPubkey(ctx context.Context, key *models.Pubkey, tag string) (string, error) {
-	ec2KeyID := fmt.Sprintf("key-%d", len(mock.Imported))
+	ec2KeyID := fmt.Sprintf("key-%d", len(mock.Imported)+1)
 	keyName := key.Name // copy the name
+	var tags []types.Tag
+	if tag != "" {
+		tags = append(tags, types.Tag{Key: ptr.To("rhhc:id"), Value: &tag})
```
adiabramovitch, line 53: I remember lzap explained it is better to initialize an array to a certain length and maybe add capacity. 
is it possible here?

internal/dao/stubs/pubkey_dao.go
```diff
@@ -64,9 +65,11 @@ func (stub *pubkeyDaoStub) Update(ctx context.Context, pubkey *models.Pubkey) er
 }
 
 func (stub *pubkeyDaoStub) GetById(ctx context.Context, id int64) (*models.Pubkey, error) {
+	var resultPk models.Pubkey
 	for _, pk := range stub.store {
 		if pk.AccountID == ctxAccountId(ctx) && pk.ID == id {
-			return pk, nil
+			resultPk = *pk // shallow copy of the struct
+			return &resultPk, nil
```
adiabramovitch, line 72: Why do you do that? Do you want to prevent change in the actual AWS result?

adiabramovitch, line 72: Why do you do that?

internal/jobs/ensure_pubkey_on_aws_job_test.go
```diff
@@ -112,6 +112,8 @@ func TestHandleEnsurePubkeyOnAWS(t *testing.T) {
 		pkrList, err := pkDao.UnscopedListResourcesByPubkeyId(ctx, pk.ID)
 		require.NoError(t, err)
 		assert.Equal(t, 1, len(pkrList))
+		assert.Equal(t, "awsPreviouslyImportedTag", pkrList[0].Tag)
+		assert.Equal(t, "key-1", pkrList[0].Handle)
```
adiabramovitch, line 116: I don't see "key-1" in the code can you please explain? :)

ezr-ondrej: the `key-1` is automatically calculated as a serie in the stub, but might be better not to rely on it :thinking:

lzap:
Linter issue, is this ready for re-review?

ezr-ondrej:
I've rebased it, but there is one issue if merging this:
We would delete any key, even if we did not created it.

internal/dao/stubs/pubkey_dao.go
lzap: This is a shallow copy: https://go.dev/play/p/0aatJFCKdnW

ezr-ondrej: because otherwise during the test we would change the same structure as we are adding to the array.
So you add stubbed pubkey with name "Pubkey", run test that is expected to change the name, but the name is changes to the initial structure in the test also.
Which is unexpected behaviour IMHO.

ezr-ondrej:
Not needed I guess, I was not able to get back to this, we will see once the app is actually used.

PR #343, opened 2022-12-15
fix(HMSPROV-390): calculate AWS fingerprint

ezr-ondrej:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_KeyPairInfo.html talks about what to expect as fingerprint format, which is different then what we store in some cases.

### DIFF
--- a/internal/models/pubkey_model.go
+++ b/internal/models/pubkey_model.go
@@ -1,5 +1,16 @@
 package models
 
+import (
+	"crypto/md5" //#nosec
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/hex"
+	"fmt"
+	"strings"
+
+	"golang.org/x/crypto/ssh"
+)
+
 // Pubkey represents SSH public key that can be deployed to clients.
 type Pubkey struct {
 	// Required auto-generated PK.
@@ -16,4 +27,40 @@ type Pubkey struct {
 
 	// Public key SHA256 fingerprint. Generated (read-only).
 	Fingerprint string `db:"fingerprint" json:"fingerprint"`
+}
+
+// FingerprintAWS calculates fingerprint used by AWS.
+// AWS calculates MD5 sums differently then all the other tools and use format DER as the base for the sub.
+// see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-keys.html for more details.
+func (pk *Pubkey) FingerprintAWS() (string, error) {
+	pkey, _, _, _, err := ssh.ParseAuthorizedKey([]byte(pk.Body))
+	if err != nil {
+		return "", fmt.Errorf("failed to calculate fingerprint for %s: %w", pk.Name, err)
+	}
+	// when rsa key
+	if pkey.Type() == "ssh-rsa" {
+		return fingerprintRsaAWS(pkey)
+	}
+	// when ed25519 key
+	// this is the same as what we store, but better be sure here
+	return strings.TrimLeft(ssh.FingerprintSHA256(pkey), "SHA256:") + "=", nil
+}
+
+// fingerprintRsaAWS calculates fingerprint for rsa keys
+func fingerprintRsaAWS(key ssh.PublicKey) (string, error) {
+	parsedCryptoKey := key.(ssh.CryptoPublicKey)
+
+	// Finally, we can convert back to an *rsa.PublicKey
+	pub := parsedCryptoKey.CryptoPublicKey().(*rsa.PublicKey)
+
+	der, err := x509.MarshalPKIXPublicKey(pub)
+	if err != nil {
+		return "", fmt.Errorf("failed to calculate rsa key fingerprint: %w", err)
+	}
+	md5sum := md5.Sum(der) //#nosec
+	hexarray := make([]string, len(md5sum))
+	for i, c := range md5sum {
+		hexarray[i] = hex.EncodeToString([]byte{c})
+	}
+	return strings.Join(hexarray, ":"), nil
 }
--- a/internal/clients/stubs/ec2_stub.go
+++ b/internal/clients/stubs/ec2_stub.go
@@ -29,11 +29,26 @@ func WithEC2Client(parent context.Context) context.Context {
 	return ctx
 }
 
+func AddStubbedEC2KeyPair(ctx context.Context, info *types.KeyPairInfo) error {
+	si, err := getEC2StubFromContext(ctx)
+	if err != nil {
+		return err
+	}
+	si.Imported = append(si.Imported, info)
+	return nil
+}
+
 func newEC2ServiceClientStubWithRegion(ctx context.Context, region string) (clients.EC2, error) {
 	return nil, nil
 }
 
 func newEC2CustomerClientStubWithRegion(ctx context.Context, _ *clients.Authentication, _ string) (si clients.EC2, err error) {
+	return getEC2StubFromContext(ctx)
+}
+
+func getEC2StubFromContext(ctx context.Context) (*EC2ClientStub, error) {
+	var si *EC2ClientStub
+	var err error
 	var ok bool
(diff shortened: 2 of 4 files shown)


ezr-ondrej:
This does not fix the issue :face_exhaling: aws is calculating the fingerprint from DER files and on top of that they don't care for duplicate fingerprints :)

ezr-ondrej:
Going to merge, I hope the test coverage is sufficient, happy to work on any code improvements in follow-up @lzap @adiabramovitch, but I'd love the app working over the next week in stage :)

lzap:
Let's not rush things please :-)

So first of, let me summarize. EC2 currently supports RSA and ED keys, I have both. Here are the public keys and their SHA256 and MD5 fingerprints:

```
% ssh-keygen -l -f id_rsa.pub 
2048 SHA256:ENShRe/0uDLSw9c+7tc9PxkD/p4blyB/DTgBSIyTAJY lzap+rsakey@redhat.com (RSA)
% ssh-keygen -l -f id_ed25519.pub 
256 SHA256:gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk lzap@box.home.lan (ED25519)

% ssh-keygen -l -E md5 -f id_rsa.pub
2048 MD5:89:c5:99:b5:33:48:1c:84:be:da:cb:97:45:b0:4a:ee lzap+rsakey@redhat.com (RSA)
% ssh-keygen -l -E md5 -f id_ed25519.pub 
256 MD5:ee:f1:d4:62:99:ab:17:d9:3b:00:66:62:32:b2:55:9e lzap@box.home.lan (ED25519)
```

After I uploaded them both into my account, I see the following fingerprints created by EC2:

* RSA: `c4:ba:72:45:16:a9:2c:39:c3:99:8d:e7:16:01:9c:77`
* ED: `gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk=`

So it looks like ED does match, however, RSA key fingerprint does NOT match for some reason.

lzap:
Holy crap, from AWS docs:

If you used [CreateKeyPair](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateKeyPair.html) to create the key pair:

* For RSA key pairs, the key fingerprint is the SHA-1 digest of the DER encoded private key.
* For ED25519 key pairs, the key fingerprint is the base64-encoded SHA-256 digest, which is the default for OpenSSH, starting with [OpenSSH 6.8](http://www.openssh.com/txt/release-6.8).

If you used [ImportKeyPair](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ImportKeyPair.html) to provide AWS the public key:

* For RSA key pairs, the key fingerprint is the MD5 public key fingerprint as specified in section 4 of RFC4716.
* For ED25519 key pairs, the key fingerprint is the base64-encoded SHA-256 digest, which is the default for OpenSSH, starting with [OpenSSH 6.8](http://www.openssh.com/txt/release-6.8).

However, I have never created a key on AWS, I have always imported my keys, yet, the fingerprint does not match.

lzap:
And quick google finds:

```
$ ssh-keygen -e -f id_rsa.pub -m pkcs8 | openssl pkey -pubin -outform der | openssl md5 -c
MD5(stdin)= c4:ba:72:45:16:a9:2c:39:c3:99:8d:e7:16:01:9c:77
```

This does indeed match my RSA key. Ok I am starting to understand what is going on here.

lzap:
Ok I think we can do better. Currently the idea was to generate fingerprint(s) at the save time and store this in the db, we currently only have one [Fingerprint](https://github.com/RHEnVision/provisioning-backend/blob/main/internal/models/pubkey_model.go#L29) column. I sort of expected that the fingerprint is the same across SSH implementations, but it looks like there is MD5 still in use and SHA256.

This patch ignores the currently stored SHA256 fingerprint and adds new code that calculates MD5 (legacy) fingerprint on the fly. I suggest to take the same approach and do this via validation/transformation hook renaming and creating a new column:

* Pubkey.FingerprintLegacy - calculated by https://pkg.go.dev/golang.org/x/crypto/ssh#FingerprintLegacyMD5
* Pubkey.Fingerprint - https://pkg.go.dev/golang.org/x/crypto/ssh#FingerprintSHA256 (already exists)
* Pubkey.FingerprintLegacyDER - calculated via the new function from this patch (see the command above)

Now we could do this on the fly too, in that case we should probably remove the existing column. But my guts tell me it is better to precalculate and store those fields because keys cannot be updated, fingerprints do not change and some crypto operations can drain the crypto random seed to the point where it could possibly block and slow the app down on some systems (perhaps not on modern Linuxes tho).

internal/models/pubkey_model.go
```diff
@@ -17,3 +28,39 @@ type Pubkey struct {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to calculate rsa key fingerprint: %w", err)
+	}
+	md5sum := md5.Sum(der) //#nosec
+	hexarray := make([]string, len(md5sum))
+	for i, c := range md5sum {
+		hexarray[i] = hex.EncodeToString([]byte{c})
+	}
+	return strings.Join(hexarray, ":"), nil
+}
```
lzap, line 66: Adding methods into models is something that I would really encourage us to avoid completely. Maybe for very tiny functions where it is obvious these are some kind of field-level small things. This is already very alarming because we now know there are at lest three possible SSH fingerprints (legacy, legacy-DER, SHA) and this can get our of control very quickly.

If we want to go through the "on the fly" generation, then I would rather see some sort of interface like:

```go
type Fingerprinter interface {
 Calculate(pubkey model.Pubkey) (string, error)
}
```

Or something similar. I feel like precalculating those and storing them in the DB is more scalable solution, tho, I realize there is some extra effort in this (we need a migration with existing keys recalculation which is a precedent on this project).

lzap:
One other thing why I prefer pre-calculating fingerprints - this allows us to enforce database integrity on fingerprint columns so no duplicate keys are ever stored.

Now, we could actually store fingerprint in a TEXT column and maybe keep some sort of FINGERPRINT_TYPE column because very likely a single key will only have a single fingerprint (either DER-legacy or SHA256 for AWS). But we could add another hyperscaler later on that could require let's say the legacy (non-DER) fingerprint for RSA keys which are not going anywhere and we would have a problem. Therefore I think it is more straightfotward to simply create three separate NULLable columns with three fingerprints on save, since we do it only once (during key upload) it is not big deal and we will have them forever enforcing integrity on the account level across all fingerprints.

internal/models/pubkey_model.go
```diff
@@ -17,3 +28,39 @@ type Pubkey struct {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to calculate fingerprint for %s: %w", pk.Name, err)
+	}
+	// when rsa key
+	if pkey.Type() == "ssh-rsa" {
+		return fingerprintRsaAWS(pkey)
+	}
+	// when ed25519 key
+	// this is the same as what we store, but better be sure here
+	return strings.TrimLeft(ssh.FingerprintSHA256(pkey), "SHA256:") + "=", nil
```
lzap, line 46: Are sou sure the padding character is always one? I mean, the hash should be constant length, I am trying to think about situation where it could actually be shorter. I am just thinking loud, this smells.

lzap:
Created a pre-calculation solution, it is a draft for now because it requires move of migrations outside of the db package: https://github.com/RHEnVision/provisioning-backend/pull/346 let me know what you think @ezr-ondrej

internal/models/pubkey_model.go
ezr-ondrej: Yeah, I agree proper calculation would be better here, I THINK it is always one from what I've read and calculated on paper, but I agree that proper calculation in code would be more appropriate here.

PR #344, opened 2022-12-15
fix(HMSPROV-352): error out jobs early

lzap:
Because the job queue implementation is generic, there is no support for cancelling dependant jobs when some fail.

For this reason, jobs must explicitly check associated Reservation field Error for non-empty value using checkExistingError helper function. If it returns an error, the job MUST immediately exit prepending "step skipped: " to the error message.

This way, the actual error message will be prefixed by N prefixes for all remaining jobs in the queue. For example, for 3 jobs when the first fails it is: "step skipped: step skipped: actual error text".

The job code currently always finishes all dependant jobs, so 3 out of 3 jobs will be done. In other words, no matter if queue of dependant jobs will succeed or fail, the counter will be equal to number of steps.

The patch also adds a new array of strings which hold a title that can be presented in a nice way in the UI. We currently do not show full list of all steps involved.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/jackc/pgx/v5 v5.0.4
 	github.com/jackc/tern/v2 v2.0.0-beta.3
 	github.com/lzap/cloudwatchwriter2 v1.0.0
-	github.com/lzap/dejq v1.1.0
+	github.com/lzap/dejq v1.2.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.13.1
 	github.com/redhatinsights/app-common-go v1.6.4
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -229,6 +229,12 @@
             "format": "int32",
             "type": "integer"
           },
+          "step_titles": {
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
+          },
           "steps": {
             "format": "int32",
             "type": "integer"
@@ -244,7 +250,9 @@
           "build_time": {
             "type": "string"
           },
-          "error": {},
+          "error": {
+            "type": "string"
+          },
           "msg": {
             "type": "string"
           },
--- a/api/openapi.gen.yaml
+++ b/api/openapi.gen.yaml
@@ -464,6 +464,10 @@ components:
         step:
           format: int32
           type: integer
+        step_titles:
+          items:
+            type: string
+          type: array
         steps:
           format: int32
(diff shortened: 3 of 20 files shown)


lzap:
We only have one reservation: AWS launch. It consists of two jobs:

* Upload public key
* Launch instance(s)

So the patch adds a code into the AWS job to check if previous job has failed, if that's the case it does not continue and skips. The error message is updated to: `step skipped: actual error text`. It is a bit dumb solution, but since folks did not like introducing slice of errors, this is the only good solution I can figure out.

A big integration test suite (4 tests in total) was added which completely test reservations and job execution against memory dejq implementation and postgres.

QA: This should fix the use case when ssh key upload fails for any reason, the expected REST response is now:

* StepTitles: ["Upload pubkey", "Launch instance(s)"]
* Steps: 2
* Step: 2
* Error: "step skipped: actual error text"
* Status: "Launched instance(s)"
* Success: false

The Status field capures current state rather than represents a status message, we should probably rename it later on, it can be confusing.

Also contains one small change: Error (in JSON payload OpenSPEC) used to be empty interface (error interface in Go), now it is a proper string which is easier to work with. Client will need to be regenerated.

ezr-ondrej (approved):
Code LGTM, I'd love if the step could not be run at all, but this is definitelly good enough and maybe even better :+1:

I didn't have time to test this, so will let at least someone else to have a look.
I'd love to test (can be done once merged - putting it to my todo): create a pubkey with name that already exists on AWS, but is a different key (fingerprint). As that should currently error out in the first step and thus I'd expect not to continue.

lzap:
One important thing: previously "memory" dejq implementation, which is the default in development mode, was panicking on error. I thought this is a good idea so we immediately see the error in the IDE/Editor, however, now that we need to work with jobs which do have errors, this is no longer the case and this patch needs newer version of the library:

* https://github.com/lzap/dejq/commit/c4b97f358b14f21184773e84d6e0ea1c52a1ec68

Make sure to do 'go get github.com/lzap/dejq' when testing this PR. I think "go build" should do this automatically.

go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/jackc/pgx/v5 v5.0.4
 	github.com/jackc/tern/v2 v2.0.0-beta.3
 	github.com/lzap/cloudwatchwriter2 v1.0.0
-	github.com/lzap/dejq v1.1.0
+	github.com/lzap/dejq v1.2.0
```
lzap, line 37: This version is needed now.

PR #346, opened 2022-12-20
fix(HMSPROV-390): code migrations and legacy fingerprint

lzap:
The last patch turned the fingerprint calculations into "on-the-fly" mode, however, the old fingerprint is still left in the database. This patch turns that around and elaborates the initial idea of precalculating fingerprints and also adds key type column.

This patch is a precedent - we need to perform a Go code migration. Luckily, tern does support OnStart Go callbacks which can be utilized for that.

Unfortunately, the migration code lives in the "db" package and this creates a cycle dependency (db-migration-dao-db) and therefore migrations needs to be moved outside the "db" package in order to prevent that.

This patch is a preliminary version - it currently does not work because refactoring (move) of migrations from internal/db/migrations to internal/migrations is needed in order to prevent Go import dependency cycle. I will only do this if we agree this is the direction we want to take.


internal/dao/dao_service_interface.go
```diff
@@ -0,0 +1,13 @@
...
+	"context"
+)
+
+var GetServiceDao func(ctx context.Context) ServiceDao
+
+// ServiceDao is used for service operations like migrations. All operations are UNSCOPED.
+// See pgx/service_pgx.go for documentation.
+type ServiceDao interface {
+	RecalculatePubkeyFingerprints(ctx context.Context) error
+}
```
lzap, line 13: This is a new special DAO only used in migrations. This allows us to write stub tests if we want to.

internal/db/migration_callbacks.go
```diff
@@ -0,0 +1,33 @@
...
+
+	return nil
+}
+
+func updateFingerprint(ctx context.Context) error {
+	// TODO creates a cyclic dependency - migration code must be moved outside of db
+	//pkd := dao.GetServiceDao(ctx)
+	//return pkd.RecalculatePubkeyFingerprints(ctx)
+	return nil
+}
```
lzap, line 33: The idea is to create internal/migrations/go and internal/migrations/sql packages where both SQL and Go code would live.

internal/models/pubkey_model.go
```diff
@@ -25,42 +14,36 @@ type Pubkey struct {
...
+	// such fingerprint: ssh-keygen -l -E md5 -f $HOME/.ssh/key.pub
+	// Example: "89:c5:99:b5:33:48:1c:84:be:da:cb:97:45:b0:4a:ee"
+	FingerprintLegacy string `db:"fingerprint_legacy" json:"fingerprint_legacy" validate:"len=48"`
+
+	// MD5 fingerprint stored as hexadecimal with colons without any prefix from key in PEM format.
+	// This format is specific to AWS. To generate such fingerprint:
+	// ssh-keygen -e -f $HOME/.ssh/key.pub -m pkcs8 | openssl pkey -pubin -outform der | openssl md5 -c
+	// Example: "c4:ba:72:45:16:a9:2c:39:c3:99:8d:e7:16:01:9c:77"
+	FingerprintPEM string `db:"fingerprint_pem" json:"fingerprint_pem" validate:"len=48"`
 }
```
lzap, line 36: This is the gist of the patch, three new columns:

* type
* legacy fingerprint (MD5 OpenSSH)
* PEM fingerprint (MD5 from PEM - speciality of AWS perhaps)

And a helper method that selects the proper fingerprint depending on type for AWS.

internal/models/transformer.go
```diff
@@ -25,15 +28,33 @@ func Transform(ctx context.Context, model interface{}) error {
...
+	md5sum := md5.Sum(der) //#nosec
+	// TODO optimize this
+	hexarray := make([]string, len(md5sum))
+	for i, c := range md5sum {
+		hexarray[i] = hex.EncodeToString([]byte{c})
+	}
+	result := strings.Join(hexarray, ":")
+
+	pk.FingerprintPEM = result
+
```
lzap, line 57: The code @ezr-ondrej wrote was moved here essentially, where it belongs if we want to go the pre-calculated path. This is where calculation of the original fingerprint was.

internal/models/transformer_test.go
```diff
@@ -7,15 +7,82 @@ import (
...
+
+	for _, td := range tests {
+		t.Run(td.name, func(t *testing.T) {
+			err := Transform(context.Background(), &td.pubkey)
+			assert.NoError(t, err)
+			assert.Equal(t, td.fingerprint, td.pubkey.FingerprintPEM,
+				"%s fingerprint %s does not match %s", td.name, td.pubkey.FingerprintPEM, td.fingerprint)
+		})
 	}
 }
```
lzap, line 88: This is a complete test of all three methods on both supported SSH key types. I tried also to add DSA keys but they do not appear to be supported by the Go crypto library so I think we will only support the two.

ezr-ondrej (commented):
Thanks for looking into this @lzap !

I agree that caching the fingerprinte might be a good idea, but I'd first like to see some data supporting this and discussion what formats we want to keep. AWS case shows us there is many fingerprint possibilities and I do not see much value in adding ALL we encounter.

There was a thought behind calculating the AWS on the fly - lets calculate on the fly until we figure what fingerprints we actually want. I'd love to keep a single `Fingerprint` with the same calculation for our database integrity and others being just a helper cache that does not really hold any functional value on our side and thus being nullable.

The migration callbacks are quite interesting concept, nice addition :+1:
Tho this should preceed a discussion about migrations being backwards compatible (meaning deployment that runs code version `n-1` is able to work with database version `n`). Which this likely breaks if I'm correct, the old code would not create these new fingerprints.

lzap:
> some data supporting this and discussion what formats we want to keep

Well it looks like SHA256 and MD5 the way OpenSSH does are the de-facto strandard, this is what Go stdlib ships with. Plus AWS has its own way of doing that. I can't tell this is all, but it is possible to add more.

> lets calculate on the fly until we figure what fingerprints we actually want

My argument would be: now we know, let's do it :-) What really bothers me is that we calculate one fingerprint and others do on the fly which feels wrong. Why to have two code places where we do this stuff, it should have been a single place. If we end up not doing precalculation, which is definitely an option, let's at least move all the fingerprint code into a separate package.

> deployment that runs code version n-1 is able to work with database version n

I am not sure what you mean by this, if you add a column via migration, there is no way that N-1 version would work, at least in our SQL scanning code.  It will be a runtime error with "unable to scan column XYZ".

lzap:
What to do with this one?

Option A is to close this PR, take the migration part into separate PR and improve the existing on-the-fly calculation to avoid doing business logic in the model: https://github.com/RHEnVision/provisioning-backend/pull/343#discussion_r1052112884 also the key type detection is a useful code too (we want to know the type in case other clouds do not support some).

Option B is to finish this PR and have fingerprints pre-calculated. I prefer this one because I think detection of key (RSA vs ED) is imporant to do anyway (in case some hyperscalers deprecate one type in the future). We currently do not do this and blindly store pubkey in the database.

I understand the argument that we do not know what comes next, a future hyperscaler might need SHA1 or something like that. While I see this as unlikely, I like the idea of doing our first code migration now before we are in production. So how about a compromise:

Option C: Update this patch to only precalculate SHA256 (existing column) and MD5 (OpenSSH RSA standard fingerprint). We would calculate Amazon PEM RSA on the fly because it feels special. I would still refactor the code so calculation is not done in the model package. @ezr-ondrej ?

lzap:
So I finished the PR (option A) with all fingerprints precalculated first, I had to move migrations out of the `db` package to avoid import cycle. I also improved integration tests, moved some functions into `factories` package so it can be reused across all tests. There is another integration package test now - migration code tests which tests the code migration.

The patch introduces the very first code migration, it is implemented as a "before migration callback", the only feature tern library have that allows this. Luckily, this is executed within the Postgres advisory lock, so it works as expected. I introduced new "service DAO" interface with migration function.

Please take a look, if you want to move PEM fingerprint to be calculated on the fly, I can rebase that. Also would appreciated testing from @adiabramovitch or @amirfefer - the migration needs to be tested as we only have "one try".

lzap:
FYI I have improved the legacy fingerprint calculation by 50%, then I realized that a similar code is also in Go x/crypto ;-) so I am sending a patch to Go:

https://go-review.googlesource.com/c/crypto/+/460675

ezr-ondrej:
After an offline discussion we have decided to go with the option B and calculate standard Fingerprints for the keys to save in DB.

lzap:
Big refactoring and rebase is up and ready for review.

I had to solve one problem tho, since PEM AWS fingerprint is now calculated on the fly, it can fail when invalid key type is stored (e.g. DSA) which is a problem - it could stop a job and fail it.

Therefore, as part of validation/transformation, the AWS PEM key is parsed and error is reported during save. This is done in transformation, because validations API does not support custom messages.

ezr-ondrej:
I guess the fixtures for pubkeys in IQE are now invalid? :thinking:

lzap:
Talked to QE guys and looks like new field can break tests. A MR was merged, thank you!

/retest

internal/ssh/fingerprint.go
```diff
@@ -0,0 +1,82 @@
...
+	fps := OpenSSHFingerprints{}
+
+	pkey, cmt, _, _, err := ssh.ParseAuthorizedKey(pubkeyBody)
+	if err != nil {
+		return OpenSSHFingerprints{}, fmt.Errorf("unable to parse public key %s: %w", pubkeyBody, err)
+	}
+
+	fps.Comment = cmt
+	fps.Type = pkey.Type()
+	fps.SHA256 = strings.TrimLeft(ssh.FingerprintSHA256(pkey), "SHA256:") + "="
```
akhil-jha, line 55: Why is `=`  added here?

lzap:
It is green!

internal/ssh/fingerprint.go
ezr-ondrej: that's just a padding `=` is an empty character included to keep a constant length, which usually is there in fingerprints, but Golang library is not including it for some reason.

lzap: Base64 can be either padded or not padded, AWS requires padding.

https://linuxhint.com/understand-base64-padding/

ezr-ondrej (approved):
I'm good here.
@akhil-jha good to merge? :)

akhil-jha (approved):

internal/ssh/fingerprint.go
mshriver: Thanks @ezr-ondrej and @lzap, confirms we're doing the 'right' thing on our end by stripping this padding when present.

PR #347, opened 2022-12-20
chore: update dependencies

lzap:
Namely cloudwatchwriter2 which now uses generics.


lzap:
Linter failed because of Go proto deprecations so I have followed https://github.com/googleapis/google-cloud-go/blob/main/migration.md as part of this patch. Only a small change.

ezr-ondrej (approved):
Nice, thanks for keeping our depts up to date! :orange_heart:

PR #348, opened 2022-12-22
feat: add account identity endpoint

croissanne:
I'm unsure if this is the right approach, but opened the PR anyway to discuss.

---

Currently it's very easy for a mismatch between user hyperscaler accounts. Users need to make sure they shared the image with the same account they configured provisioning for.  To avoid this mismatch, image builder should offer to share the image with a source rather than asking for an account number, guaranteeing  that the image was shared with the correct account.

### DIFF
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -112,6 +112,27 @@ paths:
                   $ref: '#/components/schemas/v1.SourceResponse'
         '500':
           $ref: "#/components/responses/InternalError"
+  /sources/{ID}/account_identity:
+    get:
+        parameters:
+        - in: path
+          name: ID
+          schema:
+            type: integer
+            format: int64
+          required: true
+          description: 'Source ID from Sources Database'
+        responses:
+          '200':
+            description: Return on success.
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/v1.AccountIDTypeResponse'
+          '404':
+            $ref: "#/components/responses/NotFound"
+          '500':
+            $ref: "#/components/responses/InternalError"
   /sources/{ID}/instance_types:
       get:
         description: 'Return a list of instance types (DEPRECATED: use /instance_types)'
--- a/internal/clients/account_identity.go
+++ b/internal/clients/account_identity.go
@@ -1,0 +1,9 @@
+package clients
+
+type AccountIdentity struct {
+	AWSDetails *AccountDetailsAWS `json:"aws,omitempty" yaml:"aws"`
+}
+
+type AccountDetailsAWS struct {
+	AccountID string `json:"account_id" yaml:"account_id"`
+}
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -27,6 +27,7 @@ const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/h
 
 type ec2Client struct {
 	ec2     *ec2.Client
+	sts     *sts.Client
 	assumed bool
 }
 
@@ -62,6 +63,10 @@ func awsConfig(ctx context.Context, region string, optFns ...func(*awsCfg.LoadOp
 }
 
 func newEC2ClientWithRegion(ctx context.Context, region string) (clients.EC2, error) {
+	if region == "" {
+		region = config.AWS.DefaultRegion
+	}
+
 	cfg, err := awsConfig(ctx, region,
 		awsCfg.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(config.AWS.Key, config.AWS.Secret, config.AWS.Session)))
 	if err != nil {
@@ -70,6 +75,7 @@ func newEC2ClientWithRegion(ctx context.Context, region string) (clients.EC2, er
 
 	return &ec2Client{
 		ec2:     ec2.NewFromConfig(*cfg),
+		sts:     sts.NewFromConfig(*cfg),
 		assumed: false,
 	}, nil
 }
@@ -99,6 +105,7 @@ func newAssumedEC2ClientWithRegion(ctx context.Context, auth *clients.Authentica
 
 	return &ec2Client{
 		ec2:     ec2.NewFromConfig(*cfg),
+		sts:     sts.NewFromConfig(*cfg),
 		assumed: true,
 	}, nil
 }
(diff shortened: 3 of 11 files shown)


internal/clients/account_identity.go
```diff
@@ -0,0 +1,9 @@
+package clients
+
+type AccountIdentity struct {
+	AWSDetails *AccountDetailsAWS `json:"aws,omitempty" yaml:"aws"`
+}
+
+type AccountDetailsAWS struct {
+	AccountID string `json:"account_id" yaml:"account_id"`
+}
```
lzap, line 9: You could also just put string directly into the payload type, we use embedding mainly for models. There is no need for intermediate type, but it is fine.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -358,3 +365,14 @@ func (c *ec2Client) parseRunInstancesResponse(respAWS *ec2.RunInstancesOutput) [
...
+func (c *ec2Client) GetAccountId(ctx context.Context) (string, error) {
+	input := &sts.GetCallerIdentityInput{}
+
+	out, err := c.sts.GetCallerIdentity(ctx, input)
+	if err != nil {
+		return "", err
+	}
+
+	return *out.Account, nil
+}
```
lzap, line 378: Since you call it from the web process, I would ask you to add telemetry:

```go
	ctx, span := otel.Tracer(TraceName).Start(ctx, "GetAccountId")
	defer span.End()
```

So in case things are slow we know for fact it was AWS API and not our app.

internal/payloads/account_identity.go
```diff
@@ -0,0 +1,30 @@
...
+			AWSDetails: &clients.AccountDetailsAWS{
+				AccountID: awsAccountId,
+			},
+		},
+	}
+}
+
+func (s *AccountIdentityResponse) Bind(_ *http.Request) error {
+	return nil
+}
```
lzap, line 26: You only need Render for responses, Bind is only needed for Requests. So this method can be deleted I think.

lzap:
Thank you for the patch, looks great. Since STS requests should be generally pretty fast, I see no reason not to include this new endpoint for image builder, at least until full sources is implemented.

To get the commit message checker, you will need a HMSPROV issue and link it: https://github.com/RHEnVision/provisioning-backend/blob/main/CONTRIBUTING.md

Also use `make fmt` to call all our formatters and linters.

cmd/spec/path.yaml
```diff
@@ -112,6 +112,27 @@ paths:
...
+          '200':
+            description: Return on success.
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/v1.AccountIDTypeResponse'
+          '404':
+            $ref: "#/components/responses/NotFound"
+          '500':
+            $ref: "#/components/responses/InternalError"
```
lzap, line 135: Maybe it you could drop full description in the OpenAPI spec about why this is needed, who needs it and that it might actually be deprecated in the future. Let's be honest to our API users from the day one :-)

croissanne:
> Since STS requests should be generally pretty fast,

So locally it took up to 1 second or so, which to me isn't super quick, but might be acceptable. For AWS we could theoretically just grab the account id by parsing it from the arn itself. 

> To get past the commit message checker, you will need a HMSPROV issue and link it: https://github.com/RHEnVision/provisioning-backend/blob/main/CONTRIBUTING.md

Ah cool, but we should discuss with @ezr-ondrej maybe in January as well.
> 
> Also use `make fmt` to call all our formatters and linters to shut those up. :-)

Ha will do.

lzap:
Feel free to add a new cache interface here: https://github.com/RHEnVision/provisioning-backend/blob/main/internal/cache/cache_interface.go

We already do caching for accounts, this is definitely a good candidate for caching. You need to implement the interface three times:

* noop (does nothing - for tests)
* memory (memory only - development stuff)
* redis (the stage/prod implementation)

internal/clients/http/ec2/ec2_client.go
```diff
@@ -358,3 +365,14 @@ func (c *ec2Client) parseRunInstancesResponse(respAWS *ec2.RunInstancesOutput) [
 	}
 	return list
 }
+
+func (c *ec2Client) GetAccountId(ctx context.Context) (string, error) {
+	input := &sts.GetCallerIdentityInput{}
+
+	out, err := c.sts.GetCallerIdentity(ctx, input)
```
adiabramovitch, line 372: if I understood correctly this endpoint is being added in order to receive the provisioning app account ID? 
Can you please add an explanation on why is it needed for your service?
(it will be great if written in the description of the PR) I wasn't aware of it

ezr-ondrej (commented):
I'm all for having this. Implementation make sense, tho I'd rather see it under non-prefixed path or some different-private prefix (for it to be hidden from CUs and only available within the Openshift cluster)

internal/clients/account_identity.go
croissanne: I wasn't sure, since this object might contains Details for azure and stuff in future too, I thought it might be useful to be able to add fields to the details at some point too.

internal/clients/http/ec2/ec2_client.go
croissanne: Updated the PR description, hope it's enough details.

lzap (approved):
All good!

akhil-jha:
I can't test this PR because there is no pr image on quay. I think it's because @croissanne is not part of RHEnVision organization. 
@lzap / @mshriver Can you add new members? Ondrej is on PTO.

Also, can you let me know the steps for testing this PR? 
I think I need to create an image prior to testing the PR? Creating image takes time, and I don't want to do a blackbox testing here. :sweat:

lzap:
@croissanne Service accounts are supposed to have some STS permissions, yes, not all. Here is our docs: https://github.com/RHEnVision/provisioning-backend/blob/main/docs/configure-amazon-role.md#service-account

@akhil-jha taking a look to add @croissanne to our GH org

lzap:
So it looks like the osbuild team has a special group called "friends of osbuild". I like that, I will create a similar group and add @croissanne in there!

https://github.com/orgs/osbuild/teams

croissanne:
> @croissanne Service accounts are supposed to have some STS permissions, yes, not all. Here is our docs: https://github.com/RHEnVision/provisioning-backend/blob/main/docs/configure-amazon-role.md#service-account

We should actually be ok, https://docs.aws.amazon.com/cli/latest/reference/sts/get-caller-identity.html requires no permissions, so as long as we can assume the role, we're fine.

lzap:
Great, you should have an invite into our "Friends" team, welcome :-)

akhil-jha:
@croissanne can you make the scope from private to public? https://github.com/orgs/RHEnVision/people

lzap:
What's that?

Snímek obrazovky 2023-01-09 v 10 41 40

lzap:
Oh it was not on me, I see public now on @croissanne membership.

lzap:
All green!

akhil-jha (approved):

akhil-jha:
@croissanne , can you rebase this? There were few PRs that were merged. We can merge after the pr_check is green.

ezr-ondrej (approved):
I see all :green_heart: so let me do the honours and :shipit:

Thanks @croissanne for implementing and @lzap, @akhil-jha and @adiabramovitch for moving this forward and figuring out all the hiccups! Great work y'all :orange_heart: :tada:

PR #349, opened 2022-12-22
feat(HMSPROV-177): Add total availability check request metric

adiabramovitch:


cmd/pbstatuser/main.go
```diff
@@ -41,7 +42,20 @@ type SourceInfo struct {
...
+const (
+	metricNameSourceAvailabilityCheckRequestTotal = "source_availability_check_request_total"
+)
+
 var (
+	totalAvailabilityCheckReqs = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name:        metricNameSourceAvailabilityCheckRequestTotal,
+			Help:        "availability Check requests count partitioned by type (aws/gcp/azure), source status and error",
+			ConstLabels: prometheus.Labels{"service": "provisioning"},
```
lzap, line 54: I suggest to also add another static label "component": "statuser" so we can tell which metrics are coming out from which components.

cmd/pbstatuser/main.go
```diff
@@ -41,7 +42,20 @@ type SourceInfo struct {
...
+		},
+		[]string{"type", "status", "error"},
+	)
+
```
lzap, line 58: I think it would be a good idea to introduce new package internal/metrics where we would have all the metrics as exported variables.

cmd/pbstatuser/main.go
```diff
@@ -112,6 +126,7 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 		// TODO: check if source is avavliable - WIP
 		sr.Status = kafka.StatusAvaliable
 		chSend <- sr
+		totalAvailabilityCheckReqs.WithLabelValues("azure", string(sr.Status), "").Inc()
```
lzap, line 129: The newly created package could expose functions which are typed and actually hide the implementation which is Prometheus today: `func IncTotalAvailablilityCheckReqs(string, StatusType, error)` for example. It would have been cleaner than exposing bare bone Prometheus metric where the caller actually must know all the labels.

cmd/pbstatuser/main.go
```diff
@@ -131,9 +146,11 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 			sr.Err = err
 			logger.Warn().Msgf("Could not get aws assumed client %s", err)
 			chSend <- sr
+			totalAvailabilityCheckReqs.WithLabelValues("aws", string(sr.Status), err.Error()).Inc()
 		} else {
 			sr.Status = kafka.StatusAvaliable
 			chSend <- sr
+			totalAvailabilityCheckReqs.WithLabelValues("aws", string(sr.Status), "").Inc()
```
lzap, line 153: Can't you call this just once passing `err` to the method? Hopefully the Prom client will not panic if error is `nil`?

lzap (changes requested):
It is okay, however, I suggest to introduce a new package dedicated to metrics so we won't have Prometheus library spread all over the application, what you think?

It makes a lot of sense, while the platform will probably not change from Prometheus, we might consider using OpenTelemetry Metrics API instead of direct Prometheus API:

https://opentelemetry.io/docs/reference/specification/metrics/

cmd/pbstatuser/main.go
adiabramovitch: When the error is nil I am getting a null pointer exception, I'll see if there is maybe a different usage that can fit.

cmd/pbstatuser/main.go
adiabramovitch: great idea, thank you!

cmd/pbstatuser/main.go
adiabramovitch: It makes sense why we need to differentiate between the cases. 
first of all the label parameters are of type string (error type is not possible). 
second,  when
 ```golang
 err != nil
```
we have an error thus a  string error when calling err.Error,  but when we do not have an error meaning 
```golang
 err = nil
``` 
We are calling nil.Error() which will obviously causes a null pointer exception.

lzap: What I mean was to use `error` type for the argument, therefore you do not need to call `err.Error()` at all.

cmd/pbstatuser/main.go
```diff
@@ -162,9 +167,11 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
 			sr.Err = err
 			logger.Warn().Msgf("Could not list gcp regions %s", err)
 			chSend <- sr
+			metrics.IncTotalAvailablilityCheckReqs("gcp", string(sr.Status), "statuser", err.Error())
 		} else {
 			sr.Status = kafka.StatusAvaliable
 			chSend <- sr
+			metrics.IncTotalAvailablilityCheckReqs("gcp", string(sr.Status), "statuser", "")
```
lzap, line 174: These conversions to string are ugly, keep the types as long as possible and do the conversion in the function itself. What I mean is use the proper type for the second argument of the function that will be definitely cleaner.

lzap (changes requested):
I would ask you to further improve the function types, but other than that it looks great!

lzap (approved):

internal/metrics/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+		Help:        "availability Check requests count partitioned by type (aws/gcp/azure), source status, component and error",
+		ConstLabels: prometheus.Labels{"service": "provisioning"},
+	},
+	[]string{"type", "status", "component", "error"},
+)
+
+func IncTotalAvailablilityCheckReqs(provider models.ProviderType, component string, StatusType kafka.StatusType, err error) {
+	errString := ""
+	if err != nil {
+		errString = err.Error()
```
ezr-ondrej, line 25: I'd be fine if this is just `"error"` for now, so we can identify errored requests, we can improve to best practice labeling in follow ups :)

adiabramovitch: So for first version is it ok to keep false/true as I mentioned on slack?

ezr-ondrej: yup :)

ezr-ondrej (approved):
Nice! Thanks @adiabramovitch ! :+1:

lzap:
I am testing this and it does not work, I do not see the metric when I start pbstatuser. Anything I am missing?

ezr-ondrej:
> Edit: Found it, was uninitialized metric (no value).

Yeah I assumed it is normal we need to have some value for the metric (first run of the statuser) for it to appear. Or shall we initiate a value for it?

lzap:
That is fine, we cannot initiate values it would skew the aggregated values so actually it makes sense.

PR #350, opened 2022-12-22
feat(HMSPROV-177): Add availability check request duration metric

adiabramovitch:
Depends on https://github.com/RHEnVision/provisioning-backend/pull/349

### DIFF
--- a/cmd/pbstatuser/main.go
+++ b/cmd/pbstatuser/main.go
@@ -108,15 +108,20 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 	defer processingWG.Done()
 
 	for s := range chAzure {
-		sr := kafka.SourceResult{
-			SourceID:     s.SourceID,
-			Identity:     s.Identity,
-			ResourceType: "Source",
-		}
-		// TODO: check if source is avavliable - WIP
-		sr.Status = kafka.StatusAvaliable
-		chSend <- sr
-		metrics.IncTotalAvailabilityCheckReqs(models.ProviderTypeAzure, "statuser", sr.Status, nil)
+		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeAzure, func() error {
+			var err error
+			sr := kafka.SourceResult{
+				SourceID:     s.SourceID,
+				Identity:     s.Identity,
+				ResourceType: "Source",
+			}
+			// TODO: check if source is avavliable - WIP
+			sr.Status = kafka.StatusAvaliable
+			chSend <- sr
+			metrics.IncTotalAvailabilityCheckReqs(models.ProviderTypeAzure, sr.Status, nil)
+
+			return fmt.Errorf("error during check %w:", err)
+		})
 	}
 }
 
@@ -125,22 +130,26 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 	defer processingWG.Done()
 
 	for s := range chAws {
-		sr := kafka.SourceResult{
-			SourceID:     s.SourceID,
-			Identity:     s.Identity,
-			ResourceType: "Source",
-		}
-		_, err := clients.GetEC2Client(ctx, &s.Authentication, "")
-		if err != nil {
-			sr.Status = kafka.StatusUnavailable
-			sr.Err = err
-			logger.Warn().Msgf("Could not get aws assumed client %s", err)
-			chSend <- sr
-		} else {
-			sr.Status = kafka.StatusAvaliable
-			chSend <- sr
-		}
-		metrics.IncTotalAvailabilityCheckReqs(models.ProviderTypeAWS, "statuser", sr.Status, err)
+		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeAWS, func() error {
+			var err error
+			sr := kafka.SourceResult{
+				SourceID:     s.SourceID,
+				Identity:     s.Identity,
+				ResourceType: "Source",
+			}
+			_, err = clients.GetEC2Client(ctx, &s.Authentication, "")
+			if err != nil {
+				sr.Status = kafka.StatusUnavailable
+				sr.Err = err
+				logger.Warn().Msgf("Could not get aws assumed client %s", err)
+				chSend <- sr
+			} else {
+				sr.Status = kafka.StatusAvaliable
+				chSend <- sr
+			}
+			metrics.IncTotalAvailabilityCheckReqs(models.ProviderTypeAWS, sr.Status, err)
+			return fmt.Errorf("error during check %w:", err)
+		})
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cmd/pbstatuser/main.go
```diff
@@ -41,7 +43,29 @@ type SourceInfo struct {
...
+			Name:        metricNameSourceAvailabilityCheckRequestTotal,
+			Help:        "availability Check requests count partitioned by type (aws/gcp/azure), source status and error",
+			ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName},
+		},
+		[]string{"type", "status", "error"},
+	)
+
+	availabilityCheckReqsDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:        metricNameSourceAvailabilityCheckRequestDuration,
+		Help:        "Availability check Request duration partitioned by type and error",
```
lzap, line 63: Prometheus is not time-aware, please always include the scale in the help - in this case milliseconds.

cmd/pbstatuser/main.go
```diff
@@ -41,7 +43,29 @@ type SourceInfo struct {
...
+		ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName},
+	},
+		[]string{"type", "error"},
+	)
```
lzap, line 67: I have the same comment as in the other PR - it is a good idea to introduce a new package called "metrics" with exported functions for each metrics which allows us to "hide" the actual implementation and protect our app code from these metric definitions.

cmd/pbstatuser/main.go
```diff
@@ -112,10 +137,13 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 		// TODO: check if source is avavliable - WIP
 		sr.Status = kafka.StatusAvaliable
 		chSend <- sr
+		totalAvailabilityCheckReqs.WithLabelValues("azure", string(sr.Status), "").Inc()
+		availabilityCheckReqsDuration.WithLabelValues("azure", "").Observe(float64(time.Since(start).Nanoseconds()) / 1000000)
```
lzap, line 141: I am wondering why you don't do:

```suggestion
		availabilityCheckReqsDuration.WithLabelValues("azure", "").Observe(float64(time.Since(start).Milliseconds()))
```

lzap: Oh I think I know why, you want float value not rounded integer.

cmd/pbstatuser/main.go
```diff
@@ -162,9 +197,13 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
 			sr.Err = err
 			logger.Warn().Msgf("Could not list gcp regions %s", err)
 			chSend <- sr
+			totalAvailabilityCheckReqs.WithLabelValues("gcp", string(sr.Status), err.Error()).Inc()
+			availabilityCheckReqsDuration.WithLabelValues("gcp", err.Error()).Observe(float64(time.Since(start).Nanoseconds()) / 1000000)
 		} else {
 			sr.Status = kafka.StatusAvaliable
 			chSend <- sr
+			totalAvailabilityCheckReqs.WithLabelValues("gcp", string(sr.Status), "").Inc()
+			availabilityCheckReqsDuration.WithLabelValues("gcp", "").Observe(float64(time.Since(start).Nanoseconds()) / 1000000)
```
lzap, line 206: Okay this is getting out of control, we see this bit so many times that I believe this definitely needs a helper function that will return you this. I think the best place for such function would be the newly created "metrics" package: `metrics.TimeSinceMs(start)` (returns float).

lzap (changes requested):
Dropped few more comments.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	TotalAvailabilityCheckReqs.WithLabelValues(provider.String(), string(StatusType), component, errString).Inc()
+}
+
+func RegisterTotalAvailablilityCheckReqs() {
+	prometheus.MustRegister(TotalAvailabilityCheckReqs)
+}
+
+func RegisterAvailabilityCheckReqsDuration() {
+	prometheus.MustRegister(AvailabilityCheckReqsDuration)
+}
```
lzap, line 57: There could be simply `RegisterAll` or whatever, can be refactored later this is fine.

lzap (approved):

cmd/pbstatuser/main.go
```diff
@@ -166,6 +175,8 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
 			sr.Status = kafka.StatusAvaliable
 			chSend <- sr
 		}
+		metrics.IncTotalAvailablilityCheckReqs(models.ProviderTypeGCP, "statuser", sr.Status, err)
+		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeGCP, start, err)
 	}
 }
```
lzap, line 181: Wait there is a bug, you start the duration at the application launch, not when you receive new variable from the channel.

cmd/pbstatuser/main.go
```diff
@@ -135,10 +141,13 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 			sr.Status = kafka.StatusAvaliable
 			chSend <- sr
 		}
+		metrics.IncTotalAvailablilityCheckReqs(models.ProviderTypeAWS, "statuser", sr.Status, err)
+		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeAWS, start, err)
 	}
 }
```
lzap, line 147: Here the same, start must be stored when a new variable is received from the channel.

cmd/pbstatuser/main.go
```diff
@@ -112,10 +115,13 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 		// TODO: check if source is avavliable - WIP
 		sr.Status = kafka.StatusAvaliable
 		chSend <- sr
+		metrics.IncTotalAvailablilityCheckReqs(models.ProviderTypeAzure, "statuser", sr.Status, nil)
+		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeAzure, start, nil)
 	}
 }
```
lzap, line 121: Here too.

lzap (changes requested):
Observed a bug in the observation code :-)

cmd/pbstatuser/main.go
lzap: Yes, if I am correct.

lzap:
FYI if you really want this to be super clean, you can create a function that takes a function, this is a common pattern in Go observability:

```go
ObserveAvailablilityCheckReqsDurationFunc(models.ProviderTypeAzure, nil, func() {
 // do stuff
})
```

No need of start/finish variables, this would be done in the function itself (using the safe `defer` keyword as well).

cmd/pbstatuser/main.go
```diff
@@ -106,15 +106,20 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
...
+				SourceID:     s.SourceID,
+				Identity:     s.Identity,
+				ResourceType: "Source",
+			}
+			// TODO: check if source is avavliable - WIP
+			sr.Status = kafka.StatusAvaliable
+			chSend <- sr
+			metrics.IncTotalAvailabilityCheckReqs(models.ProviderTypeAzure, "statuser", sr.Status, nil)
+
+			return fmt.Errorf("An error has occured %w:", err)
```
lzap, line 121:
```suggestion
			return fmt.Errorf("an error has occurred %w:", err)
```

Tho I would suggest something more descriptive: `error during check`

cmd/pbstatuser/main.go
```diff
@@ -123,22 +128,26 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
...
+				sr.Status = kafka.StatusUnavailable
+				sr.Err = err
+				logger.Warn().Msgf("Could not get aws assumed client %s", err)
+				chSend <- sr
+			} else {
+				sr.Status = kafka.StatusAvaliable
+				chSend <- sr
+			}
+			metrics.IncTotalAvailabilityCheckReqs(models.ProviderTypeAWS, "statuser", sr.Status, err)
+			return fmt.Errorf("An error has occured %w:", err)
```
lzap, line 149:
```suggestion
			return fmt.Errorf("an error has occurred %w:", err)
```

cmd/pbstatuser/main.go
```diff
@@ -147,30 +156,34 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
...
+				sr.Err = err
+				logger.Warn().Msgf("Could not list gcp regions %s", err)
+				chSend <- sr
+			} else {
+				sr.Status = kafka.StatusAvaliable
+				chSend <- sr
+			}
+			metrics.IncTotalAvailabilityCheckReqs(models.ProviderTypeGCP, "statuser", sr.Status, err)
+
+			return fmt.Errorf("An error has occured %w:", err)
```
lzap, line 185:
```suggestion
			return fmt.Errorf("an error has occurred %w:", err)
```

(The linter should complain)

internal/metrics/metrics.go
```diff
@@ -1,32 +1,58 @@
...
+	start := time.Now()
+	err := ObservedFunc()
+	errString := "false"
+	if err != nil {
+		errString = "true"
+	}
+
+	defer func() {
+		AvailabilityCheckReqsDuration.WithLabelValues(provider.String(), errString).Observe(float64(time.Since(start).Nanoseconds()) / 1000000)
+	}()
```
lzap, line 45: You need to move the defer call right behind `time.Now()` call otherwise it will not be used when error occurs. The correct ordering is:

```go
errString := "false"
start := time.Now()
defer func() { ... }

err := ObservedFunc()
if err != nil {
	errString = "true"
}
```

lzap (changes requested):
IMHO this looks perfect and much more readable and also safe, great work! Rebase is needed.

I have one suggestion tho, I believe these two metrics will never be used from any other component than "statuser", can you remove the function argument and put this into the constant labels? I think 90% of all our metrics will only be in one component so we can use constant labels for all of that. This will further simplify the code.

internal/metrics/metrics.go
lzap: No it is deferred execution, it will be _registered_ for execution before the func call, but _executed_ only and only the function (where `defer` is present) exists (e.g. via regular code path or panic).

adiabramovitch: Thank you lzap. 
I think that these examples you shared with me emphasize the need here of an anonymous function: 
https://go.dev/play/p/GFdfLjoYK1X
https://go.dev/play/p/UWBX8324Z9U

lzap:
Merging there is nothing to test for observability at the moment.

PR #351, opened 2023-01-02
fix(HMSPROV-392): Check if image is an original or a shared one

adiabramovitch:
Images can be shared to different regions in Image Builder. 
shared images are not visible under /compose endpoint in IB but under /clone.
Therefore if an image is not visible under /compose,  before erroring out we should search for the image under /clone. 
Only if the image is not preset neither under compose nor clone we should error out with the message that the image is not available.

### DIFF
--- a/internal/clients/http/image_builder/image_client.go
+++ b/internal/clients/http/image_builder/image_client.go
@@ -105,8 +105,25 @@ func (c *ibClient) GetGCPImageName(ctx context.Context, composeID string) (strin
 }
 
 func (c *ibClient) fetchImageStatus(ctx context.Context, composeID string) (*UploadStatus, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "fetchImageStatus")
+	defer span.End()
 	logger := logger(ctx)
 	logger.Trace().Msgf("Fetching image status %v", composeID)
+
+	composeResp, err := c.checkCompose(ctx, composeID)
+	if err != nil {
+		cloneResp, err := c.checkClone(ctx, composeID)
+		if err != nil {
+			return nil, fmt.Errorf("could not find image neither in compose nor in clones: %w", err)
+		}
+		return cloneResp, nil
+	}
+	return composeResp, nil
+}
+
+func (c *ibClient) checkCompose(ctx context.Context, composeID string) (*UploadStatus, error) {
+	logger := logger(ctx)
+	logger.Trace().Msgf("Fetching image status %v from composes", composeID)
 
 	resp, err := c.client.GetComposeStatusWithResponse(ctx, composeID, headers.AddImageBuilderIdentityHeader)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ezr-ondrej (commented):
Few nits, but nothing major.
Given we are calling Image builder twice now, I'd think about adding OpenTelemetry tracing around it and measure those calls in their own span, but that should be (if we decide to do it) a follow up anyway :)

adiabramovitch:
> Few nits, but nothing major. Given we are calling Image builder twice now, I'd think about adding OpenTelemetry tracing around it and measure those calls in their own span, but that should be (if we decide to do it) a follow up anyway :)

Fixed and added :)

internal/clients/http/image_builder/image_client.go
```diff
@@ -105,9 +105,26 @@ func (c *ibClient) GetGCPImageName(ctx context.Context, composeID string) (strin
...
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "fetchImageStatus")
+	defer span.End()
 	logger := logger(ctx)
 	logger.Trace().Msgf("Fetching image status %v", composeID)
 
+	composeResp, err := c.checkCompose(ctx, composeID)
+	if err != nil {
+		cloneResp, err := c.checkClone(ctx, composeID)
+		if err != nil {
+			return nil, fmt.Errorf("could not find image neither in compose nor in clones %w", err)
```
lzap, line 117: Mind adding colon, error could be misleading without it:

```suggestion
			return nil, fmt.Errorf("could not find image neither in compose nor in clones: %w", err)
```

internal/clients/http/image_builder/image_client.go
```diff
@@ -123,8 +140,35 @@ func (c *ibClient) fetchImageStatus(ctx context.Context, composeID string) (*Upl
...
+	resp, err := c.client.GetCloneStatusWithResponse(ctx, composeID, headers.AddImageBuilderIdentityHeader)
+	if err != nil {
+		logger.Warn().Err(err).Msg("Failed to fetch image status from image builder")
+		return nil, fmt.Errorf("cannot get compose status: %w", err)
+	}
+
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
+			return nil, fmt.Errorf("fetch image status call: %w", http.ComposeNotFoundErr)
```
lzap, line 163: This is `checkClone` so:

```suggestion
			return nil, fmt.Errorf("fetch image status call: %w", http.CloneNotFoundErr)
```

lzap (changes requested):
Two error handling comments.

lzap (approved):
Leaving this on @amirfefer since he is working on this IIRC.

amirfefer (approved):
Looks good, tested multiple image's clones 
Thanks @adiabramovitch 👍

PR #352, opened 2023-01-02
fix(HMSPROV-401): replace dejq with taskq

lzap:
Introduction

When we started working on provisioning, we had little to none knowledge about how app-interface work and what are the options. Therefore, the idea was to use image-builder's jobqueue implementation for our task queue. A new small library dejq was created as an adapter between jobqueue and in-memory. Later on, simple SQS and Redis implementations were added.

However, as time for production approached, I started looking into the codebase to clean it up and iron out details, it turns out that there are multiple drawbacks with the dejq library.

More about this change in a technical debt document.

Do not do full review yet, this is a POC code. I am happy to hear comments tho.


ezr-ondrej (commented):
I'd be more then happy to use external library if there is one that accomodates our needs and as time shown, our needs are quite simple and this PR proves that our current needs are easily met, so :+1:

lzap:
I have figured out how to fix dejq Redis implementation: https://github.com/lzap/dejq/blob/master/redis/redis_client.go

I also implemented monitoring of queue size: https://github.com/RHEnVision/provisioning-backend/pull/353

I would like to put this refactoring on hold, taskq is much more code than we actually need to have. Most of its complexity comes from the delaying feature and rescheduling of failed jobs. None of that we actually need, therefore, the dejq implementation can stay very simple (atomic linked list with polling). I might resurrect this effort in the future tho.

PR #353, opened 2023-01-03
Update dejq to 1.3.1, increase workers to 2 and add job queue metric

lzap:
I noticed a bug in dejq library, it used pub/sub mechanism to deliver jobs which incorrectly executed them on ALL workers instead just exactly one. I fixed this by using atomic linked list and bumped the version to 1.3.0+.

In the clowder config, I am increasing number of workers to 2 to actually test this on stage - we should have at least two workers to see if this works correctly.

This PR contains a separate commit that bumps all dependencies to the up to date version.

Finally, it fixes a ticket - we want to have a total number of pending jobs in the (Redis) queue to be exported as Prometheus metric. There is another commit that does that.

I have opened discussion about completely dropping dejq and replacing it with taskq, however, I would still like to fix dejq first as that effort might be longer and it is just an idea for now: https://github.com/RHEnVision/provisioning-backend/pull/352


adiabramovitch (approved):
LGTM

PR #354, opened 2023-01-05
chore: use one API worker

lzap:
The recent patch increased number of API and backend workers to 2. It is fine to test multiple workers, but for API workers it is inconvenient to have multiple workers on stage environment as you need to switch between k8s log consoles constantly.

This brings the number back to 1 but it keeps 2 workers.

@mshriver you mentioned some Splunk URL, do we have some Stage log monitoring database? I would love to have access to that.


lzap:
We are seeing issues on stage when UI is not reaching the backend, I have a suspicion that the increase of web workers was the reason. Merging.

PR #356, opened 2023-01-06
docs: adding info about trust relationship

lzap:
When setting up a stage account, trust relationship must be entered. This was not clear at all from our page.


lzap:
Doco only, tested.

PR #357, opened 2023-01-06
chore: simplify replica config and set all back to 1

lzap:
According to docs, minReplicas was deprecated, we should only use replicas field. Also there is no need to create ENV variables for these. The patch simplifies this.

Since I have tested that multiple backend workers work fine via Redis on stage, I am bringing the number back to 1 for everything. Looking up logs with multiple pods is a bad experience, we can increase this once we setup CloudWatch which is my priority for this sprint.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -24,7 +24,7 @@ objects:
         iqePlugin: provisioning
       deployments:
         - name: worker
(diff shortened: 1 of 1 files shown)


lzap:
"Failed: `response` not updated with instances" in IQE trying again:

/retest

lzap:
All good!

PR #358, opened 2023-01-09
Improve arch mismatch error (HMSPROV-340)

lzap:
This patch makes use of the new "msg" flag we have in the error playload. The UI should show it in a nicely manner, this is currently not implemented so when testing this you will see the full error.

The architecture is currently hardcoded, I filed an issue for image builder to add an architecture field into the compose detail endpoint, it is currently not available.

The patch also refreshes aws and azure instance types, I improved instance generation to print those instance types which have same name but different architectures. Turns out that on AWS most instance types share name for i386 and x86_64 therefore i386 are filtered out. For Azure there are couple of differences in types but all of them are minor difference in ephemeral disk size. Those are printed out during generation just for the record. Other than that this change enforces that instance type name is unique across architectures.

This change was needed because the new function FindInstanceType for AWS needs to find instance type by name to perform architecture check.


internal/db/seeds/dev_small.sql
```diff
@@ -4,18 +4,18 @@
...
 
+-- Seed some pubkeys, feel free to add your own key and associate it with your account
 INSERT INTO pubkeys(id, account_id, name, body, fingerprint)
-VALUES (1, 1, 'lzap-ed25519-2021',
-        'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap',
-        'SHA256:gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk')
+VALUES
+(1, 3, 'lzap-ed25519-2021',
+ 'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap', 'SHA256:gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk')
 ON CONFLICT DO NOTHING;
```
lzap, line 20: I cleaned the seed script a bit removing those arbitrary accounts and only keeping 13/13 and NULL/42. Then I moved my own pubkey to my stage account, feel free to add your own stage numbers and pubkeys if you want them after seed.

adiabramovitch (approved):
LGTM

cmd/typesctl/providers/interface.go
```diff
@@ -1,9 +1,13 @@
...
+
 type TypeProvider struct {
 	PrintRegisteredTypes      func(string)
 	PrintRegionalAvailability func(string, string)
 	GenerateTypes             func() error
 }
 
 var TypeProviders = make(map[string]TypeProvider)
+
+var ValidArchitectures = regexp.MustCompile(`^(x86[_-]64|aarch64|arm64)$`)
```
lzap, line 13: For the record, the regexp makes sure we only have ARM64 and Intel64 types.

cmd/typesctl/providers/ec2_types.go
```diff
@@ -40,25 +40,34 @@ func generateTypesEC2() error {
...
 		client, regionErr := clients.GetServiceEC2Client(ctx, region.String())
 		if regionErr != nil {
-			return fmt.Errorf("unable to list EC2 regions: %w", regionErr)
+			return fmt.Errorf("unable to get regional EC2 client: %w", regionErr)
 		}
 		instTypes, regionErr := client.ListInstanceTypesWithPaginator(ctx)
 		if regionErr != nil {
-			return fmt.Errorf("unable to list EC2 instance types: %w", regionErr)
+			// skip
+			fmt.Printf("unable to list EC2 instance types (region STS not enabled?): %s\n", regionErr.Error())
```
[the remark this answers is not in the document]
lzap: Because previously when it encountered a "disabled datacenter" configuration (see the big comment in the file for more details), it stopped the import process. And there is currently a new datacenter I cannot enable on my account for some reason (south africa 2) that makes import process impossible. Now, this only prints a warning and continues importing other DCs.

PR #359, opened 2023-01-09
fix(HMSPROV-407): enable cloudwatch in clowder

lzap:
Looks like everything is ready to enable cloudwatch API on stage, this is the only change that is needed to export logs via the API.

To get access to ELK/CW on AWS open the [HMSPROV-407](https://issues.redhat.com/browse/HMSPROV-407) ticket and follow the instructions to setup permissions for you account. You will need a GPG public key for a password to be emailed to you.

Testing: Not sure how to test this in ephemeral, but ideally let us merge this sometime in the morning hours so we can check stage. I am still waiting for my permissions to be approved.

### DIFF
--- a/internal/services/aws_reservation_service.go
+++ b/internal/services/aws_reservation_service.go
@@ -19,7 +19,7 @@ import (
 )
 
 func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
-	logger := ctxval.Logger(r.Context())
+	logger := *ctxval.Logger(r.Context())
 
 	var accountId int64 = ctxval.AccountId(r.Context())
 
@@ -68,6 +68,7 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 		renderError(w, r, payloads.NewDAOError(r.Context(), "create reservation", err))
 		return
 	}
+	logger = logger.With().Int64("reservation_id", reservation.ID).Logger()
 	logger.Debug().Msgf("Created a new reservation %d", reservation.ID)
 
 	// Get Sources client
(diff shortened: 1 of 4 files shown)


lzap:
Rebased to solve panic in ephemeral, added small logging enhancements in regard to reservation id logging.

lzap:
Failed: `response` not updated with instances

This is because of https://issues.redhat.com/browse/HMSPROV-414

I am going to merge this since cloudwatch/kibana will make investigation much easier.

internal/services/aws_reservation_service.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
-	logger := ctxval.Logger(r.Context())
+	logger := *ctxval.Logger(r.Context())
```
adiabramovitch, line 22: Just wondering, we didn't use the '*' until now, why did you add it? 
it seems like the compiler knows to reference it like the logger object and not as a pointer.

adiabramovitch (approved):
Just a small question for general knowledge

internal/services/aws_reservation_service.go
```diff
@@ -68,6 +68,7 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 		renderError(w, r, payloads.NewDAOError(r.Context(), "create reservation", err))
 		return
 	}
+	logger = logger.With().Int64("reservation_id", reservation.ID).Logger()
```
lzap, line 71: It is because of this line, I would have to introduce a helper variable here.

internal/services/aws_reservation_service.go
lzap: That is indeed suspicions, I did this on purpose (see my comment below).

It would have been dangerous, but I know for fact that this function never returns `nil`: https://github.com/RHEnVision/provisioning-backend/blob/main/internal/ctxval/context_values.go#L34

PR #360, opened 2023-01-10
fix(HMSPROV-414): start dequeue loop in api only for memory

lzap:
So the job queue is designed to run in two modes. When memory is configured, the pbapi (api worker) should automatically start goroutine that is responsible for performing background jobs. The communication is through local memory queue, this is great for development and it is the default mode when you don't provide any setting.

When in redis mode, the api worker was not supposed to start any goroutines, however, due to overlook it started handlers but did not register any functions. This meant that in stage, 4 workers are started (2 pbworkers, 2 pbapi) but two of them (pbapi) did not have registered any handler functions. Meaning the jobs were silently dropped, this was a bug in dejq redis implementation which has been fixed (error is now logged in that case).

This patch fixes the problem, it is a one-liner, however while working on that I improved dejq API by adding a job UUID. This now greatly increases logging and searching within job messages.

I am also increasing replicas back to 2/2 (api/workers) since decreasing to 1 did no do anything in kubernetes, I do not know how to decrease it at the moment and cloudwatch/kibana logging is now ready.


internal/config/helpers.go
```diff
@@ -80,5 +82,5 @@ func DumpConfig(logger zerolog.Logger) {
 	configCopy.Azure.ClientSecret = replacement
 	configCopy.GCP.JSON = replacement
 	configCopy.Unleash.Token = replacement
-	logger.Info().Msgf("Configuration: %+v", configCopy)
+	logger.Info().Str("binary", path.Base(os.Args[0])).Msgf("Configuration: %+v", configCopy)
 }
```
lzap, line 86: I want the binary name to be reported in logs because that is our cloudwatch stream name.

internal/jobs/queue/queue_interface.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 type Enqueuer interface {
-	Enqueue(ctx context.Context, jobs ...dejq.PendingJob) error
+	Enqueue(ctx context.Context, jobs ...dejq.PendingJob) ([]string, error)
 }
```
lzap, line 11: This change is needed because dejq v1.4 changed its API. And yes, I did a breaking change in a minor release, there are no other users of dejq than me :-)

deploy/clowdapp.yaml
```diff
@@ -258,9 +252,6 @@ parameters:
   - description: Determines Clowder deployment
     name: CLOWDER_ENABLED
     value: "true"
-  - description: AWS CloudWatch logging integration
-    name: CLOUDWATCH_ENABLED
-    value: "false"
```
lzap, line 263: I noticed this, this is now unused. Cloudwatch is configured directly from clowder.

ezr-ondrej (approved):
Needs rebase but apart of it LGTM :+1:

ezr-ondrej:
On related note, but just a thought not blocking the PR: if we are keeping the `dejq` I'd vote for merging it in the repository. Given we are the only user, it would at least be clear that this codebase relies on it and we would keep track of changes here. I guess that keeping changelong of external library for no reason makes no sense and it could also get more eyes. Just a thought tho :)

lzap:
I thought about this too, makes sense tho I would propose to postpone this until I have more time. I would probably drop dbjobqueue as well and simplify the interface too.

lzap:
What? I did not hit the merge button!!! WTF

lzap:
GH automatically merged this after I rebased the PR while it was still running tests!

Snímek obrazovky 2023-01-12 v 8 29 18

PR #361, opened 2023-01-11
fix(HMSPROV-407): improve logging of cw config

lzap:
Cloudwatch is not getting configured by clowder, I am shooting in the dark. Trying to figure out this, this PR should help to find some more info. We cannot dump the whole clowder config into logs that is unsafe.


internal/logging/cwl_client.go
```diff
@@ -7,15 +7,14 @@ import (
...
 		config.Cloudwatch.Key, config.Cloudwatch.Secret, config.Cloudwatch.Session))
 
-	CWL = cloudwatchlogs.New(cloudwatchlogs.Options{
+	cwClient := cloudwatchlogs.New(cloudwatchlogs.Options{
 		Region:      config.Cloudwatch.Region,
 		Credentials: cache,
 	})
+
+	return cwClient
 }
```
lzap, line 20: This file was moved from clients package into logging package where it belongs.

lzap:
Merging this is just logging during initialization I need to see this NOW.

PR #362, opened 2023-01-11
fix(HMSPROV-407): further improve logging of cw config

lzap:
Turns out there is a typo in our codebase in `InProdClowder` so the variables were not printed correctly. This fixes it.


lzap:
Again, a small change, this can be only tested in clowder (stage). Merging.

PR #365, opened 2023-01-12
fix(HMSPROV-416): simple launch statistics

lzap:
We need to explore launch statistics, this is currently in the design phase, but I wanted to take a look how statistics implemented as prometheus or logs (or both in this case) would look like.

We can only do dumb "counting" that will give us just a number of uses. However, it would make sense to collect also account numbers and then we could filter out red hat or testing accounts as a post processing step (in PowerBI or Excel). That would require more capacity.

So I think we have several options:

* Store data into Prometheus, good for short term graphing or export. Retention policy might be quite small (days), however, for counter-only approach this is good enough as counters are monotonic.
* Store data in logs, good exporting possibilities via Kibana (e.g. Excel), retention policy is still small (weeks). We could export data every month into Excel and this could be a way to get very detailed statistics for a whole year. Chances are it could be automated.
* Create our own table and track statistics there. I do not like this approach as it is the most complex and data are locked down for engineering in stage/prod environments.

I propose a combination of both A and B.

We need to gather info about instance types and other parameters form the main wizard (count perhaps). Monitoring can give us counters or histograms of usage. Additionally, creating INFO level logs with field "statistics" allows us to export data from Kibana before retention (which appears to be a week).

This implementation will be quite simple, if we encounter any issues with these stats we can do proper implementation of statistics via a database table or external system.

* https://issues.redhat.com/browse/HMSPROV-416
* https://issues.redhat.com/browse/HMSPROV-385

Consider this a draft, tho, if this gets accepted I would encourage to test this both with AWS and GCP, thanks.

### DIFF
--- a/cmd/pbapi/main.go
+++ b/cmd/pbapi/main.go
@@ -74,6 +74,7 @@ func main() {
 	tel := telemetry.Initialize(&log.Logger)
 	defer tel.Close(ctx)
 	metrics.RegisterApiMetrics()
+	metrics.RegisterApiStatistics()
 
 	// initialize the rest
 	err = db.Initialize(ctx, "public")
--- a/cmd/pbstatuser/main.go
+++ b/cmd/pbstatuser/main.go
@@ -320,6 +320,7 @@ func main() {
 	}()
 
 	metrics.RegisterStatuserMetrics()
+	metrics.RegisterStatuserStatistics()
 
 	// start processing goroutines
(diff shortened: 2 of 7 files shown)


lzap:
We have a project called floorist that could help us with the task.

lzap:
It looks like we have the floorist tool, so I will explore it first as it looks like much better way of doing what we need.

PR #366, opened 2023-01-12
fix(HMSPROV-407): fix blank logic in cw initialization

lzap:
Turns out logic algebra can still bite. Also, I confirmed that the "cloudwatch type" clowder field is suppoed to be set to "app-interface" but it is actually blank. It's a known issue.

This patch should finally grab CW configuration from clowder and finally start logging. You cannot test this in ephemeral, this must be merged.

### DIFF
--- a/internal/config/helpers.go
+++ b/internal/config/helpers.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	"os"
-	"path"
 	"strings"
 
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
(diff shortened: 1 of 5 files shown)


internal/config/helpers.go
```diff
@@ -30,12 +28,12 @@ func RedisHostAndPort() string {
...
 func InStageClowder() bool {
-	return clowder.IsClowderEnabled() && strings.HasPrefix(*clowder.LoadedConfig.Metadata.EnvName, "env-stage")
+	return clowder.IsClowderEnabled() && strings.HasPrefix(*clowder.LoadedConfig.Metadata.EnvName, "stage")
 }
 
 // InProdClowder returns true, when the app is running in production clowder environment.
 func InProdClowder() bool {
-	return clowder.IsClowderEnabled() && strings.HasPrefix(*clowder.LoadedConfig.Metadata.EnvName, "env-prod")
+	return clowder.IsClowderEnabled() && strings.HasPrefix(*clowder.LoadedConfig.Metadata.EnvName, "prod")
 }
```
lzap, line 37: This was confirmed by Gábor.

internal/config/helpers.go
```diff
@@ -99,5 +97,5 @@ func DumpConfig(logger zerolog.Logger) {
 	configCopy.Azure.ClientSecret = replacement
 	configCopy.GCP.JSON = replacement
 	configCopy.Unleash.Token = replacement
-	logger.Info().Str("binary", path.Base(os.Args[0])).Msgf("Configuration: %+v", configCopy)
+	logger.Info().Msgf("Configuration: %+v", configCopy)
 }
```
lzap, line 101: I moved this field to generic initializer so all pods get it, it is fairly useful.

PR #367, opened 2023-01-12
fix(HMSPROV-389): drop memcache count

lzap:
Previously we used memory caching on stage, we now use Redis. It has its own statistics which we can utilize, no need to track Redis. And we only use the memory implementation to cache the app ID (single value), so this value was always zero.

This patch drops the metrics, it is not very useful now.


ezr-ondrej (approved):
Looks reasonable to me, will let @adiabramovitch to have a look as well tho :)

adiabramovitch (approved):
Make sense, Thanks

PR #368, opened 2023-01-12
fix(HMSPROV-407): disable cw for migrations

lzap:
Turns out cloudwatch is not available for init containers on stage, so it panics. I was hoping to have migration logs in ELK stack for a longer time so we can investigate weeks after, well, not gonna happen!


lzap:
Some quay.io problem.

/retest

lzap:
Stage is blocked and cannot proceed, pushing new version to fix it.

PR #370, opened 2023-01-12
chore: use reservation_id instead reservation

lzap:
SSIA, we had both in Kibana now it is just one field.


ezr-ondrej (approved):
Thanks @lzap

PR #371, opened 2023-01-13
chore: drop jobqueue references

lzap:
It turns out we do not need too complicated jobs, a simple "background job" is sufficient for our use case. Current implementation uses simple Redis queue (list) and the interface also has a second option AWS SQS. Let's drop tables from the database we will not use dbjobqueue from image-builder.


lzap:
I am going to do this as part of bigger PR - moving dejq from library to the project.

PR #372, opened 2023-01-13
fix: unique index on pubkey_resource

ezr-ondrej:
Pubkey resource unique index was not changed after addition of source and region.

Refs HMSPROV-415


lzap (approved):
LGTM

PR #374, opened 2023-01-16
fix(HMSPROV-387): use time-based offset for statuser

lzap:
Last week, I enabled a Kafka consumer group which solved the problem of tracking offsets. While this worked fine locally, it does not on stage. I am not sure what is wrong, however, I figured out a different solution.

Statuser is only one instance at the moment, however, turns out that offsets can be also set based on time. When statuser is restarted, only newly created messages can be read. This patch does exactly that.

### DIFF
--- a/internal/kafka/kafka.go
+++ b/internal/kafka/kafka.go
@@ -138,12 +138,11 @@ func newContextErrLogger(ctx context.Context) func(msg string, a ...interface{})
 }
 
 // NewReader creates a reader. Use Close() function to close the reader.
-func (b *kafkaBroker) NewReader(ctx context.Context, topic, group string) *kafka.Reader {
+func (b *kafkaBroker) NewReader(ctx context.Context, topic string) *kafka.Reader {
 	return kafka.NewReader(kafka.ReaderConfig{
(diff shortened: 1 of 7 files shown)


internal/kafka/kafka.go
```diff
@@ -163,11 +162,16 @@ func (b *kafkaBroker) NewWriter(ctx context.Context) *kafka.Writer {
...
+func (b *kafkaBroker) Consume(ctx context.Context, topic string, since time.Time, handler func(ctx context.Context, message *GenericMessage)) {
 	logger := ctxval.Logger(ctx)
-	r := b.NewReader(ctx, topic, group)
+	r := b.NewReader(ctx, topic)
 	defer r.Close()
 
+	err := r.SetOffsetAt(ctx, since)
+	if err != nil {
+		logger.Warn().Err(err).Msg("Unable to set initial offset")
+	}
```
lzap, line 173: This should do it, hopefully.

ezr-ondrej: Lets see, sounds more like it :)

ezr-ondrej (approved):
Nice, sounds like the offset should do the job :)

PR #375, opened 2023-01-17
fix(HMSPROV-387): filter out noisy kafka logs

lzap:
While the last patch fixed the kafka reader problem, I noticed that if the goroutine finishes execution for whatever reason, the app stays active. This patch fixes that, adds some extra logging and exits the application gracefully, so k8s can restart the container and notify us about it.

On top of that, I noticed some noisy debug messages, the patch removes those. I still want to keep debug mode on for some time, but the amount of these polling messages is flooding our logs.

### DIFF
--- a/internal/kafka/kafka.go
+++ b/internal/kafka/kafka.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"regexp"
 	"strings"
 	"time"
 
@@ -123,8 +124,22 @@ func NewKafkaBroker(ctx context.Context) (Broker, error) {
 	}, nil
 }
 
+// kafka library has some noisy debug messages
+var ignoredMsg *regexp.Regexp
+
+func init() {
+	var err error
+	ignoredMsg, err = regexp.Compile("^no messages received from kafka within the allocated time.*")
+	if err != nil {
+		panic(err)
+	}
+}
+
 func newContextLogger(ctx context.Context) func(msg string, a ...interface{}) {
 	return func(msg string, a ...interface{}) {
+		if ignoredMsg.MatchString(msg) {
+			return
+		}
 		logger := ctxval.Logger(ctx)
 		logger.Debug().Bool("kafka", true).Msgf(msg, a...)
 	}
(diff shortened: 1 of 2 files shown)


ezr-ondrej:
Do we need to filter it? It's a debug message, I guess we can just bump the log level to info if we want it to be less noisy? :thinking:

lzap:
It is there every few seconds:

image

internal/kafka/kafka.go
```diff
@@ -123,8 +124,22 @@ func NewKafkaBroker(ctx context.Context) (Broker, error) {
...
+
+func init() {
+	var err error
+	ignoredMsg, err = regexp.Compile("^no messages received from kafka within the allocated time.*")
+	if err != nil {
+		panic(err)
+	}
+}
+
 func newContextLogger(ctx context.Context) func(msg string, a ...interface{}) {
```
adiabramovitch, line 138: I do not see any usage of `newContextLogger` function, to ignore those messages the function needs to be used throughout the code.

lzap: It is here:

```
// NewReader creates a reader. Use Close() function to close the reader.
func (b *kafkaBroker) NewReader(ctx context.Context, topic string) *kafka.Reader {
	return kafka.NewReader(kafka.ReaderConfig{
		Brokers:     config.Kafka.Brokers,
		Dialer:      b.dialer,
		Topic:       topic,
		StartOffset: kafka.LastOffset,
		Logger:      kafka.LoggerFunc(newContextLogger(ctx)),
		ErrorLogger: kafka.LoggerFunc(newContextErrLogger(ctx)),
	})
}

// NewWriter creates synchronous writer created from the pool. It does not have associated any topic with it,
// therefore topic must be set on the message-level. Make sure to close it with Close() function.
func (b *kafkaBroker) NewWriter(ctx context.Context) *kafka.Writer {
	return &kafka.Writer{
		Addr:        kafka.TCP(config.Kafka.Brokers...),
		Transport:   b.transport,
		Logger:      kafka.LoggerFunc(newContextLogger(ctx)),
		ErrorLogger: kafka.LoggerFunc(newContextErrLogger(ctx)),
	}
}
```

lzap:
> I guess we can just bump the log level to info if we want it to be less noisy?

This kafka library does not have any logging levels, it only supports "stdout" and "stderr" output.

internal/kafka/kafka.go
adiabramovitch: oh, missed it, it showed only one result in the search.

PR #377, opened 2023-01-17
chore: move EnforceIdentity so it logs through our logger

ezr-ondrej:


lzap:
Linter.

internal/middleware/enforce_identity.go
```diff
@@ -0,0 +1,89 @@
...
+	logger.Warn().Msgf("Failed to enforce the Identity header: %s", reason)
+	http.Error(w, reason, code)
+}
+
+// EnforceIdentity extracts the X-Rh-Identity header and places the contents into the
+// request context.  If the Identity is invalid, the request will be aborted.
+func EnforceIdentity(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		logger := ctxval.Logger(r.Context())
+		rawHeaders := r.Header["X-Rh-Identity"]
```
lzap, line 30: Can you please change all log lines in this file to read "X-Rh-Identity"? Because this is a map, it must be capitalized, yet we log "x-rh-identity" which is an invalid value I believe.

internal/middleware/enforce_identity.go
```diff
@@ -0,0 +1,89 @@
+package middleware
```
lzap, line 1: Can you add a comment where this code comes from and why we moved it here please?

internal/middleware/enforce_identity.go
```diff
@@ -0,0 +1,89 @@
...
+
+		topLevelOrgIDFallback(&jsonData)
+
+		err = checkHeader(r.Context(), &jsonData, w)
+		if err != nil {
+			return
+		}
+
+		ctx := context.WithValue(r.Context(), identity.Key, jsonData)
+		next.ServeHTTP(w, r.WithContext(ctx))
```
lzap, line 61: Can we use our own `ctxval` setter and getter with our own key then? For consistency, I never liked that :-)

ezr-ondrej: We would be polluting more code with this... but I guess in case it stays...

ezr-ondrej: btw we should not use our own key. some other services might want to share the header, I'd keep it under globally shared key, just in case there is more shared code?

ezr-ondrej: If you'd not mind, I'd not do this, I'd love to keep it as much unchanged from the original as possible.

lzap: This will stay forever, trust me. We will never touch this again if it works. I can do it later tho.

internal/middleware/enforce_identity.go
lzap: There is one more.... gosh. Ok. Fine

lzap (approved):
Se skřípěním zubů...

PR #378, opened 2023-01-17
chore: improve logger middleware to better visibility

ezr-ondrej:


internal/middleware/logger.go
```diff
@@ -31,6 +31,8 @@ func LoggerMiddleware(rootLogger *zerolog.Logger) func(next http.Handler) http.H
 				Logger()
 			t1 := time.Now()
 
+			logger.Debug().Msg("Started request")
```
ezr-ondrej, line 34: I think method and URL is in the annotation vars already, so nothing on top of that is needed I guess.

lzap, line 34:
```suggestion
			logger.Debug().Msgf("Started %s request %s", r.Method, r.URL.Path)
```

internal/middleware/logger.go
```diff
@@ -49,7 +51,7 @@ func LoggerMiddleware(rootLogger *zerolog.Logger) func(next http.Handler) http.H
 
 				log.Info().
 					Int("status", ww.Status()).
-					Msg(fmt.Sprintf("Completed %s request %s in %s ms with %d",
+					Msg(fmt.Sprintf("Completed %s request %s in %s with %d",
```
lzap, line 54:
```suggestion
					Msgf("Completed %s request %s in %s with %d",
```

internal/middleware/logger.go
lzap: Yes it is and the general idea of structured logging is to only put message without copies of these fields, yet, I would suggest to format it the same way as in "Completed..." message, I find it very useful when searching logs by eyes without all the fields enabled in Kibana.

lzap: I find it very useful, this is subjective I have to admit:

image

ezr-ondrej: What for? we have these in the logger itself :thinking:

ezr-ondrej: see my comment ^^ :)

lzap: I pasted an example screenshot above, see how useful this is. I would argue that we already have this for "Completed" message, for consistency let's please keep it the same way. I find it very useful myself.

In general, I agree this should not be the way we create logs, but this middleware is special - all requests have these logs.

internal/middleware/logger.go
```diff
@@ -31,6 +30,9 @@ func LoggerMiddleware(rootLogger *zerolog.Logger) func(next http.Handler) http.H
 				Logger()
 			t1 := time.Now()
 
+			lHeaders := logger.With().Fields(r.Header).Logger()
+			lHeaders.Debug().Msg("Started request with above headers")
```
lzap, line 34: There is no need to copy logger just to use it once, why don't you just do:

```
logger.Debug().Fields(r.Header).Msgf("Started %s request %s", r.Method, r.URL.Path")
```

I am asking once again for the same formatting as we do for the "Completed" message so I can "pair" the two by eyes in Kibana.

lzap: I do not know how to create a submap, I think this is okay.

internal/middleware/logger.go
ezr-ondrej: oki, you're right it doesn't hurt :)

internal/middleware/logger.go
ezr-ondrej: oh I missed this comment, there is usage of method on a pointer and for some reason it doesn't work in-line (no idea what went wrong there, didn't investigate).

lzap (approved):
Thanks, I appreciate having method and path in both messages, this allows me to quickly find proper requests without bringing additional columns in Kibana.

PR #379, opened 2023-01-17
chore: cleanup of error payloads and ctxval package

lzap:
Also simplify error payload context logger.


internal/ctxval/context_values.go
```diff
@@ -20,26 +20,35 @@ const (
...
+	value := ctx.Value(identityCtxKey)
+	if value == nil {
+		return identity.XRHID{}
+	}
+	return value.(identity.XRHID)
+}
+
 func WithIdentity(ctx context.Context, id identity.XRHID) context.Context {
-	return context.WithValue(ctx, identity.Key, id)
+	return context.WithValue(ctx, identityCtxKey, id)
```
ezr-ondrej, line 51: I liked to use the shared key, so we could pass our context to other shared code, but if you insist :shrug: :)

ezr-ondrej:
Tests are failing

internal/ctxval/context_values.go
lzap: Context should not be used to pass arguments through API boundaries, I have never seen that either. It is a small change for consistency, if we ever hit this issue we can copy identity with the needed key.

lzap:
Fixed tests.

internal/ctxval/context_values.go
adiabramovitch: I remember that it is an issue to use a different key than the shared one because sources look for the identity in the shared key. 
Have you checked that the workflow stayed intact? 
When I checked your PR I got `11:48AM ERR service returned unauthorized ` while on the main branch the launch instance job has succeeded.

lzap: Yeah it looks like some bug. But the key should not matter, it is our own context for our own app only. I bet it is something else, let me take a look.

internal/ctxval/context_values.go
```diff
@@ -54,19 +63,16 @@ func WithIdentityFrom64(ctx context.Context, id string) (context.Context, error)
 		return nil, fmt.Errorf("could not unmarshal json %w", err)
 	}
 
-	return context.WithValue(ctx, identity.Key, jsonData), nil
-}
```
lzap, line 58: I overlooked this line, it was still using the old key.

lzap:
Found the issue hopefully, I missed one replacement.

By the way, I filed a big refactoring of the identity package which will allow us to drop a lot of code from our app, both `WithIdentityFrom64` and the whole middleware could be dropped and new exported functions used instead. I will file a followup PR once this gets in: https://github.com/RedHatInsights/platform-go-middlewares/pull/45

lzap:
Interesting, it does not work, therefore I am going to ditch this - I do not need this hard, I was just thinking it is nicer to have all keys owned by us. No biggie.

lzap:
Ok I removed all the identity key stuff, left only the cleanup portions of the code, I have broken up the ctxval into individual commits the file was too long for my taste. Review commit by commit (3 of them).

PR #380, opened 2023-01-17
fix: filtering Provisioning auth for Source

ezr-ondrej:
The code assumSources respond for every source with only single authentication. This does not hold true as some Sources can have multiple applications defined.

But luckily authtype i sstable field and always holds provisionig-arn for Provisioning auth. This change makes use of this field and filters the Provisioning auth with certainit. It allows us to remove verification of Auth's Application as now we should be sure it's Provisioning.

Fixes HMSPROV-426

### DIFF
--- a/internal/clients/http/sources/sources_client.go
+++ b/internal/clients/http/sources/sources_client.go
@@ -135,32 +135,8 @@ func (c *sourcesClient) GetAuthentication(ctx context.Context, sourceId clients.
 	// Sources API currently does not provide a good server-side filtering.
 	auth, err := filterSourceAuthentications(*resp.JSON200.Data)
 	if err != nil {
-		logger.Warn().Msgf("Sources replied with more than one authentications for source: %s", sourceId)
+		logger.Warn().Msgf("Sources did not return any Provisioning authentication for source: %s", sourceId)
 		return nil, err
-	}
-
-	// Get the resource_id which equals to application_id
-	// and check that application_type_id in /applications/<app_id> equals to provisioning id
-	res, err := c.client.ShowApplicationWithResponse(ctx, *auth.ResourceId, headers.AddSourcesIdentityHeader)
-	if err != nil {
-		return nil, fmt.Errorf("cannot list source authentication: %w", err)
-	}
-
-	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
-	if err != nil {
-		if errors.Is(err, clients.NotFoundErr) {
-			return nil, fmt.Errorf("get source ARN call: %w", http.ApplicationNotFoundErr)
-		}
-		return nil, fmt.Errorf("get source ARN call: %w", err)
-	}
-
-	appTypeId, err := c.GetProvisioningTypeId(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("cannot get provisioning app type: %w", err)
-	}
-
-	if *res.JSON200.ApplicationTypeId != appTypeId {
-		return nil, fmt.Errorf("%w for source id %s", http.AuthenticationSourceAssociationErr, sourceId)
 	}
 
 	authentication := clients.NewAuthenticationFromSourceAuthType(ctx, *auth.Username, string(*auth.Authtype))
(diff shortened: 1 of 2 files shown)


internal/clients/http/sources/sources_client_test.go
```diff
@@ -29,4 +30,28 @@ func TestSourcesClient_GetAuthentication(t *testing.T) {
...
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusOK)
+			_, err := io.WriteString(w, `{"data":[{"id":"1","authtype":"access_key_secret_key","username":"JDOEAWSUSER","availability_status":"in_progress","resource_type":"Source","resource_id":"1"},{"id":"2","authtype":"arn","username":"arn:aws:iam::123456789999:role/redhat-cost-management-role-0f60c5c","availability_status":"in_progress","resource_type":"Application","resource_id":"1"},{"id":"3","authtype":"provisioning-arn","username":"arn:aws:iam::123456789999:role/redhat-provisioning-role-2f6d01c","availability_status":"in_progress","resource_type":"Application","resource_id":"2"},{"id":"4","authtype":"cloud-meter-arn","username":"arn:aws:iam::123456789999:role/redhat-cloud-meter-role-6331a17","availability_status":"in_progress","resource_type":"Application","resource_id":"3"}],"meta":{"count":4,"limit":100,"offset":0},"links":{"first":"/api/sources/v3.1/sources/1/authentications?limit=100\u0026offset=0","last":"/api/sources/v3.1/sources/1/authentications?limit=100\u0026offset=100"}}`)
+			require.NoError(t, err, "failed to write http body for stubbed server")
+		})
```
lzap, line 43: Nice, I like how you test this, well done. I am just curious, this is an account that has four sources, three are application types:

* arn
* provisioning-arn
* cloud-meter-arn

What is "arn"? I guess that is the first app that was introduced so it does not have a prefix?

internal/clients/http/sources/sources_client.go
```diff
@@ -219,7 +195,7 @@ func (c *sourcesClient) loadAppId(ctx context.Context) (string, error) {
 
 func filterSourceAuthentications(authentications []AuthenticationRead) (AuthenticationRead, error) {
 	for _, auth := range authentications {
-		if *auth.ResourceType == "Application" {
+		if *auth.ResourceType == "Application" && *auth.Authtype == "provisioning-arn" {
 			return auth, nil
 		}
 	}
```
lzap, line 201: Love these kind of patches, we need more.

internal/clients/http/sources/sources_client.go
```diff
@@ -135,34 +135,10 @@ func (c *sourcesClient) GetAuthentication(ctx context.Context, sourceId clients.
...
-
-	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
-	if err != nil {
-		if errors.Is(err, clients.NotFoundErr) {
-			return nil, fmt.Errorf("get source ARN call: %w", http.ApplicationNotFoundErr)
-		}
-		return nil, fmt.Errorf("get source ARN call: %w", err)
-	}
-
-	appTypeId, err := c.GetProvisioningTypeId(ctx)
```
lzap, line 157: This deletes one of the two `GetProvisioningTypeId` calls, I wonder if we could remove also the other one that is in `ListProvisioningSources`. I think that is a different case tho.

lzap (approved):
Tests are failing tho. This PR must be fully green.

internal/clients/http/sources/sources_client_test.go
adiabramovitch: Why four sources?
If I understood correctly, this is only one source that is linked to four apps/resources.

adiabramovitch (approved):
LGTM
It is great to improve `sources` code after gaining more knowledge.

internal/clients/http/sources/sources_client_test.go
ezr-ondrej: This is an actual data snippet, with changed account numbers and other specifics, but should be almost real data :)

@lzap: yeah, `arn` = cost-management, exactly as you're saying - the first app introduced in sources, so it got the unprefixed arn :)

@adiabramovitch: exactly, I believe one of the links is to Source itself and then there are three actual apps :)

internal/clients/http/sources/sources_client.go
ezr-ondrej: Yeah, unfortunatelly that one is different case, I was also thinking about it, but no luck (at least no simple way to do it). I'll take a look once we get this PR in :)

ezr-ondrej:
> Tests are failing tho. This PR must be fully green.

Agreed, turns out I forgot GCP and Azure, which we now also need to worry about for the availability check :)

akhil-jha (approved):

lzap:
Nice find!

PR #382, opened 2023-01-18
fix: image builder clowder config

ezr-ondrej:
Image builder defines it's backend container as service not svc. This is fixing the typo for picking up that config.

Fixes HMSPROV-421


ezr-ondrej:
I guess the value is picked up from

https://github.com/osbuild/image-builder/blob/c5bbe9beaa06f5ac4008c640ef5825863dbe340a/templates/image-builder-clowder.yml#L16

vs 

https://github.com/RedHatInsights/sources-api-go/blob/feec5b404801e1a64caa037cd720bf5a2255d292/deploy/clowdapp.yaml#L172

lzap:
Irrelevant failures?

ezr-ondrej:
> Irrelevant failures?

We were setting Sources URL instead of IB URL, fixed and thus the failures should be fixed :)

akhil-jha (approved):

PR #384, opened 2023-01-19
docs: Safe zero downtime upgrades

ezr-ondrej:
Refs HMSPROV-386


CONTRIBUTING.md
```diff
@@ -60,6 +60,39 @@ Here are few points before you start contributing:
...
+
+All changes need to be implemented with zero downtime upgrades in mind.
+There will be for a brief moment version N and version N+1 containers running over a single database during upgrade.
+
+This affects migration policies. All migrations need to be backward compatible with code running without the migration.
+If there is a need for a breaking change, it needs to happen over course of two deployments.
+
+Common examples:
+
+* **Change a column name** from "pubkey_resource" to simply "pubkey".
```
lzap, line 73:
```suggestion
* **Change a column name**
```

Everyone knows what that means :-)

CONTRIBUTING.md
```diff
@@ -60,6 +60,39 @@ Here are few points before you start contributing:
...
+      3. Drop the now unused "pubkey_resource" column
+
+* **Introduce non-nullable column** without default value
+   1. Merge/Deploy PR that would
+      1. Introduce the column nullable with NULL default
+      2. Adjust code to be adding data to the new column
+   2. Merge/Deploy followup PR that would
+      1. migrate data for the empty rows of the column
+      2. changed the column to non-nullable
+* **Drop table/column**
```
lzap, line 90: Empty line to be consistent?

CONTRIBUTING.md
```diff
@@ -60,6 +60,39 @@ Here are few points before you start contributing:
...
+   1. Merge/Deploy PR that would
+      1. Stop using the said table/column
+   2. Merge/Deploy followup PR
+      1. With migration dropping the table/column
+
```
lzap, line 95: Database migration is not the only problem, job queue arguments is another thing to consider.

* When new field is added, make sure the code is ready to process data when the value is unset (zero or `nil`)
* Rename or delete is similar, there can be jobs enqueued by the old version so code must account for that.
* Additionally, some "version" flag can be added to the arg struct for easier handling of these issues.

lzap (changes requested):
I agree in general, however, at this point I think we can be simply upgrading until we are out of beta.

ezr-ondrej:
> I agree in general, however, at this point I think we can be simply upgrading until we are out of beta.

Sure, we can have exceptions to these rules, but we should think about it at least and we are required to think about it prior deploying to prod cluster.
So this is me mainly starting the thinking of Upgrades and introducing the issue to all contributors :)
Rules are there to be broken if needed tho :shrug:

ezr-ondrej:
I've incorporated all your comments and rebased :)

lzap (approved):
lgtm

PR #387, opened 2023-01-23
fix(HMSPROV-433): change resource type to application

adiabramovitch:
The source's status is not being updated to "Available" or  "Unavailable" 
According to sources [integrate with us](https://consoledot.pages.redhat.com/docs/dev/services/sources/integration.html#_work_on_our_side) documentation the resource type is supposed to be "Application"

### DIFF
--- a/cmd/pbstatuser/main.go
+++ b/cmd/pbstatuser/main.go
@@ -39,7 +39,7 @@ const ChannelBuffer = 32
 type SourceInfo struct {
 	Authentication clients.Authentication
 
-	SourceID string
+	SourceApplictionID string
 
 	Identity identity.XRHID
 }
@@ -85,9 +85,9 @@ func processMessage(ctx context.Context, message *kafka.GenericMessage) {
 	}
 
 	s := SourceInfo{
-		Authentication: *authentication,
-		SourceID:       sourceId,
-		Identity:       ctxval.Identity(ctx),
+		Authentication:     *authentication,
+		SourceApplictionID: authentication.SourceApplictionID,
+		Identity:           ctxval.Identity(ctx),
 	}
 
 	switch authentication.ProviderType {
@@ -110,9 +110,9 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeAzure, func() error {
 			var err error
 			sr := kafka.SourceResult{
-				SourceID:     s.SourceID,
+				ResourceID:   s.SourceApplictionID,
 				Identity:     s.Identity,
-				ResourceType: "Source",
+				ResourceType: "Application",
 			}
 			// TODO: check if source is avavliable - WIP
 			sr.Status = kafka.StatusAvaliable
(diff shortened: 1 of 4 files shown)


cmd/pbstatuser/main.go
```diff
@@ -134,7 +134,7 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 			sr := kafka.SourceResult{
 				SourceID:     s.SourceID,
```
ezr-ondrej, line 135: I believe we also need to change the ID to application ID :thinking:

ezr-ondrej (changes requested):

cmd/pbstatuser/main.go
ezr-ondrej: I believe it is returned from sources as ResourceID here: https://github.com/RHEnVision/provisioning-backend/blob/0c4591ecb299ea9e42c3036508bf74145966c427/internal/clients/http/sources/sources_client.go#L136 
Tho we would need to pass it down to Authentication to get it here I guess :thinking:

adiabramovitch: I have the JSON [label](https://github.com/RHEnVision/provisioning-backend/blob/0c4591ecb299ea9e42c3036508bf74145966c427/internal/kafka/source_result.go#L17 ) set to `resoure_id`, is it enough or do you prefer changing the field to ResourceID instead of SourceID?

adiabramovitch: We can add the ResourceID to the authentication [struct](https://github.com/RHEnVision/provisioning-backend/blob/0c4591ecb299ea9e42c3036508bf74145966c427/internal/clients/authentication.go#L11) and then use it here: https://github.com/RHEnVision/provisioning-backend/blob/0c4591ecb299ea9e42c3036508bf74145966c427/cmd/pbstatuser/main.go#L89. 
Does it make sense?

ezr-ondrej: Yeah, that is the best way I've though of as well :+1:

internal/clients/authentication.go
```diff
@@ -9,6 +9,7 @@ import (
 )
 
 type Authentication struct {
+	ResourceID   string              `json:"resource_id"`
```
ezr-ondrej, line 12: Could we name it ApplicationID or even SourcesApplictionID?
If you don't agree, I'd at least document it, ResourceID is very general, which is intentional in Sources, where it can be ID of "anything", but for us it should always be ApplicationID, right?

ezr-ondrej (commented):
Just a naming nit, but looks good apart of that :+1:

internal/clients/authentication.go
adiabramovitch: sure, ill change the name :)

adiabramovitch: What about the ResourceID in SourcesInfo, is that OK or do you think I should change that as well?

ezr-ondrej: Ideally I'd change that as well, but that one lives just in that single file, so that one is not as important to change, but sure, once we are at it :+1:
In the result, it should be ResourceID tho, as that's what we send to sources.

ezr-ondrej (approved):
LGTM :+1: lets :shipit: and see how it works :)

PR #389, opened 2023-01-24
chore: run migrations for statuser

lzap:
We run migrations when api and worker pods start up, but not when statuser pod starts up. This should fix this.

It looks like during todays redeplyoment, k8s started redeployment with statuser first leading to booting up our app with old database.


ezr-ondrej (approved):
:+1: Thanks

PR #390, opened 2023-01-24
fix(HMSPROV-390): unscoped update pubkey fix

lzap:
SSIA

### DIFF
--- a/internal/migrations/code/update_fingerprints_test.go
+++ b/internal/migrations/code/update_fingerprints_test.go
@@ -15,12 +15,18 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func newContext(t *testing.T) context.Context {
+func newContextWithAccount(t *testing.T) context.Context {
 	return identity.WithTenant(t, context.Background())
 }
 
+func newContext(_ *testing.T) context.Context {
+	return context.Background()
+}
+
 func TestMigrationUpdateFingerprint(t *testing.T) {
-	ctx := newContext(t)
+	testCtx := newContext(t)
+	ctx := newContextWithAccount(t)
+
 	pkDao := dao.GetPubkeyDao(ctx)
 	defer reset()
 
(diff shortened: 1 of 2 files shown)


ezr-ondrej:
I guess you meant to use this in migrations?

lzap:
Damn it is not used :-))))

lzap:
I was rushing it because of a meeting, fixed, rebased.

lzap:
Alternatively, I could simply write SQL in the migration itself, which seems more appropriate I think?

lzap:
I have a better idea, do not merge!

ezr-ondrej:
Let's remove https://github.com/RHEnVision/provisioning-backend/blob/f5b2e1992b1987bdb7c95685f7ae2e607a420990/internal/migrations/code/update_fingerprints_test.go#L19 from the tests, as we do not have that in prod code :)

lzap:
Good point, I cannot remove the context with account because tests need to create testing records, but I am introducing new context that is only used for the actual test (the function that is being tested). That must not have account, others do require it. Hope it makes sense.

internal/migrations/code/update_fingerprints_test.go
```diff
@@ -15,12 +15,18 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func newContextWithAccount(t *testing.T) context.Context {
+	return identity.WithTenant(t, context.Background())
+}
+
 func newContext(t *testing.T) context.Context {
 	return identity.WithTenant(t, context.Background())
```
ezr-ondrej, line 23: We have the context here as well, is that intentinal? I thought we wanted clean context for this one? :thinking:

lzap: DAMN

lzap:
Rebased OMG when one's in a rush, this is the end result.

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #391, opened 2023-01-25
feat: minimal PoC Azure implementation

ezr-ondrej:
Implements Azure client minimal version of CreateVM with most values staticly defined. Implements direct call from a service to the Azure client, which bypasses the job, which will need to be defined in followup.

Refs HMS-1058

### DIFF
--- a/internal/clients/http/azure/create_vm.go
+++ b/internal/clients/http/azure/create_vm.go
@@ -1,0 +1,433 @@
+// strongly inspired by https://github.com/Azure-Samples/azure-sdk-for-go-samples/tree/main/sdk/resourcemanager/compute/create_vm
+// from commit 2ea51d9744d5d4680b4983b2e292a7b1d0085ff4
+
+package azure
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/ptr"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/codes"
+)
+
+var subscriptionId string
+
+const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/http/azure"
+
+const (
+	resourceGroupName = "redhat-deployed"
+	vmName            = "redhat-vm"
+	vnetName          = "redhat-vnet"
+	subnetName        = "redhat-subnet"
+	nsgName           = "redhat-nsg"
+	nicName           = "redhat-nic"
+	diskName          = "redhat-disk"
+	publicIPName      = "redhat-public-ip"
+	location          = "eastus"
+	adminUsername     = "azureuser"
+	vpnIPAddress      = "172.22.0.0/16"
+)
+
+func (c *client) CreateVM(ctx context.Context, imageID string, pubkey *models.Pubkey, instanceType clients.InstanceTypeName) (*string, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "CreateVM")
+	defer span.End()
+
+	logger := logger(ctx)
+	logger.Debug().Msg("Creating Azure VM instance")
+
+	resourceGroup, err := c.getOrCreateResourceGroup(ctx, resourceGroupName, location)
+	if err != nil {
+		span.SetStatus(codes.Error, "cannot create resource group")
+		logger.Error().Err(err).Msgf("cannot create resource group")
+		return nil, err
+	}
+	logger.Trace().Msgf("Using resource group id=%s", *resourceGroup.ID)
+
+	virtualNetwork, err := c.createVirtualNetwork(ctx, resourceGroupName, vnetName)
+	if err != nil {
+		span.SetStatus(codes.Error, "cannot create virtual network")
+		logger.Error().Err(err).Msgf("cannot create virtual network")
+		return nil, err
+	}
+	logger.Trace().Msgf("Using virtual network id=%s", *virtualNetwork.ID)
+
+	subnet, err := c.createSubnets(ctx, resourceGroupName, vnetName, subnetName)
+	if err != nil {
+		span.SetStatus(codes.Error, "cannot create subnet")
+		logger.Error().Err(err).Msgf("cannot create subnet")
+		return nil, err
+	}
+	logger.Trace().Msgf("Using subnet id=%s", *subnet.ID)
+
+	publicIP, err := c.createPublicIP(ctx, location, resourceGroupName, publicIPName)
+	if err != nil {
+		span.SetStatus(codes.Error, "cannot create public IP address")
+		logger.Error().Err(err).Msgf("cannot create public IP address")
+		return nil, err
+	}
+	logger.Trace().Msgf("Using public IP address id=%s", *publicIP.ID)
+
+	// network security group
+	nsg, err := c.createNetworkSecurityGroup(ctx, location, resourceGroupName, nsgName)
+	if err != nil {
+		span.SetStatus(codes.Error, "cannot create network security group")
+		logger.Error().Err(err).Msgf("cannot create network security group")
+		return nil, err
+	}
+	logger.Trace().Msgf("Using network security group id=%s", *nsg.ID)
+
+	networkInterface, err := c.createNetworkInterface(ctx, location, resourceGroupName, subnet, publicIP, nsg, nicName)
+	if err != nil {
+		span.SetStatus(codes.Error, "cannot create network interface")
+		logger.Error().Err(err).Msgf("cannot create network interface")
+		return nil, err
+	}
+	logger.Trace().Msgf("Using network interface id=%s", *networkInterface.ID)
+
+	vmParams := c.prepareVirtualMachineParameters(location, armcompute.VirtualMachineSizeTypes(instanceType), networkInterface, imageID, pubkey.Body, diskName)
+	virtualMachine, err := c.createVirtualMachine(ctx, resourceGroupName, vmName, vmParams)
+	if err != nil {
+		span.SetStatus(codes.Error, "cannot create virtual machine")
+		logger.Error().Err(err).Msgf("cannot create virtual machine")
+		return nil, err
+	}
+	logger.Debug().Msgf("Created virtual machine id=%s", *virtualMachine.ID)
+
+	return virtualMachine.ID, nil
+}
+
+func (c *client) getOrCreateResourceGroup(ctx context.Context, name string, location string) (*armresources.ResourceGroup, error) {
+	resourceGroupClient, err := c.newResourceGroupsClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	logger := logger(ctx)
+	getResp, err := resourceGroupClient.Get(ctx, name, nil)
+	if err == nil {
+		return &getResp.ResourceGroup, nil
+	}
+	if err != nil {
+		var azErr *azcore.ResponseError
+		if errors.As(err, &azErr) && azErr.StatusCode == http.StatusNotFound {
+			logger.Debug().Msgf("resource group %s not found, creating", name)
+			// 404 is expected, continue
+		} else {
+			return nil, fmt.Errorf("failed to fetch resource group: %w", err)
+		}
+	}
+
+	parameters := armresources.ResourceGroup{
+		Location: ptr.To(location),
+		// TODO tag the resource group by some RH identifier
+		// Tags:     map[string]*string{"sample-rs-tag": ptr.To("sample-tag")}, // resource group update tags
+	}
+
+	resp, err := resourceGroupClient.CreateOrUpdate(ctx, name, parameters, nil)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create resource group: %w", err)
+	}
+
+	return &resp.ResourceGroup, nil
+}
+
+func (c *client) createVirtualNetwork(ctx context.Context, resourceGroupName string, name string) (*armnetwork.VirtualNetwork, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "createVirtualNetwork")
+	defer span.End()
+
+	vnetClient, err := c.newVirtualNetworksClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	logger := logger(ctx)
+	getResp, err := vnetClient.Get(ctx, resourceGroupName, name, nil)
+	if err == nil {
+		return &getResp.VirtualNetwork, nil
+	}
+	if err != nil {
+		var azErr *azcore.ResponseError
+		if errors.As(err, &azErr) && azErr.StatusCode == http.StatusNotFound {
+			logger.Debug().Msgf("virtual network %s not found, creating", name)
+			// 404 is expected, continue
+		} else {
+			return nil, fmt.Errorf("failed to fetch virtual network: %w", err)
+		}
+	}
+
+	parameters := armnetwork.VirtualNetwork{
+		Location: to.Ptr(location),
+		Properties: &armnetwork.VirtualNetworkPropertiesFormat{
+			AddressSpace: &armnetwork.AddressSpace{
+				AddressPrefixes: []*string{
+					to.Ptr(vpnIPAddress),
+				},
+			},
+			//Subnets: []*armnetwork.Subnet{
+			//	{
+			//		Name: to.Ptr(subnetName+"3"),
+			//		Properties: &armnetwork.SubnetPropertiesFormat{
+			//			AddressPrefix: to.Ptr("10.1.0.0/24"),
+			//		},
+			//	},
+			//},
+		},
+	}
+
+	pollerResponse, err := vnetClient.BeginCreateOrUpdate(ctx, resourceGroupName, name, parameters, nil)
+	if err != nil {
+		return nil, fmt.Errorf("create of virtual network failed to start: %w", err)
+	}
+
+	resp, err := pollerResponse.PollUntilDone(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to poll for create virtual network result: %w", err)
+	}
+
+	return &resp.VirtualNetwork, nil
+}
+
+func (c *client) createSubnets(ctx context.Context, resourceGroupName string, vnetName string, name string) (*armnetwork.Subnet, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "createSubnets")
+	defer span.End()
+
+	subnetClient, err := c.newSubnetsClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	parameters := armnetwork.Subnet{
+		Properties: &armnetwork.SubnetPropertiesFormat{
+			// the subnet takes the full vpn address scope
+			AddressPrefix: to.Ptr(vpnIPAddress),
+		},
+	}
+
+	pollerResponse, err := subnetClient.BeginCreateOrUpdate(ctx, resourceGroupName, vnetName, name, parameters, nil)
+	if err != nil {
+		return nil, fmt.Errorf("create of subnet failed to start: %w", err)
+	}
+
+	resp, err := pollerResponse.PollUntilDone(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to poll for create subnet result: %w", err)
+	}
+
+	return &resp.Subnet, nil
+}
+
+func (c *client) createPublicIP(ctx context.Context, location string, resourceGroupName string, name string) (*armnetwork.PublicIPAddress, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "createPublicIP")
+	defer span.End()
+
+	publicIPAddressClient, err := c.newPublicIPAddressesClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	parameters := armnetwork.PublicIPAddress{
+		Location: to.Ptr(location),
+		Properties: &armnetwork.PublicIPAddressPropertiesFormat{
+			PublicIPAllocationMethod: to.Ptr(armnetwork.IPAllocationMethodStatic), // Static or Dynamic
+		},
+	}
+
+	pollerResponse, err := publicIPAddressClient.BeginCreateOrUpdate(ctx, resourceGroupName, name, parameters, nil)
+	if err != nil {
+		return nil, fmt.Errorf("create of public IP address failed to start: %w", err)
+	}
+
+	resp, err := pollerResponse.PollUntilDone(ctx, nil)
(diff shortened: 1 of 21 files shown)


internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,395 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	parameters := armnetwork.SecurityGroup{
+		Location: to.Ptr(location),
+		Properties: &armnetwork.SecurityGroupPropertiesFormat{
+			SecurityRules: []*armnetwork.SecurityRule{
+				// Windows connection to virtual machine needs to open port 3389,RDP
+				// inbound
```
lzap, line 233: Irrelevant comment I guess :-D

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,395 @@
...
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "CreateVM")
+	defer span.End()
+
+	logger := logger(ctx)
+	logger.Debug().Msg("Creating Azure VM instance")
+
+	resourceGroup, err := c.createResourceGroup(ctx, resourceGroupName, location)
+	if err != nil {
+		span.SetStatus(codes.Error, "cannot create resource group")
+		logger.Error().Err(err).Msgf("cannot create resource group")
```
lzap, line 46: Idea: This could easily be a `ctxval.LogErrorWithSpan(ctx, span, err, "msg")` and that function would:

* Pull the logger from the context
* Set span status
* Log the error

I really appreciate amount of logging and tracing! Well done.

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,395 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/codes"
+)
+
+var subscriptionId string
+
+const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/http/azure"
+
+const (
+	resourceGroupName = "redhat-deployed2"
```
lzap, line 24: Lighthouse supports attaching to either subscription or resource group. Are you testing this with subscription attachement or resource one? I am not sure if RGs can form a tree: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/manage-resource-groups-portal

lzap: I think we talked about this and agreed we would enforce Lighthouse on a resource group, just for cleaner environment, therefore this would probably not work I guess?!?

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,395 @@
...
+		return nil, err
+	}
+
+	parameters := armnetwork.VirtualNetwork{
+		Location: to.Ptr(location),
+		Properties: &armnetwork.VirtualNetworkPropertiesFormat{
+			AddressSpace: &armnetwork.AddressSpace{
+				AddressPrefixes: []*string{
+					to.Ptr("10.42.0.0/16"),
+				},
```
lzap, line 137: I think for V1 this could be fine, but I suggest to select a "random constant" subnet that is the most unlikely to hit existing subnets. I have a gut feeling that 10.0.0.0 space will be the most popular one, it is also the one that Azure documentation recommends. But there are others:

```
The Internet Assigned Numbers Authority (IANA) has reserved the
   following three blocks of the IP address space for private internets:

     10.0.0.0        -   10.255.255.255  (10/8 prefix)
     172.16.0.0      -   172.31.255.255  (172.16/12 prefix)
     192.168.0.0     -   192.168.255.255 (192.168/16 prefix)
```

How about selecting `172.16.0.0/16` prefix that gives:

image

That is 64k hosts (16 bits) for each which is good enough for most customers, yet, we can select one of the 172.16 - 172.31 subnets randomly (constant). For example `172.22.0.0/16` is a nice one to remember.

Now this is a wild guess, but I believe this one is less likely to collide with customers existing networks. Of course, this is a temporary solution until we implement templates.

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,395 @@
...
+			},
+			//Subnets: []*armnetwork.Subnet{
+			//	{
+			//		Name: to.Ptr(subnetName+"3"),
+			//		Properties: &armnetwork.SubnetPropertiesFormat{
+			//			AddressPrefix: to.Ptr("10.1.0.0/24"),
+			//		},
+			//	},
+			//},
```
lzap, line 146: Why is this commented out?

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,395 @@
...
+
+	pollerResponse, err := vnetClient.BeginCreateOrUpdate(ctx, resourceGroupName, name, parameters, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := pollerResponse.PollUntilDone(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
```
lzap, line 158: Yeah this will be super slow, Azure is way slower than EC2. We need to move this into a job ASAP.

I am fine merging this as-is tho.

internal/clients/http/azure/create_vm.go
lzap: Hmmm it looks like you need to define address space first, then networks. I guess for V1 (POC deployments) the subnet can be actually the (whole) address space. We will not utilize multiple subnets.

https://learn.microsoft.com/en-us/azure/virtual-network/concepts-and-best-practices

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,395 @@
...
+	subnetClient, err := c.newSubnetsClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	parameters := armnetwork.Subnet{
+		Properties: &armnetwork.SubnetPropertiesFormat{
+			AddressPrefix: to.Ptr("10.42.42.0/24"),
+		},
+	}
```
lzap, line 176: This should be the whole address space, why are you throwing out 8 bits for nothing?

For example, if we would go with my "wild guess" `172.22.0.0/16` this should be also the subnet.

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,395 @@
...
+	subnetName        = "redhat-subnet"
+	nsgName           = "redhat-nsg"
+	nicName           = "redhat-nic"
+	diskName          = "redhat-disk"
+	publicIPName      = "redhat-public-ip"
+	location          = "eastus"
+	adminUsername     = "azureuser"
+)
+
+func (c *client) CreateVM(ctx context.Context, imageID string, sshKeyBody string) (*armcompute.VirtualMachine, error) {
```
lzap, line 36: Now I know this is very early draft, but could you start using our models early? I mean:

* Pubkey
* AzureInstanceType

internal/services/azure_reservation_service.go
```diff
@@ -0,0 +1,71 @@
...
+	var azureImageID string
+	if strings.HasPrefix(payload.ImageID, "/subscriptions/") {
+		// Direct AMI ID was provided, no need to call image builder
+		azureImageID = payload.ImageID
+	} else {
+		// TODO fetch image ID from Image Builder
+	}
+
+	azureClient.CreateVM(r.Context(), azureImageID, pk.Body)
+}
```
lzap, line 71: Well that is ugly, the resource is reservation but it launches it directly. I guess we need the job for the first version :-)

internal/clients/http/azure/create_vm.go
ezr-ondrej: Ah! I did not remember us deciding to force it on ResourceGroup, I've tested with attaching to sub, but with this, it looked clumsy, so I agree attaching to ResourceGroup makes more sense :)

internal/clients/http/azure/create_vm.go
ezr-ondrej: I'll move it to job in a second commit, it should not be as hard, I just wanted to have a way of testing the client, but I'll move the changes to the service to separate commit :)

internal/clients/http/azure/create_vm.go
ezr-ondrej: Yeah I guess you're right, there is no reason yet to have multiple subnets.

internal/clients/http/azure/create_vm.go
ezr-ondrej: sure :+1: will do ;)

internal/clients/http/azure/create_vm.go
ezr-ondrej: Yeah, leftover from the sample cleanup :)

internal/services/azure_reservation_service.go
lzap: Sure, I get that. My point is do we want to merge this as is or refactor this to a background job for the first version? I lean towards the latter.

internal/clients/http/azure/create_vm.go
lzap: I wonder how are we going to enforce this, sources only stores UUID, do they? We won't be able to tell to which resource it was approved. We need to test this.

internal/clients/http/azure/create_vm.go
lzap: For the record, AWS VPC has a default range `172.31.0.0/16`: https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc.html

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,398 @@
...
+
+	pollerResponse, err := vmClient.BeginCreateOrUpdate(ctx, resourceGroupName, vmName, parameters, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := pollerResponse.PollUntilDone(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
```
lzap, line 395: I think this is reasonable for the first version. More info: https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-core-concepts

internal/clients/interface.go
```diff
@@ -89,6 +90,12 @@ var GetServiceAzureClient func(ctx context.Context) (ServiceAzure, error)
 
 type Azure interface {
 	ClientStatuser
+
+	// CreateVM creates Azure virtual machine.
+	// Most of the parameters are constant for now.
+	// imageID is expected in format /subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/images/{imageID}
+	// sshKeyBody should be full public key body
+	CreateVM(ctx context.Context, imageID string, pubkey *models.Pubkey, instanceType InstanceTypeName) (*armcompute.VirtualMachine, error)
```
lzap, line 98: One important thing - I would like us not to cross the API boundary, can you return just instance IDs like we do in AWS? If we need more, new type must be introduced.

internal/dao/pgx/reservation_pgx.go
```diff
@@ -80,6 +80,47 @@ func (x *reservationDao) CreateAWS(ctx context.Context, reservation *models.AWSR
...
+		reservationQuery := `INSERT INTO reservations (provider, account_id, steps, status)
+		VALUES ($1, $2, $3, $4) RETURNING id, created_at`
+		err := db.Pool.QueryRow(ctx, reservationQuery,
+			reservation.Provider,
+			reservation.AccountID,
+			reservation.Steps,
+			reservation.Status).Scan(&reservation.ID, &reservation.CreatedAt)
+		if err != nil {
+			return fmt.Errorf("pgx error: %w", err)
+		}
```
lzap, line 98: This could be extracted into an un-exported function I guess?

internal/dao/pgx/reservation_pgx.go
```diff
@@ -80,6 +80,47 @@ func (x *reservationDao) CreateAWS(ctx context.Context, reservation *models.AWSR
...
+
+		gcpQuery := `INSERT INTO azure_reservation_details (reservation_id, pubkey_id, source_id, image_id, detail)
```
lzap, line 100:
```suggestion
		azureQuery := `INSERT INTO azure_reservation_details (reservation_id, pubkey_id, source_id, image_id, detail)
```

internal/jobs/launch_instance_azure.go
```diff
@@ -0,0 +1,83 @@
...
+	azureClient, err := clients.GetAzureClient(ctx, args.Subscription)
+	if err != nil {
+		return fmt.Errorf("cannot create new Azure client: %w", err)
+	}
+
+	// TODO create multiple and save ids to DB
+	_, err = azureClient.CreateVM(ctx, args.AzureImageID, pubkey, clients.InstanceTypeName(reservation.Detail.InstanceSize))
+	if err != nil {
+		return fmt.Errorf("cannot create Azure instance(s): %w", err)
+	}
```
lzap, line 80: Yeah need to store instance IDs as strings. I think this is worth doing in the first cut so we have consistent database on stage. Looks like we are not deleting reservations anytime soon.

internal/models/reservation_model.go
```diff
@@ -123,6 +123,38 @@ type GCPReservation struct {
 	Detail *GCPDetail `db:"detail" json:"detail"`
 }
 
+type AzureDetail struct {
+	Location string `json:"location"`
+
+	// Instance name
+	Name string `json:"name"`
+
+	// GCP Machine type.
```
lzap, line 132:
```suggestion
	// Azure instance size.
```

Weird name tho, is this really how this is called on Azure? Looks like it: https://learn.microsoft.com/en-us/azure/virtual-machines/sizes

They call it sometimes Instances and sometimes Virtual Machines. So I like what you have, that is reasonable wording.

internal/services/azure_reservation_service.go
```diff
@@ -0,0 +1,130 @@
...
+	var azureImageID string
+	if strings.HasPrefix(payload.ImageID, "/subscriptions/") {
+		// Direct AMI ID was provided, no need to call image builder
+		azureImageID = payload.ImageID
+	} else {
+		// TODO fetch image ID from Image Builder
+	}
+
+	supportedArch := "x86_64"
+	it := types.FindInstanceType(clients.InstanceTypeName(payload.InstanceSize))
```
lzap, line 71: Could be:

```suggestion
	it := types.FindInstanceSize(clients.InstanceTypeName(payload.InstanceSize))
```

to match your terminology. I am good tho if you like this one.

internal/services/azure_reservation_service.go
```diff
@@ -0,0 +1,130 @@
...
+	if it == nil {
+		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), fmt.Sprintf("unknown instance size: %s", payload.InstanceSize), UnknownInstanceTypeNameError))
+		return
+	}
+	if it.Architecture.String() != supportedArch {
+		renderError(w, r, payloads.NewWrongArchitectureUserError(r.Context(), ArchitectureMismatch))
+		return
+	}
```
lzap, line 79: We should also check location and aws region if it is valid. Not here tho, created: https://issues.redhat.com/browse/HMS-1128

internal/services/azure_reservation_service.go
```diff
@@ -0,0 +1,130 @@
...
+	err = rDao.CreateAzure(r.Context(), reservation)
+	if err != nil {
+		renderError(w, r, payloads.NewDAOError(r.Context(), "create Azure reservation", err))
+		return
+	}
+	logger.Debug().Msgf("Created a new reservation %d", reservation.ID)
+
+	launchJob := worker.Job{
+		Type: jobs.TypeLaunchInstanceAzure,
+		Args: &jobs.LaunchInstanceAzureTaskArgs{
```
lzap, line 108: This must not be a pointer:

```suggestion
		Args: jobs.LaunchInstanceAzureTaskArgs{
```

lzap (changes requested):
Looks great.

internal/clients/http/azure/create_vm.go
ezr-ondrej: The Lighthouse template allows you to chose and it will force CU to choose ResourceGroup if we choose to bind to ResourceGroup.

internal/models/reservation_model.go
ezr-ondrej: yeah the terminology for Azure is not very well set, happy to change it if there is another suggestion, but in client I've seen InstanceSize, so I've used it.

internal/jobs/launch_instance_azure.go
ezr-ondrej: Done :)

internal/clients/http/azure/create_vm.go
ezr-ondrej: Ugh, it actually will Create statically named resource group :face_exhaling:

internal/clients/http/azure/create_vm.go
```diff
@@ -0,0 +1,430 @@
...
+func (c *client) createVirtualMachine(ctx context.Context, location string, resourceGroupName string, instanceType armcompute.VirtualMachineSizeTypes, networkInterface *armnetwork.Interface, imageID string, sshKeyBody string, vmName string, diskName string) (*armcompute.VirtualMachine, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "createVirtualMachine")
+	defer span.End()
+
+	vmClient, err := c.newVirtualMachinesClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	parameters := armcompute.VirtualMachine{
```
avitova, line 370: This could be split so that the parameters variable is set before calling createVirtualMachine. We would get rid of few parameters and it would be a bit more readable.
Tbh I do not now how to decompose this in any other way. We could call
```
parameters := getParameters(location, armcompute.VirtualMachineSizeTypes(instanceType), networkInterface, imageID, pubkey.Body, vmName, diskName)
virtualMachine, err := c.createVirtualMachine(ctx, resourceGroupName, vmName, parameters)
```

internal/clients/stubs/sources_stub.go
```diff
@@ -18,33 +19,78 @@ type SourcesIntegrationStub struct {
...
-func (mock *SourcesClientStub) GetAuthentication(ctx context.Context, sourceId sources.ID) (*clients.Authentication, error) {
-	return clients.NewAuthentication("arn:aws:iam::230214684733:role/Test", models.ProviderTypeAWS), nil
+func (stub *SourcesClientStub) GetAuthentication(ctx context.Context, sourceId sources.ID) (*clients.Authentication, error) {
+	if sourceId == "1" {
+		return clients.NewAuthentication("arn:aws:iam::230214684733:role/Test", models.ProviderTypeAWS), nil
+	}
+	if auth, ok := stub.auths[sourceId]; ok {
+		return auth, nil
+	}
+	return nil, SourceAuthenticationNotFound
```
avitova, line 93: Happy code path.
```suggestion
	if auth, ok := stub.auths[sourceId]; !ok {
		return nil, SourceAuthenticationNotFound
	}
	return auth, nil
```

avitova (commented):
These are just a few nitpicks, but since this is such a huge PR, I do not want to block this.

ezr-ondrej:
> These are just a few nitpicks, but since this is such a huge PR

Style :policewoman: ! :smile: 

Readability review, :heart: it! More of those (especially on huge PRs) :+1:
Fixed all of your suggestions :)

internal/payloads/reservation_payload.go
```diff
@@ -76,6 +76,37 @@ type AWSReservationResponsePayload struct {
...
+	// The ID of the image from which the instance is created.
+	ImageID string `json:"image_id"`
+
+	// Optional name of the instance(s).
+	Name string `json:"name"`
+
+	// Immediately power off the system after initialization
+	PowerOff bool `json:"poweroff"`
+
+	// Instance IDs, only present for finished reservations
```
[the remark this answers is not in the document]
lzap: @avitova is reading my mind! :-)

internal/jobs/launch_instance_azure.go
```diff
@@ -0,0 +1,94 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("cannot create instance reservation for id %d: %w", instanceId, err)
+	}
+	logger.Debug().Msgf("Created new instance (%s) via Azure CreateVM", *instanceId)
+
+	logger.Debug().Msg("Finished launch instance Azure job")
+
+	return nil
+}
```
lzap, line 94: This function so desperately call for using our "steps" feature :-)

internal/payloads/reservation_payload.go
```diff
@@ -134,6 +163,30 @@ type AWSReservationRequestPayload struct {
...
+	ImageID string `json:"image_id"`
+
+	// Azure Location to deploy into.
+	Location string `json:"location"`
+
+	// Azure Instance type.
+	InstanceSize string `json:"instance_size"`
+
+	// Amount of instances to provision of type: Instance type.
+	Amount int64 ` json:"amount"`
```
lzap, line 181:
```suggestion
	Amount int64 `json:"amount"`
```

internal/services/azure_reservation_service.go
```diff
@@ -0,0 +1,133 @@
...
+		Name:         name,
+	}
+	reservation := &models.AzureReservation{
+		PubkeyID: payload.PubkeyID,
+		SourceID: payload.SourceID,
+		ImageID:  payload.ImageID,
+		Detail:   detail,
+	}
+	reservation.Steps = 1
+	reservation.StepTitles = []string{"Launch instance(s)"}
```
lzap, line 98: This hurts, please add all of the 51 steps that are required in order to launch Azure VM so users get some idea on what is going on when something is not responding.

scripts/rest_examples/reservation-create-azure.http
```diff
@@ -0,0 +1,14 @@
...
+
+{
+  "name": "azure-linux-us-east",
+  "source_id": "5",
+  "image_id": "/subscriptions/4b9d213f-712f-4d17-a483-8a10bbe9df3a/resourceGroups/oezr/providers/Microsoft.Compute/images/composer-api-00e391e7-e6c8-4283-b94a-ea719a58bc05",
+  "amount": 1,
+  "instance_size": "Standard_B1ls",
+  "pubkey_id": 2,
+  "poweroff": true
+}
```
lzap, line 14: Can you please parametrize source_id (already there) and pubkey_id (not yet present [in the parameters](https://github.com/RHEnVision/provisioning-backend/blob/main/scripts/rest_examples/http-client.env.json) I think it is worth adding). Maybe image_id as well?

lzap (changes requested):
An extra space and one big ask: can you add steps when creating reservation and then increase counters at appropriate places with some status texts. With the incoming changes (progress bar replaced with item list) this will also look very fancy and it will be very useful for users to actually see what is happening. Azure isn't as fast as AWS EC2 so this will be great to have.

No other comments, looks great.

avitova:
By the way, let's be careful with this and #403, the AzureTaskArgs have AccountId. :)

internal/services/azure_reservation_service.go
ezr-ondrej: mhm... could we do it in a follow up? That will require a LOT of work as the separate functions are not exported...

ezr-ondrej:
I totaly agree with the steps ask, tho it is a lot of work in already huge PR. IMHO that can be a separate feature.

ezr-ondrej:
> By the way, let's be careful with this and #403, the AzureTaskArgs have AccountId. :)

rebased on top of that.

internal/services/azure_reservation_service.go
ezr-ondrej: Will I get away with https://issues.redhat.com/browse/HMS-1165 ? :)

lzap: Sure, I was thinking in this PR because it is a new code so no problem, but we already had a conflict so...

lzap (approved):
Ok let’s do a followup.

lzap:
Thanks, the third hyperscaler is in.

PR #392, opened 2023-01-25
build: Change unleash prefix to only app name

ezr-ondrej:
Common practice for unleash feature flags on ConsoleDot platform is to prefix the flag with the name of the app only, we initaly drafted the prefix to be `app.provisioning`, but to conform with the practice we should only have `provisioning`.


lzap (approved):
Sure, thanks.

PR #394, opened 2023-01-25
chore: add stacktrace to logging for API and worker

lzap:
During todays debugging session I noticed that when error happens, we do not print stacktrace. I added this to all error payloads (HTTP API pods) and worker pods. Not adding this to statuser where it is pretty obvious the codebase is small there.

Also refactored payload creation a bit, now it is a bit cleaner.


avitova (approved):
LGTM, I do not use redis. Therefore I am not sure about the functionality, errors look good, though.

adiabramovitch:
TY

PR #395, opened 2023-01-26
chore: fix padding and job pointer to body

lzap:
Turns out that arguments must not be pointer because then the type assertion will fail when memory implementation is used:

https://github.com/RHEnVision/provisioning-backend/blob/main/internal/jobs/launch_instance_aws.go#L46-L50

For redis implementation it worked because gob decoder dereferences pointers.

Also added a missing padding in the dev seed script in another commit.

Both changes does not affect stage/prod it was a problem in local development mode (memory only job queue), therefore, I am not creating an issue.

### DIFF
--- a/internal/services/gcp_reservation_service.go
+++ b/internal/services/gcp_reservation_service.go
@@ -104,7 +104,7 @@ func CreateGCPReservation(w http.ResponseWriter, r *http.Request) {
 
 	launchJob := worker.Job{
 		Type: jobs.TypeLaunchInstanceGcp,
-		Args: &jobs.LaunchInstanceGCPTaskArgs{
+		Args: jobs.LaunchInstanceGCPTaskArgs{
 			AccountID:     accountId,
 			ReservationID: reservation.ID,
 			Zone:          reservation.Detail.Zone,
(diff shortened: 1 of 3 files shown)


internal/services/gcp_reservation_service.go
```diff
@@ -104,7 +104,7 @@ func CreateGCPReservation(w http.ResponseWriter, r *http.Request) {
 
 	launchJob := worker.Job{
 		Type: jobs.TypeLaunchInstanceGcp,
-		Args: &jobs.LaunchInstanceGCPTaskArgs{
+		Args: jobs.LaunchInstanceGCPTaskArgs{
 			AccountID:     accountId,
 			ReservationID: reservation.ID,
 			Zone:          reservation.Detail.Zone,
```
lzap, line 110: Both GCP and AWS are affected, obviously, we have tests for Noop job only so this is where it was correct.

avitova (approved):
Great! I do not have redis, but on my setup, this works as expected now. LGTM.

ezr-ondrej:
Interesting, there is https://github.com/RHEnVision/provisioning-backend/commit/2a560d524040312ba3984197b8902dce3f1b1007 deployed in prod and the reservations seems to work, any idea why @lzap ? do we need to deploy this?

lzap:
No need, as I state in the description:

> Both changes does not affect stage/prod it was a problem in local development mode (memory only job queue), therefore, I am not creating an issue.

PR #397, opened 2023-01-29
refactor: Add additional logs to statuser and add invalid requests metric

adiabramovitch:


cmd/pbstatuser/main.go
```diff
@@ -90,6 +90,8 @@ func processMessage(ctx context.Context, message *kafka.GenericMessage) {
 		Identity:           ctxval.Identity(ctx),
 	}
 
+	logger.Trace().Msgf("received a message from sources to be processed with source application id %s", s.SourceApplictionID)
```
ezr-ondrej, line 93: Can we put this higher? Most failures I've seen seemed to be in the fetching of the authentication, so those would not get to this line IMO

cmd/pbstatuser/main.go
```diff
@@ -73,6 +73,8 @@ func processMessage(ctx context.Context, message *kafka.GenericMessage) {
 		return
 	}
 
+	logger.Trace().Msgf("received a message from sources to be processed with source application id %s", s.SourceApplictionID)
```
ezr-ondrej, line 76: We don't have app id yet, but I guess we have source id, which should be enough? :)

adiabramovitch: Yes sorry, it was a mistake. 
That is why I kept it on draft mode. 
Sorry

internal/services/availability_status_service.go
```diff
@@ -24,6 +23,5 @@ func AvailabilityStatus(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	background.EnqueueAvailabilityStatusRequest(&msg)
-	metrics.IncTotalReceivedAvailabilityCheckReqs(err)
 	writeOk(w, r)
 }
```
lzap, line 27: Are we giving up on this counter? Why?

lzap: I mean we could still count all kafka messages received to get what we want.

In addition, we need to add more metrics in order to count success/failure rate without these incorrect source IDs, as Ondra mentioned.

adiabramovitch: this metric is identical to this one and thus redundant: 
provisioning_http_request_total{exported_service="provisioning", path="/api/provisioning/v1/availability_status/sources/"}

deploy/dashboards/grafana-dashboard-provisioning.configmap.yaml
```diff
@@ -599,7 +599,7 @@ data:
                 "type": "prometheus",
                 "uid": "PDD8BE47D10408F45"
               },
-              "expr": "sum(rate(provisioning_source_availability_check_request_total{error=\"false\"}[$__range]))/sum(rate(provisioning_source_availability_check_request_total[$__range]))",
+              "expr": "sum(rate(provisioning_source_availability_check_request_total{exported_service="provisioning"}[$__range]))/sum(rate(provisioning_http_request_total{exported_service="provisioning", path="/api/provisioning/v1/availability_status/sources/"}[$__range]))",
```
lzap, line 602: The service label does not need to be used anywhere. The metric name itself is unique I believe. A nitpick.

internal/services/availability_status_service.go
```diff
@@ -1,29 +1,50 @@
...
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
+			logger.Warn().Msg("Irrelevant request from sources has received")
+			renderError(w, r, payloads.NewNotFoundError(r.Context(), "Not found error", err))
+			return
+		}
+		renderError(w, r, payloads.NewClientError(r.Context(), err))
+		return
+	}
+
```
lzap, line 41: I would *really* like to keep this service ultra-fast just enqueing messages into Kafka, can we do this in the background job?

internal/services/availability_status_service.go
lzap: Oh right, I missed this one.

internal/services/availability_status_service.go
ezr-ondrej: I liked this approach as we could use the HTTP status code to filter out the incoming requests, but if we add a metric in the background job, we are fine I guess :)

lzap: I am fine if you want it this way, it is okay too. It is just it can keep our REST workers more busy than necessary but it's fine.

adiabramovitch: I moved the check to the background job, I wanted to move the `IsProvisioningSource` function to be a part of the sources interface but a cyclic import is present that way. 
Do you have any preferences or suggestions? |
I divided the change into four commits, one of them is the relevant one

lzap:
Changelog problem: https://github.com/RHEnVision/changelog/pull/4

internal/background/availability_queue.go
```diff
@@ -59,8 +85,26 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
...
+
+	_, err = sc.GetAuthentication(ctx, info.asm.SourceID)
+	if err != nil {
+		if errors.Is(err, clients.NotFoundErr) {
+			return nil, fmt.Errorf("could not get source authentication: %w", err)
+		}
+		return nil, fmt.Errorf("could not get source authentication: %w", err)
+	}
+	return &msg, nil
+}
```
lzap, line 110: Why this function is named IsProvisioningSource while it also has a second responsibility of converting a message?

It is more clear to first convert the message at the caller site and then pass the required input (source id) as an argument. Then it becomes a single responsibility function that really just checks the source.

cmd/pbstatuser/main.go
```diff
@@ -162,6 +165,7 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
 	defer processingWG.Done()
 
 	for s := range chGcp {
+		ctxval.Logger(ctx).Trace().Msgf("checking GCP source availability status %s", s.SourceApplicationID)
 		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeGCP, func() error {
 			var err error
 			sr := kafka.SourceResult{
```
lzap, line 171: I’d suggest to extract logger into variable.

lzap (commented):
I am okay, the function with two responsibilities is not ideal tho.

adiabramovitch:
@lzap I have added your suggestions, thank you for the review.

lzap (approved):
TY @ezr-ondrej re-review please

internal/background/availability_queue.go
```diff
@@ -2,20 +2,29 @@ package background
...
 // buffered channel for incoming requests
-var kafkaAvailabilityRequest = make(chan *kafka.GenericMessage, availabilityStatusBatchSize)
+var kafkaAvailabilityRequest = make(chan availabilityRequestInfo, availabilityStatusBatchSize)
 
 // EnqueueAvailabilityStatusRequest prepares a status request check to be sent in the next
 // batch to the platform kafka. Messages can be delayed up to several seconds until sent.
 // The function can block if the enqueueing channel is full.
-func EnqueueAvailabilityStatusRequest(msg *kafka.GenericMessage) {
-	kafkaAvailabilityRequest <- msg
+func EnqueueAvailabilityStatusRequest(identity string, asm *kafka.AvailabilityStatusMessage) {
```
ezr-ondrej, line 26: Why aren't we passing ctx and getting the identity from context here in the function? I feel like it would hide bit of a complexity from the consumer (the service)

internal/background/availability_queue.go
```diff
@@ -31,12 +40,36 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
...
+		logger.Warn().Msg("Could not get sources client")
+		return
+	}
 
 	for {
 		select {
-		case msg := <-kafkaAvailabilityRequest:
-			messageBuffer = append(messageBuffer, msg)
+		case info := <-kafkaAvailabilityRequest:
+			ctx, err = ctxval.WithIdentityFrom64(ctx, info.identity)
```
ezr-ondrej, line 52: We are redefining the context every time, I don't think it is what we want, I'd rather have a new variable for this context here, otherwise the identity can leak into other cases in this function and I don't think it is waht we want.

internal/background/availability_queue.go
```diff
@@ -31,12 +40,36 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
...
+
+			msg, err := info.asm.GenericMessage(ctx)
+			if err != nil {
+				logger.Warn().Msgf("could not convert to generic message %s", err.Error())
+				continue
+			}
+
+			err = IsProvisioningSource(ctx, info, sc)
+			if err != nil {
+				logger.Warn().Msgf("Provisioning app is not enabled in source %s", info.asm.SourceID)
```
ezr-ondrej, line 66: Can we include the Error here? I think there is valuable info in it, I believe there is the the info about what application_types there was found, so we can figure out what we get send and maybe limit the miss send checks in the future.

```suggestion
				logger.Warn().Err(err).Msgf("Provisioning app is not enabled in source %s", info.asm.SourceID)
```

deploy/dashboards/grafana-dashboard-provisioning.configmap.yaml
```diff
@@ -599,7 +599,7 @@ data:
                 "type": "prometheus",
                 "uid": "PDD8BE47D10408F45"
               },
-              "expr": "sum(rate(provisioning_source_availability_check_request_total{error=\"false\"}[$__range]))/sum(rate(provisioning_source_availability_check_request_total[$__range]))",
+              "expr": "sum(rate(provisioning_source_availability_check_request_total[$__range]))/sum(rate(provisioning_http_request_total{path="/api/provisioning/v1/availability_status/sources/"}[$__range]))",
```
ezr-ondrej, line 602: I guess we decided not to use the request count as the total, right?
```suggestion
              "expr": "sum(rate(provisioning_source_availability_check_request_total[$__range]))/sum(rate(provisioning_source_availability_check_request_enqueued_total[$__range]))",
```

internal/metrics/metrics.go
```diff
@@ -14,20 +14,18 @@ import (
...
 		Name:        "provisioning_source_availability_check_request_total",
-		Help:        "availability check requests count partitioned by type (aws/gcp/azure), source status, component and error",
+		Help:        "availability check requests processed successfully or not partitioned by type (aws/gcp/azure), source status, component and error",
 		ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName, "component": "statuser"},
 	},
 	[]string{"type", "status", "error"},
 )
 
-var TotalReceivedAvailabilityCheckReqs = prometheus.NewCounterVec(
+var TotalEnqueueAvailabilityCheckReqs = prometheus.NewCounter(
```
ezr-ondrej, line 23: Could you pls explain to us noobs what is a difference between `CounterVec` and `Counter`? :pleading_face:

internal/services/availability_status_service.go
```diff
@@ -18,12 +18,8 @@ func AvailabilityStatus(w http.ResponseWriter, r *http.Request) {
...
-	msg, err := asm.GenericMessage(r.Context())
-	if err != nil {
-		renderError(w, r, payloads.NewRenderError(r.Context(), "cannot construct message", err))
-		return
-	}
-	background.EnqueueAvailabilityStatusRequest(&msg)
-	metrics.IncTotalReceivedAvailabilityCheckReqs(err)
+
+	identity := identity.GetIdentityHeader(r.Context())
+	background.EnqueueAvailabilityStatusRequest(identity, &asm)
```
ezr-ondrej, line 23: It is a minor thing, but I'd say context and extract identity in the `EnqueueAvailabilityStatusRequest` body.
Just a feeling, no blocker :)

ezr-ondrej (dismissed):
I've got few nits, main thing is the dashboard query :)

internal/background/availability_queue.go
adiabramovitch: sure, I thought it was better to pass the identity according to a similar discussion I had with @lzap  regarding passing the identity/context through a channel. 
If you think it is better to pass the context, ill change it :)

ezr-ondrej: I am thinking passing ctx to the method, method pulls the identity and passes identity to the channel.
I'm 95% behind not sending whole context through channels :)

ezr-ondrej: But in the future we might want to send the identity in different format and if we form it in this file, it's local and much easier to deal with :)

adiabramovitch: OH sorry! I completely missed your point! of course, we should send the context in the function! 
I thought you meant to save the context in the struct 
It was a mistake passing the identity xD, fixing, thank you for noticing

internal/metrics/metrics.go
adiabramovitch: just the labels, I didn't think of helpful labels to use in this new metric so we need only one counter and not a vector of counters for each label. 

`CounterVec is a Collector that bundles a set of Counters that all share the same Desc, but have different values for their variable labels.`

ezr-ondrej: Ah! right, make sense! Thanks for explaining :)

internal/background/availability_queue.go
lzap: Yeah in general, context is great as the first function argument to be passed from call to call. However saving it to database, serializing to struct. sending over channel or over network or saving to file - that is not what you generally want to do.

In that case, you want to save the individual values from context and then create a new context when you receive the value. This is what we do in jobs for example.

There are cases there this might be an overkill, for example when you have 4 goroutines which communicates via channels - the availability check pipeline. But this function is our own API and it should be consistent, so here it should definitely be a context I think.

internal/background/availability_queue.go
```diff
@@ -31,16 +42,40 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
...
 
 	for {
 		select {
-		case msg := <-kafkaAvailabilityRequest:
-			messageBuffer = append(messageBuffer, msg)
+		case info := <-kafkaAvailabilityRequest:
+			newCtx, err := ctxval.WithIdentityFrom64(ctx, info.identity)
+			if err != nil {
+				logger.Warn().Msg("Could not add identity header to ctx")
+				return
```
ezr-ondrej, line 57: Should this be "continue"? Would it be helpful to include the error here as well? It will be easier to debug later on ;)
```suggestion
				logger.Warn().Err(err).Msg("Could not add identity header to ctx")
				continue
```

cmd/pbstatuser/main.go
```diff
@@ -64,6 +64,7 @@ func processMessage(origCtx context.Context, message *kafka.GenericMessage) {
 		logger.Warn().Err(err).Msgf("could not get availability status message %s", err)
 		return
 	}
+	logger.Trace().Msgf("received a message from sources to be processed with source id %s", asm.SourceID)
```
lzap, line 67: I know this will be pesky but:

* errors all lowercase
* logs start uppercase

If only I could change this and keep logs lowercase too.

```suggestion
	logger.Trace().Msgf("Received a message from sources to be processed with source id %s", asm.SourceID)
```

cmd/pbstatuser/main.go
```diff
@@ -109,9 +110,11 @@ func processMessage(origCtx context.Context, message *kafka.GenericMessage) {
 }
 
 func checkSourceAvailabilityAzure(ctx context.Context) {
+	logger := ctxval.Logger(ctx)
 	defer processingWG.Done()
 
 	for s := range chAzure {
+		logger.Trace().Msgf("checking Azure source availability status %s", s.SourceApplicationID)
```
lzap, line 117:
```suggestion
		logger.Trace().Msgf("Checking Azure source availability status %s", s.SourceApplicationID)
```

cmd/pbstatuser/main.go
```diff
@@ -134,6 +137,7 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 	defer processingWG.Done()
 
 	for s := range chAws {
+		logger.Trace().Msgf("checking AWS source availability status %s", s.SourceApplicationID)
```
lzap, line 140:
```suggestion
		logger.Trace().Msgf("Checking AWS source availability status %s", s.SourceApplicationID)
```

internal/background/availability_queue.go
```diff
@@ -31,16 +42,40 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
...
+			}
+
+			msg, err := info.asm.GenericMessage(newCtx)
+			if err != nil {
+				logger.Warn().Msgf("could not convert to generic message %s", err.Error())
```
ezr-ondrej, line 62:
```suggestion
				logger.Warn().Err(err).Msgf("could not generate generic Kafka message")
```

cmd/pbstatuser/main.go
```diff
@@ -162,6 +166,7 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
 	defer processingWG.Done()
 
 	for s := range chGcp {
+		logger.Trace().Msgf("checking GCP source availability status %s", s.SourceApplicationID)
```
lzap, line 169:
```suggestion
		logger.Trace().Msgf("Checking GCP source availability status %s", s.SourceApplicationID)
```

Ok I am giving up. :-D

ezr-ondrej (commented):
Perfect! Last few nits I've noticed.
Mostly just debugging helpers (we should make sure debugging of any issues in Prod is as easy as possible :) )

internal/background/availability_queue.go
```diff
@@ -31,16 +42,40 @@ func sendAvailabilityRequestMessages(ctx context.Context, batchSize int, tickDur
 	logger := ctxval.Logger(ctx)
 	ticker := time.NewTicker(tickDuration)
 	messageBuffer := make([]*kafka.GenericMessage, 0, batchSize)
+	sc, err := clients.GetSourcesClient(ctx)
+	if err != nil {
+		logger.Warn().Msg("Could not get sources client")
+		return
+	}
```
lzap, line 49: Why are we performing any sort of work in the API? This can be done in the background statuser without putting any stress on our API pods. Kafka can handle twice or three times of message every hour, that is exactly why Kafka was designed for - bunch of messages sent or streamed.

This is actually serious design problem because now we have a goroutine running in each API worker pod that is now responsible from doing some REST calls. Because it is just one goroutine (and HTTP handlers spawn new goroutine for each request), it can be overloaded and lagging behind. So we should probably start monitoring the lag and errors (what if sources is down). And this is a chicken and egg problem.

Please can we keep this simple so the API workers just enqueue all checks via Kafka without doing any extra logic. I think the design is clear here - we want to do all the work in statuser so we can throttle it down when we hit any performance or AWS API limits issues.

lzap (changes requested):
I need to repeat my concern about doing any logic in the sender goroutine - that is not a safe thing to do going forward.

adiabramovitch:
@lzap, @ezr-ondrej  Dropped all the previous code and just kept the addition of logs in statuser and changed the metric to count invalid requests.

ezr-ondrej (commented):
Looks great! I believe we've iterated to what we need.

One comment on the new counter, I belive the label is just leftover and we don't need it in this case (for this counter)?
 
@lzap could you also give it a look?

ezr-ondrej (approved):
LGTM :+1:

lzap (approved):
Thanks!

PR #398, opened 2023-01-29
chore: log auth types for Source without provisioning type

ezr-ondrej:


adiabramovitch (approved):

lzap:
Guys we need to be better in giving actual descriptions, specifically to changes without issues. I would appreciate a sample of log line before and after :-)

I think I understand this one tho, thanks :-D

PR #399, opened 2023-01-29
feat(HMSPROV-1133): add ec2 instance description endpoint

amirfefer:
This PR adds  `DescribeInstances` to ec2 client and a corresponding endpont, in order to fetch further information after a successful launch. used here - https://github.com/RHEnVision/provisioning-frontend/pull/188

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -309,6 +309,25 @@ func (c *ec2Client) ListInstanceTypesWithPaginator(ctx context.Context) ([]*clie
 	return instances, nil
 }
 
+func (c *ec2Client) ListInstancesDescription(ctx context.Context, InstanceIds []string) ([]*clients.InstanceDescription, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "istInstancesDescription")
+	defer span.End()
+
+	input := &ec2.DescribeInstancesInput{
+		InstanceIds: InstanceIds,
+	}
+	resp, err := c.ec2.DescribeInstances(ctx, input)
+	if err != nil {
+		if isAWSUnauthorizedError(err) {
+			err = clients.UnauthorizedErr
+		}
+		span.SetStatus(codes.Error, err.Error())
+		return nil, fmt.Errorf("cannot fetch instances description: %w", err)
+	}
+	instancesList := c.parseDescribeInstances(resp)
+	return instancesList, nil
+}
+
 func (c *ec2Client) RunInstances(ctx context.Context, name *string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error) {
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "RunInstances")
 	defer span.End()
@@ -362,6 +381,19 @@ func (c *ec2Client) parseRunInstancesResponse(respAWS *ec2.RunInstancesOutput) [
 	list := make([]*string, len(instances))
 	for i, instance := range instances {
 		list[i] = instance.InstanceId
+	}
+	return list
+}
+
+func (c *ec2Client) parseDescribeInstances(respAWS *ec2.DescribeInstancesOutput) []*clients.InstanceDescription {
+	instances := respAWS.Reservations[0].Instances
+	list := make([]*clients.InstanceDescription, len(instances))
+	for i, instance := range instances {
+		list[i] = &clients.InstanceDescription{
+			ID:   instance.InstanceId,
+			IPV4: instance.PublicIpAddress,
+			DNS:  instance.PublicDnsName,
+		}
 	}
 	return list
 }
--- a/internal/clients/instance_description.go
+++ b/internal/clients/instance_description.go
@@ -1,0 +1,13 @@
+package clients
+
+// InstanceInfo defines a model for an instance description
+type InstanceDescription struct {
+	// The id of the instance
+	ID *string `json:"id,omitempty" yaml:"id"`
+
+	// The public ipv4 dns of the instance
+	DNS *string `json:"dns,omitempty" yaml:"dns"`
+
+	// the public ipv4 of the instance
+	IPV4 *string `json:"ipv4,omitempty" yaml:"ipv4"`
+}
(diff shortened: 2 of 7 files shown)


internal/services/instance_description_service.go
```diff
@@ -0,0 +1,87 @@
...
+	reservation, err := rDao.GetById(r.Context(), id)
+	if err != nil {
+		renderNotFoundOrDAOError(w, r, err, "get reservation detail")
+		return
+	}
+
+	switch reservation.Provider {
+	case models.ProviderTypeUnknown, models.ProviderTypeNoop:
+		if err := render.Render(w, r, payloads.NewReservationResponse(reservation)); err != nil {
+			renderError(w, r, payloads.NewRenderError(r.Context(), "unable to render reservation", err))
```
ezr-ondrej, line 31:
```suggestion
			renderError(w, r, payloads.NewRenderError(r.Context(), "unable to fetch instance information for unknown provider", err))
```

internal/services/instance_description_service.go
```diff
@@ -0,0 +1,87 @@
...
+		reservation, err := rDao.GetAWSById(r.Context(), id)
+		if err != nil {
+			message := fmt.Sprintf("get AWS reservation with id %d", id)
+			renderNotFoundOrDAOError(w, r, err, message)
+			return
+		}
+		instances, err := rDao.ListInstances(r.Context(), id)
+		if err != nil {
+			message := fmt.Sprintf("get reservation with id id %d", id)
+			renderNotFoundOrDAOError(w, r, err, message)
```
ezr-ondrej, line 44:
```suggestion
			renderNotFoundOrDAOError(w, r, err, "failed to fetch instances info")
```

ezr-ondrej (commented):
Mostly nits, tbut the code looks pretty clean overall :)

internal/clients/http/ec2/ec2_client.go
```diff
@@ -366,6 +385,19 @@ func (c *ec2Client) parseRunInstancesResponse(respAWS *ec2.RunInstancesOutput) [
 	return list
 }
 
+func (c *ec2Client) parseDescribeInstances(respAWS *ec2.DescribeInstancesOutput) []*clients.InstanceDescription {
+	instances := respAWS.Reservations[0].Instances
```
lzap, line 389: Can you check bounds here to avoid runtime panic?

internal/clients/instance_description.go
```diff
@@ -0,0 +1,13 @@
...
+type InstanceDescription struct {
+	// The id of the instance
+	ID *string `json:"id,omitempty" yaml:"id"`
+
+	// The public ipv4 dns of the instance
+	DNS *string `json:"dns,omitempty" yaml:"dns"`
+
+	// the public ipv4 of the instance
+	IPV4 *string `json:"ipv4,omitempty" yaml:"ipv4"`
+}
```
lzap, line 13: I would suggest to name both flags "PublicXXX" just in case we want to add private info later on.

And maybe `IPv4` looks slightly better :-)

Just a thought - if IPv6 is also available, I would show it too. By default it will be likely empty because IPv6 you need to enable manually on AWS EC2.

internal/payloads/instance_description_payload.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/go-chi/render"
+)
+
+type InstanceDescriptionResponse struct {
+	*clients.InstanceDescription
+}
+
+func (s *InstanceDescriptionResponse) Bind(_ *http.Request) error {
+	return nil
+}
```
lzap, line 16: If there is no payload for request, you do not need any `Bind` method. You can just delete this.

internal/routes/all_routes.go
```diff
@@ -79,8 +79,11 @@ func MountAPI(r *chi.Mux) {
 			// Different types do have different payloads, therefore TYPE must be part of
 			// URL and not a URL (filter) parameter.
 			r.Route("/{TYPE}", func(r chi.Router) {
-				r.Get("/{ID}", s.GetReservationDetail)
 				r.Post("/", s.CreateReservation)
+				r.Route("/{ID}", func(r chi.Router) {
+					r.Get("/", s.GetReservationDetail)
+					r.Get("/instances", s.ListInstancesDescription)
+				})
```
lzap, line 86: This will need an OpenAPI spec refresh.

I vote for adding some "This endpoint is temporary" documentation so we are able to remove it in the future, I feel like making our service to be inventory is not the direction we should be going forward long-term.

adiabramovitch: Can you please explain why is it a temporary endpoint?

lzap: Sure, we believe that our service should not serve as an inventory. This information (public IP, DNS) should come from subscription manager directly into the inventory, it already sends AWS account and instance ID. I will send you a document in PM.

internal/services/instance_description_service.go
```diff
@@ -0,0 +1,87 @@
...
+		// TODO: move to aws_instance_description service
+		reservation, err := rDao.GetAWSById(r.Context(), id)
+		if err != nil {
+			message := fmt.Sprintf("get AWS reservation with id %d", id)
+			renderNotFoundOrDAOError(w, r, err, message)
+			return
+		}
+		instances, err := rDao.ListInstances(r.Context(), id)
+		if err != nil {
+			message := fmt.Sprintf("get reservation with id id %d", id)
```
avitova, line 43:
```suggestion
			message := fmt.Sprintf("get reservation with id %d", id)
```

internal/services/instance_description_service.go
```diff
@@ -0,0 +1,87 @@
...
+	id, err := ParseInt64(r, "ID")
+	if err != nil {
+		renderError(w, r, payloads.NewURLParsingError(r.Context(), "unable to parse ID parameter", err))
+		return
+	}
+
+	rDao := dao.GetReservationDao(r.Context())
+	reservation, err := rDao.GetById(r.Context(), id)
+	if err != nil {
+		renderNotFoundOrDAOError(w, r, err, "get reservation detail")
```
avitova, line 24: Maybe let's add the id for clarity?

avitova (commented):
These are not blockers.

amirfefer:
close in favor of #448

PR #401, opened 2023-01-31
refactor: Regenerate HTTP clients

adiabramovitch:
In order to enable source filtering by provider type, I wanted to check if sources have fixed their filtering bug. 
In order to do that, I need to regenerate sources' client. 
We need to regenerate the clients as they had a few changes, I have fixed IB conflicts. 
Moreover, the ID type definition was duplicated therefore I added ID to the excluded schemas and checked the launch instance workflow works, as usual, to make sure the change didn't break anything. 
also, I have removed the compatibility section because it was not needed. 

If you can, please check this as well before merging.


ezr-ondrej (approved):
nice! Thanks @adiabramovitch ! :orange_heart:
Did sources fixed the filter issue? :)

adiabramovitch:
> nice! Thanks @adiabramovitch ! orange_heart Did sources fixed the filter issue? :)

Doesnt seems so, for now, we will fetch all of  provisioning sources and filter by provider type on our side

ezr-ondrej:
> Doesnt seems so, for now, we will fetch all of provisioning sources and filter by provider type on our side

That's a bummer, but at least we've updated the clients :)

lzap:
> In order to do that, I need to regenerate sources' client.

GENIUS :-)

PR #403, opened 2023-02-01
feat(HMS-953): Put account id into the context for worker.

avitova:
This PR refers to HMS-953.
There are some additional changes:
AccountId is now in **Job** rather than in TaskArgs.
I have moved the contextLogger to the worker, we have most of the needed information in the **Job**.
We are able to set the reservationID in handlers, we have been calling the contextLogger more times, while the only changed value was, in fact, ReservationID.


internal/jobs/launch_instance_aws.go
```diff
@@ -73,9 +71,7 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
 		return fmt.Errorf("step skipped: %w", err)
 	}
 
-	ctx = ctxval.WithAccountId(ctx, args.AccountID)
-	logger := ctxLogger.With().Int64("reservation_id", args.ReservationID).Logger()
-	logger.Info().Interface("args", args).Msg("Processing pubkey upload AWS job")
+	ctxLogger.Info().Interface("args", args).Msg("Processing pubkey upload AWS job")
```
lzap, line 74: Now that the temporary variable `logger` is gone, could you also rename all `ctxLogger` in all jobs to just `logger`? That is more natural to use, we use this everywhere. The only reason for that was the intermediate variable - Go does not support `variable = &FunctionWhichDoesNotReturnAPointer()` although `to.Ptr` utility function could be used to workaround that.

internal/jobs/launch_instance_aws.go
```diff
@@ -49,7 +46,8 @@ func HandleLaunchInstanceAWS(ctx context.Context, job *worker.Job) {
 		return
 	}
 
-	ctx = contextLogger(ctx, job, args.AccountID, args.ReservationID)
+	logger := ctxval.Logger(ctx).With().Int64("reservation_id", args.ReservationID).Logger()
+	ctx = ctxval.WithLogger(ctx, &logger)
```
lzap, line 50: This could have been written with `ptr.To` but this is okay too we use this all the time.

internal/jobs/launch_instance_gcp.go
```diff
@@ -56,10 +53,7 @@ func HandleLaunchInstanceGCP(ctx context.Context, job *worker.Job) {
 func DoLaunchInstanceGCP(ctx context.Context, args *LaunchInstanceGCPTaskArgs) error {
 	ctxLogger := ctxval.Logger(ctx)
 	ctxLogger.Debug().Msg("Started launch instance GCP job")
-
-	ctx = ctxval.WithAccountId(ctx, args.AccountID)
-	logger := ctxLogger.With().Int64("reservation_id", args.ReservationID).Logger()
-	logger.Info().Interface("args", args).Msg("Processing launch instance GCP job")
+	ctxLogger.Info().Interface("args", args).Msg("Processing launch instance GCP job")
```
lzap, line 56: Ditto, just use `logger` instead of `ctxLogger`.

internal/services/aws_reservation_service.go
```diff
@@ -24,7 +24,7 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 	logger := *ctxval.Logger(r.Context())
 
 	var accountId int64 = ctxval.AccountId(r.Context())
-	var identity identity.XRHID = ctxval.Identity(r.Context())
+	var id identity.XRHID = ctxval.Identity(r.Context())
```
lzap, line 27: 👍🏻

internal/services/aws_reservation_service.go
```diff
@@ -126,10 +126,10 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
 	}
 
 	launchJob := worker.Job{
-		Type:     jobs.TypeLaunchInstanceAws,
-		Identity: identity,
+		Type:      jobs.TypeLaunchInstanceAws,
+		Identity:  id,
+		AccountID: accountId,
```
lzap, line 131: The `Enqueue` function could theoretically set identity and account id automatically from the context when it is blank (or zero). But this is fine.

lzap (changes requested):
Nice job, I only have variable rename request for consistency but this is great.

Now, normally we should introduce a job version flag, but since we are only in prod-beta and there is handful of users it is very unlikely that there would be jobs in queue during upgrade of the app. Therefore, I suggest to merge as-is without special version treatment. @ezr-ondrej

avitova:
Thank you for the review. I have changed the `ctxLogger` variable as recommended. :+1:

lzap (approved):
Nice and clean.

Since this is in the critical path, I would like someone else to give this a read and then merge it. Thanks!

lzap:
QA: No changes, internal only improvement.

ezr-ondrej (approved):
Nice work :+1: Seeing this, I believe it might be good idea (in a follow up) to have `NewJob` function, that would take care of this (so we wouldn't forget to add Identity or AccountID) :)

lzap:
Alternatively, as I said in a comment, the `Enqueue` function could detect blank values and put them into place if needed. Whatever you like is a valid choice.

PR #405, opened 2023-02-02
chore: silence switch-case linter

lzap:
Quick reminder - it is fine to silence linters at some times where it makes no sense. I find this one quite useful because I already forgot some case statements myself that could lead to dangerous code, but here it is not very useful.

Also my version of golint already warns about the out-of-package reassignment, it is okay as Zerolog library works this way.


adiabramovitch (approved):

ezr-ondrej:
> but here it is not very useful.

I'd say it was useful :) It was exhaustive, but if there is additional type in Sources client we know about it. It might be other app, but it might be another provisioning provider.

That's why I've added it, it was not because I didn't know I can disable the lint rule, but because the rule made me realize why it's useful to keep it :)

PR #406, opened 2023-02-02
fix(HMS-1181): ignore pubkey resource deletion without SA

lzap:
We currently error out with "authentication for source not found" but there is nothing we can do when source is gone. Therefore we will just print warning in log instead and return 200.

https://issues.redhat.com/browse/HMS-1181

### DIFF
--- a/internal/services/pubkey_service.go
+++ b/internal/services/pubkey_service.go
@@ -1,10 +1,12 @@
 package services
 
 import (
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


lzap:
When testing this, sources once returned 404 and for the same account/pubkey it returned 500. I am unable to find further details about the transaction in logs in sources, because there is not transaction id, so I am making both errors to be a warning in our log. It is not big deal.

PR #407, opened 2023-02-02
refactor: Add step in Azure job

ezr-ondrej:
Split Resource group ensure into a job step.

Refs HMS-1165

This is a first split, I'll continue with this once we settle this is how we want to do it.

### DIFF
--- a/internal/clients/interface.go
+++ b/internal/clients/interface.go
@@ -90,11 +90,14 @@ var GetServiceAzureClient func(ctx context.Context) (ServiceAzure, error)
 type Azure interface {
 	ClientStatuser
 
+	// EnsureResourceGroup makes sure that group with give name exists in a location
+	EnsureResourceGroup(ctx context.Context, name string, location string) (*string, error)
+
 	// CreateVM creates Azure virtual machine.
 	// Most of the parameters are constant for now.
 	// imageID is expected in format /subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/images/{imageID}
 	// sshKeyBody should be full public key body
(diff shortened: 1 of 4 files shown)


internal/services/azure_reservation_service.go
```diff
@@ -94,8 +94,8 @@ func CreateAzureReservation(w http.ResponseWriter, r *http.Request) {
 		ImageID:  payload.ImageID,
 		Detail:   detail,
 	}
-	reservation.Steps = 1
-	reservation.StepTitles = []string{"Launch instance(s)"}
+	reservation.Steps = int32(len(jobs.LaunchInstanceAzureSteps))
+	reservation.StepTitles = jobs.LaunchInstanceAzureSteps
```
lzap, line 98: Yeah Steps is sort of useless now since I introduced StepTitles.

internal/clients/interface.go
```diff
@@ -90,11 +90,14 @@ var GetServiceAzureClient func(ctx context.Context) (ServiceAzure, error)
 type Azure interface {
 	ClientStatuser
 
+	// EnsureResourceGroup makes sure that group with give name exists in a location
+	EnsureResourceGroup(ctx context.Context, name string, location string) (*string, error)
+
```
lzap, line 95: Yeah bummer to do that for all steps, maybe you can figure out a solution (passing some interface in), but that really sounds clunky this is the cleanest approach I think.

lzap:
This is okay, you want me to merge or wait for other steps?

internal/clients/interface.go
ezr-ondrej: Yeah, I'll see how it goes with other steps, we might want to put some steps together, maybe "Prepare networking resources" etc :)

lzap (approved):
Ok thanks.

PR #409, opened 2023-02-02
docs: Document Azure setup

ezr-ondrej:
Add docs on how to setup Azure account for testing the deployment.

Fixes HMS-1147

### DIFF
--- a/docs/configure-azure.md
+++ b/docs/configure-azure.md
@@ -1,0 +1,51 @@
+# Configuring Azure deployment
+
+The provisioning service uses Azure SDK to perform all provisioning tasks.
+We follow the best possible security standard and practices.
+In this article, we will describe how to create necessary service account setup and configure this application with it.
+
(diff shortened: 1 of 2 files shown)


lzap (changes requested):
It is fine, small wording.

Are you planning to also create stage/prod JSON files and links which can be used from this document directly? I am assuming a lot of stage users will have to go through this ugly JSON editing. It is perfectly safe to publish our tenant UUIDs.

I guess we don’t have the app ID created yet, so later I guess. That is fine.

ezr-ondrej:
> Are you planning to also create stage/prod

I think we will need to publish them on some formal URL provided by RH for stage and prod, so I expect to have the URL directly in the Sources definition once we have account.

This is just dev docs, how would one define the JSON, but I don't expect our stage/prod users to ever doing it.

lzap:
Thanks!

PR #412, opened 2023-02-07
Support for AWS Launch Templates

lzap:
This patch introduces new client type:

```go
package clients

// LaunchTemplate represents a generic launch template for a hyperscaler.
type LaunchTemplate struct {
	// ID is an identifier, for example "lt-94397398248932342" for AWS EC2.
	ID string

	// Name describes the launch template, user defined.
	Name string
}
```

I am aiming to have a generic LaunchTemplate, the terminology we use for generic types is AWS EC2. I am assuming other hyperscalers have similar templating, well, except Azure where it is JSON. There is a new endpoint and OpenAPI spec update which lists those. The UI is supposed to load this once region is selected since region is required URL parameter (or error). The ID field must be then provided as a new field "LaunchTemplateID" in an AWS reservation detail or set to blank when no template should be used.

The RunInstances function was extended with new "LaunchTemplateID" string argument, when empty no template will be used. The function has always accepted empty instance type, no change here. A new check is added in the service function and error is returned when user provides empty string for BOTH instance type and template id. It is technically possible to create a template that does nothing (carries zero information), we currently do not check for this so it is possible that the launch will error out with insufficient data (instance type not set explicitly and also in a template).

[HMS-1209](https://issues.redhat.com/browse/HMS-1209)

### DIFF
--- a/cmd/typesctl/providers/ec2_types.go
+++ b/cmd/typesctl/providers/ec2_types.go
@@ -56,7 +56,7 @@ func generateTypesEC2() error {
 		if regionErr != nil {
 			return fmt.Errorf("unable to get regional EC2 client: %w", regionErr)
 		}
-		instTypes, regionErr := client.ListInstanceTypesWithPaginator(ctx)
+		instTypes, regionErr := client.ListInstanceTypes(ctx)
 		if regionErr != nil {
 			// skip
 			fmt.Printf("unable to list EC2 instance types (region STS not enabled?): %s\n", regionErr.Error())
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -279,8 +279,8 @@ func (c *ec2Client) ListAllZones(ctx context.Context, region clients.Region) ([]
 	return result, nil
 }
 
-func (c *ec2Client) ListInstanceTypesWithPaginator(ctx context.Context) ([]*clients.InstanceType, error) {
-	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListInstanceTypesWithPaginator")
+func (c *ec2Client) ListInstanceTypes(ctx context.Context) ([]*clients.InstanceType, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListInstanceTypes")
 	defer span.End()
 
 	input := &ec2.DescribeInstanceTypesInput{MaxResults: ptr.ToInt32(100)}
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -309,7 +309,37 @@ func (c *ec2Client) ListInstanceTypes(ctx context.Context) ([]*clients.InstanceT
 	return instances, nil
 }
 
-func (c *ec2Client) RunInstances(ctx context.Context, name *string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error) {
+func (c *ec2Client) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchTemplate, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListLaunchTemplates")
+	defer span.End()
+
+	input := &ec2.DescribeLaunchTemplatesInput{MaxResults: ptr.ToInt32(100)}
+	pag := ec2.NewDescribeLaunchTemplatesPaginator(c.ec2, input)
+
+	var res []*clients.LaunchTemplate
+	for pag.HasMorePages() {
+		resp, err := pag.NextPage(ctx)
+		if err != nil {
+			if isAWSUnauthorizedError(err) {
+				err = clients.UnauthorizedErr
+			}
+			span.SetStatus(codes.Error, err.Error())
+			return nil, fmt.Errorf("cannot list launch templates: %w", err)
+		}
+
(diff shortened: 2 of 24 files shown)


cmd/typesctl/providers/ec2_types.go
```diff
@@ -56,7 +56,7 @@ func generateTypesEC2() error {
 		if regionErr != nil {
 			return fmt.Errorf("unable to get regional EC2 client: %w", regionErr)
 		}
-		instTypes, regionErr := client.ListInstanceTypesWithPaginator(ctx)
+		instTypes, regionErr := client.ListInstanceTypes(ctx)
 		if regionErr != nil {
 			// skip
 			fmt.Printf("unable to list EC2 instance types (region STS not enabled?): %s\n", regionErr.Error())
```
lzap, line 62: This rename was done in a separate commit, I wanted to do this because the newly introduced function also uses paginator, but I do not think this is something that should be exposed through the API.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -309,7 +309,37 @@ func (c *ec2Client) ListInstanceTypesWithPaginator(ctx context.Context) ([]*clie
...
+			t := clients.LaunchTemplate{
+				ID:   ptr.From(awsTemplate.LaunchTemplateId),
+				Name: ptr.From(awsTemplate.LaunchTemplateName),
+			}
+			res = append(res, &t)
+		}
+	}
+
+	return res, nil
+}
```
lzap, line 340: New function, returns just a list of templates, easy.

internal/clients/launch_template.go
```diff
@@ -0,0 +1,10 @@
+package clients
+
+// LaunchTemplate represents a generic launch template for a hyperscaler.
+type LaunchTemplate struct {
+	// ID is an identifier, for example "lt-94397398248932342" for AWS EC2.
+	ID string `json:"id" yaml:"id"`
+
+	// Name describes the launch template, user defined.
+	Name string `json:"name" yaml:"name"`
+}
```
lzap, line 10: The new type which I hope it can be generic (except Azure).

internal/services/aws_reservation_service.go
```diff
@@ -35,23 +35,38 @@ func CreateAWSReservation(w http.ResponseWriter, r *http.Request) {
...
+		if it.Architecture.String() != supportedArch {
+			renderError(w, r, payloads.NewWrongArchitectureUserError(r.Context(), ArchitectureMismatch))
+			return
+		}
+	}
+
+	launchString := "Launch instance"
+	if payload.Amount > 1 {
+		launchString = "Launch instance(s)"
 	}
```
lzap, line 62: Sneaking a small change - singulár vs plural as we know the amount. But this could actually be bad as we discussed during scrum today - these will be likely constants for i18n with descriptions. Therefore I think I will remove this.

scripts/rest_examples/http-client.env.json
```diff
@@ -6,7 +6,9 @@
     "prefix": "api/provisioning/v1",
     "identity": "eyJpZGVudGl0eSI6IHsidHlwZSI6ICJVc2VyIiwgImFjY291bnRfbnVtYmVyIjoiMTMiLCAiaW50ZXJuYWwiOnsib3JnX2lkIjoiMDAwMDEzIn19fQo=",
     "source_id": "1",
+    "pubkey_id": "1",
     "region": "us-east-1",
+    "launch_template_id": "",
     "reservation-get-id": "1"
   }
 }
```
lzap, line 14: Generalized pubkey id too, that is useful as 1 is my key, others should use their own keys (and add them to the seed script).

lzap:
Some edge API errors.

/retest

lzap:
Tests are now passing as expected.

akhil-jha:
Probably needs a rebase as well :)

lzap:
The missing step_titles is a regression which I am fixing here: https://github.com/RHEnVision/provisioning-backend/pull/425

I fixed the missing info in the reservation detail. Rebased.

akhil-jha (approved):

ezr-ondrej (approved):
I'm fine with it, but given the amount of refactoring, would you also double check it pls @adiabramovitch ? :)

ezr-ondrej:
@lzap we do want to hide it only in the UI behind the flag we said, right? :thinking:
Should we fail passing in the `launch_template` if the flag is disabled here as well? :thinking:

lzap:
> @lzap we do want to hide it only in the UI behind the flag we said, right?

I feel like this is sufficient, yet. Tell me if you think otherwise.

> Should we fail passing in the launch_template if the flag is disabled here as well? 🤔

Nah, I mean is that a really problem that someone tries our own (documented) API for a feature that hasn’t been properly implemented in the UI? It should work, dunno.

One idea - I think it is reasonable to keep the list of flags small, if we end up with 1000 combinations of feature flags then it will be nightmare to test all states. Also once we enable something on our RH account we start to see a different code path than our prod customers.

I think what could be reasonable here is "if I am running on prod, error out" type of explicit check in the code if you want this.

ezr-ondrej:
I believe we should keep the feature flags short lived. IMHO once the
feature is tested and documented, feature flag should be removed and the
feature should be available to all.

Dne čt 16. 2. 2023 14:16 uživatel Lukáš Zapletal ***@***.***>
napsal:

ezr-ondrej (approved):
Thanks @lzap ! :orange_heart:

PR #413, opened 2023-02-07
feat(HMS-1110): Get RoleName and add iam client

avitova:
This prepares iam service, as we will need this to call policy-relates methods. Knowing a "friendly role name" is also required to call some of these methods. It is possible to get ARN from the role name but not the other way around, so there is a getter to parse ARN. If we have a way to get a role name, let me know, but I did not find a way to do that.
There are some go.mod changes as we did not have iam before; I'd be happy if this does not break anything.


lzap:
We are taking a look on the IQE failures, this is irrelevant to your changes. Please fix linter tho :-)

internal/clients/http/ec2/ec2_client.go
```diff
@@ -28,6 +30,7 @@ const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/h
 type ec2Client struct {
 	ec2     *ec2.Client
 	sts     *sts.Client
+	iam     *iam.Client
 	assumed bool
 }
```
lzap, line 35: This seems unused, would you mind to continue and add the next step in this PR?

internal/clients/http/ec2/ec2_client.go
```diff
@@ -378,3 +383,18 @@ func (c *ec2Client) GetAccountId(ctx context.Context) (string, error) {
...
+	}
+	roleName := strings.Split(arnParts[len(arnParts)-1], "/")
+	if len(roleName) != 2 {
+		return "", fmt.Errorf("%w: ARN has incorrect syntax: %s rolename parsing result: %s",
+			http.ARNParsingError, arn, roleName)
+	} else if roleName[0] != "role" {
+		return "", fmt.Errorf("%w: ARN does not have any role: %s", http.ARNParsingError, arn)
+	}
+	return roleName[1], nil
+}
```
lzap, line 400: Don’t export the function if you are not going to use it externally.

lzap (approved):
It’s all good, but it is so tiny patch and iam variable is unused, could you continue and add more functionality? Just a step further, don’t do too big changes.

internal/clients/http/ec2/ec2_client.go
avitova: Sure, I had a follow-up prepared, but I was not sure if this would not be too much.

internal/services/aws_iam_service.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/payloads"
+	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/render"
+)
+
+func ValidatePermissions(w http.ResponseWriter, r *http.Request) {
```
avitova, line 14: This function currently only lists attached role policies, but the idea is to do the rest for validation here. I do not expect we would need the endpoint to use only ListAttachedPolicies, but I left the function call there so that we'd see the outcome.

internal/clients/interface.go
```diff
@@ -79,6 +80,9 @@ type EC2 interface {
 	RunInstances(ctx context.Context, name *string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error)
 
 	GetAccountId(ctx context.Context) (string, error)
+
+	// ListAttachedRolePolicies returns list of all policies for the given arn role
+	ListAttachedRolePolicies(ctx context.Context, auth *Authentication) ([]*iamTypes.AttachedPolicy, error)
 }
```
lzap, line 86: We must not expose internal types through our EC2 interface, it is here to hide implementation details. Similarly as in other fucntions, please create a new type in this `clients` package and copy relevant values you want to carry over to the caller. Typically it is just minimum set of fields you need to do what you want to achieve. Perhaps some sort of "id" in this case.

But before you do that, consider moving the logic from service package in here. In that case a simple function would do the check and return some struct with result and error (for runtime errors):

```go
struct PolicyCheckResult struct {
  Valid bool
  Missing []string
}

func CheckPolicy(ctx) (PolicyCheckResult, error)
```

That is a different approach, not that flexible, however, it will save you time creating all those custom types. Whatever you like.

internal/services/aws_iam_service.go
```diff
@@ -0,0 +1,58 @@
...
+		renderError(w, r, payloads.NewAWSError(r.Context(), "unable to get AWS EC2 client", err))
+		return
+	}
+
+	logger.Info().Msgf("Listing policies attached to the role.")
+	policies, err := ec2Client.ListAttachedRolePolicies(r.Context(), authentication)
+	if err != nil {
+		renderError(w, r, payloads.NewAWSError(r.Context(), "unable to list AWS EC2 instances", err))
+		return
+	}
```
lzap, line 52: Consider moving this code logic from here (service package) to the EC2 client (clients package) if you want to avoid creating too many custom types.

Or if you like it this way, I am okay, it is just more work. It is better because then we can re-use these functions later on. That is fine too, just little bit more work.

lzap (changes requested):
Nice I have one request on the custom type and API boundary leak.

Plus one additional thing to consider: make your pick and I think the scope is okay this can be merged then.

internal/services/aws_iam_service.go
adiabramovitch: Is this endpoint going to serve validation to Azure and GCP as well?

internal/payloads/policies_payload.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"net/http"
+
+	iamTypes "github.com/aws/aws-sdk-go-v2/service/iam/types"
+
+	"github.com/go-chi/render"
+)
+
+// Prevzato z Instance types payload
```
adiabramovitch, line 11: What is Prevzato? :)

internal/services/aws_iam_service.go
lzap: That is my thinking going forward, yes. One of the reasons why I would prefer a simple string-based interface.

internal/clients/interface.go
avitova: I like the idea of returning an array of missing permissions. We should not expose anything we do not want. Let me know if the current approach is what we want.

internal/payloads/permissions_payload.go
```diff
@@ -0,0 +1,28 @@
...
+	MissingPolicies []string
+}
+
+func (s *PermissionsResponse) Bind(_ *http.Request) error {
+	return nil
+}
+
+func (s *PermissionsResponse) Render(_ http.ResponseWriter, _ *http.Request) error {
+	return nil
+}
```
lzap, line 20: Response needs Render. Reply needs Bind. I think we got that incorrectly in other parts of the app too, no big deal.

internal/payloads/permissions_payload.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"net/http"
+
+	"github.com/go-chi/render"
+)
+
+type PermissionsResponse struct {
+	Valid           bool
+	MissingPolicies []string
```
lzap, line 11: The only thing is this name - I am not sure if on GCP these will be policies. I suggest to rename this to MissingEntities or just Errors where you put strings like "missing policy XXX", "missing policy YYY" or something like that.

For example in Azure it is called Roles and Permissions.

lzap (changes requested):
One small field rename and I like it.

I think we can merge, @adiabramovitch ?

ezr-ondrej:
oh, sorry I've merged other PRs and now we have conflict :(

internal/payloads/permissions_payload.go
avitova: Okay, let's get rid of Bind.

lzap (approved):
I am good. @adiabramovitch

adiabramovitch (approved):
looks great 
Just one small question 
Does the validation go to take place in checkPermissions function? are the policies going to be compared there?

avitova:
> looks great Just one small question Does the validation go to take place in checkPermissions function? are the policies going to be compared there?

Great question. I intend to do it this way, yes. :)

PR #414, opened 2023-02-08
feat(HMS-969): List and filter sources by their provider

adiabramovitch:
We are receiving a list of sources from sources team, due to a bug we cannot filter the sources according to their source types. 
This patch adds the hyperscaler query param support, fetching the source types supported by the sources team and filtering only the relevant for our service. 
I also added tests to sources service and sources client.


internal/clients/http/sources/sources_client_test.go
```diff
@@ -74,3 +74,71 @@ func TestSourcesClient_GetAuthentication(t *testing.T) {
...
+			require.NoError(t, err, "failed to write http body for stubbed server")
+		})
```
adiabramovitch, line 123: The response is huge, should I extract the responses to a different file?
 maybe create a fixtures folder for those responses.

internal/clients/http/sources/sources_client_test.go
```diff
@@ -74,3 +74,71 @@ func TestSourcesClient_GetAuthentication(t *testing.T) {
...
+		require.NoError(t, err, "failed to write http body for stubbed server")
+	})
+
+	mux.HandleFunc(fmt.Sprintf("/application_types/%s/sources", "5"), func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		_, err := io.WriteString(w, `{"data":[{"availability_status":"in_progress","id":"1","created_at":"2023-02-02T16:06:00Z","updated_at":"2023-02-02T16:06:00Z","name":"Amazon source","uid":"384ccb67-5838-4571-811f-d725a0a57b6f","app_creation_workflow":"manual_configuration","source_type_id":"1"},{"availability_status":"in_progress","id":"3","created_at":"2023-02-02T16:06:00Z","updated_at":"2023-02-02T16:06:00Z","name":"Google source","uid":"cb6fa946-395a-4650-ae6c-83501775ca4e","app_creation_workflow":"manual_configuration","source_type_id":"6"}],"meta":{"count":3,"limit":100,"offset":0},"links":{"first":"/api/sources/v3.1/sources?limit=100&offset=0","last":"/api/sources/v3.1/sources?limit=100&offset=100"}}`)
```
lzap, line 90: Assuming this is copy pasted from real database, in the example above the "links":"first" path matches the handler path, but here I see a difference:

* handler path is `/application_types/%s/sources`
* and links is `/api/sources/v3.1/sources?limit=100&offset=0`

Just checking if this is just some typo in the test path or you pasted a wrong line.

internal/clients/http/sources/sources_client_test.go
lzap: Good point, I think this passed the readability threshold for me and I would appreciate separate files for all cases, ideally with JSON-formatted content so we can easily track diffs in git when Sources update their API.

internal/services/sources_service.go
```diff
@@ -9,16 +9,31 @@ import (
...
+
 	client, err := clients.GetSourcesClient(r.Context())
 	if err != nil {
 		renderError(w, r, payloads.NewClientError(r.Context(), err))
 		return
 	}
-	sourcesList, err := client.ListProvisioningSources(r.Context())
-	if err != nil {
-		renderError(w, r, payloads.NewClientError(r.Context(), err))
+	hyperscaler := r.URL.Query().Get("hyperscaler")
```
lzap, line 19: We already use "Provider" term in multiple places in our OpenAPI, let’s not introduce another term for the same thing.

```suggestion
	hyperscaler := r.URL.Query().Get("provider")
```

internal/services/sources_service.go
```diff
@@ -9,16 +9,31 @@ import (
...
+	if hyperscaler == "gcp" || hyperscaler == "azure" || hyperscaler == "aws" {
```
lzap, line 20: See my comment about clean API, here you could just use our own `ProviderType` type, then you can use existing function named `ProviderTypeFromString` to actually do this check in a more elegant way:

    if models.ProviderTypeFromString(param) != models.ProviderTypeUnknown {

internal/services/sources_service.go
```diff
@@ -9,16 +9,31 @@ import (
...
+			return
+		}
+	} else if hyperscaler == "" {
+		sourcesList, err = client.ListAllProvisioningSources(r.Context())
+		if err != nil {
+			renderError(w, r, payloads.NewClientError(r.Context(), err))
+			return
+		}
+	} else {
+		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), "hyperscaler is not supported", clients.UnknownHyperscalerErr))
```
lzap, line 33:
```suggestion
		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), fmt.Sprintf("unknown provider: %s", param), clients.UnknownHyperscalerErr))
```

lzap (changes requested):
Very nice, I have some comments about clean API but other than that, good.

internal/clients/http/sources/sources_client_test.go
adiabramovitch: Yes a wrong copy-pasting, fixing

lzap (approved):
Nice and clean :-) Thanks.

PR #416, opened 2023-02-08
fix: allow setting proxy per client

ezr-ondrej:
We are sharing the transport, so if the proxy is set for one client, it got set to all.

This isolates the clients by creating new http.Transport if a proxy is set for given client.

This should affect only devel instances so Production should still utilize caching.

Fixes HMS-1227


lzap (approved):
Okay what was the issue? I mean, if it works, I am fine. Just curious.

ezr-ondrej:
> Okay what was the issue? I mean, if it works, I am fine. Just curious.

As described above :) If I set Proxy for Image builder but not for Sources, it uses the proxy for both anyway :)
I guess setting different proxies would not work eighter.

lzap:
Ah sorry I missed that, nasty bug, sorry about that :-)

PR #417, opened 2023-02-08
feat: respect Amount in Azure deployments

ezr-ondrej:
Deploy multiple Azure instances at once.
We need to generate unique names for the instances, so we generate uuid suffix for each of them.

Fixes HMS-1146


lzap (approved):
Yeah thanks.

PR #418, opened 2023-02-09
chore: enable passing Azure params

ezr-ondrej:
This is a temporary workaround until we have secrets in place for all environments.


lzap (approved):
Merging.

PR #422, opened 2023-02-10
chore: add edge request id into logger

lzap:
This adds edge request id from headers into logs. This should be according to folks on the forum slack channel the best request id to log along for correlation with the edge API.

Also adds documentation link to the README.

### DIFF
--- a/internal/ctxval/edge_id.go
+++ b/internal/ctxval/edge_id.go
@@ -1,0 +1,17 @@
+package ctxval
+
+import "context"
+
+// EdgeRequestId returns edge API (3Scale) request id or an empty string when not set.
+func EdgeRequestId(ctx context.Context) string {
+	value := ctx.Value(edgeRequestIdCtxKey)
+	if value == nil {
+		return ""
+	}
+	return value.(string)
+}
+
+// WithEdgeRequestId returns context copy with trace id value.
+func WithEdgeRequestId(ctx context.Context, id string) context.Context {
+	return context.WithValue(ctx, edgeRequestIdCtxKey, id)
+}
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Provisioning backend service for cloud.redhat.com.
 
 Requirements: Go 1.18+
 
+[![Go Reference](https://pkg.go.dev/badge/github.com/RHEnVision/provisioning-backend.svg)](https://pkg.go.dev/github.com/RHEnVision/provisioning-backend)
+
 ## Components
 
 * pbapi - API backend service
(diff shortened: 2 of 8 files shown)


ezr-ondrej:
Could you pls follow up on what's edge id? is that the edge service, or what edge are we talking and why do we want it's id in logs as "request id"? :open_mouth:

lzap:
Sure, this is an id generated by our edge API (3scale). It appears that the OpenTelemetry effort is not moving forward too much, so I would like our app to write it to logs and also contribute the same into sources/ib too. This will be much easier to get merged, for example the otel PR in sources seems to be stuck.

internal/ctxval/edge_id.go
```diff
@@ -0,0 +1,17 @@
+package ctxval
+
+import "context"
+
+// EdgeRequestId returns request id or an empty string when not set.
+func EdgeRequestId(ctx context.Context) string {
```
ezr-ondrej, line 6: Could we pls document here that it is coming from edge gateway? :)

ezr-ondrej (commented):
So do we want to add it as otel correlation id as well and pass the `X-Rh-Edge-Request-Id` header to requests within our platform?

internal/ctxval/edge_id.go
lzap: Done

lzap:
Yeah I am thinking both, in dev setup it will be obviously empty.

ezr-ondrej:
> Yeah I am thinking both

So it will be done in follow-up PRs? :)
I'm happy to merge this as is if that's the case :+1:

ezr-ondrej:
/retest
well the bot ain't happy for some reason

ezr-ondrej (approved):

lzap:
> So it will be done in follow-up PRs? :)

What I mean is I would like to log both trace_id (OpenTel id) and request_id (Edge id) in all logs (if available).

But do not merge yet, I forgot to add important part - passing the edge request id to sources/ib HTTP headers! That is primary reason for this patch.

lzap:
Added the header passing.

ezr-ondrej:
Can't we use the edge as request id instead of generating it?

lzap:
Trace id is required and it must have specific properties defined by OpenTelemetry spec. Edge api is generated by 3scale and does not conform to this specs. I refactored the code that if there is no edge API (ephemeral, dev) nothing is logged or passed actually. Would that work? I would really like to have both options for searching.

ezr-ondrej:
> Would that work? I would really like to have both options for searching.

Sure, sure, I :heart: the addition, I just did not understand it and then wanted to make sure we are getting the best out of it :+1:

ezr-ondrej (approved):
Loving it :+1:
Do we want to add this to Image builder as well? That would be awesome! :orange_heart:

lzap:
That is my plan as a follow up yes.

lzap:
https://github.com/osbuild/image-builder/pull/628

PR #423, opened 2023-02-10
feat: Azure lighthouse offering template

ezr-ondrej:
Might not be ideal solution to the problem , I feel it being bit nasty, but it gives us quite some benefits:
- We are testing the template we use in documentation, cool :tada:
- We can't forget change the tenant in template, because it's dynamic
- We can change the template in code. Cool.

Happy to hear improvements on the technical solution tho :)
I would whole heartedly agree it has quite some quirks :stuck_out_tongue: 

-----

Adds lighthouse offering template rendering.
This makes the template available on a public URL.
Reuses the documentation template.

Also improves the documentation of Why in the Azure devel doc.

Refs HMS-1148

-----

Add lighthouse template to the container image during build.
Fixes the build container on 1.18 tag
Adds latest to ubi-minimal

### DIFF
--- a/config/api.env.example
+++ b/config/api.env.example
@@ -41,7 +41,7 @@
 #   TELEMETRY_ENABLED bool
 #     	open telemetry collecting (default "false")
 #   CLOUDWATCH_ENABLED bool
-#     	cloudwatch logging exporter (default "false")
+#     	cloudwatch logging exporter (enabled in clowder) (default "false")
 #   CLOUDWATCH_REGION string
 #     	cloudwatch logging AWS region (default "")
 #   CLOUDWATCH_KEY string
@@ -66,14 +66,16 @@
 #     	AWS service account logging (verbose) (default "false")
 #   AZURE_TENANT_ID string
 #     	Azure service account tenant id (default "")
-#   AZURE_SUBSCRIPTION_ID string
-#     	Azure service account subscription id (default "")
 #   AZURE_CLIENT_ID string
 #     	Azure service account client id (default "")
 #   AZURE_CLIENT_SECRET string
 #     	Azure service account client secret (default "")
+#   AZURE_CLIENT_NAME string
+#     	Azure service account display name (default "RH HCC")
 #   AZURE_DEFAULT_REGION string
 #     	Azure region when not provided (default "eastus")
+#   AZURE_SUBSCRIPTION_ID string
+#     	Azure service account subscription id (default "")
 #   GCP_PROJECT_ID string
 #     	GCP service account project id (default "")
 #   GCP_JSON string
@@ -88,14 +90,32 @@
 #     	open telemetry HTTP context pass and trace (default "true")
 #   WORKER_QUEUE string
 #     	job worker implementation (memory, redis, sqs, postgres) (default "memory")
-#   WORKER_CONCURRENCY int
-#     	number of goroutines handling jobs (default "50")
-#   WORKER_HEARTBEAT int64
-#     	heartbeat interval (time interval syntax) (default "30s")
-#   WORKER_MAX_BEATS int
-#     	maximum amount of heartbeats allowed (default "10")
+#   WORKER_POLL_INTERVAL int64
+#     	polling interval (network timeout) (default "5s")
+#   WORKER_MAX_THREADS int
+#     	maximum allowed amount of polling goroutines (CPUs + 1) (default "64")
+#   UNLEASH_ENABLED bool
+#     	unleash service (feature flags) (default "false")
+#   UNLEASH_ENVIRONMENT string
+#     	unleash environment (default "")
+#   UNLEASH_PREFIX string
+#     	unleash flag prefix (default "provisioning")
+#   UNLEASH_URL string
+#     	unleash service URL (default "http://localhost:4242")
+#   UNLEASH_TOKEN string
+#     	unleash service client access token (default "")
+#   KAFKA_ENABLED bool
+#     	kafka service enabled (default "false")
+#   KAFKA_BROKERS slice
+#     	kafka hostname:port list of brokers (default "localhost:9092")
+#   KAFKA_AUTH_TYPE string
+#     	kafka authentication type (mtls, sasl or empty) (default "")
+#   KAFKA_CA_CERT string
+#     	kafka TLS CA certificate path (default "")
 #   APP_CACHE_TYPE string
 #     	application cache (none, memory, redis) (default "none")
+#   APP_CACHE_EXPIRATION int64
+#     	expiration for both memory and Redis (time interval syntax) (default "1h")
 #   TELEMETRY_JAEGER_ENABLED bool
 #     	open telemetry jaeger exporter (default "false")
 #   TELEMETRY_JAEGER_ENDPOINT string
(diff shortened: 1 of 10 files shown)


config/api.env.example
```diff
@@ -88,14 +90,32 @@
...
+#   UNLEASH_TOKEN string
+#     	unleash service client access token (default "")
+#   KAFKA_ENABLED bool
+#     	kafka service enabled (default "false")
+#   KAFKA_BROKERS slice
+#     	kafka hostname:port list of brokers (default "localhost:9092")
+#   KAFKA_AUTH_TYPE string
+#     	kafka authentication type (mtls, sasl or empty) (default "")
+#   KAFKA_CA_CERT string
+#     	kafka TLS CA certificate path (default "")
```
ezr-ondrej, line 114: We should check these by GH actions... the refresh was not executed for some time.

lzap (changes requested):
Are you sure this will work on our authorization-only HCCP API? I think Cost management is publishing the JSON in a different way.

ezr-ondrej:
You're right, that needs to be answered first, lets convert to draft for now :)

ezr-ondrej:
We've got ourselves an answer and it is possible and there is no better solution for now even if we do the template static.
So I propose to role with this one.
@lzap WDYT?

build/RedHat.dockerfile
```diff
@@ -1,11 +1,13 @@
...
 USER 0
 RUN mkdir /build
 WORKDIR /build
 COPY . .
 RUN make prep build strip
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+RUN mkdir /docs
+COPY docs/lighthouse_template.json /docs
```
lzap, line 10: Please use Go file [embedding feature](https://pkg.go.dev/embed). We do this for everything - migrations and preload data. Then there is no reason to copy any files.

config/api.env.example
lzap: Agreed. :-)

docs/lighthouse_template.json
```diff
@@ -7,34 +7,34 @@
...
+        "roleDefinitionId": "9980e02c-c2be-4d73-94e8-173b1dc7cf3c"
       },
       {
-        "principalId": "<EnterpriseAppID>",
-        "principalIdDisplayName": "<EnterpriseAppName>",
+        "principalId": "{{.EnterpriseAppID}}",
+        "principalIdDisplayName": "{{.EnterpriseAppName}}",
         "roleDefinitionId": "b24988ac-6180-42a0-ab88-20f7382dd24c"
       }
     ]
```
lzap, line 40: I suggest to rename this file to `xxx.tmpl.json` to indicate use of Go templating. But this is fine as well.

internal/services/azure_offering_template_service.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"net/http"
+	"text/template"
+
+	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/payloads"
+)
+
+var LighthouseTemplateFilePath = "./docs/lighthouse_template.json"
```
lzap, line 11: Here you can just embed this. [Example](https://github.com/hardcaporg/hardcap/blob/main/internal/srv/kickstart.go#L45) from my prototype.

internal/services/azure_offering_template_service.go
```diff
@@ -0,0 +1,49 @@
...
+	// EnterpriseAppID of the App that will act as an offering Principal
+	EnterpriseAppID string
+
+	// EnterpriseAppName of the offering principal - the display name
+	EnterpriseAppName string
+}
+
+func AzureOfferingTemplate(w http.ResponseWriter, r *http.Request) {
+	params := azureOfferingParams{
+		OfferingDefaultName:        "Red Hat Hybrid cloud console",
```
lzap, line 32: I think product names should be capitalized (just my guts):

```suggestion
		OfferingDefaultName:        "Red Hat Hybrid Cloud Console",
```

internal/services/azure_offering_template_service.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		renderError(w, r, payloads.NewRenderError(r.Context(), "failed to fetch Azure template", err))
+		return
+	}
+
+	if err = template.Execute(w, params); err != nil {
+		renderError(w, r, payloads.NewRenderError(r.Context(), "failed to render the Azure template", err))
+		return
+	}
+}
```
lzap, line 49: Can you also log an INFO message that we served this? Just curious, there will not be much traffic but it is a public service so I am interested in tracking this.

lzap (changes requested):
It is fine except that we should really use embedding instead of file copy since we use it everywhere.

build/RedHat.dockerfile
```diff
@@ -1,11 +1,13 @@
-FROM registry.access.redhat.com/ubi8/go-toolset:latest as build
+FROM registry.access.redhat.com/ubi8/go-toolset:1.18 as build
```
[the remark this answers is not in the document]
ezr-ondrej: no, it is irrelevant, I'll move it to another PR to discuss it properly, but I think upgrade to 1.19 should be concious choice, currently we would move automatically...

internal/services/azure_offering_template_service.go
ezr-ondrej: Isn't this logged automatically by the logger middleware? :thinking:

build/RedHat.dockerfile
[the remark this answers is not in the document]
ezr-ondrej: I've opened a PR for it: https://github.com/RHEnVision/provisioning-backend/pull/449

internal/services/azure_offering_template_service.go
ezr-ondrej: I would like to keep it in the docs, I don't have a strong preference for it, but my gut feeling was docs should be first class citizen. But if we have strong preference for embedding, I'm not against, here it make sense IMHO, just the placing of it is bit of a blocker in my mind, but not necessarily ofc :)
Link in the docs still does the trick.

lzap: Can we please use embedding? If you really want this to be physically present also in the docs folder, then make a relative symlink.

internal/services/azure_offering_template_service.go
lzap: Yes if you keep this in the API mount function, I was sort of assuming this needs to go to the RootApi mount function where there are no middlewares (I think).

ezr-ondrej:
I've rebased to use embeding, I love the symlink idea, don't know why it did not come up with that :facepalm:
I believe this is ready.

lzap:
One more rebase please.

lzap:
Other than that looks great.

deploy/clowdapp.yaml
```diff
@@ -78,6 +78,8 @@ objects:
                     name: provisioning-azure-acc
                     key: client_secret
                     optional: true
+              - name: AZURE_CLIENT_NAME
+                value: ${AZURE_CLIENT_NAME}
```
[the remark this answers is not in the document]
ezr-ondrej: Right, it was comming from parameter, but you're right it should come from the secret.
I've made it come from the secret and added a default value in a code - this is just a display name, so we might decide to keep it blank in the secret, so the default would be used.

lzap (approved):
Why this is a secret? I mean it is a display name, I would be fine with keeping this just in app-interface YAML. Either merge if you want this to be secret or rebase and merge if you also feel like me this is an extra step that is unnecessary.

ezr-ondrej:
> Why this is a secret?

I've ended up with secret, because you just include it in the json that is already created for Azure, so it's IMHO actually easier then passing a new clowder parameter just for this.

PR #425, opened 2023-02-13
fix(HMS-1240): add step titles back

lzap:
This fixes a regression introduced by Azure refactoring: https://github.com/RHEnVision/provisioning-backend/commit/02b55c39af439a06db029ec0fb7a02a1286406c4

The step_titles column somehow dropped from the code, I assume you had the branch from earlier than it was introduced.

This is a regression. QA scenario:

* Create an AWS reservation.
* Get reservation details through `reservations/aws/ID` (not `reservations/ID` that returns only basic data)
* Check if it contains all the generic fields (success, status, steps, step_titles...)
* Check if it contains all the AWS fields (instance id, amount, instance type...)

### DIFF
--- a/internal/dao/pgx/reservation_pgx.go
+++ b/internal/dao/pgx/reservation_pgx.go
@@ -127,12 +127,13 @@ func (x *reservationDao) createGenericReservation(ctx context.Context, reservati
 	reservation.AccountID = ctxval.AccountId(ctx)
 	reservation.Status = "Created"
 
-	reservationQuery := `INSERT INTO reservations (provider, account_id, steps, status)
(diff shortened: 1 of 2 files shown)


ezr-ondrej:
Can we also cover it with unit tests ? Might be pretty straight forward

lzap:
Great idea, added integration assertions.

ezr-ondrej (approved):
Nice :+1:

ezr-ondrej:
Reopened to wake up GH actions, the IQE bot is just dead, so will ignore for such a small patch and :crossed_fingers: 
Today is not a bot day apparently.

PR #426, opened 2023-02-13
fix(HMS-1242): atomically read statistics

lzap:
I noticed the read atomic overlook, a variable that must be threadsafe must be read and written via `atomic` function. This could possibly lead to stale value read.

Additionally, I noticed that the Redis client creates connection pool of CPUs * 10. We know for fact how big the pool should be - it needs to be numPollers + 1 + some room for statistics calls (usually just 1 but to give it some room 10 should be a good limit).

I renamed some variables to better reflect their purpose.

### DIFF
--- a/pkg/worker/redis.go
+++ b/pkg/worker/redis.go
@@ -35,7 +35,7 @@ type RedisWorker struct {
 
 	// polling and wait groups
 	pollInterval time.Duration
-	pollMax      int
+	numPollers   int
 	loopWG       sync.WaitGroup
 	handleWG     sync.WaitGroup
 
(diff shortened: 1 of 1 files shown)


pkg/worker/redis.go
```diff
@@ -219,6 +220,6 @@ func (w *RedisWorker) Stats(ctx context.Context) (Stats, error) {
 
 	return Stats{
 		EnqueuedJobs: uint64(count),
-		InFlight:     w.inFlight,
+		InFlight:     atomic.LoadInt64(&w.inFlight),
```
lzap, line 223: This is the bugfix, the rest is just renaming of variables and setting up PoolSize to more reasonable number.

ezr-ondrej: I thought you actually need to define it as atomic... ILT

ezr-ondrej:
/retest
but the bot needs to learn as well

ezr-ondrej:
/retest
The IQE pod failed to start now :thinking:

ezr-ondrej (approved):

PR #427, opened 2023-02-14
build: Use sonarqube RH container for builds

ezr-ondrej:


.rhcicd/pr_check.sh
```diff
@@ -23,6 +23,8 @@ curl -s $CICD_URL/bootstrap.sh > .cicd_bootstrap.sh && source .cicd_bootstrap.sh
 # This script is used to build the image that is used in the PR Check
 source $CICD_ROOT/build.sh
 
+source $WORKSPACE/.rhcicd/sonarqube.sh
```
mshriver, line 26: I'm not 100% sure this will resolve correctly, absolutely worth a try.

mshriver (approved):

PR #428, opened 2023-02-15
Update job queue metrics

lzap:
This patch updates the job queue size metric and adds new metric that represents in-flight tasks: number of tasks which are currently in processing. Note that while job queue size is a global value (Gauge) - this is actually a redis queue length, the in-flight number is different for each consumer/worker (GaugeVec, vectorized by worker unique hostname).

For monitoring and alerts, we are interested in the following:

* Job queue size (one number). When it hits > 1000 we should alert that workers are too slow picking up new jobs (not running for example).
* In-flight count (sum). When the sum is > 500 we should alert that workers are probably too busy (can hit CPU/memory limits at some point).

I am making these numbers up these are educated guesses, feel free to tune them up. We have 3 workers which are each capable of executing hundreds of jobs in parallel each. So that's why 1000/500.

Now, the implementation is a bit tricky. I refactored the job queue size from a callback to just a regular "observation" function which is prefixed with Set as that is the function name Prometheus use for gauge metrics. There is now a new background goroutine which is spawned for each worker process. Every 30 seconds, it performs statistics read which in fact makes a call to Redis to get the queue length. It also picks up the in-flight number - this one is different for each worker.

Now, because we have 3 workers and we do not need this metric every 5 seconds, each worker does this. Ideally only one worker should be doing job queue length and all workers inflight, but since it is difficult to find out which is "number one" container in k8s, I am going for the easy solution - all do the same, the polling interval is 30 seconds and the start is randomized so approximetly every 10 seconds we get the updated gauge values.

During work, I noticed a typo in a metric name so I fixed it in a separate commit.

I also noticed that my IDE really wants me to use `x/exp/rand` package which has its own global pseudo random generator. This led to situation when my patch was always generating the same values, until Go 1.20 random generators are seeded with 1 therefore they always return the same numbers. Although this changes in 1.20, I am making a small change - our `random.InitializeSeed` function seeds both `math/rand` and `x/exp/rand` just in case someone else accidentally imports the experimental random package (which has the same interface but unseeded global generator).

### DIFF
--- a/cmd/pbstatuser/main.go
+++ b/cmd/pbstatuser/main.go
@@ -116,7 +116,7 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 
 	for s := range chAzure {
 		logger.Trace().Msgf("Checking Azure source availability status %s", s.SourceApplicationID)
-		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeAzure, func() error {
+		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAzure, func() error {
 			var err error
 			sr := kafka.SourceResult{
 				ResourceID:   s.SourceApplicationID,
@@ -139,7 +139,7 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 
 	for s := range chAws {
 		logger.Trace().Msgf("Checking AWS source availability status %s", s.SourceApplicationID)
-		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeAWS, func() error {
+		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAWS, func() error {
 			var err error
 			sr := kafka.SourceResult{
 				ResourceID:   s.SourceApplicationID,
@@ -168,7 +168,7 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
 
 	for s := range chGcp {
 		logger.Trace().Msgf("Checking GCP source availability status %s", s.SourceApplicationID)
-		metrics.ObserveAvailablilityCheckReqsDuration(models.ProviderTypeGCP, func() error {
+		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeGCP, func() error {
 			var err error
 			sr := kafka.SourceResult{
 				ResourceID:   s.SourceApplicationID,
--- a/cmd/pbstatuser/main.go
+++ b/cmd/pbstatuser/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/ptr"
+	"github.com/RHEnVision/provisioning-backend/internal/random"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 
 	// Clients
@@ -54,6 +55,10 @@ var (
 	processingWG = sync.WaitGroup{}
 	senderWG     = sync.WaitGroup{}
 )
+
+func init() {
+	random.SeedGlobal()
+}
 
 func processMessage(origCtx context.Context, message *kafka.GenericMessage) {
 	logger := ctxval.Logger(origCtx)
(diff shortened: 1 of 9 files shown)


cmd/pbstatuser/main.go
```diff
@@ -55,6 +56,10 @@ var (
 	senderWG     = sync.WaitGroup{}
 )
 
+func init() {
+	random.SeedGlobal()
+}
```
lzap, line 61: This is needed in all commands, we were missing this.

With Go 1.20 we can remove this. But we are still on 1.18 in prod.

internal/background/initialize.go
```diff
@@ -13,11 +13,22 @@ import (
...
+
+// InitializeWorker starts background goroutines for worker processes.
+// Use context cancellation to stop them.
+func InitializeWorker(ctx context.Context, workerName string) {
+	logger := ctxval.Logger(ctx).With().Bool("background", true).Logger()
+	ctx = ctxval.WithLogger(ctx, &logger)
+
+	// start job queue telemetry
+	go jobQueueMetricLoop(ctx, 30*time.Second, workerName)
+}
```
lzap, line 34: Background package now have two functions:

* InitializeApi - for REST API processes
* InitializeWorkers - for worker processes

If you need any background goroutine, add it here.

internal/random/random.go
```diff
@@ -7,23 +7,31 @@ package random
...
-	_ = binary.Read(crand.Reader, binary.LittleEndian, &rngSeed)
-	rand.Seed(rngSeed)
+	var seedInt64 int64
+	_ = binary.Read(crand.Reader, binary.LittleEndian, &seedInt64)
+	mrand.Seed(seedInt64)
+
+	var seedUint64 uint64
+	_ = binary.Read(crand.Reader, binary.LittleEndian, &seedUint64)
+	erand.Seed(seedUint64)
 }
```
lzap, line 30: This is a defensive change - we already have x/exp/rand in our dependencies and GoLand tends to prefer this package or it was just a misclick by me. But using unseeded random generator is dangerous.

We can remove seeding completely with 1.20 Go upgrade.

pkg/worker/redis.go
```diff
@@ -136,7 +136,7 @@ func (w *RedisWorker) dequeueLoop(ctx context.Context, i, total int) {
 
 	// spread polling intervals
 	delayMs := (int(w.pollInterval.Milliseconds()) / total) * (i - 1)
-	logger.Info().Msgf("Worker start delay %dms", delayMs)
+	logger.Debug().Msgf("Worker start delay %dms", delayMs)
```
lzap, line 139: This is a similar approach so I am syncing the logger level to debug for both messages.

internal/random/random.go
adiabramovitch: Can you explain the need for that random generator? I am not familiar with that.

lzap: I meant seed as in [PRNG](https://en.wikipedia.org/wiki/Pseudorandom_number_generator).

internal/background/initialize.go
```diff
@@ -13,11 +13,22 @@ import (
...
 	ctx = ctxval.WithLogger(ctx, &logger)
 
 	// start availability request batch sender
 	go sendAvailabilityRequestMessages(ctx, availabilityStatusBatchSize, 5*time.Second)
 }
+
+// InitializeWorker starts background goroutines for worker processes.
+// Use context cancellation to stop them.
+func InitializeWorker(ctx context.Context, workerName string) {
+	logger := ctxval.Logger(ctx).With().Bool("background", true).Logger()
```
ezr-ondrej, line 29:
```suggestion
	logger := ctxval.Logger(ctx).With().Bool("worker", true).Logger()
```

ezr-ondrej (approved):
I'd distinguish worker and the background goroutine in logs ideally :)

internal/background/initialize.go
lzap: I know this is confusing, but I would like to keep this as "background" to indicate this comes from the "background" goroutine. We have two now:

* Enqueuer for Kafka messages for checks, runs within "api" pods.
* This for jq metrics, runs within "worker" pods.

The package is named "background" so that was the idea.

ezr-ondrej: oki :shrug: :+1:

lzap:
I was hoping to get review from @adiabramovitch feel free to do a post-merge review. I can file another PR if needed.

PR #431, opened 2023-02-17
chore: update more deps

lzap:
TIL: `go get -u all` is not the same as `go get -u ./...` and the latter is what we want to use. This patch fixes that, I am not entirely what "all" exactly does but it did not work.


internal/config/config.go
```diff
@@ -236,8 +236,8 @@ func Initialize(configFiles ...string) {
 					config.Kafka.CACert = *b.Cacert
 				}
 				if b.Sasl != nil {
-					if b.Sasl.SecurityProtocol != nil && *b.Sasl.SecurityProtocol != "" {
-						config.Kafka.SASL.SecurityProtocol = *b.Sasl.SecurityProtocol
+					if b.SecurityProtocol != nil && *b.SecurityProtocol != "" {
+						config.Kafka.SASL.SecurityProtocol = *b.SecurityProtocol
```
lzap, line 240: This was deprecated in clowder and moved up one level, linter errors on deprecations.

internal/telemetry/otel.go
```diff
@@ -52,17 +52,11 @@ func Initialize(rootLogger *zerolog.Logger) *Telemetry {
...
-			semconv.ServiceVersionKey.String(version.OpenTelemetryVersion),
-		),
+	res := resource.NewWithAttributes(
+		semconv.SchemaURL,
+		semconv.ServiceNameKey.String("provisioning"),
+		semconv.ServiceVersionKey.String(version.OpenTelemetryVersion),
 	)
-	if err != nil {
-		panic(err)
-	}
```
lzap, line 65: The new OpenTelemetry panics with schema merge problem, this is how you do it now.

lzap:
Fixed tests.

lzap:
Tests are passing!

ezr-ondrej (approved):
:exploding_head: no idea what's going on with `all` vs `./...`, but looks reasonable :shrug: :)

lzap:
Even folks on the Go Slack channel were unable to tell, the conclusion was: use `./...`

PR #432, opened 2023-02-17
fix(HMS-1260): document pubkey and template behavior

lzap:
I propose to insist on pubkeys even when launch templates are required. We have foreign keys and we would need to introduce NULLs and do OUTER JOINs just for this small use case. For now I think this should be reasonable approach.

So this patch changes nothing effectively, but documents this in the OpenAPI spec.

While working on this, I noticed that the transaction code tries to commit after rollback which is leading to incorrect error - the original root cause is thrown away and "cannot commit: transaction is closed" is returned to the client. This fixes it.

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -559,7 +559,7 @@
     },
     "/reservations/aws": {
       "post": {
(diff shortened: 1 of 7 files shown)


akhil-jha (approved):

ezr-ondrej (approved):
QE is happy, I'm happy :+1:
Thanks @lzap :+1:

PR #433, opened 2023-02-20
fix(HMS-951): add region validation

lzap:
We want to test if region is actually supported prior launch, meaning we do have preloaded instance type information. This is for API users who would submit with an unknown region.

This PR contains two commits. The first one is big and it is refactoring and it will be hard to review because it also contains changes - it was not possible to just move YAML files and Go source files into a single package, because it was written as a bunch of copy-and-paste functions which cannot live in a single package (redefinition problem).

So instead, I am introducing a new type, interface and a common code that is shared across all three cloud providers. This is much more clean now.

Please review the refactoring first, then I will refresh data for each cloud provider in separate commits to verify generation is not broken by the refactoring and finally I will implement the functionality for HMS-951.

Note for myself: the code lives here: https://gist.github.com/lzap/929617e87dcbd05dc3086912af6dce3a


lzap:
I dropped the unused interface because actually all the code is shared in the common type. Also improved loading function, it is a method now.

.gitignore
```diff
@@ -4,6 +4,7 @@
 /.fleet/
 /cmd/*/__debug_bin
 /generate_*_types
+/print_*_types
```
ezr-ondrej, line 7: why are the two here? was it just rename or are both relevant?

ezr-ondrej (commented):
Quite hard to review, but the pipeline is green, so I'm ok :+1:

I'd have prefered two PR as you've guessed :stuck_out_tongue:

internal/preload/instance_type.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (p *instanceType) Load() error {
+	// load instance types
+	typesBuf, err := fsTypes.ReadFile(p.filename)
+	if err != nil {
+		return fmt.Errorf("unable to read instance types %s: %w", p.filename, err)
+	}
+
+	err = p.typeInfo.RegisteredTypes.Load(typesBuf)
```
adiabramovitch, line 25: Just wondering here: 
Isn't it a problem that the RegisteredTypes is nil? in `ec2_types.go` we are not initializing the typeInfo field.

internal/preload/instance_type.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to load instance types %s: %w", p.filename, err)
+	}
+
+	// load availability information
+	err = p.typeInfo.RegionalAvailability.Load(fsTypes, p.path)
```
adiabramovitch, line 31: ditto

adiabramovitch (commented):
looks good, I would love to have the workflow documented. I don't think it is trivial.

.gitignore
lzap: That is just how GoLand stores binaries, I could not figure out how to tell it to store them elsewhere.

internal/preload/azure_types.go
```diff
@@ -0,0 +1,15 @@
...
+
+func init() {
+	AzureInstanceType = instanceType{
+		filename: "azure_types.yaml",
+		path:     "azure_availability",
+		etagName: "azure-types",
+	}
+	err := AzureInstanceType.Load()
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
lzap: You could look up in the `Load` function, it shows: `fmt.Errorf("unable to read instance types %s: %w", p.filename, err)`

I believe this is sufficient, it contains the error message, filename and root error (e.g. "not found").

Note I use `panic` because this is an `init` function. The app hasn’t been initialized yet, there is no logging, no configuration. Therefore I use panic to simply exit immediately. Since these YAML files are embedded into the application itself, this will always work (or always fail if a file was renamed or something).

If you still want more detailed error message, let me know, what exactly should I put here. Panic also shows stacktrace, should be pretty obvious where it failed.

internal/preload/instance_type.go
lzap: It is not, `Load` calls `yaml.Unmarshal` which apparently works with `nil` values. In fact, since you pass a pointer, it is actually not important what value it contains because it will be overwritten and garbage collected.

internal/preload/azure_types.go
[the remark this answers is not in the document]
lzap: How about this: `panic(fmt.Errorf("cannot preload ec2 types: %w", err))`

lzap:
Fixed failed tests. Added more detailed error message for the "init" functions.

@ezr-ondrej it is actually pretty easy to review, just review the first and the last commits, skip the region refresh. I had to do this in a single PR to ensure the generation works, there were some changes needed after refactoring so in this case I did the right thing.

Agreed with the documentation, I will start working on it shortly.

lzap:
When I was writing the documentation, I noticed that InstanceTypeInfo type is still in `clients` package. I should definitely move it.

Also I see the old `supported` package which was meant to be replaced by this. I think it is time to remove it, but not in this PR. I will do a followup.

ezr-ondrej (approved):
I'm ok here :+1:
@adiabramovitch will you finish this one? :)

lzap:
Correction: I am not moving from `clients` because the main interfaces use these types. So keeping them in, no changes. I am working on documentation but I can add it in separate PR.

adiabramovitch:
@lzap Thanks!

PR #434, opened 2023-02-20
fix(HMS-1259): Add adjustable datasource and SLOs panels

adiabramovitch:
Also added two panels to show the HTTP request latency and Source availability latency according to our SLOs

### DIFF
--- a/deploy/dashboards/grafana-dashboard-provisioning.configmap.yaml
+++ b/deploy/dashboards/grafana-dashboard-provisioning.configmap.yaml
@@ -35,8 +35,8 @@ data:
       "editable": true,
       "fiscalYearStartMonth": 0,
       "graphTooltip": 0,
-      "id": 52102,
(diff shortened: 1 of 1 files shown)


adiabramovitch:
> 
 Fixing it, be ready soon

adiabramovitch:
> There are conflicts (?).
> 
> How do I test this? :-)

In order to test it you can copy the JSON itself (from line 12) and paste it instead of the JSON [here](https://grafana.stage.devshift.net/d/22222/provisioning?editview=dashboard_json):

lzap (approved):
It works. Thanks.

ezr-ondrej:
Did not have time to review, but could we also make the production datasource selected by default? :)

lzap:
I think the production cluster was the default, yes. Now that my Grafana remembered my selection, I cannot confirm.

PR #439, opened 2023-02-22
fix(HMS-879): workers metrics

lzap:
This adds a new duration metric which captures duration of every job processed from the background redis queue. It contains "type" label which is the job name, e.g. `launch_instances_aws` for AWS EC2 launch job. Since this is the first metric we have in workers, the patch adds a new prometheus endpoint in the main function.

### DIFF
--- a/cmd/pbapi/main.go
+++ b/cmd/pbapi/main.go
@@ -160,10 +160,10 @@ func main() {
 		signal.Notify(sigint, os.Interrupt, syscall.SIGTERM)
 		<-sigint
 		if err := apiServer.Shutdown(context.Background()); err != nil {
-			log.Fatal().Err(err).Msg("Main service shutdown error")
+			log.Warn().Err(err).Msg("Main service shutdown error")
 		}
 		if err := metricsServer.Shutdown(context.Background()); err != nil {
-			log.Fatal().Err(err).Msg("Metrics service shutdown error")
+			log.Warn().Err(err).Msg("Metrics service shutdown error")
 		}
 		close(waitForSignal)
 	}()
@@ -181,7 +181,7 @@ func main() {
 
 	if err := apiServer.ListenAndServe(); err != nil {
 		if !errors.Is(err, http.ErrServerClosed) {
-			log.Fatal().Err(err).Msg("Main service listen error")
+			log.Warn().Err(err).Msg("Main service listen error")
 		}
 	}
 
--- a/cmd/pbstatuser/main.go
+++ b/cmd/pbstatuser/main.go
@@ -121,7 +121,7 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 
 	for s := range chAzure {
 		logger.Trace().Msgf("Checking Azure source availability status %s", s.SourceApplicationID)
-		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAzure, func() error {
+		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAzure.String(), func() error {
 			var err error
 			sr := kafka.SourceResult{
 				ResourceID:   s.SourceApplicationID,
@@ -131,7 +131,7 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 			// TODO: check if source is avavliable - WIP
 			sr.Status = kafka.StatusAvaliable
 			chSend <- sr
-			metrics.IncTotalSentAvailabilityCheckReqs(models.ProviderTypeAzure, sr.Status, nil)
+			metrics.IncTotalSentAvailabilityCheckReqs(models.ProviderTypeAzure.String(), sr.Status.String(), nil)
 
 			return fmt.Errorf("error during check: %w", err)
 		})
@@ -144,7 +144,7 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 
 	for s := range chAws {
 		logger.Trace().Msgf("Checking AWS source availability status %s", s.SourceApplicationID)
-		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAWS, func() error {
+		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAWS.String(), func() error {
 			var err error
 			sr := kafka.SourceResult{
 				ResourceID:   s.SourceApplicationID,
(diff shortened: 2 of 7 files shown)


cmd/pbapi/main.go
```diff
@@ -181,7 +181,7 @@ func main() {
 
 	if err := apiServer.ListenAndServe(); err != nil {
 		if !errors.Is(err, http.ErrServerClosed) {
-			log.Fatal().Err(err).Msg("Main service listen error")
+			log.Warn().Err(err).Msg("Main service listen error")
```
lzap, line 184: I noticed that we call Fatal when there is a shutdown error. Fatal immediately exists the application will kill other goroutines. We should never use Fatal unless we want to exit immediately (e.g. in migrations or during startup).

lzap:
Rebased to solve the import cycle problem.

internal/metrics/metrics.go
```diff
@@ -3,8 +3,6 @@ package metrics
 import (
 	"time"
 
-	"github.com/RHEnVision/provisioning-backend/internal/kafka"
-	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/version"
 	"github.com/prometheus/client_golang/prometheus"
```
lzap, line 7: This was the culprit of import cycle problems. The metrics package will be in wide use througout the whole app and it is important to keep the imports at minimum. Therefore instead of `models.ProviderType` function argument, I refactored it to simply accept `string` because in the end, we convert the provider type to string anyway. This removes the need for import. And the same applies for everything, well, except `version` which is just a simple package with constants (no metrics there).

cmd/pbstatuser/main.go
```diff
@@ -144,7 +144,7 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 
 	for s := range chAws {
 		logger.Trace().Msgf("Checking AWS source availability status %s", s.SourceApplicationID)
-		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAWS, func() error {
+		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAWS.String(), func() error {
```
adiabramovitch, line 147: Why did you add `.String()`?
it is not needed, it is present [here](https://github.com/RHEnVision/provisioning-backend/blob/bad43f81bf343b9f3898ab69ded82b6198804bd2/internal/metrics/metrics.go#L55)

`provisioning_source_availability_check_request_duration_ms_bucket{component="statuser",error="true",service="provisioning",type="aws",le="5000"} 2`

internal/metrics/metrics.go
```diff
@@ -48,25 +46,44 @@ var AvailabilityCheckReqsDuration = prometheus.NewHistogramVec(
...
+		Buckets:     []float64{5, 10, 50, 100, 250, 500, 1000, 2500, 5000, 10000},
+	},
+	[]string{"type"},
+)
+
+func ObserveAvailabilityCheckReqsDuration(provider string, observedFunc func() error) {
 	errString := "false"
 	start := time.Now()
 	defer func() {
-		AvailabilityCheckReqsDuration.WithLabelValues(provider.String(), errString).Observe(float64(time.Since(start).Nanoseconds()) / 1000000)
```
adiabramovitch, line 55: Why?
I think it is cleaner to put it here

internal/metrics/metrics.go
```diff
@@ -88,3 +105,7 @@ func RegisterStatuserMetrics() {
 func RegisterApiMetrics() {
 	prometheus.MustRegister(JobQueueSize)
 }
+
+func RegisterWorkerMetrics() {
+	prometheus.MustRegister(JobQueueSize)
```
adiabramovitch, line 110: it is supposed to be `BackgroundJobDuration` isnt it?

pkg/worker/redis.go
```diff
@@ -205,7 +206,9 @@ func (w *RedisWorker) processJob(ctx context.Context, job *Job) {
 	logger.Info().Msg("Dequeued job from Redis")
 	ctx = contextLogger(ctx, job)
 	if h, ok := w.handlers[job.Type]; ok {
-		h(ctx, job)
+		metrics.ObserveBackgroundJobDuration(job.Type.String(), func() {
```
ezr-ondrej, line 209: ouch. we are using internal package function in `pkg`. Can't we do it in `job/common.go` somewhere?
Or just abandone the idea of it being under `pkg`.
Not a blocker, more of a food for thought, given we already import other internal packages :shrug:

internal/metrics/metrics.go
lzap: Damn good eyes! I tested only that the metrics appear in general, but it actually DOES NOT WORK without this.

pkg/worker/redis.go
lzap: Yeah this is not the first time. Tho I think it is sort of okay I guess as long as you so not return an "internal" type. Unless you prove otherwise, I think this is okay.

I think somebody from the content team already asked about our Redis queue and I gave them link. Maybe this would get used by someone else?!?

We can always move this into the internal package later.

lzap:
Changed one line which caused this patch to, ehm, not work. :-) Copy and paste...

internal/metrics/metrics.go
adiabramovitch: What about registering `JobQueueInFlight`?

lzap:
Also fixed queue size which was not working (it was in the wrong initialization function).

internal/metrics/metrics.go
```diff
@@ -86,5 +103,10 @@ func RegisterStatuserMetrics() {
 }
 
 func RegisterApiMetrics() {
+	// no metrics
+}
+
+func RegisterWorkerMetrics() {
 	prometheus.MustRegister(JobQueueSize)
+	prometheus.MustRegister(BackgroundJobDuration)
```
adiabramovitch, line 111: Just for consistency
```suggestion
	prometheus.MustRegister(JobQueueSize, BackgroundJobDuration)
```

ezr-ondrej: @adiabramovitch could you do a follow-up for it? If it's not a blocker, I'd like to get this in, so we start capturing the metrics and Lukas will not get to it of a week now :)

adiabramovitch (approved):

internal/metrics/metrics.go
ezr-ondrej: Addressed in #443

PR #441, opened 2023-02-23
feat(HMS-1110): Add source permission validation check endpoint

avitova:
From my point of view, this should be a final PR concerning [HMS-1110](https://issues.redhat.com/browse/HMS-1110), so if you have any more ideas on how to make this different, I am happy to do it now rather than later.


internal/clients/http/ec2/ec2_client.go
```diff
@@ -36,6 +38,47 @@ type ec2Client struct {
...
+		"ec2:DescribeLaunchTemplateVersions",
+		"ec2:DescribeRegions",
+		"ec2:DescribeSecurityGroups",
+		"ec2:DescribeSnapshotAttribute",
+		"ec2:DescribeTags",
+		"ec2:ImportKeyPair",
+		"ec2:RunInstances",
+		"ec2:StartInstances",
+	},
+}
```
lzap, line 80: This could be `const` instead of `var`.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -442,6 +485,52 @@ func getRoleName(arn string) (string, error) {
 	return roleName[1], nil
 }
 
+func getStatementFromAWSDocument(document string) ([]Statement, error) {
+	unescapedDoc, err := url.QueryUnescape(document)
+	if err != nil {
+		return nil, fmt.Errorf("Could not parse url from the Document: %w", err)
```
lzap, line 491: All errors must start with lowercase (our linter should enforce it not sure why it passes). On the other hand, logs start with capital letter. If I could change this when I started with capitals I would do this. The latter is not enforced tho.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -450,6 +539,7 @@ func (c *ec2Client) listAttachedRolePolicies(ctx context.Context, auth *clients.
 		return nil, fmt.Errorf("unable to parse ARN: %w", err)
 	}
 
+	logger.Info().Msgf("Listing attached role policies.")
```
lzap, line 542: We tend to avoid dots, these are statements not full sentences.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -464,17 +554,63 @@ func (c *ec2Client) listAttachedRolePolicies(ctx context.Context, auth *clients.
...
 
-	// TODO: GetPolicyVersion for policies.
+	statements, err := listStatements(versions)
+	if err != nil {
+		return nil, fmt.Errorf("Could not list statements: %w", err)
+	}
 
-	return nil, nil
+	return listMissingPermissions(statements, expectedStatement), nil
 }
```
lzap, line 616: It looks great, I would maybe consider moving this from already crowded `ec2_client.go` into a separate file.

internal/clients/http/ec2/ec2_client_test.go
```diff
@@ -4,6 +4,61 @@ import (
...
+	shouldBeEmpty := listMissingPermissions(expectedStatementList, expected)
+	if len(shouldBeEmpty) != 0 {
+		t.Errorf(`correct missing permissions listing failed`)
+	}
+
+	shouldBeEmpty = listMissingPermissions(duplicityPermissions, expected)
+	if len(shouldBeEmpty) != 0 {
+		t.Errorf(`missing permission listing with duplicities failed`)
+	}
+}
```
lzap, line 60: I suggest to use `require` package from the `testify` library, tests are much shorter.

lzap:
Just few random thoughts. It looks great.

Now, once this is merged, you can continue with refactoring our statuser to use this new function to do full check of account, when permissions are missing just log the missing elements as a warning into logs (we cannot send this back to sources there is no field in the payload. Then you can implement caching, it should be done here in the EC2 client so both statuser and the endpoint can utilize it.

The final integration of this into the app can be done via the endpoint - UI should add an extra step before SSH key is uploaded (or after) and perform a call and stop the reservation in case there actually is a missing permission. Alternatively, this could even be done in the job as a first step (then endpoint would not be necessary at all). Both are valid solutions.

So there is plenty of work on this one :-) Cheers.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -442,6 +485,52 @@ func getRoleName(arn string) (string, error) {
...
+	var policyVersion Version
+	err = json.Unmarshal([]byte(unescapedDoc), &policyVersion)
+	if err != nil {
+		return nil, fmt.Errorf("Could not unmarshal json: %w", err)
+	}
+	return policyVersion.Statement, nil
+}
+
+func listStatements(versions []*iamTypes.PolicyVersion) ([]Statement, error) {
+	var result []Statement
```
adiabramovitch, line 503: Do we want `result` to be null if empty or an empty array? 
https://stackoverflow.com/questions/25543520/declare-slice-or-make-slice

internal/clients/http/ec2/ec2_client.go
```diff
@@ -442,6 +485,52 @@ func getRoleName(arn string) (string, error) {
...
+		}
+
+		result = append(result, policyVersionStatements...)
+	}
+	return result, nil
+}
+
+func listMissingPermissions(statements []Statement, expected Statement) []string {
+	presentPermissions := make(map[string]struct{})
+	var missing []string
```
adiabramovitch, line 517: ditto

internal/clients/http/ec2/ec2_client.go
```diff
@@ -442,6 +485,52 @@ func getRoleName(arn string) (string, error) {
...
+	for _, statement := range statements {
+		if statement.Effect != expected.Effect {
+			continue
+		}
```
adiabramovitch, line 521: Can you please add a description or links to the PR explaining the effect, and statements logic?

internal/clients/http/ec2/ec2_client.go
adiabramovitch: if we are moving this from the ec2_client.go file I would expect to have an EC2 folder with all the relevant related functions. And the functions with shared logic are stored in the same file.

internal/clients/http/ec2/ec2_client.go
avitova: Good point. I have decided to use a slice in this case rather than an array. AWS does not really have the best way of returning policy Document. We have to decode the url and unmarshal the json. Therefore we do not know the length of the resulting slice/array before we decode and unmarshal it.
In this case I thought it is more convenient to use a slice, as it is more dynamic and does not require memory allocation for each append.

internal/clients/http/ec2/ec2_client.go
avitova: Same, we do not know how many permissions would be missing. Therefore a slice seemed more convenient. This should work the same if the nil is returned instead of an empty array.
Let me know if this makes sense. I am happy to reconsider.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -36,6 +38,47 @@ type ec2Client struct {
...
+		"ec2:DescribeKeyPairs",
+		"ec2:DescribeLaunchTemplates",
+		"ec2:DescribeLaunchTemplateVersions",
+		"ec2:DescribeRegions",
+		"ec2:DescribeSecurityGroups",
+		"ec2:DescribeSnapshotAttribute",
+		"ec2:DescribeTags",
+		"ec2:ImportKeyPair",
+		"ec2:RunInstances",
+		"ec2:StartInstances",
```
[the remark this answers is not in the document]
avitova: I am unsure what you mean, but I will look into this. These are values against which we check whether the user's account (role, to be exact) is correctly set. I am happy to discuss this further and see more ideas concerning this. :)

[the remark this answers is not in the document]
adiabramovitch: I thought we could add it to our `.env` files as we do for the GCP_JSON for example, But I might be wrong here. 
@ezr-ondrej What do you think?

internal/clients/http/ec2/ec2_client.go
adiabramovitch: I totally agree that we should use slices, I was referring to what is the correct way to initialize it. 
I was asking if we prefer `result` to be null if empty or if we want it to be an empty slice (that is the discussion I have pasted in the comment above)

internal/clients/http/ec2/ec2_client.go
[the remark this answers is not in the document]
ezr-ondrej: I'd not include it in the config, this should not be configurable IMHO, having it in separate file and embedding it, could be useful tho.
(having json file in this folder and using https://pkg.go.dev/embed to load it into the code)
That way we could easily reference it from a documentation.

internal/clients/http/ec2/ec2_client_test.go
```diff
@@ -4,6 +4,61 @@ import (
...
+	missingStatements := listMissingPermissions(missingLastPermission, expected)
+	if len(missingStatements) != 1 || missingStatements[0] != expected.Action[len(expected.Action)-1] {
+		t.Errorf(`missing permission listing failed`)
+	}
+
+	expectedStatementList := []Statement{expected}
+	shouldBeEmpty := listMissingPermissions(expectedStatementList, expected)
+	if len(shouldBeEmpty) != 0 {
+		t.Errorf(`correct missing permissions listing failed`)
+	}
```
ezr-ondrej, line 54: ditto
```suggestion
	assert.Equal(t, 0, len(shouldBeEmpty), "Missing permission listing with duplicities failed")
```

ezr-ondrej (changes requested):
I'd like to use embedding for the json from a separate file, it would help us to use it in documentation, or when we will need to reference the file from other places.

I've got some improvements for the tests, I believe readable test suite is almost as good as developement documentation, so I'm bit pedantic, tho you do not need to follow my suggestions, I'd love improving the readability of the tests a bit :)

internal/clients/http/ec2/ec2_client.go
lzap: I am not sure I follow, we already have an `ec2` folder and it seems pretty empty so far:

image

avitova:
This solution should be more flexible for different ways of defining the permissions for AWS roles.

internal/clients/http/ec2/iam_client.go
```diff
@@ -0,0 +1,304 @@
...
+				switch statementElement := v.(type) {
+				case interface{}:
+					result = append(result, getPermissionsFromAllowed(ctx, statementElement)...)
+				}
+			}
+		case interface{}:
+			result = append(result, getPermissionsFromAllowed(ctx, statementField)...)
+
+		}
+	}
```
lzap, line 139: Why do you do this empty interface parsing? You should be able to unmarshal directly into a type, Go JSON parser is able to ignore unknown fields AFAIK and the format [seems to be documented](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements.html).

lzap (commented):
Nicely done, great work.

I dropped just some random thoughts.

internal/clients/http/ec2/iam_client.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+type Policy struct {
+	Statement []Statement `json:"Statement"`
+}
+
+func expectedStatement() Statement {
+	return Statement{
+		Sid:    "RedHatProvisioning",
+		Effect: "Allow",
```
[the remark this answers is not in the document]
avitova: Yes, and I do not think we'd be able to determine the priority of policies, as some 'Allow' might discard the 'Deny' policy elsewhere.

docs/configure-amazon-role.md
```diff
@@ -136,7 +136,8 @@ Create a new policy for the Tenant account:
                 "ec2:DescribeTags",
                 "ec2:ImportKeyPair",
                 "ec2:RunInstances",
-                "ec2:StartInstances"
+                "ec2:StartInstances",
+                "ec2:ListRolePolicies",
```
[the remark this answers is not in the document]
lzap: Created https://issues.redhat.com/browse/HMS-1415 and https://issues.redhat.com/browse/HMS-1414

internal/clients/http/ec2/iam_client.go
lzap: Okay then this is indeed correct.

internal/clients/http/ec2/iam_client.go
```diff
@@ -0,0 +1,304 @@
...
+type Statement struct {
+	Sid    string
+	Effect string
+	Action []string
+}
+
+type Policy struct {
+	Statement []Statement `json:"Statement"`
+}
+
```
adiabramovitch, line 27: I think documentation is missing here as we do for other types: https://github.com/RHEnVision/provisioning-backend/blob/20294330f9b82e703c799d7e87643e263210c816/internal/clients/instance_type.go#L17

ezr-ondrej (approved):
Nice ! Looks great.

I'd love to add tracing for it in follow up PRs, so we know the timings it adds, but that's for sure out of scope for now :+1:

internal/clients/http/ec2/iam_client.go
avitova: I think this is a valid point. Even though the structure can be seen below, it might be a good practice. I've left a few comments.

lzap (approved):

adiabramovitch:
Thanks @avitova 🥇

PR #444, opened 2023-02-27
feat(HMS-894): Add GCP reservation test

adiabramovitch:


ezr-ondrej (commented):
Is there some validation that is expected to go wrong? If so, we might want to check for these error paths as well, but if the service is simple enough, no need I guess :)

We might want to also check, the job got queued in the stub client and maybe even check, the image name got resolved correctly for the job params.

WDYT?

adiabramovitch:
> 
Currently, this is the only check we do in other providers' services, **but** now we have the region validation, so we can check that case across all providers' tests. 

I agree, we should test those cases as well, I'll add it to the AWS, GCP, and Azure.

ezr-ondrej:
We might do it in a follow up, here I'd just add the check for the id to name translation and be ok with it for the scope of this PR, but up to you ^_^

ezr-ondrej (approved):
Great start, at least we see if the service is not completely broken, lets add another asserts as follow up :)

PR #445, opened 2023-03-01
fix: allow dynamic naming for Azure disk, IP and NIC

ezr-ondrej:
We were forcing a disk name, we can just leave disk name blank and it will be named with the VM name prefix and randomized suffix.

Refs HMS-1146


akhil-jha (approved):

ezr-ondrej:
We will need to revisit the naming once we are reviewing the full workflow with UX, so we can address any potentional comments there :)

PR #446, opened 2023-03-02
feat: Add sentry writer to zerolog 

ezr-ondrej:
Refactors the initialization of Logger, we still have two functions, but only one is used for initialization.
We can choose to initialize only stdout, which is used in unit/db tests and migrations.

We  do use other initializer for everything else.

Last commit adds zerolog writer for sentry, meaning all `Fatal()` and `Error()` calls are automatically sent to sentry.

Alternative for #440

Fixes HMS-851


internal/logging/zerolog.go
```diff
@@ -81,6 +83,20 @@ func cloudwatchWriter() (io.Writer, func(), error) {
...
+// sentryWriter creates a zerolog writer for sentry.
+// Uses github.com/archdx/zerolog-sentry which is very simple wrapper.
+func sentryWriter(dsn string) (io.Writer, func(), error) {
+	wr, err := zlogsentry.New(dsn)
+	if err != nil {
+		return nil, func() {}, fmt.Errorf("cannot initialize sentry: %w", err)
+	}
+	sentry.ConfigureScope(func(scope *sentry.Scope) {
+		scope.SetTag("stream", config.BinaryName())
+	})
```
ezr-ondrej, line 95: This doesn't seem to work :/

ezr-ondrej: I've needed to copy the writer in our project to use `sentry.Init`
I've sent https://github.com/archdx/zerolog-sentry/pull/5 once that's in, we can switch to using that package.

internal/logging/sentry.go
```diff
@@ -0,0 +1,24 @@
...
+	wr, err := sentry_writer.New(dsn)
+	if err != nil {
+		return nil, func() {}, fmt.Errorf("cannot initialize sentry: %w", err)
+	}
+	sentry.ConfigureScope(func(scope *sentry.Scope) {
+		scope.SetTag("stream", config.BinaryName())
+	})
+	// close function returns error, but it's always nil, so we are ignoring it here
+	return wr, func() { _ = wr.Close() }, nil
+}
```
lzap, line 24: Not a huge fan of this approach which serializes the events into JSON and then parsing them back into memory. But as long as this is only done for error/fatal events, I am fine. I will be looking for a proof this is not done for info and lower-level events tho.

I was hoping we could use zerolog's hook system instead.

internal/logging/sentry_writer/writer.go
```diff
@@ -0,0 +1,262 @@
...
+	return config{
+		levels: []zerolog.Level{
+			zerolog.ErrorLevel,
+			zerolog.FatalLevel,
+			zerolog.PanicLevel,
+		},
+		sampleRate:   1.0,
+		flushTimeout: 3 * time.Second,
+	}
+}
```
lzap, line 262: Why are you making a copy? I mean, it is fine to simply use the library if you want to use it "as is". I was thinking loud in the other PR that maybe creating these sentry events via zerolog hook could be so simple that we would not need any dependancy at all.

If you really want to use the parsing approach, then just remove this copy and use the library.

lzap (changes requested):
Great refactoring and I like the idea of using Sentry API.

However, I would like not to see our app parsing every single log line. It looks fine probably just for errors/traces but what if we change that in the future.

I am really hoping utilizing zerolog hook API to do similar integration as sentry-go library shows for the logrus (which also have a hook API):

* https://github.com/getsentry/sentry-go/blob/master/_examples/logrus/main.go
* https://github.com/rs/zerolog#hooks

Now I don’t want you to push into this, what you have is probably fine. But I am going to do a refactoring followup if you don’t have time to do this :-)

If you want the parsing approach, just drop the copy of the library, which is I think misunderstanding. It is okay to add dependencies, as long as they are healthy and simple enough for the job. Which is this case I think.

internal/logging/sentry_writer/writer.go
ezr-ondrej: As I explained offline, I needed https://github.com/archdx/zerolog-sentry/pull/5, but I'll try to implement it with a hook and let see, I agree that if that works, we are bit more efficient and we will need no dependency :+1:

ezr-ondrej:
Extracting the logger initialization into separate PR, so we can get that in or not separately and then decide on the sentry approach :)
https://github.com/RHEnVision/provisioning-backend/pull/453

lzap:
Merged the other one, appreciated walking the extra mile for the cleaner approach, thanks!

internal/logging/sentry.go
ezr-ondrej: So it looks like there is no other way for zerolog :(

lzap (approved):
Nice and clean. I don’t have local dev sentry setup yet, so let’s just merge and you iron out the details if any.

PR #448, opened 2023-03-02
feat(HMS-761): add instance description to aws job

amirfefer:
This adda fetching instance description to aws job, stores the instance's public ip and dns into reservation_instances table

I keep this PR as draft until UI is ready

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -313,6 +313,29 @@ func (c *ec2Client) ListInstanceTypes(ctx context.Context) ([]*clients.InstanceT
 	return instances, nil
 }
 
+func (c *ec2Client) DescribeInstanceDetails(ctx context.Context, InstanceIds []string) ([]*clients.InstanceDescription, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "DescribeInstanceDetails")
+	defer span.End()
+
+	input := &ec2.DescribeInstancesInput{
+		InstanceIds: InstanceIds,
+	}
+	resp, err := c.ec2.DescribeInstances(ctx, input)
+	if err != nil {
+		if isAWSUnauthorizedError(err) {
+			err = clients.UnauthorizedErr
+		}
+		span.SetStatus(codes.Error, err.Error())
+		return nil, fmt.Errorf("cannot fetch instances description: %w", err)
+	}
+	instanceDetailList, err := c.parseDescribeInstances(resp)
+	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
+		return nil, fmt.Errorf("failed to fetch instances description: %w", err)
+	}
+	return instanceDetailList, nil
+}
+
 func (c *ec2Client) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchTemplate, error) {
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListLaunchTemplates")
 	defer span.End()
@@ -409,6 +432,22 @@ func (c *ec2Client) parseRunInstancesResponse(respAWS *ec2.RunInstancesOutput) [
 		list[i] = instance.InstanceId
 	}
 	return list
+}
+
+func (c *ec2Client) parseDescribeInstances(respAWS *ec2.DescribeInstancesOutput) ([]*clients.InstanceDescription, error) {
+	if len(respAWS.Reservations) == 0 {
+		return nil, http.NoReservationErr
+	}
+	instances := respAWS.Reservations[0].Instances
+	list := make([]*clients.InstanceDescription, len(instances))
+	for i, instance := range instances {
+		list[i] = &clients.InstanceDescription{
+			ID:         *instance.InstanceId,
+			PublicIPv4: *instance.PublicIpAddress,
+			PublicDNS:  *instance.PublicDnsName,
+		}
+	}
+	return list, nil
 }
 
 func (c *ec2Client) GetAccountId(ctx context.Context) (string, error) {
--- a/internal/clients/instance_description.go
+++ b/internal/clients/instance_description.go
@@ -1,0 +1,13 @@
+package clients
+
+// InstanceInfo defines a model for an instance description
+type InstanceDescription struct {
+	// The id of the instance
+	ID string `json:"id,omitempty" yaml:"id"`
+
+	// The public ipv4 dns of the instance
+	PublicDNS string `json:"dns,omitempty" yaml:"dns"`
+
+	// the public ipv4 of the instance
+	PublicIPv4 string `json:"ipv4,omitempty" yaml:"ipv4"`
+}
--- a/internal/jobs/launch_instance_aws.go
+++ b/internal/jobs/launch_instance_aws.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/RHEnVision/provisioning-backend/internal/clients/http"
@@ -59,6 +60,11 @@ func HandleLaunchInstanceAWS(ctx context.Context, job *worker.Job) {
 	}
 
 	jobErr = DoLaunchInstanceAWS(ctx, &args)
+	if jobErr != nil {
+		finishWithError(ctx, args.ReservationID, jobErr)
+		return
+	}
+	jobErr = FetchInstancesDescriptionAWS(ctx, &args)
 
 	finishJob(ctx, args.ReservationID, jobErr)
 }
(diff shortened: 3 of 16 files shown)


internal/jobs/launch_instance_aws.go
```diff
@@ -60,6 +61,12 @@ func HandleLaunchInstanceAWS(ctx context.Context, job *worker.Job) {
 
 	jobErr = DoLaunchInstanceAWS(ctx, &args)
 
+	if jobErr != nil {
```
ezr-ondrej, line 64: Can we move this to the above statement so it is clear we are checking the error of that statement? It's common practice not to divide by new line for these cases.

OTH I'd add new line after this if, as the following statement is not related :)

internal/jobs/launch_instance_aws.go
```diff
@@ -212,3 +219,47 @@ func DoLaunchInstanceAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
...
+		instancesIDList[i] = instance.InstanceID
+	}
+	ec2Client, err := clients.GetEC2Client(ctx, args.ARN, args.Region)
+	if err != nil {
+		return fmt.Errorf("cannot create new ec2 client from config: %w", err)
+	}
+
+	time.Sleep(6000 * time.Millisecond)
+
+	instancesDescriptionList, err := ec2Client.ListInstancesDescription(ctx, instancesIDList)
```
ezr-ondrej, line 252: This forces the 6 second wait, shall we instead wait a second, check if ready and repeat six times?
I believe usually AWS is much faster, so there is not much reason to wait 6 seconds for all reservations, or is there some research you've done how long it takes?

internal/clients/http/ec2/ec2_client.go
```diff
@@ -414,6 +438,22 @@ func (c *ec2Client) parseRunInstancesResponse(respAWS *ec2.RunInstancesOutput) [
 	return list
 }
 
+func (c *ec2Client) parseDescribeInstances(respAWS *ec2.DescribeInstancesOutput) ([]*clients.InstanceDescription, error) {
+	if len(respAWS.Reservations) == 0 {
+		return nil, errors.New("no reservation has found in AWS response")
```
ezr-ondrej, line 443:
```suggestion
		return nil, errors.New("no reservation found in AWS response")
```

ezr-ondrej: Also as linter says, this error should be a constant variable.

ezr-ondrej (commented):
Great, this looks on the right track, I'd really love if we could add some polling instead of static wait time tho...

internal/clients/http/ec2/ec2_client.go
```diff
@@ -316,6 +317,29 @@ func (c *ec2Client) ListInstanceTypes(ctx context.Context) ([]*clients.InstanceT
...
+		span.SetStatus(codes.Error, err.Error())
+		return nil, fmt.Errorf("cannot fetch instances description: %w", err)
+	}
+	instancesList, err := c.parseDescribeInstances(resp)
+	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
+		return nil, fmt.Errorf("failed to fetch instances description: %w", err)
+	}
+	return instancesList, nil
+}
```
lzap, line 341: I would probably keep the naming convention from the AWS API here, even if it returns a slice. The call is named "DescribeInstances" so we could have the same name in our own interface too. Something like "DescribeInstanceDetails" maybe? (I like "Details" more than "Description" so the struct type could be `InstanceDetail`)

internal/clients/instance_description.go
```diff
@@ -0,0 +1,13 @@
...
+type InstanceDescription struct {
+	// The id of the instance
+	ID *string `json:"id,omitempty" yaml:"id"`
+
+	// The public ipv4 dns of the instance
+	PublicDNS *string `json:"dns,omitempty" yaml:"dns"`
+
+	// the public ipv4 of the instance
+	PublicIPv4 *string `json:"ipv4,omitempty" yaml:"ipv4"`
+}
```
lzap, line 13: Can we not use string pointers? Pointers are bad for strings (string is already a pointer itself) and the only reason AWS SDK use them is because the types are generated from XML. Just use strings and empty value means "nothing". You will have to convert them, we have a package `ptr` with helper functions for that.

internal/clients/http/ec2/ec2_client.go
lzap: Another idea for the name as I commented down below: `ReservationInstance`

lzap:
Dropped some comments, I think we should use JSONB for these fields as they are only "informative" and very likely not something we would need to create database index on. We do the same for AWS details.

internal/migrations/sql/018_add_description_reservation_instances.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE reservation_instances ADD COLUMN
+  public_ipv4 TEXT NOT NULL DEFAULT '';
+
+ALTER TABLE reservation_instances ADD COLUMN
+  public_dns TEXT NOT NULL DEFAULT '';
```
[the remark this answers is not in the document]
lzap: I forgot the yaml tags I think these are needed for OpenAPI, but you get the idea.

internal/jobs/launch_instance_aws.go
avitova: Also, there we are using the first page of the DescribeInstanceOutput, and only the [first item](https://github.com/RHEnVision/provisioning-backend/blob/67fef988a14ec93adf037e27eef77cada6d1f32d/internal/clients/http/ec2/ec2_client.go#L445) in the array.
There could be something in the NextToken field. As well as the reservation does not have to be on this index in the reservation array?
I wonder, is it safe since there is at least 6-second gap between the launching and obtaining the description?

lzap:
Just for the record, what am I suggesting is to store new data as JSON in the database in a single column with JSONB type. Postgres allows selecting data from JSON directly and even indexing, although we will never need to utilize indexing hopefully.

image

internal/jobs/launch_instance_aws.go
amirfefer: @avitova, as far as I understand, no need to iterate the rerservations array, because all of the instances are related to the same reservation, so only one reservation should return. 
Regarding pagination, I guess for now it is not needed, the max instances count for now is 45

amirfefer:
Thanks @lzap , @ezr-ondrej , @avitova  for the review
Hopefully I've addressed them all :)

internal/jobs/launch_instance_aws.go
```diff
@@ -212,3 +218,50 @@ func DoLaunchInstanceAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
...
+		for _, instance := range instancesDescriptionList {
+			err := rDao.UpdateReservationInstance(ctx, args.ReservationID, instance)
+			if err != nil {
+				return fmt.Errorf("cannot update instance description: %w", err)
+			}
+		}
+	}
+
+	return nil
+}
```
lzap, line 267: Note for me or @ezr-ondrej - the https://github.com/RHEnVision/provisioning-backend/pull/461 requires all functions to use the new `nilUnlessTimeout` here so if this gets merged before this needs to be updated.

internal/migrations/sql/019_add_detail_reservation_instances.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE reservation_instances ADD COLUMN
+  detail JSONB NOT NULL DEFAULT '{}'::jsonb
```
lzap, line 2: We do just `detail JSONB` which is nullable with no default value in the other table but this is fine.

lzap:
It’s perfect, going to test it now.

lzap:
Two problems before I can start testing tho revealed by our CI:

* You need to regenerate OpenAPI since you are adding new fields to the type.
* IQE now fails with `iqe_provisioning_api.exceptions.ApiTypeError: Invalid type for variable '0'. Required value type is str and passed type was dict at ['received_data']['instances'][0]`

This is because you are changing a type of existing field. I think it is fine since we are not V1 API yet, opinions @ezr-ondrej @akhil-jha ?

internal/payloads/reservation_payload.go
```diff
@@ -75,8 +79,8 @@ type AWSReservationResponsePayload struct {
 	// Immediately power off the system after initialization
 	PowerOff bool `json:"poweroff"`
 
-	// Instance IDs, only present for finished reservations
-	Instances []string `json:"instances,omitempty"`
+	// Instances array, only present for finished reservations
+	Instances []InstanceResponse `json:"instances,omitempty"`
 }
```
lzap, line 84: FYI this is what is causing IQE test break, this is a reasonable change so my proposal is to fix OpenAPI and wait for @akhil-jha to prepare IQE PR so we can merge this at once.

akhil-jha: Sure, tag me in the updated PR

lzap:
Looks like OpenAPI is now ready, @akhil-jha thanks!

internal/payloads/reservation_payload.go
```diff
@@ -42,6 +42,10 @@ type GenericReservationResponsePayload struct {
 	Success *bool `json:"success" nullable:"true"`
 }
 
+type InstanceResponse struct {
+	*models.ReservationInstance
+}
```
lzap, line 47: Nitpick: since this is named "instances" in JSON this could be also named "Instance".

internal/jobs/launch_instance_aws.go
```diff
@@ -212,3 +218,51 @@ func DoLaunchInstanceAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
...
+		if len(instancesDescriptionList) == 0 {
+			continue
+		}
+		for _, instance := range instancesDescriptionList {
+			err := rDao.UpdateReservationInstance(ctx, args.ReservationID, instance)
+			if err != nil {
+				return fmt.Errorf("cannot update instance description: %w", err)
+			}
+		}
+		break
```
amirfefer, line 264: forgot the `break`... now it much faster :)

ezr-ondrej (approved):
Changing the OpenAPI field is fine IMHO :+1:

internal/jobs/launch_instance_aws.go
lzap: Quite honestly, now that you brought my attention, this for-continue-break is something that I see maybe for the first time in my life. I would always write:

```
for ... {
  // request
  if there is a response {
    // update and break out
  }
}
```

This is more natural use of for loop which by design - loops so there is no need to do continue on each cycle. What you have is little bit more clunky, it is acceptable I guess :-)

internal/payloads/reservation_payload.go
lzap, line 47: Please do not use embedding here since it also pulls `instance_id` unwanted field (makes no sense to expose it through the API). Just copy the fields and in the copy function copy one by one.

lzap (changes requested):
I need to ask for one small change because OpenAPI now exposes an unwanted field.

Thanks.

internal/payloads/reservation_payload.go
amirfefer: why the instance_id is unwanted?

internal/jobs/launch_instance_aws.go
amirfefer: I agree, it's more clunky :)
I removed the `continue` and using just the break now

internal/payloads/reservation_payload.go
ezr-ondrej: I'd agree with amir here, we are exposing the instance_id in the UI, so we should include it.
I'd back the copying is bit better because its bit more clear what data are we exposing.
I believe the field we don't want is `reservation_id`, as that's already included in the response.

ezr-ondrej (approved):
LGTM, I'd probably not nest the dns and ipv4 in details, but that's just my preference, this is fine.
Great job @amirfefer :+1:

lzap (approved):
Perfect, thank you! Very nice job.

PR #449, opened 2023-03-03
build: Use static go build version

ezr-ondrej:
Use static version for go build container.
We want to migrate to new versions consciously.

OTH we want to use the latest ubi version for the final container.


lzap:
Can you drop a comment in the [go.mod](https://github.com/RHEnVision/provisioning-backend/blob/main/go.mod#L3) to instruct anyone who is modifying the value to also change this line here? Other than that 👍🏻

ezr-ondrej:
Good idea, I've done that :)

lzap (approved):
TY

PR #450, opened 2023-03-06
chore: update dependencies

lzap:
Tern v2 is finally out: https://github.com/jackc/tern/issues/66

However, due to a dependency (puddle/v2) requiring 1.19 version of Go and Red Hat Go Runtime container being still at 1.18, we need to keep that pin for a while until we are able to move away from the beta.

This patch removes the pin, however, we still need to stay on beta. Updates also few other deps.


ezr-ondrej (approved):
sure, all seems minor bumps :+1:

PR #453, opened 2023-03-06
Logger initialization refactor

ezr-ondrej:
Pulls initialization in a single function, `InitializeStdout` is used just in tests, where we know we need only stdout logger.


lzap (approved):
Much cleaner thanks.

PR #454, opened 2023-03-07
refactor: Add test for valid region/zone/location

adiabramovitch:
Adding a test for each provider's reservation checking we are failing the test if the region/zone/location provided are not supported.

### DIFF
--- a/internal/services/aws_reservation_service_test.go
+++ b/internal/services/aws_reservation_service_test.go
@@ -19,7 +19,6 @@ import (
 )
 
 func TestCreateAWSReservationHandler(t *testing.T) {
-	var err error
 	var json_data []byte
 	ctx := stubs.WithAccountDaoOne(context.Background())
 	ctx = identity.WithTenant(t, ctx)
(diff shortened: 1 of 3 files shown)


internal/services/gcp_reservation_service_test.go
```diff
@@ -29,33 +28,65 @@ func TestCreateGCPReservationHandler(t *testing.T) {
...
+		require.Equal(t, http.StatusOK, rr.Code, "Handler returned wrong status code")
+
+		stubCount := stubs.GCPReservationStubCount(ctx)
+		assert.Equal(t, 1, stubCount, "Reservation has not been Created through DAO")
+	})
+
+	t.Run("failed reservation with invalid zone", func(t *testing.T) {
+		var err error
+		require.NoError(t, err, "failed to generate pubkey")
+		source, err := Clientstubs.AddSource(ctx, models.ProviderTypeGCP)
```
avitova, line 67: I do not think there is a need to check for the NoError here, as you declare a new err variable the line before. The err should have the default error value (nil).

internal/services/aws_reservation_service_test.go
```diff
@@ -28,30 +27,59 @@ func TestCreateAWSReservationHandler(t *testing.T) {
...
+		req.Header.Add("Content-Type", "application/json")
+
+		rr := httptest.NewRecorder()
+		handler := http.HandlerFunc(services.CreateAWSReservation)
+		handler.ServeHTTP(rr, req)
+
+		require.Equal(t, http.StatusOK, rr.Code, "Handler returned wrong status code")
+
+		stubCount := stubs.AWSReservationStubCount(ctx)
+		assert.Equal(t, 1, stubCount, "Reservation has not been Created through DAO")
```
avitova, line 57:
```suggestion
		assert.Equal(t, 1, stubCount, "Reservation has not been created through DAO")
```

internal/services/gcp_reservation_service_test.go
```diff
@@ -29,33 +28,65 @@ func TestCreateGCPReservationHandler(t *testing.T) {
...
+		req.Header.Add("Content-Type", "application/json")
+
+		rr := httptest.NewRecorder()
+		handler := http.HandlerFunc(services.CreateGCPReservation)
+		handler.ServeHTTP(rr, req)
+
+		require.Equal(t, http.StatusOK, rr.Code, "Handler returned wrong status code")
+
+		stubCount := stubs.GCPReservationStubCount(ctx)
+		assert.Equal(t, 1, stubCount, "Reservation has not been Created through DAO")
```
avitova, line 61: ditto

avitova (commented):
Even more tests, nice!
Left some nitpicks, but it looks good. By the way, is there a reason for differences in a message starting with a capital letter vs no capital letter?

internal/services/gcp_reservation_service_test.go
adiabramovitch: Was a leftover, great catch :)

ezr-ondrej (approved):
Looks great to me, great to cover these corner cases in our testing as its fast and integration tests won't be able to cover all of them, so huge fan :+1:

@lzap you've added the funcionality that's tested, could you also peek if the coverage makes sense? :)

Side note: I think that validating the substring is not perfect, but great first step, we could figure out a tooling to parse errors to json struct to verify the json structure is valid. That's beyond the scope of this PR tho.

lzap (approved):
Yeah this is good, thanks.

lzap:
@avitova please merge once you are happy.

avitova:
LGTM:) Thank you!

PR #455, opened 2023-03-08
chore: Update Makefile to work for Go 1.18

avitova:
Update of the make install-tools to work for our required version of Go.


lzap:
For the record, the reason is that one of the dependencies now requires Go 1.19 and we are not there yet.

PR #456, opened 2023-03-08
feat: shorten Azure polling intervals

ezr-ondrej:
Many resources are created faster then the default 30 second interval. This makes in the worst case the whole proces up to 30 times # of resources slower.

We want to shorten this inter

Fixes HMS-1404


lzap (approved):
Perfect.

PR #457, opened 2023-03-08
fix: cascade delete of pubkey to Azure details

ezr-ondrej:
This is what we do for other providers.
We loose all the details about the reservation.

Fixes HMS-1402


akhil-jha (approved):
Works for me!

PR #459, opened 2023-03-08
test: Allow collecting planned jobs in tests

ezr-ondrej:
Adds two structs in tests, one that drops all the jobs, second that stores them in the context.
Second one needs to be initialized in the context first.

Shows the usage of it on an Azure service.
Needs to have the context passed for the GetEnqueuer, but given it's called just few times over our code I guess its acceptable change.


avitova (approved):
More tests! LGTM

lzap (approved):
Sure, thanks.

PR #460, opened 2023-03-09
feat: Create Azure VMs in paralel

ezr-ondrej:
Splits Azure launch start and polling for status.
Now we start the launch process for all instances first, only then we start polling for the creation result.

Fixes HMS-1407

### DIFF
--- a/internal/clients/http/azure/create_vm.go
+++ b/internal/clients/http/azure/create_vm.go
@@ -18,7 +18,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
-	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/codes"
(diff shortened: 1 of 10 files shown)


akhil-jha:
So I merged the azure test(count=1), the whole api tests in pr_check took about 15mins.

akhil-jha:
Testing with count=n locally

internal/clients/interface.go
```diff
@@ -109,6 +109,17 @@ type Azure interface {
...
+	// location - to deploy into
+	// imageName - the imageID will be inferred as /subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/images/{imageName}
+	// returns a resumeToken
+	BeginCreateVM(ctx context.Context, location string, resourceGroupName string, imageName string, pubkey *models.Pubkey, instanceType InstanceTypeName, vmName string) (string, error)
+
+	// WaitForVM starts polling of status of Azure virtual machine creation and wait until it's done.
+	// needs resumeToken from BeginCreateVM operation.
+	WaitForVM(ctx context.Context, resumeToken string) (*string, error)
+
 	// CreateVM creates Azure virtual machine.
```
lzap, line 123: I understand why you want to split begin and wait, but do you really need to expose both via the interface? I mean, these both could be unexported functions and callers would continue using `CreateVM` as usual?

ezr-ondrej: Sure, we can start with that and if we want to expose this to multiple steps, we can get back to this solution :+1:

lzap: Thanks I would appreciate the interface to be simple, because something tells me we might be removing the waiting pretty soon. :-D

ezr-ondrej:
Alright, I've hidden the Waiting to inside of the client, but kept it as Exported function in case we want to get back to waiting in second job step :)

I've rebased this and extracted the Virtual machine parameters to a struct, so it's easier to adjust later on.

internal/clients/http/azure/create_vm.go
```diff
@@ -38,24 +37,104 @@ const (
 	vmPollFrequency       = 10 * time.Second
 )
 
-func (c *client) CreateVM(ctx context.Context, location string, resourceGroupName string, imageName string, pubkey *models.Pubkey, instanceType clients.InstanceTypeName, vmName string, userData []byte) (*string, error) {
+func (c *client) BeginCreateVM(ctx context.Context, vmParams clients.AzureInstanceParams, vmName string) (string, error) {
```
adiabramovitch, line 40: I see that you have removed vmName as well, should I remove it as well for consistency?

ezr-ondrej: Yeah, i guess so. That's specific to a single machine i feel like... But happy to discuss ofc :)

internal/clients/http/azure/create_vms.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (c *client) CreateVMs(ctx context.Context, vmParams clients.AzureInstanceParams, amount int64, vmNamePrefix string) ([]*string, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "CreateVMs")
+	defer span.End()
+
+	logger := logger(ctx)
+	logger.Debug().Msgf("Started creating %d Azure VM instances", amount)
+
+	vmIds := make([]*string, amount)
+	var resumeTokens []string
```
lzap, line 21: Why vmIds is preallocated while tokens are not? They should be both the same, I mean if you skip a VM because of error, then it makes sense to not to preallocate vmids.

Also, vmIds are pointers so they may contain nils. Why is that? If I launch 5 instances and one is skipped, receiving slice of 4 values and one randomly placed nil is not a great experience.

I suggest to make this consistent and when VMs were skipped for some reason, return a slice that has a lower length. We cannot do much about it, there is no transactions or anything :-)

internal/clients/interface.go
```diff
@@ -111,10 +111,11 @@ type Azure interface {
...
-	// Most of the parameters are constant for now.
-	// location - to deploy into
-	// imageName - the imageID will be inferred as /subscriptions/{subscriptionID}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/images/{imageName}
-	CreateVM(ctx context.Context, location string, resourceGroupName string, imageName string, pubkey *models.Pubkey, instanceType InstanceTypeName, vmName string, userData []byte) (*string, error)
+	CreateVM(ctx context.Context, params AzureInstanceParams, vmName string) (*string, error)
+
+	// CreateVMs creates multiple Azure virtual machines
+	// Returns array of instance IDs and error if something went wrong
+	CreateVMs(ctx context.Context, instanceParams AzureInstanceParams, amount int64, vmNamePrefix string) (vmIds []*string, err error)
 }
```
lzap, line 119: Thanks for not exposing wait functions. I will say that CreateVM = CreateVMs with amount=1 so we might not want to expose the CreateVM at all. Remember, more functions = more tests = more states to cover.

internal/clients/stubs/azure_stub.go
```diff
@@ -2,16 +2,19 @@ package stubs
...
 	"strconv"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
-	"github.com/RHEnVision/provisioning-backend/internal/models"
-	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 )
 
+var NotStartedVMError = errors.New("the VM under given resumeToken not started")
```
lzap, line 14:
```suggestion
var ErrNotStartedVM = errors.New("the VM under given resumeToken not started")
```

Prefix Err for values, Error suffix for types. We must enable linter once we fix them all :-)

internal/clients/stubs/azure_stub.go
```diff
@@ -41,21 +44,64 @@ func (stub *AzureClientStub) Status(ctx context.Context) error {
...
+	}
+	for i = 0; i < amount; i++ {
+		vmIds[i], err = stub.WaitForVM(ctx, resumeTokens[i])
+		if err != nil {
+			return vmIds, err
+		}
+	}
+
+	return vmIds, nil
+}
```
lzap, line 79: I think since you dropped exporting begin/wait functions, you can also simplify the stub  :-D

If you drop the other function and only keep CreateVMs then you can further simplify it.

Also, you do not have to really begin/wait in a stub, stub can simply return "yep, all were launched" :-)

lzap:
Dropped few random comments, also rebase needed.

internal/clients/stubs/azure_stub.go
ezr-ondrej: I kept the stub ready for what I anticipate will happen - exposing the begin function and don't wait for the machines to come up.
We should discuss unifying this, as for AWS we just get ack from AWS "we are working on it" and we are not waiting for the fullfilment and that's IMHO the right approach, so I'd discuss it first and simplify the stub once we are sure we won't want only the "BeginCreate" :)
But if you don't like it I can save it in local branch ofc xD I just thought why not mimic the real thing, when it's already done :)

internal/clients/http/azure/create_vms.go
ezr-ondrej: Pushed a new commit for easier review with strongly typing vmIDs and preallocated resumeTokens.

internal/clients/interface.go
ezr-ondrej: in a new commit, I've removed the `CreateVM` function all together, I believe it is not very usefull anymore.

internal/clients/stubs/azure_stub.go
lzap: I don’t see us not waiting for VM to launch, these are very cheap jobs basically polling HTTP endpoint, workers can handle thousands of them at one time. I don’t see this as a problem. We might update our job to send a notification or something and update our UI but I think an API with just "CreateVMs" is okay.

ezr-ondrej: I was thinking exactly of sending memo to UI and I think that easiest way to do this is to split it in the client API and be able to figure out in the job where we are at and that can be done (cleanly) only if we surface the split of planning the launch and waiting for it ;)

lzap:
One more rebase, when rebasing please make sure to avoid `logger.Error().Err(err).Msgf("xxx %s", err)`: https://github.com/RHEnVision/provisioning-backend/pull/488

ezr-ondrej:
Rebased and I believe I'm not adding any offensive Error logs :+1:

internal/clients/http/azure/create_vm.go
```diff
@@ -38,68 +37,113 @@ const (
...
 
-	virtualNetwork, err := c.createVirtualNetwork(ctx, location, resourceGroupName, vnetName)
+	logger := logger(ctx)
+	imageID := c.getImageId(ctx, vmParams.ResourceGroupName, vmParams.ImageName)
+
+	virtualNetwork, err := c.createVirtualNetwork(ctx, vmParams.Location, vmParams.ResourceGroupName, vnetName)
 	if err != nil {
 		span.SetStatus(codes.Error, "cannot create virtual network")
-		logger.Error().Err(err).Msg("cannot create virtual network")
+		logger.Error().Err(err).Msgf("cannot create virtual network")
```
lzap, line 109: Why do you do that? :-) `Msg` is actually correct, I am not sure if `Msgf` would even work. Probably yes.

You have this at multiple places.

lzap (commented):
Other than that I am okay.

internal/clients/http/azure/create_vm.go
ezr-ondrej: no idea, bad rebase I'm guessing... let me fix that :)

internal/clients/http/azure/create_vms.go
```diff
@@ -0,0 +1,49 @@
...
+func (c *client) CreateVMs(ctx context.Context, vmParams clients.AzureInstanceParams, amount int64, vmNamePrefix string) ([]clients.AzureInstanceID, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "CreateVMs")
+	defer span.End()
+
+	logger := logger(ctx)
+	logger.Debug().Msgf("Started creating %d Azure VM instances", amount)
+
+	vmIds := make([]clients.AzureInstanceID, amount)
+	resumeTokens := make([]string, amount)
+	var i int64
```
lzap, line 22: Very optimistic about Azure API capabilities and ARM throughput :-)

lzap (approved):
Merging since the holidays are long.

PR #461, opened 2023-03-10
fix(HMS-1403): timeout for job queue

lzap:
This implements a relatively long (default) timeout for job queue. The purpose of this is when there is a bug that makes jobs to get stuck, resources (CPU/memory) can be kept on a good value until we figure out what is going on. I propose 30 minutes as a safe timeout value.

The implementation uses context deadline and adds checks to every job queue "step" (function) which is using the reservation status update so jobs are cancelled early. Of course, both AWS and Azure SDKs do support context cancellation, so in practice, job should be cancelled immediately once it reaches the deadline. The `nilUnlessDeadline` function serves a safety measure in case we copy-paste existing code and we use a library that does not support context deadlines.

Initially, I used `panic` to handle timeouts but this was very hard to test. This new version use explicit checking via the newly introduced function. The PR is separated into three commits:

* the timeout implementation
* new integration test and refactoring needed in order to be able to introduce Redis
* enforcement of config and make documentation via CI job @ezr-ondrej you really wanted this


lzap:
Do not merge, using `panic` is causing serious issues in tests.

lzap:
Finally this is ready.

adiabramovitch:
@lzap SOmething weird why are the tests still running?

lzap:
That is sort of expected because I renamed "Go tests" to more informative "Unit tests". Also renamed "DAO tests" to "Integration tests". Not sure why "Go tests" triggered while "DAO tests" already disappeared. To to double check, run them locally please.

ezr-ondrej:
> Also renamed "DAO tests" to "Integration tests". Not sure why "Go tests" triggered while "DAO tests" already disappeared.

Both are there (under scroll bar), because both are required checks, will change it once this is merged :+1:

mk/z_common.mk
```diff
@@ -3,5 +3,5 @@
 #
 
 .PHONY: validate
-validate: validate-spec validate-clients
+validate: validate-spec validate-clients validate-example-config
```
ezr-ondrej, line 6: :orange_heart:

pkg/worker/redis.go
```diff
@@ -206,8 +209,10 @@ func (w *RedisWorker) processJob(ctx context.Context, job *Job) {
 	logger.Info().Msg("Dequeued job from Redis")
 	ctx = contextLogger(ctx, job)
 	if h, ok := w.handlers[job.Type]; ok {
+		cCtx, cFunc := context.WithTimeout(ctx, config.Worker.Timeout)
+		defer cFunc()
 		metrics.ObserveBackgroundJobDuration(job.Type.String(), func() {
-			h(ctx, job)
+			h(cCtx, job)
 		})
```
lzap, line 216: These three lines are actually the change + configuration option.

All the rest are just tests :-)

pkg/worker/redis.go
```diff
@@ -158,10 +159,12 @@ func (w *RedisWorker) dequeueLoop(ctx context.Context, i, total int) {
 func recoverAndLog(ctx context.Context) {
 	if rec := recover(); rec != nil {
 		logger := ctxval.Logger(ctx).Error().Stack()
-		if err, ok := rec.(error); ok {
+		err, ok := rec.(error)
+		if ok {
 			logger = logger.Err(err)
```
lzap, line 164: Ah this is not even needed, the initial version contained some changes in this function but I dropped it (it was really hard to test for `panic`). I can remove this hunk.

internal/testing/integration/db_env.go
```diff
@@ -4,33 +4,19 @@ import (
...
+	err := db.Initialize(context.Background(), "integration")
 	if err != nil {
 		panic(fmt.Errorf("cannot connect to database: %w (integration schema)", err))
 	}
 }
 
-func CloseEnvironment(ctx context.Context) {
+func CloseDbEnvironment(ctx context.Context) {
 	db.Close()
 	jq.StopDequeueLoop(ctx)
```
ezr-ondrej, line 21: Probably a leftover?

pkg/worker/redis.go
ezr-ondrej: Shall we add this to memory implementation as well?
You know, just for consistency...

internal/queue/tests/main_test.go
```diff
@@ -0,0 +1,41 @@
...
+func reset() {
+	integration.DbSeed()
+}
+
+func TestMain(t *testing.M) {
+	ctx := context.Background()
+	ctx = integration.InitConfigEnvironment(ctx, "../../../config/test.env")
+	// override the default value to reasonable timeout for testing
+	config.Worker.Queue = "redis"
+	config.Worker.Timeout = 1 * time.Second
```
ezr-ondrej, line 29: I'm really nitpicking here, pls feel free to ignore me, but this would look nice in a function (`configureJobqueue` for example).
But I'm all for merging as is and addressing this in a follow-up, if I get around to this, such a global setting would look nice in a config file, that we would create for the GH actions (definitelly out of scope this PR).

ezr-ondrej (commented):
Very nice job indeed, I love the tests :orange_heart: and basically all about this PR :+1:
I've got just one comment I'd like to see addressed, I think we have a leftover there) all the other comments are not blocking :+1:

ezr-ondrej:
I'll force merge once ready btw and rename the required checks right after :+1:

pkg/worker/redis.go
lzap: Yes this was a leftover, I will amend this change which just improves the error handling - when the panic was caused by error we use the native stack tracing and error reporting of zerolog. Otherwise we just do manual stacktrace:

```
		if err, ok := rec.(error); ok {
			logger.Err(err).Stack().Msgf("Job queue panic: %s", err.Error())
		} else {
			logger.Msgf("Error during job handling: %v, stacktrace: %s", rec, debug.Stack())
		}
```

pkg/worker/redis.go
lzap: Memory is only used for development, I don’t see any benefit but why not.

internal/queue/tests/main_test.go
lzap: Don’t exactly understand what you mean.

lzap:
Pushed changes.

ezr-ondrej:
The new validation is failing as GCP PR had broken it :(

lzap:
What was failing? Docs? I could have fixed that, but thanks.

lzap:
NP did not notice it.

ezr-ondrej (approved):
Thanks ! :orange_heart:

PR #464, opened 2023-03-14
chore: deployment is 3-3-1 replicas

lzap:
SRE bot informs us about a check, we need to turn it off for statuser.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -17,6 +17,8 @@ objects:
       labels:
         app: provisioning-backend
         service: provisioning
+      annotations:
+        ignore-check.kube-linter.io/minimum-three-replicas: "statuser pod runs in a single instance"
     spec:
       envName: ${ENV_NAME}
       featureFlags: true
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
Potentionaly we could try adding annotation to the given deployment (statuser), it allows annotations, but hard to say if it would work adding this specific annotation just for the given deployment.

deploy/clowdapp.yaml
```diff
@@ -104,7 +106,7 @@ objects:
                 cpu: ${CPU_REQUESTS}
                 memory: ${MEMORY_REQUESTS}
         - name: statuser
-          # statuser pod must have exactly one replicas
+          # statuser pod is just one replica for easier throttling
           replicas: 1
```
ezr-ondrej, line 110: I'd love to try the annotations here, but up to you :)
```suggestion
          replicas: 1
          metadata:
            annotations:
              ignore-check.kube-linter.io/minimum-three-replicas: "statuser pod runs in a single instance"
```

lzap: Let’s try: https://github.com/RHEnVision/provisioning-backend/pull/467

Let’s merge whatever works.

lzap:
I am going to merge this because stage is stuck and Ondra is offline.

lzap:
I am merging back to 2-2-1 because we were hitting CPU limits. We will disable the linter to stop notifications about 3 minimum pods.

PR #465, opened 2023-03-14
chore: Update Kafka deprecated version

avitova:
Kafka 3.3.1 version leads to 404. The current version is higher.
As the 3.3.1 was not redirected, the setup script would create the kafka.tgz on the first time trying with the content of 404 request. As the file kafka.tgz would exist after this failed attempt, the second time running the setup would end up as "Kafka already downloaded".


ezr-ondrej (approved):
I don't use kafka locally, but indeed the download folder for 3.3.1 is gone.
Thanks @avitova ! :+1:

PR #468, opened 2023-03-15
feat(HMS-1429): Add get gcp reservation by id 

adiabramovitch:
This PR is Adding get GCP reservation by id as we have for Azure and for AWS.

### DIFF
--- a/internal/payloads/reservation_payload.go
+++ b/internal/payloads/reservation_payload.go
@@ -142,6 +142,9 @@ type GCPReservationResponsePayload struct {
 
 	// Immediately power off the system after initialization
 	PowerOff bool `json:"poweroff"`
+
+	// Instances IDs, only present for finished reservations.
+	Instances []InstanceResponse `json:"instances,omitempty"`
 }
 
 type NoopReservationResponsePayload struct {
@@ -306,7 +309,15 @@ func NewAzureReservationResponse(reservation *models.AzureReservation, instances
 	return &response
 }
 
-func NewGCPReservationResponse(reservation *models.GCPReservation) render.Renderer {
+func NewGCPReservationResponse(reservation *models.GCPReservation, instances []*models.ReservationInstance) render.Renderer {
+	instanceIds := make([]InstanceResponse, len(instances))
+	for iter, inst := range instances {
+		instanceIds[iter] = InstanceResponse{
+			InstanceID: inst.InstanceID,
+			Detail:     inst.Detail,
+		}
+	}
+
 	response := GCPReservationResponsePayload{
 		PubkeyID:         reservation.PubkeyID,
 		ImageID:          reservation.ImageID,
@@ -317,6 +328,7 @@ func NewGCPReservationResponse(reservation *models.GCPReservation) render.Render
 		GCPOperationName: reservation.GCPOperationName,
 		ID:               reservation.ID,
 		PowerOff:         reservation.Detail.PowerOff,
+		Instances:        instanceIds,
 	}
 	return &response
 }
--- a/docs/configure-gcp.md
+++ b/docs/configure-gcp.md
@@ -38,10 +38,8 @@ Create a new Service account:
 7. Choose the permissions the provisioning app needs (TBC):
 
 Roles:
- - Service Account User
-  
-Permissions:
-  - compute.instances.insert
+   - Service Account User
+   - Compute Admin (We are allowing here more permissions than needed)
 
 8. Click CONTINUE
 9.  Click DONE 
@@ -61,11 +59,8 @@ This is a setup in the account in which the service shall deploy the actual inst
 5. Fill in the Service Account e-mail under New principals.
 6. Choose the permissions the provisioning app needs (TBC): 
 Roles:
- - Service Account User
-  
-Permissions:
-  - compute.instances.insert
-
+   - Service Account User
+   - Compute Admin (We are allowing here more permissions than needed)
 7. Click SAVE.
 
 #### Authenticating as the service account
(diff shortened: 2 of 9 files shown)


internal/payloads/reservation_payload.go
```diff
@@ -134,6 +134,9 @@ type GCPReservationResponsePayload struct {
 
 	// Immediately power off the system after initialization
 	PowerOff bool `json:"poweroff"`
+
+	// Instances IDs, only present for finished reservations.
+	Instances []string `json:"instances,omitempty"`
```
lzap, line 139: I will make sense to create struct here instead of string. It is very likely that we will want to also fetch DNS and IP of the launched instances alongside id and refactoring later is painful, as you can see here: https://github.com/RHEnVision/provisioning-backend/pull/448/files#diff-93db714a65f393cc33692e57cab38ad7ae1b8b12b3a0cefb1769b435b747a0d1

For now, I suggest to create a struct type with just a string as the single field so we can add IP/DNS later.

lzap (changes requested):
Perfect, tho, to prevent a problem with future string to struct change I suggest to create a struct already now so we can add the remaining fields later in the future.

internal/payloads/reservation_payload.go
adiabramovitch: I see that Amir has added `InstanceResponse` type for AWS. 
I guess I should wait until his PR will be merged?

internal/payloads/reservation_payload.go
```diff
@@ -42,6 +42,10 @@ type GenericReservationResponsePayload struct {
 	Success *bool `json:"success" nullable:"true"`
 }
 
+type InstanceResponse struct {
+	*models.ReservationInstance
+}
+
```
lzap, line 48: This will cause conflict with @amirfefer 's PR but this will likely go in earlier because there is no IQE test breakage here. Sorry :-)

api/openapi.gen.yaml
```diff
@@ -558,6 +604,59 @@ components:
...
+        image_id:
+          type: string
+        instances:
+          items:
+            properties:
+              instance_id:
+                type: string
+              reservation_id:
+                format: int64
+                type: integer
```
lzap, line 642: This field is exposed for no reason tho, I mean this is reservation response, there is no need to copy reservation_id N times, it has zero information since the user performed GET call with reservation ID already.

Now since any refactoring would cause pains, what I suggest is to merge this as is and ask @amirfefer to hide this field (by simply copying only relevant fields into the payload type from model type).

lzap (approved):
TY I will explain to Amir what to do.

lzap (changes requested):
Ok since instance detail is higher priority than GCP, I take back what I said - I will ask Amir to do this small change first, merge his PR since IQE tests are almost ready and then I will ask you to rebase on top of that. Will be just a tiny conflict for the `InstanceResponse` type.

lzap:
@adiabramovitch we have merged @amirfefer 's PR. In case you are wondering what was the change, it was simple: https://github.com/RHEnVision/provisioning-backend/compare/d53b532f924f75bd0198456b7d36c81bf7857843..27bcce0c7cffb1dcdd20e3df3b23e9eb39b46e5d

Please rebase this PR is ready IMHO. Thank you.

ezr-ondrej (approved):
Yup, looks good :+1:
I guess this should wait for @akhil-jha to prepare IQE tests (btw we can help if that would help, given we already have credentials).

Thanks for fixing the Azure wording :orange_heart:

internal/payloads/reservation_payload.go
```diff
@@ -42,6 +42,10 @@ type GenericReservationResponsePayload struct {
 	Success *bool `json:"success" nullable:"true"`
 }
 
+type InstanceResponse struct {
+	*models.ReservationInstance
```
ezr-ondrej, line 46: We should probably do the same what we did in Amir's PR and for the Instances response (copy the instance details), but I guess that will be done while rebasing.

adiabramovitch: rebased

lzap:
Re: Unleash. I think it is fine not to have it any only hide the UI part, @ezr-ondrej

lzap:
A conflict again. Can you rebase?

Shall we just merge this since GCP is not yet exposed to the users without automation? What harm can it done, GCP code is already in. @ezr-ondrej @akhil-jha

lzap:
Errors with:

AttributeError: 'ProvisionedHost' object has no attribute 'client'

lzap:
Fails for my PR as well, pinged QA folks what is going on :-)

lzap (approved):
All good.

I suggest to merge this PR which introduces OpenAPI for GCP even if we are not yet ready to write a IQE test. If we end up need to push to prod without GCP, we can add an unleash flag so the reservation returns an error. @akhil-jha

cmd/spec/main.go
```diff
@@ -23,6 +23,8 @@ func addPayloads(gen *APISchemaGen) {
 	gen.addSchema("v1.AWSReservationResponse", &payloads.AWSReservationResponsePayload{})
 	gen.addSchema("v1.AzureReservationRequest", &payloads.AzureReservationRequestPayload{})
 	gen.addSchema("v1.AzureReservationResponse", &payloads.AzureReservationResponsePayload{})
+	gen.addSchema("v1.GCPReservationRequest", &payloads.GCPReservationRequestPayload{})
+	gen.addSchema("v1.GCPReservationResponse", &payloads.GCPReservationResponsePayload{})
```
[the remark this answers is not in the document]
lzap: Correction: I suggest to split OpenAPI from this PR and merge without it.

PR #472, opened 2023-03-16
feat(HMS-1001): Add created instances ids to GCP reservation

adiabramovitch:
After an investigation, I have not found a better way to fetch the information of the instances created under a single reservation. 

This PR: 
- Generates a uuid for each bulkInsert request and  will be used as a label
-  Fetches the created instances corresponding to bulkInsert requests' uuid 
-  saves the instances ids in the db and also share them under GetGCPByID

**Things I would like to get feedback on:** 
- Does the usage of uuid make sense like that? 
-  Did I use otel correctly and did not miss something?

### DIFF
--- a/internal/clients/http/gcp/gcp_client.go
+++ b/internal/clients/http/gcp/gcp_client.go
@@ -4,16 +4,20 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 
-	"cloud.google.com/go/compute/apiv1/computepb"
-	"github.com/RHEnVision/provisioning-backend/internal/config"
-	"google.golang.org/api/iterator"
-	"google.golang.org/api/option"
+	guuid "github.com/google/uuid"
 
 	compute "cloud.google.com/go/compute/apiv1"
+	"cloud.google.com/go/compute/apiv1/computepb"
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
 	"github.com/RHEnVision/provisioning-backend/internal/ptr"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/codes"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
 )
 
 type gcpClient struct {
@@ -24,6 +28,8 @@ type gcpClient struct {
 func init() {
 	clients.GetGCPClient = newGCPClient
 }
+
+const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/http/gcp"
 
 // GCP SDK does not provide a single client, so only configuration can be shared and
 // clients need to be created and closed in each function.
@@ -46,6 +52,9 @@ func (c *gcpClient) Status(ctx context.Context) error {
 }
 
 func (c *gcpClient) ListAllRegions(ctx context.Context) ([]clients.Region, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListAllRegions")
+	defer span.End()
+
 	client, err := compute.NewRegionsRESTClient(ctx, c.options...)
 	if err != nil {
 		return nil, fmt.Errorf("unable to create GCP regions client: %w", err)
@@ -66,6 +75,7 @@ func (c *gcpClient) ListAllRegions(ctx context.Context) ([]clients.Region, error
 			break
 		}
 		if err != nil {
+			span.SetStatus(codes.Error, err.Error())
 			return nil, fmt.Errorf("iterator error: %w", err)
 		}
 		regions = append(regions, clients.Region(*region.Name))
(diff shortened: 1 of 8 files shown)


internal/clients/interface.go
```diff
@@ -143,5 +143,5 @@ type GCP interface {
 	ListAllRegions(ctx context.Context) ([]Region, error)
 
 	// InsertInstances launches one or more instances
```
ezr-ondrej, line 145: would be nice to document the return parameters as there is more now, it might be handy to know what those are.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -133,8 +150,43 @@ func (c *gcpClient) InsertInstances(ctx context.Context, namePattern *string, im
...
+				instances := pair.Value.Instances
+				for _, o := range instances {
+					idAsString := strconv.FormatUint(o.GetId(), 10)
+					ids = append(ids, &idAsString)
+				}
+			}
+		}
+		return ids, ptr.To(op.Name()), nil
+	}
+	return nil, ptr.To(op.Name()), nil
```
lzap, line 191: Since you are changing the field to be struct rather than just string, I think it is worth storing both our UUID tag as well as instance id (alongside DNS/IP in the future).

Not now of course, just an idea to have the UUID stored in the reservations details JSON as well just in case we need it sometimes.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -133,8 +150,43 @@ func (c *gcpClient) InsertInstances(ctx context.Context, namePattern *string, im
...
+	if op.Done() {
+		ids := make([]*string, 0)
+		instances := client.AggregatedList(ctx, lstReq)
+		for {
+			pair, err := instances.Next()
+			if errors.Is(err, iterator.Done) {
+				log.Trace().Msg("Instances iterator has finished")
+				break
+			} else if err != nil {
+				log.Trace().Msgf("An error has occured %s", err.Error())
```
lzap, line 178: Log level. Also I would also try to explain when this happened, e.g. "Error during fetching instance ids" or whatever.

internal/config/config.go
```diff
@@ -92,7 +92,7 @@ var config struct {
 	GCP struct {
 		ProjectID   string `env:"PROJECT_ID" env-default:"" env-description:"GCP service account project id"`
 		JSON        string `env:"JSON" env-default:"e30K" env-description:"GCP service account credentials (base64 encoded)"`
-		DefaultZone string `env:"DEFAULT_ZONE" env-default:"us-east1" env-description:"GCP region when not provided"`
+		DefaultZone string `env:"DEFAULT_ZONE" env-default:"us-central1-a" env-description:"GCP region when not provided"`
```
lzap, line 95: Why? HCCP is deployed in us-east1 I would expect the default to align with that (as well as the AWS default).

internal/jobs/launch_instance_gcp.go
```diff
@@ -83,5 +83,17 @@ func DoLaunchInstanceGCP(ctx context.Context, args *LaunchInstanceGCPTaskArgs) e
...
+	for _, instanceId := range instances {
+		err = rDao.CreateInstance(ctx, &models.ReservationInstance{
+			ReservationID: args.ReservationID,
+			InstanceID:    *instanceId,
+		})
+		if err != nil {
+			return fmt.Errorf("cannot create instance reservation for id %d: %w", instanceId, err)
+		}
+		logger.Info().Str("instance_id", *instanceId).Msgf("Created new instance via GCP reservation %s", *opName)
+	}
```
lzap, line 96: We need to do this in a DB transaction, I suggest to refactor the `CreateInstance` function to `CreateInstances` accepting slice of `ReservationInstance` and then using `dao.WithTransaction` to perform the operation atomically.

We can do this in a followup too, that is fine. We currently do this in AWS too.

lzap (changes requested):
Just few nitpicks, nice solution indeed. I would probably choose `rhhcc-rid` to indicate reservation id just in case we later introduce some kind of "instance id" that could be possibly `rhhcc-id`.

internal/clients/http/gcp/gcp_client.go
ezr-ondrej: also we could use the `.Err(err)` for the error logging.

internal/clients/http/gcp/gcp_client.go
ezr-ondrej, line 191: This case seems to be accessible only when the operation is not done, but given we were waiting for it, I'd consider it a failed state.

This could be IMHO done by

```go
if !op.Done() {
  return nil, ptr.To(op.Name()), <error explaining the failure - the operation name ..... has failed to finish within expected time>
}
```

internal/config/config.go
adiabramovitch: us-east1 is a region and not a zone is GCP: 
https://cloud.google.com/compute/docs/regions-zones
What zone do you want to be the default?

internal/clients/http/gcp/gcp_client.go
adiabramovitch: You are right, it is a failure. fixing

internal/jobs/launch_instance_gcp.go
adiabramovitch: I think it is out of scope for this PR, no problem doing it as a follow up

internal/config/config.go
ezr-ondrej: I've not found in docs what zone the image gets built into, but IMHO, we should use the same zone.

ezr-ondrej:
> Depends on #468

Why? :thinking:
I believe the spec should go in last and ideally include changes in this PR :thinking: 
@akhil-jha will be adding tests according to the spec and this PR changes it again.

If we could, I'd do the oposite - merge this one and have the spec being final with changes from this PR included in the spec already.

internal/config/config.go
ezr-ondrej: @adiabramovitch it's https://github.com/osbuild/image-builder/blob/c02a1be7dea2252927b684a06e1feebc47de42b4/templates/image-builder.yml#L232 `us-east4` region for image builder (so zone would be us-east4-a)?

IMHO going forward, we should have a region selection and zone would be selected within that region on the background.

Tho for now I'd be ok to have default to same what image builder has and lets merge :)

ezr-ondrej (approved):
the zone would be ideal to keep in sync with Image Builder.

Apart of it, its ready to go IMO :+1:

ezr-ondrej:
> * Did I use otel correctly and did not miss something?

Btw otel usage feels right to me, thanks for adding that! :+1:

PR #473, opened 2023-03-19
refactor: Simplify OpenAPI generator

ezr-ondrej:
Moves all type registration to function to isolate the exact place where contributor should add their code.

Removes additional struct used just for generating.
Simplifies generating and removes need for conversion from JSON to YAML.
Moves to standard go yaml package, thanks to no need for the conversion.


cmd/spec/main.go
```diff
@@ -1,23 +1,51 @@
...
 	"encoding/json"
-	"io/ioutil"
+	"os"
 	"reflect"
 
 	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/getkin/kin-openapi/openapi3gen"
-	"github.com/ghodss/yaml"
+	"gopkg.in/yaml.v2"
```
ezr-ondrej, line 11: I wanted to move to `gopkg.in/yaml.v3` that we already use, but that reformats the whole yaml spec, which makes it quite hard to review.

adiabramovitch (approved):
LGTM
Do we need QE ack?

lzap (approved):
Nice I have similar changes locally but never got to finishing it because at the time I wanted to bump version of the generator the upstream was broken. Nice.

lzap:
> Do we need QE ack?

This change must not change anything in the OpenAPI. I don’t think it does, do you see anything? It is just a different order of keys because of a different parser I believe.

api/openapi.gen.yaml
```diff
@@ -696,10 +678,28 @@ components:
...
+          schema:
+            $ref: '#/components/schemas/v1.ResponseError'
 servers:
-- description: Local development
-  url: http://0.0.0.0:{port}/api/{applicationName}
+- url: http://0.0.0.0:{port}/api/{applicationName}
+  description: Local development
   variables:
     applicationName:
       default: provisioning
```
lzap, line 705: This all really looks like just a different ordering of keys.

cmd/spec/main.go
```diff
@@ -1,23 +1,51 @@
 package main
 
 import (
-	"bytes"
 	"encoding/json"
-	"io/ioutil"
+	"os"
```
lzap, line 5: We have few more in the codebase I think but this is a great start :-)

PR #475, opened 2023-03-22
fix(HMS-1105): add reservation counters

lzap:
I noticed that the existing reservation job metrics are missing the prefix.

Also we need to track success and failure rate of jobs/reservations. This cannot be done on the job level as background jobs do not have any sort of error handling, this must be all done on the reservation level when we finish reservation either with error or success flag.

This PR adds a new metric which I will use on the dashboard for the main SLO.

### DIFF
--- a/internal/metrics/registration.go
+++ b/internal/metrics/registration.go
@@ -1,0 +1,15 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+func RegisterStatuserMetrics() {
+	prometheus.MustRegister(TotalSentAvailabilityCheckReqs, AvailabilityCheckReqsDuration, TotalInvalidAvailabilityCheckReqs)
+}
+
+func RegisterApiMetrics() {
+	// no metrics
+}
+
+func RegisterWorkerMetrics() {
+	prometheus.MustRegister(JobQueueSize, JobQueueInFlight, BackgroundJobDuration, ReservationCount)
+}
--- a/internal/ctxval/copy.go
+++ b/internal/ctxval/copy.go
@@ -1,0 +1,14 @@
+package ctxval
+
+import "context"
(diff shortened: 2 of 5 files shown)


internal/metrics/registration.go
```diff
@@ -0,0 +1,15 @@
...
+	prometheus.MustRegister(TotalSentAvailabilityCheckReqs, AvailabilityCheckReqsDuration, TotalInvalidAvailabilityCheckReqs)
+}
+
+func RegisterApiMetrics() {
+	// no metrics
+}
+
+func RegisterWorkerMetrics() {
+	prometheus.MustRegister(JobQueueSize, JobQueueInFlight, BackgroundJobDuration, ReservationCount)
+}
```
lzap, line 15: Moving registration functions outside of the main file, it is starting to be a bit messy.

adiabramovitch:
@lzap works as expected, thank you! 
Can you check out why the tests are failing? 
image

lzap:
Yeah I knew but had to chat with Go nerds on Gophers Slack about it. The problem is: when timeout occurs, context is cancelled, meaning if we pass it anywhere context-aware it will error out as it should. Problem is, when timeout occurs, we want to "finish the reservation" and do one last final SQL UPDATE to store the end result (which was error because timeout). And pgx was erroring out with "context cancelled" error.

I thought for a moment I could create a new context with a new deadline, but that does not work - once context is cancelled there is no way of reversing that. So I created `ctxval.Copy` function which copies context values to a new context, that is passed into DAO to perform the reservation table update (which requires account_id to be present in the context).

This should solve the tests.

adiabramovitch (approved):
Thanks!

ezr-ondrej:
Nice ! :orange_heart:

PR #476, opened 2023-03-22
fix(HMS-1105): add reservations and jobs to dash

lzap:
Adds reservation and job metrics.

image

Description:

* AWS / Azure reservation success (28d) - this is our main SLO, that is `sum(rate(provisioning_reservation_count{result="success",type="aws"}[28d])) / sum(rate(provisioning_reservation_count{type="aws"}[28d]))` I hardcopied 28 days because this is our SLO. Currently reservations only occurs few times a day and these metrics will show "NaN" all the time.
* Reservation success rate - as a graph per reservation type. This graph will show us rate of successful or error reservation in time, it is continious so it works even for sparse reservations on stage.
* Job latency 95th percentile - pretty clear what it does.
* Job queue size - amount of jobs stuck in queue.
* Job inflight - amount of processed jobs at that time.


adiabramovitch:
I don't think you should hard code 28 days' value. 
https://service.pages.redhat.com/dev-guidelines/docs/appsre/onboarding/creating-slos/#defining-slos

lzap:
The thing is, we have so small amount of reservation per day that this will almost always get you NaN result with typical settings we use (1h, 6h etc). It is practically useless at this point, as I explained in the description.

I was hoping to get this merged with a month window fixed and as our app gets more use we can change it back to `$__interfal`.

Let me play around with Grafana more, there must be something.

lzap:
I have removed the fixed 28d for `$__interval`. Just make sure to change the interval and window to see the stat.

In the second commit, I am adding a makefile target `generate-dashboard` which allows us to easily work with the JSON file. It appends a template and JSON, additionally it reformats it with `json_pp` (an old Perl command) and indents it with four spaces. The reason is now you can simply open up the JSON file and copy/paste it all easily.

ezr-ondrej (commented):
I'm fine to merge and try it out, but what makes me bit worried is we have the template in the `dashboards` folder and our configuration is set to deploy everything within that table, won't it try to deploy the template as well?

I've got no idea here, just asking if you think it's safe.

adiabramovitch (approved):

lzap:
Thanks the dashboard on stage looks okay.

PR #477, opened 2023-03-22
refactor: Create struct for InsertInstances request

adiabramovitch:
Just an idea to simplify and have a cleaner interface
Would like to hear your opinion
Of course that this should be done for all providers and maybe for the responses as well.

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -366,7 +366,7 @@ func (c *ec2Client) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchT
 	return res, nil
 }
 
-func (c *ec2Client) RunInstances(ctx context.Context, launchTemplateID string, name *string, amount int32, instanceType types.InstanceType, AMI string, keyName string, userData []byte) ([]*string, *string, error) {
+func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanceParams, amount int32, name *string) ([]*string, *string, error) {
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "RunInstances")
 	defer span.End()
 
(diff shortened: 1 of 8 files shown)


avitova:
I only have one comment. I think it might be worth considering removing the Amount parameter outside the struct. I am thinking about the possible reusability of the struct. If there are values that are common for some other GCP requests, it might be worth considering not incorporating all of the values there. Tbh, I do not know Azure very well, so this is just an observation.
Otherwise, I really like the idea. :cool: It definitely looks cleaner like this. I'd love to see this for other providers. :+1:

adiabramovitch:
> I only have one comment. I think it might be worth considering removing the Amount parameter outside the struct. I am thinking about the possible reusability of the struct. If there are values that are common for some other GCP requests, it might be worth considering not incorporating all of the values there. Tbh, I do not know Azure very well, so this is just an observation. Otherwise, I really like the idea. cool It definitely looks cleaner like this. I'd love to see this for other providers. +1

Thank you for your comment! 
I am not sure why the amount should be outside of the struct and not machine type for example. 
I agree that we would want to reuse the struct when possible but machine type, zone, and even name pattern can be different from request to request.

adiabramovitch:
> I'd agree with Anna here, that the amount could be split up, as for azure for instance, we have call per instance, so it would be consistent cross providers.
> 
> The method would read "create instace with these parameters `Amount` times".
> 
> It sounds like a bit cleaner method API to me as well, but no strong preference.

Ok it makes sense because of Azure use case :) 
Will move amount outside of the struct

ezr-ondrej:
I've stolen the idea for Azure, so you can check out why it might be a good idea to split the amount for Azure (I know it's not the same case here, but it would keep the consistency): https://github.com/RHEnVision/provisioning-backend/pull/460

avitova:
I do not have a strong opinion on the response struct, to be honest. I would say that both current and your approach is okay. If that makes sense for you and you feel like we would need this in the future, I'd be happy if you did it. :)

ezr-ondrej (approved):
Great simplification! :orange_heart: LGTM :+1:

PR #478, opened 2023-03-24
feat: details of Azure source

ezr-ondrej:
Add enpoint to list details of Azure source.

Returns subscription ID, tenant ID and list of resource groups.

Fixes HMS-1509

This is needed for first version of IB sources integration, where IB will just list azure sources and prefil the values from source, and give user selection of their user groups.

FYI @croissanne


ezr-ondrej:
@akhil-jha are you ok to merge this without test coverage (these endpoinds shall not be used by end users - are not meant to). IMHO the unit tests that I'm adding for it would suffice, but ofc up to you :)

akhil-jha (approved):
I have added HMS-1514 in backlog for later.

lzap (changes requested):
Seed of an ugly API, we need to act now! :-)

internal/routes/all_routes.go
```diff
@@ -63,7 +63,8 @@ func MountAPI(r *chi.Mux) {
 				r.Get("/instance_types", s.ListInstanceTypes)
 
 				r.Get("/launch_templates", s.ListLaunchTemplates)
-				r.Get("/account_identity", s.GetAccountIdentity)
+				r.Get("/account_identity", s.GetAWSAccountIdentity)
+				r.Get("/azure_detail", s.GetAzureSourceDetails)
```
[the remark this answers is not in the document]
ezr-ondrej: I'm happy to do whatever we decide in this PR.

I don't like having it under single endpoint much tho. What about having `sources/<id>/details/<provider>`?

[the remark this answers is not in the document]
croissanne: Can a single source have multiple provider types in them? Didn't realize that when I wrote the account_identity endpoint :(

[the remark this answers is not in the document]
ezr-ondrej: no, single source = single provider.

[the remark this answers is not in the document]
ezr-ondrej: No worries, we are in luck as we did not stabilize the Spec yet (didn't GA) so we still have time to change this ;)

ezr-ondrej:
This is green and has two acks, merging.

internal/services/sources_service.go
```diff
@@ -42,3 +44,104 @@ func ListSources(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	authentication, err := sourcesClient.GetAuthentication(r.Context(), sourceId)
+	if err != nil {
+		renderError(w, r, payloads.NewClientError(r.Context(), err))
+		return
+	}
+
+	payload := payloads.SourceUploadInfoResponse{Provider: models.ProviderTypeAzure.String()}
```
adiabramovitch, line 95: Why is that? I just notice it in our code base.

internal/services/sources_service.go
```diff
@@ -42,3 +44,104 @@ func ListSources(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	if err := render.Render(w, r, payloads.NewAccountIdentityResponse(uploadInfo)); err != nil {
+		renderError(w, r, payloads.NewRenderError(r.Context(), "unable to render account id", err))
+		return
+	}
+}
+
+func GetSourceUploadInfo(w http.ResponseWriter, r *http.Request) {
+	sourceId := chi.URLParam(r, "ID")
```
adiabramovitch, line 81: Why are we getting a source id as a parameter? 
I suggest we fetch all the sources from sources, loop over them, get the information from each source and send all this information once to the frontend. 
That way, we could solve the bug we encounter, avoid back and forth between the browser and our backend, and also we can filter out empty sources (which are valid according to sources)

lzap: I think that the endpoint is used by image builder and this PR was refactoring of existing one in order to bring Azure. So that is why it exists.

Now, the new UI source step check using it is another story and we should probably introduce a new one that can be more resilient. But I would love to hear from @ezr-ondrej when he gets to this.

ezr-ondrej: yeah, I have kept it per one source, because it makes more sense in the API contract, but to fit our needs it would be better to have that for all sources, yes :)

lzap: Just for the context: we already talked about this and agreed that we might need another endpoint. That was the last week when you were away. Now that I have little bit more info about how it works, I think we should consider maybe only frontend fix, in case this page turns out to be temporary. Let’s be just careful with that because history shows temporary solutions tend to stick :-)

PR #479, opened 2023-03-27
Update clients and deps

lzap:
Two commits.

First, I bump dependencies.

Second, I regenerate both clients. There are issues for both.

Sources, the constants strings are no longer in the spec and we need to hardcode strings in our app, the documentation comment in Sources OpenAPI says we should use listing endpoint to get those. Seems okay and compatible change.

Image builder, there is now a type change which is a breaking change, it was previously string (UUID) now it is a native OpenAPI UUID type. This looks scary at first sight, but it appears to be just a type format hint, not a full type change. It is still a string and this is how the binding sends it.

So overall, I think this is a safe to merge.

### DIFF
--- a/config/ib_api.yaml
+++ b/config/ib_api.yaml
@@ -45,6 +45,10 @@ paths:
       responses:
         '200':
           description: returns this document
+          content:
+            application/json:
+              schema:
+                type: object
   /distributions:
     get:
       summary: get the available distributions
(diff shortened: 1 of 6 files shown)


config/ib_api.yaml
```diff
@@ -174,6 +184,7 @@ paths:
           name: composeId
           schema:
             type: string
+            format: uuid
             example: '123e4567-e89b-12d3-a456-426655440000'
```
lzap, line 188: These all changes are just format hint, in Go this maps to Google's UUID so I think this should be safe to merge and it appears to be a compatible change.

PR #481, opened 2023-03-29
fix(HMS-1105): improve reservations dashboard

lzap:
This is a big update of the jobs/reservations panes on our dashboard. Now it shows proper numbers even when there is no data, I learned this from IB team and the trick is to use `on() vector(0)`.

image

But the big new feature is (also learned from IB dashboard) a new variable called Hyperscaler where you can switch between azure, aws and gcp to drill down the data. Hard to show on an image so here is a short demo:

https://www.youtube.com/watch?v=X7M7WxkBLHc

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -48,10 +48,10 @@ var AvailabilityCheckReqsDuration = prometheus.NewHistogramVec(
 
 var BackgroundJobDuration = prometheus.NewHistogramVec(
 	prometheus.HistogramOpts{
-		Name:        "provisioning_background_job_duration_ms",
-		Help:        "task queue job duration (ms) by type",
+		Name:        "provisioning_background_job_duration",
+		Help:        "task queue job duration (in seconds) by type",
 		ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName, "component": "worker"},
-		Buckets:     []float64{50, 100, 250, 500, 1000, 2500, 6000},
+		Buckets:     []float64{1, 4, 5, 10, 30, 60 * 2, 60 * 5, 60 * 10, 60 * 20, 60 * 30, 60 * 60},
 	},
 	[]string{"type"},
 )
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
I love it! :orange_heart:

internal/metrics/metrics.go
```diff
@@ -51,7 +51,7 @@ var BackgroundJobDuration = prometheus.NewHistogramVec(
 		Name:        "provisioning_background_job_duration_ms",
 		Help:        "task queue job duration (ms) by type",
 		ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName, "component": "worker"},
-		Buckets:     []float64{50, 100, 250, 500, 1000, 2500, 6000},
+		Buckets:     []float64{1000, 3000, 7000, 20000, 60000, 60000 * 5, 60000 * 10, 60000 * 20, 60000 * 40},
```
ezr-ondrej, line 54: Sounds reasonable, will merge, but FYI @adiabramovitch in case you'd not agree lets discuss post fact :)

lzap: Maybe this is a chance to change this to seconds before we fully commit to SLI/SLOs? Easy change I could amend.

lzap:
I took the opportunity to change job duration (provisioning_background_job_duration) to second. I renamed it so there is no _ms in the name because description nicely shows it. And I also found why the graph in grafana was not working - it was using the old name without provisioning_ prefix :-) Now it is perfect I hope.

lzap:
Fixed also coloring scheme on the job duration graph. It was too wild.

Note: when testing this, rename the job duration metric to the old name (ms) and update the graph unit to milliseconds in order to see the correct numbers.

lzap:
That is still the old page, I also see the old one.

lzap:
Found the reason: https://github.com/RHEnVision/provisioning-backend/pull/482

PR #482, opened 2023-03-29
fix(HMS-1105): check for make dashboard target

lzap:
Forgot to call `make` this will not happen again since this PR is also adding a make check that will be executed by our CI.


ezr-ondrej:
Nice, the board is up and working! :tada:

PR #484, opened 2023-03-29
feat: allow Azure group principal in Lighthouse

ezr-ondrej:
Azure supports permissions on Groups and it is a safer bet. We should support this in offering template.
It also allows for Enterprise Apps to be used properly as Principal ID is different to Client ID.

Refs HMS-1148

We finally have Azure stage and prod secrets, this gets ready to consume them.
Principal is only for the Lighthouse offering template, so now its values are set only on the API through clowdapp.


lzap (approved):
Sounds right, I haven’t checked this in Azure docs so I believe it is the right call.

PR #485, opened 2023-03-30
fix(HMS-1453): platform RBAC support

lzap:
This implements RBAC as a 3rd service alongside sources and image builder. It uses the same techniques and packages, so OpenAPI spec is downloaded and client generated, then an interface is available and there are three implementations. The interface is very simple:

```go
// Rbac interface provides access to the RBAC backend service API. Each action that needs to
// check must provide resource (e.g. pubkey) and action (e.g. write) in order to check permission
// presence for principal that is in the identity headers. Definition of permissions and default
// roles are at https://github.com/RedHatInsights/rbac-config (app named "provisioning").
type Rbac interface {
	// GetPrincipalAccess return an ACL object that can be used to check permissions
	GetPrincipalAccess(ctx context.Context) (RbacAcl, error)

	// Ready returns readiness information
	Ready(ctx context.Context) error
}

type RbacAcl interface {
	IsAllowed(res, verb string) bool
}
```

The main function is named `IsAllowed` and depending on permission and verb it returns a boolean. The package where all permission checks are done is `service` where a helper function is available, so checks are pretty easy:

```go
func GetPubkey(w http.ResponseWriter, r *http.Request) {
	if CheckPermissionAndRender(w, r, "pubkey", "read") != nil {
		return
	}
	// ...
}
```

There are few new settings namely:

```
REST_ENDPOINTS_RBAC_URL=https://xxxx.redhat.com/api/rbac/v1
REST_ENDPOINTS_RBAC_USERNAME=xxxx
REST_ENDPOINTS_RBAC_PASSWORD=xxxx
REST_ENDPOINTS_RBAC_PROXY_URL=http://squid.xxxxx.redhat.com:3128

APP_RBAC_ENABLED=true
```

By default, RBAC is disabled meaning the implementation which always return `true` is used therefore all checks succeeds immediately. It is possible to configure the application against stage instance, proxy must be used similarly as with existing endpoints. Currently permissions are not yet deployed on stage, the PR is pending review: https://github.com/RedHatInsights/rbac-config/pull/359/files

The patch also adds a new metric that will show duration of RBAC calls, this service is absolutely crucial for our app to operate so I want to show this on our dashboard. Also small update in makefile, changed `wget` to more widely available `curl`.

Note this patch does not make any use of the already existing Go RBAC client because I simply don’t like it and it seems to be a WIP. We already have a better way of dealing with OpenAPI clients, so my approach is better fit for our app. https://github.com/RedHatInsights/rbac-client-go

I want to do a followup PR and implement caching as well, but that is not necessary in the initial phase. Also, there are no integration tests and I wonder if I should be writing ones because for development and unit teststing, ACLs are turned off and IQE will cover positive and negative tests in ephemeral.

@akhil-jha this PR also enables RBAC service on ephemeral and enables RBAC support in the provisioning app, therefore, tests will now all fail. We need to figure way how to create user and assign them roles in ephemeral.


docs/make.md
```diff
@@ -30,6 +30,10 @@ Building
 Image building
   build-podman          Build container image using Podman
 
+Dashboard
+  generate-dashboard    Generate dashboard
+  validate-dashboard    Compare dashboard configmaps with git
+
```
lzap, line 36: Interesting, this passed CI/CD but now it bugs me. Probably there is a bug in the checker.

ezr-ondrej (commented):
Very nice and clean! :orange_heart:
I was surprised by us fetching all the user perms instead of just asking for a direct check for given resource & action directly, but if that's what RBAC wants us to do I'm all for it :+1:

internal/config/config.go
ezr-ondrej, line 1: We should consume the URL from clowder config: https://github.com/RHEnVision/provisioning-backend/blob/18db8e87c7ea137decf1f121574815d006016dff/internal/config/config.go#L292

lzap:
Added clowder config dependency URL, nice catch. I also added some logging in case I will not be able to guess the "service" hash key. It appears to be a bit random.

lzap:
So it works, all requests are failing with 403 and RBAC service is being called, it uses "jdoe" user account which apparently don’t have the required roles. I am unsure what to do at this point, RBAC documentation is sparse on this topic.

internal/metrics/metrics.go
```diff
@@ -27,15 +27,24 @@ var TotalInvalidAvailabilityCheckReqs = prometheus.NewCounter(
...
 
+var RbacAclFetchDuration = prometheus.NewHistogram(
+	prometheus.HistogramOpts{
+		Name:        "provisioning_rbac_acl_request_duration",
+		Help:        "duration of ACL request to RBAC service (ms)",
+		ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName, "component": "api"},
+		Buckets:     []float64{50, 100, 250, 500, 1000, 2500, 6000},
+	},
+)
+
```
lzap, line 47: TODO: I forgot to register the metric. It is not recognized by Prometheus.

lzap:
> Have you thought of using middleware

Ha, I just came here to write a TODO item, yeah, makes a lot of sense.

internal/services/permissions.go
adiabramovitch, line 1: Can we change the file name to `permissions_service.go` just for consistency?

adiabramovitch (commented):
Super clean and readable! 
Just a small nit and your comments still stand (regarding the metrics)

internal/services/permissions.go
lzap: We actually have services suffixes only for service functions, there is a parameters.go file that does not have this suffix:

image

I can rename if you wish, no biggie.

lzap:
Ok middleware was actually huge refactoring which took me almost whole day of work. I got bunch of depcycle problems. Please re-review carefully.

The way this is implemented now is that we have a middleware and that works great. However, for reservation we have TYPE URL parameter which cannot be checked on the middleware level. Therefore, the middleware stores the RBAC ACL object in the context so it can be later checked in the service. I added a new permission "reservation" which acts as a "base" permission just in case we later on forget do perform a check in a newly created endpoint. I encourage this in the documentation of the function.

Tests now actually fail because in the previous PR I actually forgot to check reservations POST request! And everyone loved the initial version, whoops :-) Will fix tests tomorrow, but please re-review as the changes are pretty large.

internal/ctxval/acl.go
```diff
@@ -0,0 +1,31 @@
...
+	if value == nil {
+		return noPermAcl{}
+	}
+	return value.(RbacAcl)
+}
+
+// WithAcl returns context copy with ACL interface.
+func WithAcl(ctx context.Context, id RbacAcl) context.Context {
+	return context.WithValue(ctx, aclCtxKey, id)
+}
```
lzap, line 31: This ctxval package is used literally EVERYWHERE so basically it cannot import pretty much anything because of import cycles.

This is the reason why I make copies of both the ACL interface and NoPermissionsAcl variable here. This is fine to do in Go, it is better to copy sometimes than to do hacks.

internal/metrics/registration.go
```diff
@@ -3,13 +3,26 @@ package metrics
...
 func RegisterWorkerMetrics() {
-	prometheus.MustRegister(JobQueueSize, JobQueueInFlight, BackgroundJobDuration, ReservationCount)
+	prometheus.MustRegister(
+		RbacAclFetchDuration,
+		JobQueueSize,
+		JobQueueInFlight,
+		BackgroundJobDuration,
+		ReservationCount,
+	)
 }
```
lzap, line 28: Reformatted and added the new metric.

internal/middleware/enforce_identity.go
```diff
@@ -65,19 +66,25 @@ func EnforceIdentity(next http.Handler) http.Handler {
...
-		doError(ctx, w, 400, "X-Rh-Identity header is missing type")
-		return MissingTypeErr
+		errRender := render.Render(w, r, payloads.NewMissingIdentityError(r.Context(), "X-Rh-Identity header is missing type", ErrIdentity))
+		if errRender != nil {
+			ctxval.Logger(ctx).Warn().Err(errRender).Msg("Cannot render permission middleware error")
+		}
+		return ErrIdentity
 	}
 
 	return nil
```
lzap, line 90: I am dropping the `doError` function, initially I thought I would reuse it in the other enforce middleware I am introducing, that did not work out, then I tried to use service function but import cycles. Then I decided to simplify it further - both middleware functions now use render package directly. It is slightly more verbose but I am not introducing middleware-service package dependency.

internal/services/gcp_reservation_service.go
```diff
@@ -20,7 +20,6 @@ import (
 
 func CreateGCPReservation(w http.ResponseWriter, r *http.Request) {
 	logger := *ctxval.Logger(r.Context())
-
 	var accountId int64 = ctxval.AccountId(r.Context())
 	var id identity.XRHID = ctxval.Identity(r.Context())
```
lzap, line 25: These are just leftovers when I added the explicit check everywhere... :-(

internal/services/permissions.go
```diff
@@ -0,0 +1,25 @@
...
+// be enforced via middleware as a bare minimum.
+func CheckPermissionAndRender(w http.ResponseWriter, r *http.Request, resource, permission string) error {
+	if !ctxval.Acl(r.Context()).IsAllowed(resource, permission) {
+		permErr := fmt.Errorf("%w: %s on %s", ErrMissingExtraPermission, permission, resource)
+		renderError(w, r, payloads.NewMissingPermissionError(r.Context(), resource, permission, permErr))
+		return permErr
+	}
+
+	return nil
+}
```
lzap, line 25: This function takes ACL from the context and performs an extra check. This is called during request so request parameters and headers are all present. This way service functions can perform permission checks like `reservation.aws` where `aws` comes from an URL param.

internal/ctxval/acl.go
ezr-ondrej: Makes me thinking: is this actually the best approach?
Shouldn't we instead make the `WithXXX` functions part of the package they store the value of?
It sounds like it would fix the issue and to me it sounds a bit cleaner.
Here we could define the context key type and so on, but storing to context would be done in each package.
We need to do it for stubs already anyway.

ezr-ondrej: I mean I'm fine here, we can decide on and do this later :+1:

lzap: Yeah that is a valid option I considered, main reason why I like centralized approach is that you don’t need to think about the package name, you just type `ctxval` and the required function will pop up. We use it everywhere in the code.

The split will be **really painful tho** :-) Not here, not now.

lzap:
The ephemeral setup was merged: https://github.com/RedHatInsights/insights-rbac/pull/848

I will rebase now to see if this fixes the tests.

lzap:
Hmmm: `make: oapi-codegen: Command not found`

This is caused by sources API client files being older than generated Go code. Weird, somebody had to touch it and forgot to generate the client.

lzap:
Found it, GoLand stripped an empty line from the client YAML configuration file which caused make to regenerate clients.

lzap:
Hmm does not help, I think we need to assign roles/permissions to jdoe or something. Not sure.

lzap:
Rebased, all is failing with access denied @akhil-jha 

From the logs it seems that the RBAC API is working, at least it responds correctly:

```
7:23AM TRC Account cache hit account=1 account_number=0369233 bytes_in=414 hostname=provisioning-backend-api-b8d676854-dhkgv method=POST org_id=3340851 remote_ip=10.131.20.163:34474 trace_id=bf57af787f593d3920863c195f102d6d url=/api/provisioning/v1/pubkeys version=34e6

7:23AM TRC otel: HTTP GET duration=31.180352 hostname=provisioning-backend-api-b8d676854-dhkgv http.flavor=1.1 http.method=GET http.url=http://rbac-service.ephemeral-2kqkgg.svc:8000/api/rbac/v1/access/?application=provisioning&limit=999 net.peer.name=rbac-service.ephemeral-2kqkgg.svc otel=true span_id=77f72c785e095818 span_id_parent=a5edb8418ecdea97 trace_id=bf57af787f593d3920863c195f102d6d version=34e6

7:23AM TRC {"meta":{"count":0,"limit":999,"offset":0},"links":{"first":"/api/rbac/v1/access/?application=provisioning&limit=999&offset=0","next":null,"previous":null,"last":"/api/rbac/v1/access/?application=provisioning&limit=999&offset=0"},"data":[]} account_id=1 account_number=0369233 bytes_in=414 content_length=240 hostname=provisioning-backend-api-b8d676854-dhkgv method=POST org_id=3340851 remote_ip=10.131.20.163:34474 response_trace=true status="200 OK" status_code=200 trace_id=bf57af787f593d3920863c195f102d6d url=/api/provisioning/v1/pubkeys version=34e6

7:23AM TRC otel: GetPrincipalAccess duration=31.456958 hostname=provisioning-backend-api-b8d676854-dhkgv otel=true span_id=a5edb8418ecdea97 span_id_parent=f106ad806e8498ef trace_id=bf57af787f593d3920863c195f102d6d version=34e6

7:23AM TRC Checking permission 'write' on 'pubkey' account_id=1 account_number=0369233 acl_permission=write acl_resource=pubkey bytes_in=414 hostname=provisioning-backend-api-b8d676854-dhkgv method=POST org_id=3340851 remote_ip=10.131.20.163:34474 trace_id=bf57af787f593d3920863c195f102d6d url=/api/provisioning/v1/pubkeys version=34e6

7:23AM WRN Access denied, missing permission write on pubkey error="missing permission: write on pubkey" account_id=1 account_number=0369233 bytes_in=414 hostname=provisioning-backend-api-b8d676854-dhkgv method=POST org_id=3340851 remote_ip=10.131.20.163:34474 trace_id=bf57af787f593d3920863c195f102d6d url=/api/provisioning/v1/pubkeys version=34e6

7:23AM INF Completed POST request /api/provisioning/v1/pubkeys in 41ms with 403 bytes_in=414 bytes_out=205 hostname=provisioning-backend-api-b8d676854-dhkgv latency_ms=41.298117 method=POST remote_ip=10.131.20.163:34474 status=403 trace_id=bf57af787f593d3920863c195f102d6d url=/api/provisioning/v1/pubkeys version=34e6
```

Here the code I wrote calls out to RBAC and then checks against the list of permissions and fails to find "write" on "pubkey". Looks like the permission is not there, I can perhaps add more debug log statements to actually print the payload that is coming from via REST_ENDPOINTS_TRACE_DATA setting to confirm that.

lzap:
Rebased with increased logging.

lzap:
So with body tracing enabled:

```
1:05PM TRC {"meta":{"count":0,"limit":999,"offset":0},"links":{"first":"/api/rbac/v1/access/?application=provisioning&limit=999&offset=0","next":null,"previous":null,"last":"/api/rbac/v1/access/?application=provisioning&limit=999&offset=0"},"data":[]} account_id=1 account_number=0369233 bytes_in=414 content_length=240 hostname=provisioning-backend-api-7987f57cd7-zw8c2 method=POST org_id=3340851 remote_ip=10.131.21.126:35466 response_trace=true status="200 OK" status_code=200 trace_id=207b07e3668b71e8452e1f8e26da8a6b url=/api/provisioning/v1/pubkeys version=9676
```

The RBAC endpoint returns empty result for "provisioning" application. Apparently, there is no data there, we need to seed it somehow. @akhil-jha @mshriver you guys know how?

lzap:
@akhil-jha what is the status? Do you need anything else from me? I am going to rebase once again.

lzap:
This needs full re-review as I had massive conflicts all around.

@akhil-jha I would appreciate if this could go through if the ephemeral setup turns out to be problem we can commit this feature but disable it via flag on all environments until IQE is ready.

deploy/clowdapp.yaml
```diff
@@ -414,6 +419,12 @@ parameters:
   - description: Postgres driver logging level (trace, debug, info, warn, error, fatal, panic)
     name: DATABASE_LOGGING_LEVEL
     value: "debug"
+  - description: RBAC checking enabled
+    name: APP_RBAC_ENABLED
+    value: "false"
```
lzap, line 424: I am proposing to re-review this patch and merge it as-is with RBAC currently disabled so tests will pass and everything will operate normally.

I will file a new one-liner PR that enables RBAC which will fail tests so @akhil-jha can work on tests.

internal/clients/http/rbac/ctx.go
```diff
@@ -0,0 +1,27 @@
...
+	if value == nil {
+		return clients.NoPermissionsRbacAcl
+	}
+	return value.(clients.RbacAcl)
+}
+
+// WithAcl returns context copy with ACL interface.
+func WithAcl(ctx context.Context, id clients.RbacAcl) context.Context {
+	return context.WithValue(ctx, aclCtxKey, id)
+}
```
lzap, line 27: This file previously had import cycles, not anymore without `ctxval`! :-)

deploy/clowdapp.yaml
ezr-ondrej: I love the approach, hope @akhil-jha is ok with that :)

ezr-ondrej (approved):
Makes sense to me :+1:
I like the super simple interface :orange_heart: 

A thought: It would be a great candidate for a library implementation in go-middlewares. :)

internal/config/config.go
```diff
@@ -99,6 +100,12 @@ var config struct {
 		Path string `env:"PATH" env-default:"/metrics" env-description:"prometheus metrics path"`
 	} `env-prefix:"PROMETHEUS_"`
 	RestEndpoints struct {
+		RBAC struct {
+			URL      string `env:"URL" env-default:"" env-description:"RBAC URL"`
+			Username string `env:"USERNAME" env-default:"" env-description:"RBAC credentials (dev only)"`
+			Password string `env:"PASSWORD" env-default:"" env-description:"RBAC credentials (dev only)"`
+			Proxy    proxy  `env-prefix:"PROXY_" env-description:"RBAC HTTP proxy (dev only)"`
```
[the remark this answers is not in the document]
lzap: Actually no, `env` is a variable name while `env-prefix` is a prefix name. This uses `proxy` type defined above.

ezr-ondrej:
I'd love re-review from @adiabramovitch and we should get ack from @akhil-jha he's good on the approach of flipping the flag in follow-up PR.

internal/clients/http/rbac/rbac_client.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
+	if err != nil {
+		return nil, fmt.Errorf("get principal: %w", err)
+	}
+
+	if resp.JSON200.Meta.Count != nil && *resp.JSON200.Meta.Count > 1 {
+		return nil, ErrPaginationUnsupported
+	}
+
```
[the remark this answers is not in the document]
lzap: This behavior is documented few lines above: `// ErrPaginationUnsupported is returned instead of paginated payload. This is unsupported in the official rbac-client-go.`

I was simply lazy, it is unlikely we will ever get more permissions than 999, the hardcoded limit. But let me do pagination it is simple. Rebased.

mayurilahane:
Hey @lzap @ezr-ondrej  is this PR still valid?

lzap:
Yes this was put on hold because it broke tests, then I added configurable option and it was never merged.

lzap:
Requesting full re-review, I got some bad conflicts so I may have messed up some hunks.

The RBAC is currently turned off by default so let’s merge this PR and I will enable it in another one (which will fail tests).

ezr-ondrej (approved):
My ack still stands :+1:

internal/clients/http/rbac/rbac_client.go
```diff
@@ -0,0 +1,119 @@
...
+	start := time.Now()
+	defer func() {
+		metrics.RbacAclFetchDuration.Observe(float64(time.Since(start).Nanoseconds()) / 1000000)
+	}()
+
+	var result clients.AccessList
+	var records int64 = math.MaxInt64
+
+	// keep fetching until we have all the records
+	for int64(len(result)) < records {
```
adiabramovitch, line 88: This condition is kinda weird, can you explain? I think this will loop forever

lzap: You are right there was a bug, I forgot to add offset so it actually stacked the first N elements to the slice over and over again. It did not loop tho, fixed. I tested it by adding some roles to my account so it returned a longer list and then I set the FetchLimit to a small number.

lzap:
Rebased: fixed a bug in pagination.

ezr-ondrej:
Linter is not happy with the `var` with initialization usage :face_exhaling:

lzap:
OMG rebased. :-D

internal/clients/http/rbac/rbac_client.go
lzap: You know what, there is a possibility for an endless loop which is really bad - if there is a bug in RBAC service and it returns "total" number that is higher than actual records and in subsequent queries it will return empty array. No results will be ever appended and it will continue looping. And it is possible to get here in case someone deletes a permission just after you receive the first page.

So I added maximum limit of queries (10), if that ever happens a warning is logged and the result is returned.

Consequence: on an account with more than 5000 permissions, some might be ignored. Meaning access denied.

adiabramovitch:
Let's check it out. 
@lzap Thank you!

ezr-ondrej:
Agreed on stand up to get this in! Great effort, thanks @lzap to push this and thanks everyone for reviews! :tada:

PR #487, opened 2023-03-31
feat: integrate Sentry via OpenTelemetry

lzap:
This is an experiment, but it appears that Sentry can be configured via OpenTelemetry API/SDK so it delivers traces into Sentry (Glitchtip in our case). I slightly updated how telemetry and sentry is configured to streamline the configuration experience and enabled this for ephemeral. Assuming the Glitchtip deployed in ephemeral is accepting traces, this should work "out of box".

I haven’t tried this locally, would like to ask @ezr-ondrej to test this first before I fully commit to elaborating this. There is one limitation at the moment - the logging sentry library we use currently does not allow to initialize Sentry and pass it in, but this should be pretty small change to either contribute or we could make a copy of that library.

But if this works out, then use of OpenTelemetry could already pay off and that is quite earlier than I expected.

### DIFF
--- a/internal/clients/http/azure/create_vm.go
+++ b/internal/clients/http/azure/create_vm.go
@@ -19,11 +19,12 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/RHEnVision/provisioning-backend/internal/ptr"
+	"github.com/RHEnVision/provisioning-backend/internal/telemetry"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/codes"
 )
 
-const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/http/azure"
+const TraceName = telemetry.TracePrefix + "internal/clients/http/azure"
 
 const (
 	vnetName              = "redhat-vnet"
--- a/internal/jobs/noop_job.go
+++ b/internal/jobs/noop_job.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
 	"github.com/RHEnVision/provisioning-backend/pkg/worker"
+	"go.opentelemetry.io/otel"
 )
 
 type NoopJobArgs struct {
@@ -20,6 +21,9 @@ var NoOperationFailure = errors.New("job failed on request")
 
 // Unmarshall arguments and handle error
 func HandleNoop(ctx context.Context, job *worker.Job) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "HandleNoop")
+	defer span.End()
+
 	args, ok := job.Args.(NoopJobArgs)
 	if !ok {
 		err := fmt.Errorf("%w: job %s, reservation: %#v", ErrTypeAssertion, job.ID, job.Args)
--- a/internal/logging/sentry.go
+++ b/internal/logging/sentry.go
@@ -12,7 +12,16 @@ import (
 // sentryWriter creates a zerolog writer for sentry.
 // Uses github.com/archdx/zerolog-sentry which is very simple wrapper.
 func sentryWriter(dsn string) (io.Writer, func(), error) {
-	wr, err := sentrywriter.New(dsn)
+	var opts []sentrywriter.WriterOption
+	if config.Sentry.Debug {
+		opts = append(opts, sentrywriter.WithDebug())
+	}
+	opts = append(opts, sentrywriter.WithSampleRate(config.Sentry.SampleRate))
+	// TODO this calls sentry.Init again (it is initialized in our telemetry package)
+	// We must be able to initialize it only once and pass it into here, or, create it
+	// here with tracing and sample rate enabled and use it also for otel.
+	// Currently it is not possible to have sentry for both errors and tracing.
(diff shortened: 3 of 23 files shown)


internal/clients/http/azure/create_vm.go
```diff
@@ -20,13 +20,14 @@ import (
...
 	"github.com/RHEnVision/provisioning-backend/internal/ptr"
+	"github.com/RHEnVision/provisioning-backend/internal/telemetry"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/codes"
 )
 
 var subscriptionId string
 
-const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/http/azure"
+const TraceName = telemetry.TracePrefix + "internal/clients/http/azure"
```
lzap, line 30: This looks better, also we already had typos in there "EnVision" vs "RhEnVision".

internal/logging/sentry.go
```diff
@@ -12,7 +12,16 @@ import (
...
+	var opts []sentrywriter.WriterOption
+	if config.Sentry.Debug {
+		opts = append(opts, sentrywriter.WithDebug())
+	}
+	opts = append(opts, sentrywriter.WithSampleRate(config.Sentry.SampleRate))
+	// TODO this calls sentry.Init again (it is initialized in our telemetry package)
+	// We must be able to initialize it only once and pass it into here, or, create it
+	// here with tracing and sample rate enabled and use it also for otel.
+	// Currently it is not possible to have sentry for both errors and tracing.
+	wr, err := sentrywriter.New(dsn, opts...)
```
lzap, line 24: Here we need to contribute new opt functions WithTracingEnabled and I think this is all. Then we would enforce that sentry logging is required if you want also to do tracing to make sure it is initialized before we configure otel.

ezr-ondrej:
Can't get that working with Glitchtip :/
Tho we can use the hub, but unfortunatelly the writer doesn't support tracing: true, so we need to initialize new client anyway:
```diff
diff --git a/internal/telemetry/otel.go b/internal/telemetry/otel.go
index 5cfc653..9fabb37 100644
--- a/internal/telemetry/otel.go
+++ b/internal/telemetry/otel.go
@@ -48,7 +48,7 @@ func Initialize(rootLogger *zerolog.Logger) *Telemetry {
 
        var exporterOption trace.TracerProviderOption
        if config.Telemetry.Type == "sentry" {
-               err := sentry.Init(sentry.ClientOptions{
+               client, err := sentry.NewClient(sentry.ClientOptions{
                        Dsn:              config.Sentry.Dsn,
                        EnableTracing:    true,
                        TracesSampleRate: config.Sentry.SampleRate,
@@ -57,6 +57,7 @@ func Initialize(rootLogger *zerolog.Logger) *Telemetry {
                if err != nil {
                        panic(err)
                }
+               sentry.CurrentHub().BindClient(client)
 
                exporterOption = trace.WithSpanProcessor(sentryotel.NewSentrySpanProcessor())
        } else if config.Telemetry.Type == "jaeger" {
@@ -81,10 +82,10 @@ func Initialize(rootLogger *zerolog.Logger) *Telemetry {
                trace.WithResource(res),
        )
        )
```

lzap:
Wait I just filed a new PR: https://github.com/archdx/zerolog-sentry/pull/7

Are you saying that we can bind multiple clients? That could solve the problem.

lzap:
Oh this does not solve it, you cannot pass an existing client into the zerolog-sentry library, that is the issue.

lzap:
So I think I have everything correctly set, yet, our GlitchTip does not take any performance data. I think it must be turned off or something, I asked on the platform channel.

So I disabled telemetry by default but I would still like to ask for a careful review and merge. I think we can revisit this later.

docs/make.md
```diff
@@ -30,6 +30,10 @@ Building
 Image building
   build-podman          Build container image using Podman
 
+Dashboard
+  generate-dashboard    Generate dashboard
+  validate-dashboard    Compare dashboard configmaps with git
+
```
adiabramovitch, line 36: Is that supposed to be in this PR?

internal/jobs/noop_job.go
```diff
@@ -19,6 +20,9 @@ var NoOperationFailure = errors.New("job failed on request")
 
 // Unmarshall arguments and handle error
 func HandleNoop(ctx context.Context, job *worker.Job) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "HandleNoop")
```
adiabramovitch, line 23: nit
Why are you using a global trace name and not s specific one regarding the noop job?

docs/make.md
lzap: No: https://github.com/RHEnVision/provisioning-backend/pull/490

lzap:
Christian Assing was able to help me to track down the 500 root cause, looks like a bug in GlitchTip. Filed an MR: https://gitlab.com/glitchtip/glitchtip-backend/-/merge_requests/338/diffs

ezr-ondrej:
so this one is ready as well? :) I've seen some traces in glitchtip already? :)

lzap:
It is not ready, I am currently doing small changes in the upstream backend project. It was brought to my attention that GlitchTip tracing capabilities are very limited and it does not scale: https://gitlab.com/glitchtip/glitchtip-backend/-/merge_requests/354#note_1379907344

So we will not be likely able to use it for tracing, but before I close this PR I want to experiment a bit maybe with sampling intervals or at least to see how it looks like in the UI.

PR #488, opened 2023-03-31
chore: avoid logging Err().Msgf("msg with error")

lzap:
Heads up: Please do not include error message in log message body, because it leads to unique messages like _cannot assume role operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 47e5734c-a37e-4be3-aba2-77f4e306a055_ which are incorrectly identified as unique errors by Sentry and creates too many error reports with the same cause.

The code which is causing this is:

    logger.Error().Err(err).Msgf("cannot assume role %s", err)

The correct way is:

    logger.Error().Err(err).Msg("Cannot assume role")

Note the error itself actually goes into logger through Err function, there is no need to copy it again. This is counter intuitive because when wrapping errors, you do exactly the opposite: fmt.Errorf("cannot assume role %w", err)

### DIFF
--- a/internal/payloads/error_payload.go
+++ b/internal/payloads/error_payload.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	httpClients "github.com/RHEnVision/provisioning-backend/internal/clients/http"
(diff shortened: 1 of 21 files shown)


internal/payloads/error_payload.go
```diff
@@ -54,7 +55,8 @@ func newResponse(ctx context.Context, status int, userMsg string, err error) *Re
 		strError = err.Error()
 	}
 	if userMsg == "" {
-		userMsg = err.Error()
+		// take only part up to the first colon to avoid unique ids (UUIDs, database IDs etc)
+		userMsg = strings.SplitN(err.Error(), ":", 2)[0]
```
lzap, line 59: There is this change.

When we get an error in service package and return it to the user as a HTTP payload, previously, we used the whole error message as a log message. This is an attempt to split these errors:

something happen: something: blah: aws error: request = UUID XXXXXXXX

to just

something happen

This will only works for errors with colons which are most of them in our app. Alternatively, we can give this a log line something like "Returned an error as payload" but that is quite lame.

PR #489, opened 2023-04-03
chore: introduce Sentry replacer

lzap:
It looks like AWS SDK does not makes use of error wrapping, we already workaround that in `isAWSOperationError` function. However, errors like `unable to initialize AWS client: cannot assume role operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 23c69a5e-69bf-439a-a1c5-c2161ce06a8e, api error AccessDenied: User: arn:aws:iam::12333232:user/crc-ephemeral is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::321321213:role/satellite-services-role` makes Sentry to identify these as unique errors.

This patch adds regular expression writer filter that filters out UUIDs and ARNs so errors are no unique by the error message. This will hopefully group up errors in Sentry.

### DIFF
--- a/internal/logging/sentry_replacer.go
+++ b/internal/logging/sentry_replacer.go
@@ -1,0 +1,85 @@
+package logging
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+	"sync"
+)
+
+var filters = []string{
+	// Example (AWS SDK): RequestID: ca767444-d1f9-11ed-afa1-0242ac120002
+	`[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}`,
+	// Example (AWS SDK): arn:aws:iam::4328974392798432:role/my-role-123
+	`arn:aws:[[:word:]]+::\d+:[[:word:]\*-]+/[[:word:]\*-]+`,
+}
+
+var replacement = []byte{'?'}
+
+type SentryReplacer struct {
+	buf    []byte
+	w      io.Writer
+	re     *regexp.Regexp
+	mu     sync.Mutex
+	closed bool
+}
+
+func NewSentryReplacer(w io.Writer) *SentryReplacer {
+	sr := SentryReplacer{
+		w:  w,
+		re: regexp.MustCompile("(" + strings.Join(filters, "|") + ")"),
+	}
+	return &sr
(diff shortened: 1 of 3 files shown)


ezr-ondrej:
I :orange_heart: you're resolving this and I guess this is the best we can do for now.
Long term, I'd love to  actually start mapping known AWS errors (the most common ones) to our own errors, but it will always be prone to errors and will never reach 100% coverage, so this will most likely be always needed :+1: 

I wonder tho, would that be possible to have the replaced tokens be `<filtered arn>` or something along those lines?
The `?` is fine, but if we could do better easily and improve readability of the errors, I'd prefer that.

lzap:
It is possible, will be slightly slower because it cannot be just a single regexp (single state machine). Do you think it is worth it? I mean, it feels okay to me:

_unable to initialize AWS client: cannot assume role operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: ?, api error AccessDenied: User: ? is not authorized to perform: sts:AssumeRole on resource: ?_

internal/logging/sentry_replacer.go
```diff
@@ -0,0 +1,85 @@
...
+	w      io.Writer
+	re     *regexp.Regexp
+	mu     sync.Mutex
+	closed bool
+}
+
+func NewSentryReplacer(w io.Writer) *SentryReplacer {
+	sr := SentryReplacer{
+		w:  w,
+		re: regexp.MustCompile("(" + strings.Join(filters, "|") + ")"),
```
lzap, line 32: This is compiled down to just a single state machine, which is much faster than going over a list of expressions. As we find more and more of these patterns it could get to the point this would be slow.

ezr-ondrej:
It's fine, I guess :+1: tho do we have these values anywhere?
It gives away some information - namely it gives away info about if it was manual ARN (mallformed ARN) or if that was automatically set up ARN. The remediation for both is different

lzap:
We will always have that in logging, it is easy find (the fields are the same).

ezr-ondrej (approved):
I'm fine with this :+1:
I'd like to do a full review tho, didn't tested it yet.

lzap:
Ok take your time, thanks.

ezr-ondrej (approved):
Totally make sense, lets try it out! :+1:
Thanks @lzap :orange_heart:

lzap:
It works, some issues get paired with older so the issue topic still contains UUIDs I think it is using stacktrace to calculate some sort of hash value.

PR #490, opened 2023-04-03
chore: fix generated content

lzap:
I don’t get why other PRs passed, I think the reason was that something passed tests before we implemented the CI check. This fixes it.


ezr-ondrej (approved):
Thanks! :+1:

PR #492, opened 2023-04-04
chore: add annotation for kube-linter statuser

lzap:
Another week, another try to silence that pesky linter.

Previous attempts failed but reason was probably something different, our pipeline got stuck so this time let’s just do this simple change alone to see if it works.

https://github.com/RHEnVision/provisioning-backend/pull/464


lzap:
Looks like it worked, meged at 17:10, last deployment on stage 17:15.

PR #495, opened 2023-04-10
feat: store Azure instance's IP address

ezr-ondrej:
Refactor Azure creation.
Extracts the common networking to be done once for all instances. Extracts Networking preparation to separate method, so we can store info about these.

Adjust Instance storing information to keep the IP address.

Refs HMS-1595

### DIFF
--- a/internal/clients/http/azure/create_vm.go
+++ b/internal/clients/http/azure/create_vm.go
@@ -23,8 +23,6 @@ import (
 	"go.opentelemetry.io/otel/codes"
 )
 
-var subscriptionId string
-
 const TraceName = "github.com/RHEnVision/provisioning-backend/internal/clients/http/azure"
 
 const (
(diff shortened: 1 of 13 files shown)


internal/clients/http/azure/create_vm.go
```diff
@@ -85,7 +87,9 @@ func (c *client) WaitForVM(ctx context.Context, resumeToken string) (clients.Azu
 		span.SetStatus(codes.Error, "polling of virtual machine creation status failed to start")
 		return "", fmt.Errorf("polling of virtual machine creation status failed to start: %w", err)
 	}
-	resp, err := pollerResponse.PollUntilDone(ctx, nil)
+	resp, err := pollerResponse.PollUntilDone(ctx, &runtime.PollUntilDoneOptions{
+		Frequency: vmPollFrequency,
```
ezr-ondrej, line 91: slipped off in previous refactoring back to default :/

scripts/rest_examples/reservation-create-azure.http
```diff
@@ -5,7 +5,7 @@ X-Rh-Identity: {{identity}}
 
 {
   "name": "azure-linux-us-east",
-  "source_id": "5",
+  "source_id": "{{azure-source-id}}",
```
lzap, line 8: This needs to be added in the `http-client-env.json` file I think GoLand will error out once this is used. But it’s fine.

lzap (approved):
Sure.

lzap:
@akhil-jha the API change is expected, instead of string we return struct with additional fields. This change was done previously for AWS.

akhil-jha (approved):

ezr-ondrej:
Nice! Thanks for quick reviews :orange_heart:

scripts/rest_examples/reservation-create-azure.http
ezr-ondrej: I've got it defined locally, there is no good default value for the comitted context IMO :thinking:

PR #497, opened 2023-04-11
chore: add trace id response header

lzap:
This adds correlation (trace) id in all response headers. This is useful when debugging success (or looks like success) requests on stage, the header immediately gives something you can search in Kibana.

Although we use trace id from open telemetry, this is not the official header which is [much more complex](https://www.w3.org/TR/trace-context/) and also does not make sense in response headers.

Edit: Initially I used `Correlation-Id` but I think we should stick to what we actually store in logs.


ezr-ondrej (approved):
Thanks :+1:

PR #498, opened 2023-04-11
refactor: Change permission check behaviour for not implemented sources

avitova:
The permission check was working only for AWS. This change should return `Valid: true` for Azure and GCP. :)


lzap:
Hmmm I was hoping to do this:

* AWS: try to connect via the credentials, check permissions, fail otherwise
* Azure: try to connect via the credentials, fail otherwise
* GCP: try to connect via the credentials, fail otherwise

We have a nice pattern here so you can utilize some sort of generalization (interface) if you think its a good fit.

lzap (approved):
You know what this is fine for now, we can improve later.

PR #500, opened 2023-04-14
fix(HMS-1616): split payload and model for better OpenAPI

lzap:
This is fairly meaty refactoring of our payload handling and openapi generation, but it is for the good.

Up until now, we embedded models into request/response payloads and generated the openapi types without any examples. Worked fine for our prototype times, however, now we are reaching our stable version and this is simply not enough.

The patch refactors pubkey and error payloads, splits the model and payload which was planned initially but we never got to it. The newly introduced package `payload/pex` contains examples which are used in the generator. What is great about this hybrid approach is that the Go compiler ensures types. Also we can use these payloads in our HTTP endpoint unit tests (aka integration tests).

This is first patch in the series which will introduce examples, I want to get this reviewed first and then followup with other models. I know it’s Summit time so this might get delayed, it is not on our top priority list.

When testing, please generate OpenAPI and check it. Here is a screenshot:

image

Aaaaand Friday! Yay. 🍻 🍻


cmd/spec/path.yaml
```diff
@@ -70,6 +73,9 @@ paths:
           application/json:
             schema:
               "$ref": "#/components/schemas/v1.PubkeyRequest"
+            examples:
+              example:
+                $ref: '#/components/examples/v1.PubkeyRequestExample'
         description: request body
         required: true
       responses:
```
lzap, line 81: Once this PR is merged, we can keep adding the following example references and example values in the `pex` package.

api/openapi.gen.json
```diff
@@ -305,21 +370,8 @@
...
-          "fingerprint": {
-            "type": "string"
-          },
-          "fingerprint_legacy": {
-            "type": "string"
-          },
-          "id": {
-            "format": "int64",
-            "type": "integer"
-          },
```
ezr-ondrej, line 317: This doesn't sound correct :)

ezr-ondrej (commented):
Very nice! :orange_heart: 

We will need to remove registration for the Pubkey and add PubkeyRequest IMO and use the PubkeyRequest instead of current one, now it generates wrong spec.

api/openapi.gen.json
lzap: It is correct actually, these are all fields generated by the server, so request payload must not contain it. Previously we did not care so it was all present (and we actually enforced an empty string on fingerprints for security reasons). This patch correctly removes this from the request payload.

lzap:
> We will need to remove registration for the Pubkey and add PubkeyRequest IMO and use the PubkeyRequest instead of current one, now it generates wrong spec.

What’s that? I don’t understand. Can you point out the code bits which are incorrect? Granted I was rushing this in during Friday evening so there will be some dragons. :-)

ezr-ondrej:
And i was rushing the review, it might be correct, I've assumed the payload
registration to the spec generator needs to change, but I'm probably wrong
and we might just need to regenerate the IQE spec for the tests to pass.

Dne po 17. 4. 2023 10:55 uživatel Lukáš Zapletal ***@***.***>
napsal:

lzap:
Good morning :-) No problem I will ask someone else for review. @avitova would you find time to do review and testing of this patch? Thanks, no rush.

internal/payloads/pubkey_payload.go
```diff
@@ -8,32 +8,54 @@ import (
...
-	// Fingerprint is read-only field
-	p.Fingerprint = ""
 	return nil
 }
 
 func (p *PubkeyResponse) Render(_ http.ResponseWriter, _ *http.Request) error {
 	return nil
 }
 
+func (p *PubkeyRequest) Model() *models.Pubkey {
```
avitova, line 36: I am not sure if Model is self-explanatory enough. The package is payloads, so we'd call this payloads.Model from outside, right? As this is exported, someone might call this from other package.

avitova (commented):
I have only one minor nitpick, but it is nothing important. Looks good. The file was generated as expected. :) But I think this looks good.

internal/payloads/pubkey_payload.go
lzap: This is a function with receiver so the use is:

```go
pr := PubkeyRequest{}
// ...
pubkeyModel := pr.Model()
```

It converts payload to model. Recommend me a better name, I did not like it either :-)

lzap:
Sorry for rebasing mid-review, I just removed "omitempty" for the ID which was incorrect and likely the reason why tests were failing.

internal/payloads/pubkey_payload.go
avitova: As I think about it, this might be the best way. I just found it confusing, that the word 'model' can be a noun, verb, or adjective. How about something like "NewModel"?

internal/payloads/pex/package.go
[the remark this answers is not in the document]
lzap: Payloads examples. I felt like `examples` is too long and also too generic.

internal/models/pubkey_model.go
```diff
@@ -10,33 +10,33 @@ import (
...
 	// typically prints the fingerprint without padding: ssh-keygen -l -f $HOME/.ssh/key.pub
 	// The length is exactly 45 characters including the padding.
 	// Example: "gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk="
-	Fingerprint string `db:"fingerprint" json:"fingerprint" validate:"omitempty,len=44"`
+	Fingerprint string `db:"fingerprint" validate:"omitempty,len=44"`
 
 	// MD5 fingerprint stored as hexadecimal with colons without any prefix. To generate
 	// such fingerprint: ssh-keygen -l -E md5 -f $HOME/.ssh/key.pub
 	// Example: "89:c5:99:b5:33:48:1c:84:be:da:cb:97:45:b0:4a:ee"
-	FingerprintLegacy string `db:"fingerprint_legacy" json:"fingerprint_legacy" validate:"omitempty,len=47"`
```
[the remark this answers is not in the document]
lzap: Because this is a model type, only pgx library needs "db" tags in order to marshal data but both json and yaml are used by Chi and OpenAPI respectively for marshalling. There will be few exceptions, those "detail" db columns which are marshalled into JSONB PostgreSQL type they will likely need also a JSON tag.

ezr-ondrej:
Thanks @lzap for doing this, it now needs rebase, but happy to get this in once we get QE on board (but should be just a matter of IQE client refresh IMO)

akhil-jha:
> Thanks @lzap for doing this, it now needs rebase, but happy to get this in once we get QE on board (but should be just a matter of IQE client refresh IMO)

no image on quay. Lets rebsae so new image is available. Don't want to retest it just yet

internal/payloads/pubkey_payload.go
lzap: Maybe `ToModel` or Ruby on Rails-like `AsModel`?

internal/payloads/pex/package.go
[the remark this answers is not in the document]
lzap: I just realized that since examples are currently only used in OpenAPI generator, I can actually move them to `/cmd/spec`. We can move them somewhere else if we ever need to use them in unit tests, which might not happen honestly.

internal/payloads/pubkey_payload.go
lzap: Going for the latter.

internal/payloads/pex/package.go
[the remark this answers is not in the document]
lzap: I am moving it there, we can move back later.

lzap:
So I rebased and here are the changes of the final version:

* Split of payloads and models.
* Added example type generation.
* Added example references into the spec.
* Added tags into the spec. This makes the left navigation way better (see Pubkeys on the left side).
* Removed the "index.html" page and replaced it with README link to ReDocly (live documentation from an arbitrary link - this is main branch not this PR tho).
* Settled on copy function name `NewXXX` that we should use to convert request payload to model. There is no interface yet, I don’t think we need one at this point.
* Exported `NewResponseError` which was previously not exported, initial version of my code used this function but then I gave up. I think it should be exported tho, somewhat unrelated yet I would like to get this in.
* Fixed IQE test failures (the code was actually returning ID=0 due to bug).

internal/payloads/error_payload.go
```diff
@@ -153,47 +151,47 @@ func NewClientError(ctx context.Context, err error) render.Renderer {
...
+func NewAzureError(ctx context.Context, message string, err error) *ResponseError {
 	message = fmt.Sprintf("Azure API error: %s", message)
 	return NewResponseError(ctx, http.StatusInternalServerError, message, err)
 }
 
-func NewGCPError(ctx context.Context, message string, err error) render.Renderer {
+func NewGCPError(ctx context.Context, message string, err error) *ResponseError {
 	message = fmt.Sprintf("Google API error: %s", message)
 	return NewResponseError(ctx, http.StatusInternalServerError, message, err)
 }
```
lzap, line 197: Oh I see that the `NewResponseError` refactoring was merged in other change, but the change of the return type to `Renderer` was a left-over (previously I planned to unexport the ReponseError type). This puts it back, I dont think we need to return `Renderer` interface at all.

avitova (commented):
LGTM. Generates correctly, and the interactive demo looks cool!

ezr-ondrej:
@lzap can you take a look at @avitova's findings so we can get this in before code freeze? IMO this would be really nice to have!

internal/payloads/pubkey_payload.go
```diff
@@ -8,32 +8,54 @@ import (
...
 func (p *PubkeyResponse) Render(_ http.ResponseWriter, _ *http.Request) error {
 	return nil
 }
 
+func (p *PubkeyRequest) NewModel() *models.Pubkey {
+	return &models.Pubkey{
+		Name: p.Name,
+		Body: p.Body,
+	}
+}
```
adiabramovitch, line 41: Just curious, why did you choose to create a new model?

internal/payloads/pubkey_payload.go
```diff
@@ -8,32 +8,54 @@ import (
 	"github.com/go-chi/render"
 )
 
-type PubkeyPayload struct {
-	*models.Pubkey
+// See models.Pubkey
+type PubkeyRequest struct {
+	Name string `json:"name" yaml:"name"`
+	Body string `json:"body" yaml:"body"`
+}
```
adiabramovitch, line 15: Again, out of curiosity, why is that way better?

cmd/spec/path.yaml
```diff
@@ -1,48 +1,60 @@
...
       operationId: getPubkeyById
+      tags:
+        - Pubkey
       description: >
         A pubkey represents an SSH public portion of a key pair with name and body.
         Pubkeys must have unique name and body (SSH public key fingerprint) per each account.
-        To find if a pubkey was uploaded to particular cloud, check PubkeyResource resource.
+        Pubkey type is detected during create operation. Currently two types are supported:
+        RSA and ssh-ed25519. Also fingerprint (SHA) and legacy (MD5) fingerprint is calculated
+        automatically and available. Fingerprints are used to check uniquness of key.
```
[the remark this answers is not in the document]
lzap: Updated to:

        A pubkey represents an SSH public portion of a key pair with name and body.
        Pubkeys must have unique name and body (SSH public key fingerprint) per each account.
        Pubkey type is detected during create operation as well as fingerprints.
        Currently two types are supported: RSA and ssh-ed25519. Also, two fingerprint
        types are calculated: standard SHA fingerprint and legacy MD5 fingerprint available
        under fingerprint_legacy field. Fingerprints are used to check uniqueness of key.

internal/payloads/pubkey_payload.go
lzap: This does not create new model, it returns a pointer of the current Pubkey model with two fields set, others are unset (thus default - empty string, 0 or nil). This is rather "conversion" function than anything else.

internal/payloads/pubkey_payload.go
lzap: This was alway the idea, in most typical HTTP apps we deal with two kind of types:

* Payloads represent the data that go in (requests) and out (responses). There are very different depending on the endpoint purpose, e.g. "create" endpoint may return nothing, the created data or just ID of the database index. Depends on what you want.
* Models represent tables in SQL, they need to be ideally 1:1 mapping to the fields.

Due to time constraints, we used models for all request/response payloads, but this created confusion - for example it generated an OpenAPI "create pubkey" which requires all fields:

* name
* body
* type
* fingerprint
* fingerprint_legacy

This was confusing, this is actually the purpose of this PR - to split payloads and models into two distinct objects. Now we can create PubkeyRequest that only have relevant fields - name and body. Everything else is calculated by the server and is "read only".

lzap:
Sorry I did not hit the summit date with this, rebased. Fixed all typos, reworded the unclear description.

avitova (approved):
LGTM:)

PR #501, opened 2023-04-19
fix: Azure image check

lzap:
Azure image IB ID check had a typo in the if statement, so it was actually never performed. I realized that when I started testing Azure on my dev setup and wanted to spawn a RHEL generic image from the Azure marketplace.

The patch also updates the http testing file with more sane defaults including the RHEL, btw image IDs are here: https://imagedirectory.cloud/browser/Azure (searching is not yet possible).

### DIFF
--- a/internal/services/azure_reservation_service.go
+++ b/internal/services/azure_reservation_service.go
@@ -3,7 +3,6 @@ package services
 import (
 	"fmt"
 	"net/http"
-	"strings"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/RHEnVision/provisioning-backend/internal/config"
@@ -16,6 +15,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/queue"
 	"github.com/RHEnVision/provisioning-backend/pkg/worker"
 	"github.com/go-chi/render"
+	"github.com/google/uuid"
 )
 
 func CreateAzureReservation(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


ezr-ondrej (commented):
I think it was correct and this fix is not.
We want UUID to get resolved by ibClient, which is now not happening.
And the direct link is working for sure as our IQE pipeline is using it :thinking:

internal/services/azure_reservation_service.go
```diff
@@ -77,14 +77,15 @@ func CreateAzureReservation(w http.ResponseWriter, r *http.Request) {
 
 	var azureImageName string
 	if strings.HasPrefix(payload.ImageID, "composer-api") {
```
ezr-ondrej, line 79: if you want the Marketplace images to work, just adjust this statement, here (the first branch of this if) is where we are handling the direct Azure images, we just expected the name to be from IB, which names the images with this prefix :)

lzap:
Oooh that is right, I thought for a moment that compose IDs start with `composer-api` followed by UUID. That is actually Azure ID, compose ID is a pure UUID. Of course, this make zero sense.

On top of that, @adiabramovitch is right that I carried over this "knowledge" also to GCP :-D

Now, the solution for what I want to achieve is a bit harder then AWS. In AWS, they have a consistent ID convention where everything stats with a prefix. In case of images, it is `ami-` and it is easy to find it. Since Azure is free-form (anything can be image ID really - I could not google any sort of documentation on this), I propose to make a UUID parsing test - if the image id is a valid UUID, then we treat it like it is compose ID. Otherwise we pass it in directly.

This allows me to test also with official RHEL images. I very often simply spawn a marketplace image to test something. I want to be able to quickly do that. Rebased.

lzap:
I also updated the model documentation. I am working on a big refactoring of OpenAPI specification which will include better examples. I will improve OpenAPI documentation once I get to this endpoint.

ezr-ondrej (approved):

ezr-ondrej:
I don't love linking fixes to epics (should have more specific target it's fixing).
I'll merge now, but we should discuss if this is important and we want to enforce it, or we do not care :)

lzap:
In the meantime, you just merged a PR that was entitled `ref: xxxx` and contained no description. Yeah, we should talk :-)

PR #502, opened 2023-04-19
chore: allow non-IB images for GCP

lzap:
This allows to launch official images not just UUIDs from image builder. Comes with an example .http file update which defaults to RHEL9 official image: https://imagedirectory.cloud/browser/GCP

### DIFF
--- a/internal/services/gcp_reservation_service.go
+++ b/internal/services/gcp_reservation_service.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/RHEnVision/provisioning-backend/internal/preload"
+	"github.com/google/uuid"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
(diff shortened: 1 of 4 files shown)


scripts/rest_examples/reservation-create-gcp.http
```diff
@@ -4,12 +4,12 @@ Content-Type: application/json
 X-Rh-Identity: {{identity}}
 
 {
-  "name": "gcp-linux-us-east4-a",
-  "zone": "us-east4-a",
+  "name": "gcp-linux-eu-west8-c",
+  "zone": "europe-west8-c",
```
lzap, line 8: east4-a was full. Since GCP has global images IIUC and most of the team is in EU zone, I think something closer might make sense.

scripts/rest_examples/reservation-create-gcp.http
```diff
@@ -4,12 +4,12 @@ Content-Type: application/json
...
   "source_id": "3",
-  "image_id": "80967e7f-efef-4eee-85b0-bd4cef4c455d",
+  "image_id": "https://www.googleapis.com/compute/v1/projects/rhel-cloud/global/images/rhel-9-v20230411",
   "amount": 1,
-  "machine_type": "n1-standard-1",
+  "machine_type": "e2-micro",
```
lzap, line 12: e2 are cheaper

internal/services/gcp_reservation_service.go
```diff
@@ -103,15 +104,22 @@ func CreateGCPReservation(w http.ResponseWriter, r *http.Request) {
...
+		name, ibErr = IBClient.GetGCPImageName(r.Context(), reservation.ImageID)
+		if ibErr != nil {
+			renderError(w, r, payloads.NewClientError(r.Context(), ibErr))
+			return
+		}
+
+		logger.Trace().Msgf("Image Name is %s", name)
+	} else {
+		// Direct image ID (e.g. from https://imagedirectory.cloud)
+		name = payload.ImageID
```
adiabramovitch, line 120: Why? I want the name, not the id. 
for example, the image id I know is e85c829f-bb14-4659-9571-25df69c0365a and the image name will not be found.

lzap:
Oooh that is right, I thought for a moment that compose IDs start with `composer-api` followed by UUID. That is actually Azure ID, compose ID is a pure UUID. Of course, this make zero sense. I carried over this from the Azure PR.

Now, the solution for what I want to achieve is similar to the Azure solution, it appears that Image IDs on GCP are in fact URLs pointing to the actual image but it can be also relative URLs (e.g. `/project/XXX`). I was hoping I could do a prefix test for "http" but that is not enough. I propose to make a UUID parsing test - if the image id is a valid UUID, then we treat it like it is compose ID. Otherwise we pass it in directly.

This allows me to test also with official RHEL images. I very often simply spawn a marketplace image to test something. I want to be able to quickly do that. Rebased.

I am also adding some improvements to the logging, I want to be able to easily search for compose IDs and image IDs.

lzap:
I also updated the model documentation. I am working on a big refactoring of OpenAPI specification which will include better examples. I will improve OpenAPI documentation once I get to this endpoint.

internal/services/gcp_reservation_service.go
ezr-ondrej: if this would be UUID, it would fall in the other branch and got resolved by IB client

ezr-ondrej (approved):
Looks good to me :+1:

PR #504, opened 2023-04-19
feat(HMS-1246): Use permission checks in statuser

avitova:
This adds permission logging in case of missing permissions. Caching should also be done for this. I feel like the source result should not be flagged as Unavailable or with Error in case of missing permission, as this is neither of these. Therefore, I thought adding the MissingPermission parameter might be suitable, but let me know, what you think.
For testing the missing permission in statuser, you could change iam_client hardcoded struct.

### DIFF
--- a/cmd/pbackend/statuser.go
+++ b/cmd/pbackend/statuser.go
@@ -130,8 +130,7 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 				Identity:     s.Identity,
 				ResourceType: "Application",
 			}
-			// TODO: check if source is avavliable - WIP
-			sr.Status = kafka.StatusAvaliable
+			sr.Status = kafka.StatusAvailable
 			chSend <- sr
 			metrics.IncTotalSentAvailabilityCheckReqs(models.ProviderTypeAzure.String(), sr.Status.String(), nil)
 
@@ -148,21 +147,28 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 		logger.Trace().Msgf("Checking AWS source availability status %s", s.SourceApplicationID)
 		metrics.ObserveAvailabilityCheckReqsDuration(models.ProviderTypeAWS.String(), func() error {
 			var err error
+			var permissions []string
 			sr := kafka.SourceResult{
 				ResourceID:   s.SourceApplicationID,
 				Identity:     s.Identity,
 				ResourceType: "Application",
 			}
-			_, err = clients.GetEC2Client(ctx, &s.Authentication, "")
+			ec2Client, err := clients.GetEC2Client(ctx, &s.Authentication, "")
 			if err != nil {
 				sr.Status = kafka.StatusUnavailable
 				sr.Err = err
 				logger.Warn().Err(err).Msg("Could not get aws assumed client")
-				chSend <- sr
 			} else {
-				sr.Status = kafka.StatusAvaliable
-				chSend <- sr
+				sr.Status = kafka.StatusAvailable
+				permissions, err = ec2Client.CheckPermission(ctx, &s.Authentication)
+				if err != nil {
+					sr.Err = err
+					sr.Status = kafka.StatusUnavailable
+					sr.MissingPermissions = permissions
+					logger.Warn().Err(err).Bool("missing_permissions", true).Str("source_id", sr.ResourceID).Msg("AWS permission check failed")
+				}
 			}
+			chSend <- sr
 			metrics.IncTotalSentAvailabilityCheckReqs(models.ProviderTypeAWS.String(), sr.Status.String(), err)
 			return fmt.Errorf("error during check: %w", err)
 		})
@@ -196,7 +202,7 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
 				logger.Warn().Err(err).Msg("Could not list gcp regions")
 				chSend <- sr
 			} else {
-				sr.Status = kafka.StatusAvaliable
+				sr.Status = kafka.StatusAvailable
 				chSend <- sr
 			}
 			metrics.IncTotalSentAvailabilityCheckReqs(models.ProviderTypeGCP.String(), sr.Status.String(), err)
--- a/internal/kafka/source_result.go
+++ b/internal/kafka/source_result.go
@@ -10,7 +10,7 @@ type StatusType string
 
 const (
 	StatusUnavailable StatusType = "unavailable"
-	StatusAvaliable   StatusType = "available"
+	StatusAvailable   StatusType = "available"
 )
 
 type SourceResult struct {
@@ -24,6 +24,8 @@ type SourceResult struct {
 	Err error `json:"error"`
 
 	Identity identity.Principal `json:"-"`
+
+	MissingPermissions []string `json:"-"`
 }
 
 func (sr SourceResult) GenericMessage(ctx context.Context) (GenericMessage, error) {
--- a/internal/clients/http/ec2/iam_client.go
+++ b/internal/clients/http/ec2/iam_client.go
@@ -3,6 +3,7 @@ package ec2
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"strings"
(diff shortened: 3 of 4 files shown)


lzap:
> I feel like the source result should not be flagged as Unavailable or with Error in case of missing permission, as this is neither of these.

That is correct, I think a warning in logs is enough. We should not see many of these, only old sources would experience this.

lzap:
Test failure seems relevant (timeout)?

cmd/pbstatuser/main.go
```diff
@@ -151,16 +151,24 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
...
 				ResourceType: "Application",
 			}
-			_, err = clients.GetEC2Client(ctx, &s.Authentication, "")
+			ec2Client, err := clients.GetEC2Client(ctx, &s.Authentication, "")
 			if err != nil {
 				sr.Status = kafka.StatusUnavailable
 				sr.Err = err
-				logger.Warn().Err(err).Msg("Could not get aws assumed client")
-				chSend <- sr
+				logger.Warn().Err(err).Msgf("Could not get aws assumed client %s", err)
```
lzap, line 158: Please do not send error messages to messages because of Sentry.

https://github.com/RHEnVision/provisioning-backend/pull/488

cmd/pbstatuser/main.go
```diff
@@ -151,16 +151,24 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
...
+				sr.Status = kafka.StatusAvailable
+			}
+			permissions, err := ec2Client.CheckPermission(ctx, &s.Authentication)
+			if err != nil {
+				sr.Err = err
+				logger.Warn().Err(err).Msgf("Could not check aws permissions %s", err)
+			}
+			sr.MissingPermissions = permissions
+			if len(permissions) != 0 {
+				logger.Warn().Err(err).Msgf("No sufficient permissions: %s", permissions)
```
lzap, line 169: This does not seem correct. CheckPermissions either returns nil result and error, or something in the result. You reported the error above, here I would simply issue a warning with the list. Also can you add a field so we can easily find all records in Kibana, for example:

    logger.Warn().Bool("missing_permission", true).Msgf("No sufficient permissions: %s", permissions)

cmd/pbstatuser/main.go
```diff
@@ -151,16 +151,24 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
...
+				logger.Warn().Err(err).Msgf("Could not get aws assumed client %s", err)
 			} else {
-				sr.Status = kafka.StatusAvaliable
-				chSend <- sr
+				sr.Status = kafka.StatusAvailable
+			}
+			permissions, err := ec2Client.CheckPermission(ctx, &s.Authentication)
+			if err != nil {
+				sr.Err = err
+				logger.Warn().Err(err).Msgf("Could not check aws permissions %s", err)
```
lzap, line 165: Ditto.

internal/kafka/source_result.go
```diff
@@ -24,6 +24,8 @@ type SourceResult struct {
 	Err error `json:"error"`
 
 	Identity identity.XRHID `json:"-"`
+
+	MissingPermissions []string `json:"missing_permissions"`
 }
```
lzap, line 29: Sources do not currently support that, so I would use `-` json tag for now. Only into logging for now.

lzap (commented):
Looks great, there is a failure can you take a look?

lzap:
_For the record, @avitova might not get to working on this PR. This is not high-priority so let’s keep it opened._

lzap:
Let me know when this is ready for re-review.

cmd/pbackend/statuser.go
```diff
@@ -131,7 +131,7 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 				ResourceType: "Application",
 			}
 			// TODO: check if source is avavliable - WIP
```
lzap, line 133: Can you drop this TODO? This patch actually implements the missing functionality.

cmd/pbackend/statuser.go
```diff
@@ -153,16 +153,24 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
...
+				sr.Status = kafka.StatusAvailable
 			}
+			permissions, err := ec2Client.CheckPermission(ctx, &s.Authentication)
+			if err != nil {
+				sr.Err = err
+				logger.Warn().Err(err).Msg("AWS permission check failed")
+			}
+			sr.MissingPermissions = permissions
+			if permissions != nil {
+				logger.Warn().Bool("missing_permission", true).Msgf("No sufficient permissions: %s", permissions)
```
lzap, line 171: Perhaps "insufficient" is better wording.

Also you can use `strings.Join` to join string slice by comma, but [this works too](https://go.dev/play/p/yVujHTNW-IL).

Finally, when `CheckPermission` returns `nil` it likely returns an error too, can you squash this into just a single warning message? You can stuff everything into one log record, me as someone who is debugging this issue would be interested in:

* source id (already present in the context logger)
* account number (could be added in the `processMessage` too as it is in the identity but perhaps out of scope of this PR)
* error message (via Err function)
* missing permissions

So basically this line contains it all, perhaps just delete the 167 and move this into the if statement above or something like that.

cmd/pbackend/statuser.go
```diff
@@ -153,16 +153,24 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
...
 				logger.Warn().Err(err).Msg("Could not get aws assumed client")
-				chSend <- sr
 			} else {
-				sr.Status = kafka.StatusAvaliable
-				chSend <- sr
+				sr.Status = kafka.StatusAvailable
 			}
+			permissions, err := ec2Client.CheckPermission(ctx, &s.Authentication)
+			if err != nil {
+				sr.Err = err
```
lzap, line 166: Here you should also set `sr.Status = kafka.StatusUnavailable` because missing permissions effectively means the source is not "all right" and the user should delete it and create it again.

cmd/pbackend/statuser.go
avitova: This is a very good point. The CheckPermission function returned only missing permissions when something was missing, and the error value was returned only when something else failed. We should return an error for missing permissions, so I changed it in this PR. I hope this now makes more sense.

ezr-ondrej:
:tada: ... no idea what was the failure before, but now it works :smirk:

cmd/pbackend/statuser.go
```diff
@@ -153,16 +152,22 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 				Identity:     s.Identity,
 				ResourceType: "Application",
 			}
-			_, err = clients.GetEC2Client(ctx, &s.Authentication, "")
+			ec2Client, err := clients.GetEC2Client(ctx, &s.Authentication, "")
 			if err != nil {
 				sr.Status = kafka.StatusUnavailable
 				sr.Err = err
 				logger.Warn().Err(err).Msg("Could not get aws assumed client")
```
ezr-ondrej, line 159: Shouldn't we make sure, that when `GetEC2Client` errors out, we do not call CheckPermission on it? that might be nil pointer reference, right?

I'd just return here honestly, we would need to bump the metric here as well, but that's just three extra lines (write to channel, bump metric and return), not a big deal. WDYT?

ezr-ondrej: I might be missing something here tho :)

avitova: That's definitely true, we should return here! TY.

lzap: Guys, please no `return` statements here, this is a channel processing loop!

One note: this is one of rare examples when `goto` statement which is [available in Go](https://go.dev/ref/spec#Goto_statements) could be reasonable, also a function could be good fit. I see some code being duplicated (creating the result, sending the result to).

ezr-ondrej: I believe @avitova solved it pretty well :)
Good compromise! :+1:

ezr-ondrej (approved):
I'm good here, thanks! :orange_heart:

cmd/pbackend/statuser.go
lzap: Oh no, did I scared you, @ezr-ondrej ? :-)

You merge me a trigger and stored SQL procedure and the next day I debate loops :-D

lzap (approved):
One detail - the error length is unlimited and in case user actually forgot to add any permission this gets rendered to screen as something very long (covering whole screen or something).

In another PR, I suggest to limit the list to some reasonable length and in these cases add ` and more` to the error.

PR #505, opened 2023-04-20
feat: add environment to error message

lzap:
I find sometimes difficult to find which Kibana should I start searching for. Error reports usually do not tell.

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -398,6 +398,12 @@
           "build_time": {
             "type": "string"
           },
+          "edge_id": {
+            "type": "string"
+          },
+          "environment": {
+            "type": "string"
+          },
           "error": {
             "type": "string"
           },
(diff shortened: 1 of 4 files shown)


lzap:
Also added Edge ID which is also propagated to other services as well (sources and image builder).

Finally, rebased to fix checks and unexported error payload which must be used via the New method to enforce that info from context is present.

avitova (commented):
So I tested it, and it looks like everything works fine. I have a minor nitpick. I am not sure if the alias InClowder() is even needed, I found it a bit confusing.

lzap:
Sure, rebased.

lzap:
Good point, so what happened was that I wanted to unexport the ErrorPayload and only have `NewErrorPayload` function to be able to create the type to ensure fields are initialized properly. But that actually removed the type from OpenAPI.

I am fixing that, I guess that is not possible, at least without some ugly hacks like user-defined JSON marshalling.

I am going to merge this since our stage is stuck somehow and I want to know if thats fixed or not.

lzap:
Rebased.

adiabramovitch:
LGTM

PR #507, opened 2023-04-23
ref: Change metrics according to app-sre notes

adiabramovitch:


ezr-ondrej:
I'd rather see a description of a change as a commit message instead of why it happend, but apart of that, LGTM :+1:

ezr-ondrej (approved):
Thanks @adiabramovitch

ezr-ondrej:
yeah, not counting 400 as error really bumped our success rate xD 74% -> 92% on stage :sunglasses:

lzap:
The only concern I have is that sources (SAC - availability checking) will likely do 90% of all traffic on prod instance, let’s start simply for now but we might split these at some point as it can easily bring the rate down a bit in case of some SAC bug (while the app might be actually 99% available).

Not now, not here tho.

PR #510, opened 2023-04-26
fix(HMS-1675): simplify SAC success metric

lzap:
The dashboard success rate sometimes display 101% and this is caused by inflight messages I think. This patch simplifies this, there is no need to make this unnecessary complex, a single counter with proper labels can give us the number.

Let’s merge the change first, then I want to wait and tomorrow I will push a grafana dashboard change, something like (untested):

SAC Success Rate:

    sum(rate(provisioning_source_availability_check{status=="ok"}[$__range]))/sum(rate(provisioning_source_availability_check[$__range]))

I am thinking introducing also number of "in flight" requests (currently enqueued and not yet processed). Basically a "kafka lag" (pseudo query I don’t know yet how to achieve this):

    provisioning_http_request_total{path="/api/provisioning/v1/availability_status/sources/"}[$__range] - provisioning_source_availability_check{status=="ok"}

And total SAC checks in the selected range:

    sum(increase(provisioning_source_availability_check[$__range]))

### DIFF
--- a/cmd/pbstatuser/main.go
+++ b/cmd/pbstatuser/main.go
@@ -87,7 +87,6 @@ func processMessage(origCtx context.Context, message *kafka.GenericMessage) {
 	// Fetch authentication from Sources
 	authentication, err := sourcesClient.GetAuthentication(ctx, sourceId)
 	if err != nil {
-		metrics.IncTotalInvalidAvailabilityCheckReqs()
 		if errors.Is(err, clients.NotFoundErr) {
 			logger.Warn().Err(err).Msg("Not found error from sources")
 			return
@@ -128,10 +127,10 @@ func checkSourceAvailabilityAzure(ctx context.Context) {
 				Identity:     s.Identity,
 				ResourceType: "Application",
 			}
-			// TODO: check if source is avavliable - WIP
+			// TODO: https://issues.redhat.com/browse/HMS-1674
 			sr.Status = kafka.StatusAvaliable
 			chSend <- sr
-			metrics.IncTotalSentAvailabilityCheckReqs(models.ProviderTypeAzure.String(), sr.Status.String(), nil)
+			metrics.IncSourceAvailabilityCheck(models.ProviderTypeAzure.String(), "ok")
 
 			return fmt.Errorf("error during check: %w", err)
 		})
@@ -157,11 +156,12 @@ func checkSourceAvailabilityAWS(ctx context.Context) {
 				sr.Err = err
 				logger.Warn().Err(err).Msg("Could not get aws assumed client")
 				chSend <- sr
+				metrics.IncSourceAvailabilityCheck(models.ProviderTypeAWS.String(), "err")
 			} else {
 				sr.Status = kafka.StatusAvaliable
 				chSend <- sr
+				metrics.IncSourceAvailabilityCheck(models.ProviderTypeAWS.String(), "ok")
 			}
-			metrics.IncTotalSentAvailabilityCheckReqs(models.ProviderTypeAWS.String(), sr.Status.String(), err)
 			return fmt.Errorf("error during check: %w", err)
 		})
 	}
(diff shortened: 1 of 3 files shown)


ezr-ondrej:
IMO `sum(rate(provisioning_source_availability_check{status=="ok"}[$__range]))/sum(rate(provisioning_source_availability_check[$__range]))` would be wrong, because you'd be counting

"How many checks resulted with available status out of all"

Which is not what we wanted and that's why the metric is so complicated, because we are trying to figure out how many of the checks were requested but were not successfuly completed.

check that was done, but resulted in "Unavailable" status, is still a successful check in my opinion and in current design of the metric.

ezr-ondrej:
That said, I'd be ok scraping that metric totaly if we do not know how to measure for it properly

lzap:
> Which is not what we wanted and that's why the metric is so complicated, because we are trying to figure out how many of the checks were requested but were not successfuly completed.

Well, yes and no. The way we do it is not the best way to measure it. Let me explain.

Imagine the current metric goes down from 100% or 101% :-D down to 60%. Can you tell, without further investigation, if this was caused by sources offline or our statuser pod offline? Or whatever reason it could be. You can't, this is not a good metric.

I want a clean and real value of success rate that is not skewed by inflight messages. That is what the newly introduced counter provides.

At the same time, I very much like the idea of ability to see when SAC HTTP reqests are being ignored. I said this twice and I will say it again: I very much like your idea of doing this. But not this way. I believe that calculating number of messages which are "stuck in the queue" for whatever reason is more straightforward solution. That is exactly why I propose to introduce a new metric for this. And no code change is actually needed for this, it will be a dashboard change which I clearly described in the PR above. Granted it is a pseudo code, I can take a look and write it and show you a full graph if you are in doubts.

This way we will have:

* clear number of real success rate
* clean number of SAC messages not being processed for whatever reason

lzap:
Just in case it was not clear, I am proposing to keep the existing metric:

    ( sum(increase(provisioning_source_availability_check_request_total[6h])) + sum(increase(provisioning_invalid_source_availability_check_request_total[6h])) )/ sum(increase(provisioning_http_request_total{path="/api/provisioning/v1/availability_status/sources/"}[6h]))

but basically simplify it (instead of two metrics we would only have one - see this patch) and rename it from "Success rate" to "Delivery rate". And introduce a new "Success rate" that is unbiased.

This is currently 1 (100%) by the way.

cmd/pbstatuser/main.go
```diff
@@ -193,11 +193,12 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
 				sr.Err = err
 				logger.Warn().Err(err).Msg("Could not list gcp regions")
 				chSend <- sr
+				metrics.IncSourceAvailabilityCheck(models.ProviderTypeGCP.String(), "err")
```
ezr-ondrej, line 196: can we make this "result=unavailable"

cmd/pbstatuser/main.go
```diff
@@ -193,11 +193,12 @@ func checkSourceAvailabilityGCP(ctx context.Context) {
...
 			} else {
 				sr.Status = kafka.StatusAvaliable
 				chSend <- sr
+				metrics.IncSourceAvailabilityCheck(models.ProviderTypeGCP.String(), "ok")
```
ezr-ondrej, line 200: can we make this "result=available"?

adiabramovitch:
I agree with Ondrej
`> I want a clean and real value of success rate that is not skewed by inflight messages. That is what the newly introduced counter provides.
`
But that is not the case here, this metric does not count how many successful requests have been done to sources, which means it does not show the success rate, the launch service success rate requesting sources. 
It checks how many available sources/valid sources we have, which I don't think this information is relevant for us. 

We wanted to check how many of our requests to sources service have gone right. 
I would say we need to find a way to get confirmation from sources (maybe using Kafka) that the process went through, if so we increase the metric, and if not we know we had a problem.

lzap:
Oh I get the point now, yeah. I did not mean to check only "ok" results, I was so focused on the 101% than I haven’t realized that todays "Source Availability Success Rate" really does mean "Successfully Delivered Available Check Rate".

I also realized that we actually don’t need in-flight requests metric because the "success rate" actually goes down when things are not being delivered. So maybe changing wording to:

**Processed Requests** as `http hits / processed total` : can be still actually 101% on message redelivery. I think this can be fixed with some sort of `max(result, 1.0)` function.

**Available Sources**: this is just informative, not anything we can do something about, but it is indicative if it goes down to 10% for a hyperscaler that it might be down. The pane can be implemented per-hyperscaler actually if we have a label for that.

I might raise a new PR we use two metrics where we could only use one + two labels. I need to check.

PR #511, opened 2023-05-03
chore: update http dashboard panels

lzap:
I took a closer look on our SLO target (200ms successful requests) and because it way way bellow 95% I investigated, looks like the query we have there is not correct. Also took the opportunity to slightly enhance all other HTTP-related panels:

* Removed HTTP Success Latency graph and created a new one named HTTP Success Latency 200ms Target which is now the "big number" with background line graph for history.
* The expression `sum(rate(provisioning_http_request_duration_ms_bucket{code=~"2.*", le="200"}[$__range])) / sum(rate(provisioning_http_request_duration_ms_count[$__range]))` does divide by all requests, but we need to divide only by successful counts. This was giving us quite low number under our current SLO (95%). The correct expression is `sum(rate(provisioning_http_request_duration_ms_bucket{code=~"2.*", le="200"}[$__range])) / sum(rate(provisioning_http_request_duration_ms_bucket{code=~"2.*", le="+Inf"}[$__range]))` which divides 200ms successful requests by all successful requests. Either infinity bucket or `_count` can be used, this is equal.
* Updated HTTP requests color to blue (it is informative only).
* Removed decimals from HTTP Success Rate, added background line graph for history.
* Renamed legends in HTTP Throughtput Rate, updated y-axis to ops/m and made line 2px width for better visibility of the red line.
* Changed the Request Latency to point graph (it was already sort of point with lines).
* Renamed Source Avail Success Rate to Source Avail Delivery rate, that represents the number in a more correct way.
* Rewritten Source Availability Check Latency to be dot graph similarly to the HTTP Request Latency.

Before the patch:

image

After the patch:

image

### DIFF
--- a/deploy/Provisioning-1674548289785.json
+++ b/deploy/Provisioning-1674548289785.json
@@ -661,7 +661,7 @@
             "axisLabel": "milliseconds",
             "axisPlacement": "auto",
             "barAlignment": 0,
-            "drawStyle": "line",
+            "drawStyle": "points",
             "fillOpacity": 36,
             "gradientMode": "scheme",
             "hideFrom": {
@@ -673,8 +673,8 @@
             "lineStyle": {
               "fill": "solid"
             },
-            "lineWidth": 3,
-            "pointSize": 5,
+            "lineWidth": 0,
+            "pointSize": 7,
             "scaleDistribution": {
               "type": "linear"
             },
@@ -746,7 +746,7 @@
               {
                 "id": "color",
                 "value": {
-                  "fixedColor": "red",
+                  "fixedColor": "orange",
                   "mode": "fixed"
                 }
               }
@@ -755,8 +755,8 @@
         ]
       },
       "gridPos": {
-        "h": 9,
-        "w": 19,
+        "h": 13,
+        "w": 24,
         "x": 0,
         "y": 26
       },
@@ -826,7 +826,7 @@
         "h": 1,
         "w": 24,
         "x": 0,
-        "y": 35
+        "y": 39
       },
       "id": 24,
       "panels": [],
@@ -847,11 +847,13 @@
         "type": "prometheus",
         "uid": "${datasource}"
       },
+      "description": "Our SLO target",
       "fieldConfig": {
         "defaults": {
           "color": {
             "mode": "thresholds"
           },
+          "decimals": 0,
           "mappings": [
             {
               "options": {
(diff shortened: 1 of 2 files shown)


deploy/Provisioning-1674548289785.json
```diff
@@ -1553,8 +1654,8 @@
       {
         "current": {
           "selected": false,
-          "text": "crcs02ue1-prometheus",
-          "value": "crcs02ue1-prometheus"
+          "text": "crcp01ue1-prometheus",
+          "value": "crcp01ue1-prometheus"
```
lzap, line 1658: Ha this was not intentional, I made the changes on prod Grafana this needs to be fixed.

deploy/Provisioning-1674548289785.json
```diff
@@ -24,6 +24,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
+  "id": 66892,
```
lzap, line 27: Not sure why "id" is being added here.

deploy/Provisioning-1674548289785.json
```diff
@@ -746,7 +747,7 @@
               {
                 "id": "color",
                 "value": {
-                  "fixedColor": "red",
+                  "fixedColor": "orange",
```
lzap, line 750: I don’t like red color for p50 dots, red means "bad" so I am proposing orange here. No other colors from the default color palettes are available.

deploy/Provisioning-1674548289785.json
ezr-ondrej: did you figure out?

ezr-ondrej (approved):
I'm guessing you need to regenerate the configmap, apart of that looks really great! thanks for doing this! :orange_heart:

Nit: did you find out why the id is added?
Nit2: why is the 99% in first pane lower line than 96% in the second pane? :thinking:

lzap:
> id

No idea why this was added. I removed it just to be sure.

> sparkline problem

No idea how to set y-axis on a sparkline, I do have "max" value set to 1 for both, yet, something is a bit off. I just noticed tho that one panel was slightly wider, changed, it is now the same. Also added sparkline to the SAC Delivery rate:

image

lzap:
Regenerated, rebased.

ezr-ondrej:
Thanks! :orange_heart:

PR #512, opened 2023-05-03
chore: improve glitchtip error handling

lzap:
I did a full review of all our Sentry errors, I noticed one that is ugly when pubkey is not valid (name/body), this fixes it returning a better error.

Also when there is a panic in worker, I saw nil exception this hopefully fixes it.


ezr-ondrej (approved):
Nice, thanks! :+1:

PR #518, opened 2023-05-05
refactor: remove compression

lzap:
We have figured out that compression for browsers are done by 3scale, there is no point having this feature for in-cluster queries. Removing.


ezr-ondrej (approved):
Thanks! :orange_heart:

PR #520, opened 2023-05-06
feat: Fetch Azure image resource group from IB

ezr-ondrej:
Support for fetching resource group the image resides in from Image Builder.
This allows for launching images that are build in a different resource group then the Launch one.

Changes the expected identifier in the Azure client.
Now we expect the full ID instead of a name, thus the resolution of name to ID is done in the service now.

Fixes HMS-1691

This was enabled by https://github.com/osbuild/image-builder/pull/707
We need to wait for IB to release for us to release this.

### DIFF
--- a/internal/clients/http/image_builder/image_client.go
+++ b/internal/clients/http/image_builder/image_client.go
@@ -10,7 +10,6 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
 	"github.com/RHEnVision/provisioning-backend/internal/headers"
-	openapi_types "github.com/deepmap/oapi-codegen/pkg/types"
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 	"go.opentelemetry.io/otel"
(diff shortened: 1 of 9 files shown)


ezr-ondrej:
I've tested this works for image build into a different Resrouce Group :tada:

internal/clients/http/image_builder/image_client.go
```diff
@@ -167,12 +175,24 @@ func (c *ibClient) checkCompose(ctx context.Context, composeID string) (*UploadS
 		return nil, fmt.Errorf("fetch image status call: %w", err)
 	}
 
-	if resp.JSON200.ImageStatus.Status != ImageStatusStatusSuccess {
+	return resp.JSON200, nil
+}
+
+func (c *ibClient) checkCompose(ctx context.Context, composeID string) (*UploadStatus, error) {
+	logger := logger(ctx)
+	logger.Trace().Msgf("Fetching image status %v from composes", composeID)
```
lzap, line 183:
```suggestion
	logger.Trace().Msgf("Fetching image status from compose %s", composeID)
```

is slightly better, not a blocker.

lzap (approved):
Great ty not testing but...

PR #521, opened 2023-05-06
Fix Go version and tools, update deps

lzap:
Regenerating clients today revealed one problem - we are all using "random" versions of the OpenAPI generator and this leads to differences in the code. Also using different version of Go leads to changes in `go.mod` as all of us use slightly different versions of dependencies.

This patch is inspired by https://github.com/osbuild/image-builder/blob/main/tools/prepare-source.sh and it introduces a new makefile target:

    make install-go

It will use the Go standard way of installing multiple versions of Go and download `go1.18.10` (configurable) into the GOROOT directory (typically `$HOME/go`). Then this stub is used to extract SDK into the same directory and this version is used for ALL make targets (build, test, etc).

In addition to that, all tools installed by `make install-tools` now install into PROJECT/bin instead of `$HOME/go/bin`. Also explicit versions are specified in the Makefile to make sure we all use the very same version.

This should avoid the mentioned problems. If you see a change in `go.sum` without changing `go.mod` during review, ask the contributor to remove that change and to use the required Go version. People can still use their IDEs for development, they just need to make sure:

* They use the correct Go version (typically `$HOME/go/go1.18.10` location).
* They use utilities from PROJECT/bin.

Finally, when building the release, we do NOT want to use Go from internet, we want to use the Go from the build container. Therefore GO variable is overriden to simply `go`.

As part of this PR I also updated dependencies and regenerated the clients as I needed to make sure the changes do work. Please review carefully, this is a big update.


internal/jobs/common.go
```diff
@@ -112,6 +112,7 @@ func nilUnlessTimeout(ctx context.Context) error {
 }
 
 // sleep with context deadline
+//
 //nolint:wrapcheck
```
lzap, line 116: Caused by the newer version of the formatter I think.

ezr-ondrej (commented):
We will need to use that version of go in our GH actions I'm guessing.

mk/0_initial.mk
```diff
@@ -2,4 +2,14 @@
...
+
+GO_VERSION?=1.18.10
+GOBIN:=$(PROJECT_DIR)/bin
+GO?=$(shell go env GOPATH)/bin/go$(GO_VERSION)
+
+GOLINT?=$(GOBIN)/golangci-lint
+GOFUMPT?=$(GOBIN)/gofumpt
+GOIMPORTS?=$(GOBIN)/goimports
+OAPICODEGEN?=$(GOBIN)/oapi-codegen
+TERN?=$(GOBIN)/tern
```
lzap, line 15: In this file, we define the required Go version. We need to regularly bump this but if we are a bit behind the RH container, it is not a big deal. We will just make sure we all are on the same page.

ezr-ondrej:
The commit message is not complient and GH actions are not happy, but apart of those issues, I'm good here :+1:

mk/deps.mk
```diff
@@ -2,26 +2,34 @@
...
+	$(GO) get github.com/jackc/puddle/v2@v2.0.0
+	# Compile errors with Go 1.18
+	$(GO) get go.opentelemetry.io/contrib@v1.15.0
+	$(GO) get go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp@v0.40.0
+	$(GO) get go.opentelemetry.io/otel@v1.14.0
+	$(GO) get go.opentelemetry.io/otel/exporters/jaeger@v1.14.0
+	$(GO) get go.opentelemetry.io/otel/metric@v0.37.0
+	$(GO) get go.opentelemetry.io/otel/sdk@v1.14.0
+	$(GO) get go.opentelemetry.io/otel/trace@v1.14.0
+	$(GO) mod tidy
```
lzap, line 32: Had to downgrade OpenTelemetry - compile error on 1.18 :-(

lzap:
> We will need to use that version of go in our GH actions I'm guessing.

Yeah ideally we could use `install-go` on GHA but let me just try to set `GO=go` meaning "use the Go you have in the container" for now. I mean if we keep the versions the same, and I created a comment, it should work.

ezr-ondrej (approved):
Nice! This resolves my issues and lets me get rid of gvm, thanks! :orange_heart:

@avitova @amirfefer @adiabramovitch pls use the new target now :)

PR #522, opened 2023-05-09
feat(HMSPROV-449): Add correlation id to logger and context

adiabramovitch:
After adding the correlation id header to the request sent from the UI, we add the correlation id to the context and the logger in this PR.

### DIFF
--- a/internal/middleware/correlation_id.go
+++ b/internal/middleware/correlation_id.go
@@ -1,0 +1,26 @@
+package middleware
+
+import (
+	"net/http"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+)
+
+func CorrelationID(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		corrId := r.Header.Get("X-Correlation-Id")
+		if corrId != "" {
+			ctx = ctxval.WithCorrelationId(ctx, corrId)
+			// Store in response headers for easier debugging
+			w.Header().Set("X-Correlation-Id", corrId)
+			logger := ctxval.Logger(ctx).With().Str("correlation_id", corrId).Logger()
+			logger.Trace().Msgf("Added correlation id %s to logger", corrId)
+			ctx = ctxval.WithLogger(ctx, &logger)
+		}
+
+		next.ServeHTTP(w, r.WithContext(ctx))
+	}
+	return http.HandlerFunc(fn)
+}
--- a/internal/middleware/logger.go
+++ b/internal/middleware/logger.go
@@ -21,6 +21,7 @@ func LoggerMiddleware(rootLogger *zerolog.Logger) func(next http.Handler) http.H
 			bytesIn, _ := strconv.Atoi(r.Header.Get("Content-Length"))
 			traceId := ctxval.TraceId(r.Context())
 			edgeId := ctxval.EdgeRequestId(r.Context())
+			corrId := ctxval.CorrelationId(r.Context())
 			lctx := rootLogger.With().
 				Timestamp().
 				Str("trace_id", traceId).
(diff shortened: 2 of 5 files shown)


adiabramovitch:
@lzap Is there any other place I should add the correlation id?

internal/middleware/correlation_id.go
```diff
@@ -0,0 +1,28 @@
...
+		// Store in response headers for easier debugging
+		w.Header().Set("Provisioning-Wizard-Id", corrId)
+		logger := ctxval.Logger(ctx).With().Str("correlation_id", corrId).Logger()
+		logger.Trace().Msgf("Added correlation id %s to logger", corrId)
+
+		ctx = ctxval.WithLogger(ctx, &logger)
+		next.ServeHTTP(w, r.WithContext(ctx))
+	}
+	return http.HandlerFunc(fn)
+}
```
lzap, line 28: The middleware and context is named "correlation", in the comment you have "edge request" and the header is named "provisioning wizard id".

I think we should be consistent, just a tad :-D

I don’t have a preference, just pick one term and use it everywhere.

Also note that custom HTTP headers should start with "X-" prefix as that namespace is reserved for custom headers. So for example if you choose this feature to be "correlation_id", then the header should be named `X-Correlation-Id`.

internal/middleware/logger.go
```diff
@@ -21,13 +21,15 @@ func LoggerMiddleware(rootLogger *zerolog.Logger) func(next http.Handler) http.H
...
+			corrId := ctxval.CorrelationId(r.Context())
 			lctx := rootLogger.With().
 				Timestamp().
 				Str("trace_id", traceId).
 				Str("remote_ip", r.RemoteAddr).
 				Str("url", r.URL.Path).
 				Str("method", r.Method).
-				Int("bytes_in", bytesIn)
+				Int("bytes_in", bytesIn).
+				Str("corr_id", corrId)
```
lzap, line 32: If you choose correlation id as the term, then I would suggest to use the full field name just to be clear:

```suggestion
				Str("correlation_id", corrId)
```

lzap (changes requested):
No, this is entirely correct.

Except naming, I do suggest to pick a consistent name-header-function-ctxval name that is consistent.

internal/middleware/correlation_id.go
```diff
@@ -0,0 +1,28 @@
...
+	"net/http"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ctxval"
+)
+
+func CorrelationID(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		// Edge request id
```
lzap, line 13: This is a left-over but no big deal.

internal/middleware/correlation_id.go
```diff
@@ -0,0 +1,28 @@
...
+		corrId := r.Header.Get("X-Correlation-Id")
+		if corrId != "" {
+			ctx = ctxval.WithCorrelationId(ctx, corrId)
+		}
+
+		// Store in response headers for easier debugging
+		w.Header().Set("X-Correlation-Id", corrId)
+		logger := ctxval.Logger(ctx).With().Str("correlation_id", corrId).Logger()
+		logger.Trace().Msgf("Added correlation id %s to logger", corrId)
```
lzap, line 22: Can you move this into the if statement above? This should not be executed when no header is present, otherwise this will add the following (empty) header to every single request:

    X-Correlation-Id:

internal/middleware/logger.go
```diff
@@ -21,13 +21,15 @@ func LoggerMiddleware(rootLogger *zerolog.Logger) func(next http.Handler) http.H
...
 				Str("trace_id", traceId).
 				Str("remote_ip", r.RemoteAddr).
 				Str("url", r.URL.Path).
 				Str("method", r.Method).
-				Int("bytes_in", bytesIn)
+				Int("bytes_in", bytesIn).
+				Str("correlation_id", corrId)
 			if edgeId != "" {
 				lctx = lctx.Str("request_id", edgeId)
 			}
```
lzap, line 35: Can you treat the newly added field the same way as edgeId? Meaning, when it is not present (blank string) we would not add the empty field into the logger.

lzap (changes requested):
Noticed few more things to make the experience cleaner.

lzap (approved):
Perfect, thank you!

PR #523, opened 2023-05-10
build: updgrade to Go 1.19.6

lzap:
SSIA

Also adds a CI check - if the Go on the CI does not match the required version, stop and fail all tests.

PR #524, opened 2023-05-10
refactor: use Gob encoding for cache

lzap:
The account cache currently use JSON encoding, but the newer version of the linter revealed that we removed JSON tags from models. Cache works, however, the linter error is still correct and I would like to keep this lint warning on.

Instead of adding JSON tags to models, I would like to change behavior of the cache, instead of JSON let’s use Gob (native Go binary encoding) which is not only faster, but it also does not require any tags. Redis is not very transparent and we unlikely need to investigate contents of cache (logging can do the same job if we need to).

The patch changes to Gob encoding, only account is currently stored in cache as a struct. Upgrade path: automatic. When previous cache entry is loaded, Gob decoding fails and NotFound error is returned thus the cache entry is overwritten (or expires).

See https://github.com/RHEnVision/provisioning-backend/pull/523 for the linter change.

This needs to be either merged first and `musttag` linter exception removed from the other patch or rebased and remove the `musttag` in this PR.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,7 +20,6 @@ linters:
     - gci # too strict
     - contextcheck # too restrictive
     - gomoddirectives # TODO remove after scany/tern v2 are released
-    - musttag # cache uses JSON for models (Account)
 linters-settings:
   govet:
     check-shadowing: true
--- a/internal/cache/redis_cache.go
+++ b/internal/cache/redis_cache.go
@@ -1,12 +1,14 @@
 package cache
 
 import (
+	"bytes"
 	"context"
-	"encoding/json"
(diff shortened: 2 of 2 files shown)


lzap:
Removed the `musttag` workaround after 1.19.6 update PR was merged. Rebased.

ezr-ondrej:
/retest
Some glitch in the pipeline

lzap:

ezr-ondrej (approved):

ezr-ondrej:
I do not fully get the reasoning TBH, but the change itself looks fine :+1:

lzap:
You will see the benefits in a moment :-)

PR #527, opened 2023-05-15
chore: update dashboard SLO latency to 500ms target

lzap:
SSIA

Also changed the defaults to 1h window last 24 hours which is a bit more relevant I think. Also faster.

The version dropped from 7 to 1 not sure why, I exported this via Grafana, I guess it is just informative number since it is managed by app-interface rather than Grafana's built-in versioning: https://grafana.com/docs/grafana/latest/dashboards/build-dashboards/manage-version-history/

### DIFF
--- a/deploy/Provisioning-1674548289785.json
+++ b/deploy/Provisioning-1674548289785.json
@@ -1001,13 +1001,13 @@
             "uid": "${datasource}"
           },
           "editorMode": "code",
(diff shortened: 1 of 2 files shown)


adiabramovitch:
You should use your  make target in order to change the dashboard :)

lzap:
Damn happens ALL the time to me! :-)

adiabramovitch (approved):

PR #529, opened 2023-05-15
build: drop GHA OpenAPI validation

lzap:
We use a GHA OpenAPI generation but then we call `make validate` to generate the same files. The use of the GHA is not needed actually.

In fact, it started failing our tests with a weird error recently:

```
Error:  Check the spelling of the generator's name and try again.
202
Error: Process completed with exit code 1.
```

This patch removes that, the `make` utility does perform generation and then `git diff` to check if there is any sort of change. This is all we need.

### DIFF
--- a/.github/workflows/gotest.yml
+++ b/.github/workflows/gotest.yml
@@ -103,9 +103,5 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: ${{ env.GO_SVR }}
-      - uses: openapi-generators/openapitools-generator-action@v1
-        with:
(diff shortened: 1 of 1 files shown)


adiabramovitch (approved):

lzap:
`Error: trying to reuse blob sha256:11af434db46f77441bdbab81c38e2a4bcd2796083a662e97f56781be401a76df at destination: Requesting bear token: invalid status code from registry 500 (Internal Server Error)`

But I am going to merge this, it fixes a GHA issue other PRs depends on.

PR #530, opened 2023-05-15
build: install go tools and cache

lzap:
I noticed that we do run `make validate` which is supposed to generate clients and then validate them. All targets are phony, except `generate-clients` which by default does not do anything since there is no change in the code base.

This patch adds the required `touch` call which causes the clients to be actually regenerated, then `git` is used to compare for changes. Since the OpenAPI generator version if pinned, it should not result in any difference. The tools must be installed from now on, it was working without them because actually nothing was happening.

In addition to that, the `bin` directory is cached so `go install-tools` will be only executed when necessary speeding up the validation.

In a separate commit, I upgraded GHA `setup-go@v4` to version 4 which automatically cache compiled code for quicker compilation. The key for the cache [is Go version and go.sum](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs) so this works out of box without any configuration.

### DIFF
--- a/.github/workflows/gotest.yml
+++ b/.github/workflows/gotest.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
           go-version: ${{ env.GO_SVR }}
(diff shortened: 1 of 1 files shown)


lzap:

ezr-ondrej (approved):
:orange_heart: it, perfect, let see the :rocket:

PR #531, opened 2023-05-15
chore: more otel for some slow calls

lzap:
Improving otel constants and adding more otel calls for sources and azure.

Also incorporated one small change from https://github.com/RHEnVision/provisioning-backend/pull/487 which I am going to close now.


ezr-ondrej (approved):
Thanks @lzap ! :orange_heart:

PR #532, opened 2023-05-15
fix(HMS-1782): cap job concurrency at 100

lzap:
Jobqueue concurrency was uncapped until now, this was a bug. I meant to cap it at 64 per pod, however, the setting was only for pollers which spawn new goroutine on each job received.

This patch removes processing in goroutines per job, every poller executes jobs in the same goroutine which leads to effective concurrency that can be capped by `WORKER_CONCURRENCY` variable (previously maxPollers which was confusing I think).

The default value is set to 33, the idea is that with 3 pods we are at concurrency of 100 (99 actually) which is easy to do math on. For example, AWS bucket for instance launch is 1000 so at "maximum speed" the service can withstand 100 seconds without throttling. We can update the number as we get access to CloudWatch AWS API logs and learn how we can scale the service.

### DIFF
--- a/config/api.env.example
+++ b/config/api.env.example
@@ -92,8 +92,8 @@
 #     	job worker implementation (memory, redis, sqs, postgres) (default "memory")
 #   WORKER_POLL_INTERVAL int64
 #     	polling interval (network timeout) (default "5s")
-#   WORKER_MAX_THREADS int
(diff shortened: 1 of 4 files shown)


lzap:

ezr-ondrej (approved):
Looks sane and much more predictable I guess :+1:

PR #533, opened 2023-05-16
fix(HMS-1784): set GOMAXPROC for API workers

lzap:
We currently leave the GOMACPROC set to default which is amount of cores + 1. In Kubernetes this is actually 3 which might be little too little. There is a library that tries to heuristically find the sweet spot. It also dumps the value to logs.

https://github.com/uber-go/automaxprocs


ezr-ondrej (approved):
I don't love magic libraries, but first look it looks safe, if you think it will help us, lets have it :shrug: :)

PR #534, opened 2023-05-16
fix(HMS-1800): refactor and add more caching

lzap:
This is complete refactoring of the caching package. Initially, there was a memory implementation, then Redis was added and also noop for tests - many interfaces and it was a bit over-engineered to my taste. Everything is now simplified: cache can be enabled (Redis) or disabled (no caching, the default). The API is vastly more simple with only few functions available and it makes use of Go generics as well as Gob binary encoding for seamless experience. It is now a type-safe cache that is easy to work with and adding a new model/type is just three lines change.

It started as adding support for upload info caching but the more I digged into the old design the more I wanted to refactor this from scratch. As part of one commit, this also adds caching for:

* Account Details for AWS (whole model)
* Azure Tenant ID (only the ID which is then used for upload info endpoint)

Finally, metrics were improved - alongside with hits and misses also errors are reported. I am going to create a new graph on the dashboard so we can take a look on cache success rate or error ratio. Maybe as part of this PR if I have chance.

Added also integration tests for Redis. No upgrade is needed, everytime a cache encounters a deserialization error, it is reported into logs and prometheus, but ErrNotFound is returned so the caller will effectively overwrite the old entry with new data.


internal/payloads/sources_payload.go
```diff
@@ -31,9 +31,9 @@ func NewListSourcesResponse(sourceList []*clients.Source) []render.Renderer {
 }
 
 type SourceUploadInfoResponse struct {
-	Provider  string                     `json:"provider"`
-	AwsInfo   *clients.AccountDetailsAWS `json:"aws" nullable:"true"`
-	AzureInfo *clients.AzureSourceDetail `json:"azure" nullable:"true"`
+	Provider  string                       `json:"provider"`
+	AwsInfo   *clients.AccountDetailsAWS   `json:"aws" nullable:"true"`
+	AzureInfo *clients.AccountDetailsAzure `json:"azure" nullable:"true"`
 }
```
lzap, line 37: Because the new cache key is based on generics and type infers the prefix, string build-in type cannot be used. So I introduced a special type here.

internal/cache/redis.go
```diff
@@ -0,0 +1,174 @@
...
+	// application id "constant" memory-only cache
+	appTypeId      *string
+	appTypeIdMutex sync.Mutex
+)
+
+const Forever time.Duration = 24 * time.Hour * 30
+
+type Cacheable interface {
+	models.Account | clients.AccountDetailsAWS | clients.AzureTenantId
+}
```
lzap, line 41: To add a new generic type to cache, change here.

internal/cache/redis.go
```diff
@@ -0,0 +1,174 @@
...
+	case *models.Account:
+		return "account", nil
+	case *clients.AccountDetailsAWS:
+		return "account_detail_aws", nil
+	case *clients.AzureTenantId:
+		return "azure_tenant_id", nil
+	default:
+		return "", ErrUnknownType
+	}
+}
```
lzap, line 92: To add a new generic type to cache, change here.

internal/cache/redis.go
```diff
@@ -0,0 +1,174 @@
...
+
+// nolint: wrapcheck
+func SetForever[T Cacheable](ctx context.Context, key string, value *T) error {
+	return SetExpires(ctx, key, value, Forever)
+}
+
+// nolint: wrapcheck
+func Set[T Cacheable](ctx context.Context, key string, value *T) error {
+	return SetExpires(ctx, key, value, config.Application.Cache.Expiration)
+}
```
lzap, line 174: The interface is simple:

* Find
* Set
* SetForever (with a very long expiration)
* SetExpires (with your own expiration duration)

I see I forgot to document these functions, will do during rebase.

internal/cache/redis.go
```diff
@@ -0,0 +1,174 @@
...
+
+func SetAppTypeId(_ context.Context, value string) error {
+	appTypeIdMutex.Lock()
+	defer appTypeIdMutex.Unlock()
+
+	appTypeId = &value
+	return nil
+}
+
+func Prefix[T Cacheable](value *T) (string, error) {
```
ezr-ondrej, line 81: Looks weird to use generics and then still switch by type.
Why is it better then having SetXXX function? :thinking:

lzap: Yeah I rebased the code to use interface normally instead of constraint. This is indeed more natural and better, good point.

lzap:
I believe

```
/usr/bin/tar -xf /home/runner/work/_temp/9d82feee-0932-4ee7-82bf-2f8f8b676785/cache.tgz -P -C /home/runner/work/provisioning-backend/provisioning-backend -z
  /usr/bin/tar: ../../../go/pkg/mod/go.opentelemetry.io/otel@v1.14.0/VERSIONING.md: Cannot open: File exists
  Error: /usr/bin/tar: ../../../go/pkg/mod/go.opentelemetry.io/otel@v1.14.0/propagation/trace_context_test.go: Cannot open: File exists
```

is caused by the fact this branch was branched off before linter/go caching upgrade. Rebased on `main` to solve that.

lzap:
Linter error solution: https://github.com/RHEnVision/provisioning-backend/pull/537

ezr-ondrej (approved):
Loving it :orange_heart:

PR #535, opened 2023-05-16
chore: add more default buckets

lzap:
I wanted to temporarily increase SLO alert to 1000ms until we push the caching of Azure to prod but it turns out we have 500ms and 5000ms, that is too large gap I guess. My bad, this adds two more LEs.

This will not help me with the current alerts situation but let’s fix this for the future.


internal/middleware/prometheus.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-var buckets = []float64{100, 200, 500, 5000}
+var buckets = []float64{100, 200, 500, 1000, 2000, 5000}
```
ezr-ondrej, line 17: Would 750 make sense here? given we are slightly over 500 on the offending endpoint, I guess it might be nice to have bit more granuality there, right?

PR #536, opened 2023-05-17
fix(HMS-1616): OpenAPI cleanup and examples

lzap:
Continuation of https://github.com/RHEnVision/provisioning-backend/pull/500 for all sources endpoints. I am keeping the deprecated endpoints untouched for now, let’s remove them in a separate PR.

I cannot continue working on upload_info until https://github.com/RHEnVision/provisioning-backend/pull/534 is merged tho, so that is TBD.

### DIFF
--- a/cmd/spec/example_source.go
+++ b/cmd/spec/example_source.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/payloads"
+)
+
+var SourceListResponse = []payloads.SourceResponse{{
+	ID:   "654321",
+	Name: "My AWS account",
+}, {
+	ID:   "543621",
+	Name: "My other AWS account",
+}}
+
+var SourceUploadInfoAWSResponse = payloads.SourceUploadInfoResponse{
+	Provider: "aws",
+	AwsInfo:  &clients.AccountDetailsAWS{AccountID: "78462784632"},
+}
+
+var SourceUploadInfoAzureResponse = payloads.SourceUploadInfoResponse{
+	Provider: "azure",
+	AzureInfo: &clients.AccountDetailsAzure{
+		TenantID:       "617807e1-e4e0-481c-983c-be3ce1e49253",
+		SubscriptionID: "617807e1-e4e0-4855-983c-1e3ce1e49674",
+		ResourceGroups: []string{"MyGroup 1", "MyGroup 42"},
+	},
+}
--- a/internal/payloads/sources_payload.go
+++ b/internal/payloads/sources_payload.go
@@ -7,11 +7,12 @@ import (
 	"github.com/go-chi/render"
 )
 
-type SourceID struct {
-	SourceId string `json:"source_id"`
-}
+// See clients.Source
 type SourceResponse struct {
-	*clients.Source
+	ID           string `json:"id" yaml:"id"`
+	Name         string `json:"name,omitempty" yaml:"name"`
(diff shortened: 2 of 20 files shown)


internal/payloads/sources_payload.go
```diff
@@ -7,11 +7,10 @@ import (
 	"github.com/go-chi/render"
 )
 
-type SourceID struct {
-	SourceId string `json:"source_id"`
-}
+// See clients.Source
 type SourceResponse struct {
```
ezr-ondrej, line 11: Would we be able to add the provider type here?
We apparently have consumer for that info and it might be helpful even in the UI in the future.

lzap: I would like to either do cleanup in this PR (removal of things) or nothing. I can probably put it back and create a ticket.

But whoever is using that number, it will change between prod/ephemeral and stage likely. It is a not good data source.

lzap: Created https://issues.redhat.com/browse/HMS-1829

This is now ready for review and merge, I want to do followup for reservations and I am done!

cmd/spec/example_source.go
```diff
@@ -0,0 +1,28 @@
...
+	ID:   "543621",
+	Name: "My other AWS account",
+}}
+
+var SourceUploadInfoAWSResponse = payloads.SourceUploadInfoResponse{
+	Provider: "aws",
+	AwsInfo:  &clients.AccountDetailsAWS{AccountID: "78462784632"},
+}
+
+var SourceUploadInfoAzureResponse = payloads.SourceUploadInfoResponse{
```
ezr-ondrej, line 21: I'm confused :thinking:
This generates example where we have key `azureinfo`, but I don't see where the type is defined, but schema defines a key `azure`, so there is discrepancy in the key names, but I don't see where it comes from. Are we using the same types for The example response and the schema generating?

akhil-jha (approved):

cmd/spec/example_source.go
ezr-ondrej: image

lzap: Yeah this is a typical error I have seen many times now:

```diff
diff --git a/internal/payloads/sources_payload.go b/internal/payloads/sources_payload.go
index cb3dd40..2da271e 100644
--- a/internal/payloads/sources_payload.go
+++ b/internal/payloads/sources_payload.go
@@ -37,9 +37,9 @@ func NewListSourcesResponse(sourceList []*clients.Source) []render.Renderer {
 }
 
 type SourceUploadInfoResponse struct {
-       Provider  string                       `json:"provider"`
-       AwsInfo   *clients.AccountDetailsAWS   `json:"aws" nullable:"true"`
-       AzureInfo *clients.AccountDetailsAzure `json:"azure" nullable:"true"`
+       Provider  string                       `json:"provider" yaml:"provider"`
+       AwsInfo   *clients.AccountDetailsAWS   `json:"aws" nullable:"true" yaml:"aws"`
+       AzureInfo *clients.AccountDetailsAzure `json:"azure" nullable:"true" yaml:"azure"`
 }
```

This should fix it, pushing.

lzap:
I noticed a TODO comment in `path.yaml` that was no longer relevant, rebased.

cmd/spec/example_source.go
ezr-ondrej: That helped, thanks! :rocket:

ezr-ondrej:
let me regenerate the spec for you, it's just the description and we are ready to :shipit:

ezr-ondrej:
Thanks!

PR #537, opened 2023-05-17
build: disable cache for golangci-lint

lzap:
CI currently fails with

```
/usr/bin/tar -xf /home/runner/work/_temp/9d82feee-0932-4ee7-82bf-2f8f8b676785/cache.tgz -P -C /home/runner/work/provisioning-backend/provisioning-backend -z
  /usr/bin/tar: ../../../go/pkg/mod/go.opentelemetry.io/otel@v1.14.0/VERSIONING.md: Cannot open: File exists
  Error: /usr/bin/tar: ../../../go/pkg/mod/go.opentelemetry.io/otel@v1.14.0/propagation/trace_context_test.go: Cannot open: File exists
```

This is because cache is stored twice, Go cache and Linter cache. The documentation for the linter explicitly disables the Go cache because it is essentially stored twice. This patch does exactly that.

During editing of the workflow I realized that we do not build the artifacts at all. So there can be a build error in `main.go` and we would not notice until container would fail to build. This adds a new job for that. Unfortunately, build cache for the Go action cannot be shared between build, test and integration test via `setup-go`, so I disabled it for now. There is a workaround but that is a lot of YAML I want rather to keep it simple.

### DIFF
--- a/.github/workflows/gotest.yml
+++ b/.github/workflows/gotest.yml
@@ -30,14 +30,51 @@ jobs:
           make check-commits
         working-directory: code
 
-  linting:
-    name: "🎯 Code format, imports and style"
+  build:
+    name: "🛠️ Go Build"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
         with:
           go-version: ${{ env.GO_SVR }}
+          cache: true
+      - name: Build all artifacts
+        run: |
+          make check-system-go build GO=go
+
+  lint:
+    name: "🎯 Code format & OpenAPI lint"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_SVR }}
+          # disabled until 'lookup-only' is implemented for setup-go or cache prefix is added
+          cache: false
+      - uses: actions/cache@v3
+        id: cache
+        with:
+          path: bin
+          key: bin-go${{ env.GO_SVR }}-tools-${{ hashFiles('mk/tools.mk') }}
+      - name: Install Go Tools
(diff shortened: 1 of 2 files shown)


lzap:
The test "Code format, imports and style" will not pass until this is merged, I renamed jobs for better consistency.

lzap:
Interesting observation:

The [golangci-lint action](https://github.com/golangci/golangci-lint-action) uses its own cache for both Go and linter themselves. On every run, cache is reused (downloaded), used and then uploaded back to the cache. It grows over time (e.g. when dependencies are bumped) the starting point being 390 MB and I saw it as big as 750 MB recently. Every 7 days the cache is deleted and new one created.

When there is no cache (or cache was deleted), golangci can actually timeout (5 minutes limit of GHA), but the cache is uploaded even after a failed run, therefore the subsequent job will likely pass as most of the files will be already in the cache.

As much as I would like to use our own golangci-lint from the `bin` directory, this is quite useful and maintaining this would probably take too much time, so let’s continue using that for now.

ezr-ondrej (approved):
Looks like it works, the reasoning make sense, I do not understand all the caches, but lets merge and see :eyes: it :)

lzap:
Just to explain and elaborate what is going on:

Action setup-go/v4 (which we upgraded to recently) has go package cache and it is enabled by default. It automatically caches compiled Go until `go.mod` is changed. With this patch, we only use this in the newly introduced "Build Go" job. Unfortunately, test and integration test cannot utilize the cache easily, so it is turned off. This is reported: [one report](https://github.com/actions/setup-go/issues/316) and [other report](https://github.com/actions/setup-go/issues/358).

Action golangci-lint uses its own cache that contains Go libraries plus linter parser cache. It collides with Go cache when it was turned on thus the error. This fixes it. It grows over time and the strategy for this action is to clean it every 7 days.

Hope this makes it more clear.

ezr-ondrej:
It does, even I understand it! :)

PR #539, opened 2023-05-18
feat(HMS-#1804): add compose file for containerizing dev env

amirfefer:
a `docker-compose.yml` file help you to roll up the Provisioning application, including frontend and other services such as sources locally for dev or demo with no extra setup.

### Install
a docker or podman (with [podman-compose](https://github.com/containers/podman-compose)) is needed, the folder structure should be:
```
.
├── provisioning-backend
├── provisioning-frontend
├── sources-api-go
└── image-builder-frontend
```
Run 
```sh
$ COMPOSE_PROFILES=migrate docker compose up 
```

### Profiles
a compose profile allows you to run a subset of containers. When no profile is given, only provisioning-backend, frontend, postgres and redis will run, in this case use staged sources.

Currently there are a few profiles:
- migrate: migrate provisioning backend, terminates after migration
- kafka: run kafka with zookeeper, register topics
- sources: run local sources with postgres db, on first use notice that you will need to run `/script/sources.seed.sh` for seeding your local sources data.

For running sources and kafka profiles, run
```sh
$ COMPOSE_PROFILES=kafka,sources docker compose up 
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ TAGS
 # envdir/envrc (https://direnv.net)
 /.envrc
 /bin
+/compose-data/*
--- a/build/Dockerfile.dev
+++ b/build/Dockerfile.dev
@@ -1,0 +1,15 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.19
+USER 0
+ENV PROJECT_DIR=/backend \
+    GO111MODULE=on \
+    CGO_ENABLED=0
+
+WORKDIR /backend
+RUN mkdir "/build"
+
+COPY . .
+RUN go get github.com/githubnemo/CompileDaemon
+RUN go install github.com/githubnemo/CompileDaemon
+RUN make prep GO=go
+
+ENTRYPOINT $HOME/go/bin/CompileDaemon -build=" go build -o /build/pbackend ./cmd/pbackend" -command="/build/pbackend api"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,174 @@
+version: '3.7'
+services:
+  provisioning-db:
+    restart: on-failure
+    image: docker.io/postgres:13.1
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
+      interval: 2s
+      timeout: 3s
+      retries: 10
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=provisioning
+    volumes:
+      - ./compose-data/postgres/backend:/var/lib/postgresql/data
+    expose:
+      - 5432
+    
+  redis:
+    image: registry.redhat.io/rhel8/redis-6
+    healthcheck:
+      test: [ "CMD", "redis-cli", "ping" ]
+      interval: 2s
+      timeout: 2s
+      retries: 10
+    expose:
+      - 6379
+    volumes:
+      - ./compose-data/redis:/data
+  
+  zookeeper:
+    profiles:
+      - "kafka"
+    container_name: zookeeper
+    image: quay.io/strimzi/kafka:latest-kafka-3.4.0
+    command: [
+      "sh", "-c",
+      "bin/zookeeper-server-start.sh config/zookeeper.properties"
+    ]
+    expose:
+      - "2181"
+    environment:
+      LOG_DIR: /tmp/logs
+    volumes:
+      - ./compose-data/zookeeper:/tmp/logs
+  
+  kafka:
+    profiles:
+     - "kafka"
+    image: quay.io/strimzi/kafka:latest-kafka-3.4.0
+    command: [
+      "sh", "-c",
+      "bin/kafka-server-start.sh config/server.properties  --override inter.broker.listener.name=$${KAFKA_INTER_BROKER_LISTENER_NAME} --override listener.security.protocol.map=$${KAFKA_LISTENER_SECURITY_PROTOCOL_MAP} --override listeners=$${KAFKA_LISTENERS} --override advertised.listeners=$${KAFKA_ADVERTISED_LISTENERS} --override zookeeper.connect=$${KAFKA_ZOOKEEPER_CONNECT}"
+    ]
+    expose:
+      - 9092
+      - 29092
+    depends_on:
+      - zookeeper
+    environment:
+      KAFKA_AUTO_CREATE_TOPICS_ENABLE: true
+      LOG_DIR: "/tmp/logs"
+      KAFKA_BROKER_ID: 1
+      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+      KAFKA_INTER_BROKER_LISTENER_NAME: DOCKER
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: DOCKER:PLAINTEXT,LOCALHOST:PLAINTEXT
+      KAFKA_LISTENERS: DOCKER://kafka:29092,LOCALHOST://localhost:9092
+      KAFKA_ADVERTISED_LISTENERS: DOCKER://kafka:29092,LOCALHOST://localhost:9092
+    volumes:
+      - ./compose-data/kafka:/tmp/logs
+
+  init-kafka:
+    profiles:
+      - "kafka"
+    image: quay.io/strimzi/kafka:latest-kafka-3.4.0
+    entrypoint: [ '/bin/sh', '-c' ]
+    command: |
+      "
+      # blocks until kafka is reachable
+      bin/kafka-topics.sh --bootstrap-server kafka:29092 --list
+
+      echo -e 'Creating kafka topics'
+      bin/kafka-topics.sh --bootstrap-server kafka:29092 --create --if-not-exists --topic platform.sources.status --replication-factor 1 --partitions 1
+      bin/kafka-topics.sh --bootstrap-server kafka:29092 --create --if-not-exists --topic platform.provisioning.internal.availability-check --replication-factor 1 --partitions 1
+
+      echo -e 'Successfully created the following topics:'
+      bin/kafka-topics.sh --bootstrap-server kafka:29092 --list
+      "
+    depends_on:
+      - kafka
+      - zookeeper
+  
+  backend:
+    env_file:
+      - scripts/compose/api.compose.env
+    depends_on:
+      migrate:
+        condition: service_completed_successfully
+      provisioning-db:
+        condition: service_healthy
+      init-kafka:
+        condition: service_completed_successfully
+    build:
+      context: .
+      dockerfile: build/Dockerfile.dev
+    ports:
+      - ${APP_PORT:-8000}:${APP_PORT:-8000}
+      - ${PROMETHEUS_PORT:-9000}:${PROMETHEUS_PORT:-9000}
+    volumes:
+      - ./:/backend
+
+  migrate: # This service is used to run migrations
+    profiles:
+      - "migrate"
+    build:
+      context: .
+      dockerfile: build/Dockerfile.dev
+    env_file:
+      - config/api.env
+    depends_on:
+      provisioning-db:
+        condition: service_healthy
+    entrypoint: [go,run,/build/cmd/pbackend migrate]
+
+  frontend:
+    profiles:
+      - "frontend"
+    environment:
+      - PROV_API_HOST=backend
+      - WATCHPACK_POLLING=true
+    build:
+      context: ../provisioning-frontend
+      dockerfile: Dockerfile.dev
+    volumes:
+      - ../provisioning-frontend/:/prov-frontend
+    ports: 
+      - 1337:1337
+    depends_on:
+    - backend
+
+  sources:
+    profiles:
+      - "sources"
+    build:
+      context: ../sources-api-go
+    ports: 
+      - 8131:8131
+    env_file:
+      - scripts/compose/sources.compose.env
+    depends_on:
+      redis:
+        condition: service_healthy
+      sources-db:
+        condition: service_healthy
+  sources-db:
+    restart: on-failure
+    profiles:
+      - "sources"
+    image: docker.io/postgres:13.1
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
+      interval: 2s
+      timeout: 2s
+      retries: 10
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=sources_devel
+    volumes:
+      - ./compose-data/postgres/sources:/var/lib/postgresql/data
+    expose:
+      - 5432
--- a/docs/dev-environment.md
+++ b/docs/dev-environment.md
@@ -126,6 +126,52 @@ Tip: Alternatively, the application supports connecting to the stage environment
 
 Because Image Builder is more complex for installation, we do not recommend installing it on your local machine right now. Configure connection through HTTP proxy to the stage environment in `config/api.env`. See [configuration example](../config/api.env.example) for an example, you will need to ask someone from the company for real URLs for the service and the proxy.
 
+## Containerized environment
+
+A `docker-compose.yml` file helps rolling up the provisioning application, including frontend and other services such as sources locally for development purpose or demo with no extra setup.
+
+Please notice that the compose file use a dedicated dev Dockerfile.dev both for backend and frontend, it uses [CompileDaemon](github.com/githubnemo/CompileDaemon) for live reloading, it watches for changes and re-build using `go build` command when a change occurs, no need to build the container after code changes.
+
+### Install
+A docker or podman (with [podman-compose](https://github.com/containers/podman-compose)) is needed, the folder structure should be:
+```
+.
+├── provisioning-backend
+├── provisioning-frontend
+├── sources-api-go
+└── image-builder-frontend
+```
+
+Edit [app.env](/config/api.env.example) to fit containerized services (i.e db, redis, kafka), these are not exposed directly to your localhost ports.
+
+Run 
+```sh
+$ COMPOSE_PROFILES=migrate docker compose up 
+```
+This command also migrates data to postgres db, using the `migrate` profile.
+
+Alternatively do:
+```sh
+$ COMPOSE_PROFILES=migrate podman-compose up 
+```
+
+
+### Profiles
+A compose profile allows you to run a subset of containers. When no profile is given, 
+the provisioning backend, postgres and redis will run by default.
+
+Currently there are a few profiles:
+- migrate: migrate provisioning backend, terminates after migration
+- kafka: run kafka with zookeeper, register topics
+- frontend: run local provisioning frontend
+- sources: run local sources with postgres db, on first use notice that you will need to run `/script/sources.seed.sh` for seeding your local sources data.
+
+For example, in order to run sources, kafka and frontend profiles, run
+```sh
+# using docker
+$ COMPOSE_PROFILES=frontend,kafka,sources docker compose up 
+```
+
 ## Writing Go code
 
 Ready to write some Go code? Read [contributing guide](../CONTRIBUTING.md).
--- a/scripts/compose/api.compose.env
+++ b/scripts/compose/api.compose.env
@@ -1,0 +1,9 @@
+# Minimum required variables for compose file
+# This file is used for local development and testing
+# Edit this file to match your environment and rename it for safety
+
+KAFKA_BROKERS="kafka:29092"
+KAFKA_ENABLED="true"
+REST_ENDPOINTS_SOURCES_URL="http://sources:8131/api/sources/v3.1"
+DATABASE_PASSWORD="postgres"
+DATABASE_HOST='provisioning-db'
--- a/scripts/compose/sources.compose.env
+++ b/scripts/compose/sources.compose.env
@@ -1,0 +1,17 @@
+# Minimum required variables for compose file
+# This file is used for local development and testing
+# It is not used for production deployment
+
+PORT=8131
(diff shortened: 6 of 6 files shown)


build/Dockerfile.dev
```diff
@@ -0,0 +1,17 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.19 as build
```
amirfefer, line 1: I created a dev dockerfile for improving the go modules caching and for future improvments such as hot reloading support.

docker-compose.yml
```diff
@@ -0,0 +1,150 @@
...
+      kafka-topics --bootstrap-server kafka:29092 --create --if-not-exists --topic platform.sources.status --replication-factor 1 --partitions 1
+      kafka-topics --bootstrap-server kafka:29092 --create --if-not-exists --topic platform.provisioning.internal.availability-check --replication-factor 1 --partitions 1
+
+      echo -e 'Successfully created the following topics:'
+      kafka-topics --bootstrap-server kafka:29092 --list
+      "
+    depends_on:
+      - kafka
+      - zookeeper    
+  backend:
```
amirfefer, line 76: we can add the latest images from quay as well

build/Dockerfile.dev
```diff
@@ -0,0 +1,17 @@
...
+COPY . .
+RUN make build strip GO=go
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+COPY --from=build /build/pbapi /pbapi
+COPY --from=build /build/pbworker /pbworker
+COPY --from=build /build/pbstatuser /pbstatuser
+COPY --from=build /build/pbmigrate /pbmigrate
+USER 1001
+CMD ["/pbapi"]
```
lzap, line 17: I think for development purposes, you do not need to use ubi-minimal at all. In fact, it is counter productive as you get rid of the Go compiler cache which is one of the reasons why Go compiler is very fast. Also you can take advantage of `go run` command. Something like (pseudo code):

```
FROM registry.access.redhat.com/ubi9/go-toolset:1.19 as build
USER 0
RUN mkdir /build
WORKDIR /build
COPY . .
RUN make prep GO=go
CMD ["go", "run", "/build/cmd/pbapi/main.go"]
```

This way you can easily just add new and new commits on top of this single container, everytime you do a change the subsequent "go run" will be fast as files will be already pre-compiled. Also you can change the entrypoint easily for worker or statuser.

build/RedHat.dockerfile
```diff
@@ -2,8 +2,11 @@ FROM registry.access.redhat.com/ubi9/go-toolset:1.19 as build
...
 WORKDIR /build
+# cache dependencies
+ADD go.mod go.sum /build/
+RUN go mod download
 COPY . .
-RUN make prep build strip GO=go
+RUN make build strip GO=go
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 COPY --from=build /build/pbapi /pbapi
```
lzap, line 12: Why are you changing the production dockerfile when you have a "dev" dockerfile?

docker-compose.yml
```diff
@@ -0,0 +1,150 @@
...
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=provisioning
+    volumes:
+      - ./postgres-data:/var/lib/postgresql/data
+    ports:
+      - 5432:5432
+    
+  redis:
+    image: docker.io/redis:4.0.11
```
lzap, line 21: Can you use the official red hat container [from RH registry](https://catalog.redhat.com/software/containers/search) instead?

docker-compose.yml
lzap: I think this could make things little bit too complex, I am fine doing dev for now.

docker-compose.yml
```diff
@@ -0,0 +1,150 @@
...
+      - "migrate"
+    build:
+      context: .
+      dockerfile: build/Dockerfile.dev
+    env_file:
+      - config/api.env
+    depends_on:
+      provisioning-db:
+        condition: service_healthy
+    entrypoint: [/pbmigrate]
```
lzap, line 103:
```suggestion
    entrypoint: ["go", "run", "/build/pbmigrate]
```

if you would stick to my suggestion to use Go run instead compilation.

docker-compose.yml
```diff
@@ -0,0 +1,150 @@
...
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=sources_devel
+    volumes:
+      - ./postgres-data-sources:/var/lib/postgresql/data
+    expose:
+      - 5432
+
+volumes:
+  go-modules: # This volume is used to cache go modules
```
lzap, line 150: Where does this volume go? Can we have it in the working directory and `.gitignore` just like postgres? I would very much like to have "everything in one place" so I can clean it up if needed.

docs/dev-environment.md
```diff
@@ -126,6 +126,38 @@ Tip: Alternatively, the application supports connecting to the stage environment
 
 Because Image Builder is more complex for installation, we do not recommend installing it on your local machine right now. Configure connection through HTTP proxy to the stage environment in `config/api.env`. See [configuration example](../config/api.env.example) for an example, you will need to ask someone from the company for real URLs for the service and the proxy.
 
+## Containerized environment
+
+a `docker-compose.yml` file help you to roll up the Provisioning application, including frontend and other services such as sources locally for dev or demo with no extra setup.
```
lzap, line 131:
```suggestion
A `docker-compose.yml` file help you to roll up the Provisioning application, including frontend and other services such as sources locally for dev or demo with no extra setup.
```

docs/dev-environment.md
```diff
@@ -126,6 +126,38 @@ Tip: Alternatively, the application supports connecting to the stage environment
...
+
+### Install
+a docker or podman (with [podman-compose](https://github.com/containers/podman-compose)) is needed, the folder structure should be:
```
lzap, line 134:
```suggestion
A docker or podman (with [podman-compose](https://github.com/containers/podman-compose)) is needed, the folder structure should be:
```

docs/dev-environment.md
```diff
@@ -126,6 +126,38 @@ Tip: Alternatively, the application supports connecting to the stage environment
...
+├── sources-api-go
+└── image-builder-frontend
+```
+Run 
+```sh
+$ COMPOSE_PROFILES=migrate docker compose up 
+```
+
+### Profiles
+a compose profile allows you to run a subset of containers, when no profile is given,
```
lzap, line 148:
```suggestion
A compose profile allows you to run a subset of containers, when no profile is given, 
```

docs/dev-environment.md
```diff
@@ -126,6 +126,38 @@ Tip: Alternatively, the application supports connecting to the stage environment
...
+```
+
+### Profiles
+a compose profile allows you to run a subset of containers, when no profile is given, 
+only provisioning backend, frontend, postgres and redis will run, in this case use staged sources.
+
+Currently there are a few profiles:
+- migrate: migrate provisioning backend, terminates after migration
+- kafka: run kafka with zookeeper, register topics
+- sources: run local sources with postgres db, on first use notice that you will need to run `/script/sources.seed.sh` for seeding your local sources data.
```
lzap, line 154: I would appreciate also a profile for frontend and backend. I will use this to run all the backend-backend services except backend

.gitignore
```diff
@@ -39,3 +39,5 @@ TAGS
 # envdir/envrc (https://direnv.net)
 /.envrc
 /bin
+/postgres-data
+/postgres-data-sources
```
lzap, line 43: I suggest to create a tree of directories for this:

```
/compose-data/postgres/backend
/compose-data/postgres/sources
/compose-data/redis/backend
/compose-data/go-cache
```

lzap (changes requested):
Nice, please explain the change of `Dockerfile.redhat` I don’t think it is necessary since you introduced a "dev" dedicated new Dockerfile. It can also be simplified, no need to use build and runtime containers. Finally, configure the compose Go cache volume into a directory the same way as Postgres.

build/RedHat.dockerfile
amirfefer: I changed it before adding the dev dockerfile... :)

build/Dockerfile.dev
amirfefer: > I think for development purposes, you do not need to use ubi-minimal at all. 

Makes sense 👍  a tiny tweak i've added is using [CompileDaemon](github.com/githubnemo/CompileDaemon) for deamomzing go and allowing live reloading when .*go files have change, saves tone of time during dev.

docker-compose.yml
```diff
@@ -0,0 +1,150 @@
+version: '3.7'
+services:
+  provisioning-db:
+    restart: on-failure
+    image: docker.io/postgres:10.5
```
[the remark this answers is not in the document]
amirfefer: I tried to use [RH postgresql]( https://catalog.redhat.com/software/containers/rhel8/postgresql-12/5db133bd5a13461646df330b)  but it  differ from the official Postgres dev image. For example, you can't use postgres user (only for interal usage), the env variables are different, and the default user (defiend by POSTGRESQL_USER) doesn't have poweruser permissions by default, unlike the official postgres image  - https://hub.docker.com/_/postgres

With no poweruser permission our migration fails, due to creating extension for `uuid-ossp`, I guess this is the reason why IB and sources use the official postgres image  for dev puporse.

docker-compose.yml
amirfefer: works 👍

docker-compose.yml
```diff
@@ -0,0 +1,150 @@
...
+      echo -e 'Creating kafka topics'
+      kafka-topics --bootstrap-server kafka:29092 --create --if-not-exists --topic platform.sources.status --replication-factor 1 --partitions 1
+      kafka-topics --bootstrap-server kafka:29092 --create --if-not-exists --topic platform.provisioning.internal.availability-check --replication-factor 1 --partitions 1
+
+      echo -e 'Successfully created the following topics:'
+      kafka-topics --bootstrap-server kafka:29092 --list
+      "
+    depends_on:
+      - kafka
+      - zookeeper
```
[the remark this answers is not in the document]
amirfefer: AMQ streams looks super cool, but it looks overkill to our dev needs, it simplfies the setup under Kubernetes env, because it built for it, for dev compose setup the configration seems harder than using kafka and zookeper.

I switched to Strimzi though 👍

docker-compose.yml
amirfefer: let's add `go-modules`  caching later on :) I removed it for now

amirfefer:
Thanks @lzap for reviewing this up

I added also live-reloading feature for backend and hot-reloading for frontend, therefore once containers are up, developers do not need to re-build them, code change is watched.

backend:

https://github.com/RHEnVision/provisioning-backend/assets/11807069/dbc33739-1ef0-455a-96ca-1f8f02cb4cc8


frontend:

https://github.com/RHEnVision/provisioning-backend/assets/11807069/679be6aa-44ba-4020-95a6-dc4fca75586a

docs/dev-environment.md
amirfefer: frontend profile is added :)

build/Dockerfile.dev
lzap: Not a big fan of that as our binary is 70MB and compilation is relatively slow but I am fine with this :-)

lzap (changes requested):
Looks good, does not work tho as I have Postgres, Redis and Kafka running on the host machine. Please use EXPOSE instead of PORT statements to prevent from exporting the ports to the host machine, it is not needed for backend services except our frontend and backend (which you want to check with curl).

For debugging you can always install curl and shell to the container to perform a test request. This will allow me to run this even on the host which has ports bound already.

build/Dockerfile.dev
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /backend
+RUN mkdir "/build"
+
+COPY . .
+RUN go get github.com/githubnemo/CompileDaemon
+RUN go install github.com/githubnemo/CompileDaemon
+RUN make prep GO=go
+
+ENTRYPOINT $HOME/go/bin/CompileDaemon -build=" go build -o /build/pbackend ./cmd/pbackend" -command="/build/pbackend api"
```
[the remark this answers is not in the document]
amirfefer: we can add env variable `LIVE_RELOAD` or something later on

amirfefer:
Thanks @lzap, fixed

- services ports are not exposed to localhost
- Using enviroment variable for backend port
- expsoing prometheus as well

docker-compose.yml
```diff
@@ -0,0 +1,167 @@
...
+    environment:
+      - PROV_API_HOST=backend
+      - WATCHPACK_POLLING=true
+    build:
+      context: ../provisioning-frontend
+      dockerfile: Dockerfile.dev
+    volumes:
+      - ../provisioning-frontend/:/prov-frontend
+    ports: 
+      - 1337:1337
```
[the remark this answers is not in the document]
amirfefer: this is the frontend port, it should be visible through localhost

lzap (changes requested):
Great I am testing on podman on a fresh machine with fresh checkouts, noticed this:

```
Trying to pull registry.redhat.io/rhel8/redis-6:latest...
Error: initializing source docker://registry.redhat.io/rhel8/redis-6:latest: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication
exit code: 125
```

Looks like you need a RHEL machine with subscription when using registry.redhat.io can you switch this to upstream instead? Unfortunately I don’t know where upstream for redhat container is, there are some old versions of Redis on https://registry.fedoraproject.org and quay.io is pretty hard to search through, si I am fine if you simply fallback to just the official docker image from Docker Hub.. Apologies I asked you to use downstream, I haven’t realized subscriptions are needed.

The same problem is with `api.env`, you mention that user should "edit it" but you don’t specify how. Isn’t it better to simply configure the application via env variables directly from compose file like you do for APP_PORT and others? I mean, users will likely have some weird customizations in this file and it will break compose. You know the hostnames and ports so just configure the app so user do not need to configure anything else. Of course enable trace logging and the usual development setup stuff. Here are all possible options (don’t add them all of course only those which are relevant): https://github.com/RHEnVision/provisioning-backend/blob/main/config/api.env.example

Also noticed an error when `sources.local.conf` file is missing. This could be the same as above - figure out some sane defaults. However, I see that sources are completely unconfigured - this is done in `sources.conf` file in our git repo. Sources will not start until you have some key options like BYPASS_RBAC or REDIS_CACHE_HOST.

Unfortunately this does not work with podman, kafka fails to initialize for me. I am not sure what is wrong, when I try profiles to only start kafka, it also starts sources and backend it is weird.

.gitignore
```diff
@@ -41,3 +41,4 @@ TAGS
 # envdir/envrc (https://direnv.net)
 /.envrc
 /bin
+/compose-data/*
```
lzap, line 44: This could be named just "data" in the new git repo.

I noticed one thing, when you use podman as non-root, postgres will change permissions of the directory to container user (random uid) and 600 permissions which makes the directory undeletable for the hosting user. It does not seem to be a problem for podman itself. There is a nice trick to fix this:

    podman unshare chmod 777 data/*

Let’s just document this.

build/Dockerfile.dev
lzap, line 15: This can be moved into the root folder of the new repository. Maybe name it `Dockerfile.backend.dev` so we can add more dockerfiles if needed. Consider adding `Dockerfile.frontend.dev` as well because I think it is beneficial to have everything under one roof, a frontend user might try to build container themselves in "provisioning-frontend" with a bad context and without documentation.

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+      timeout: 3s
+      retries: 10
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=provisioning
+    volumes:
+      - ./compose-data/postgres/backend:/var/lib/postgresql/data
+    expose:
+      - 5432
```
lzap, line 18: You name this "provisioning-db" and dependnency is named "backend". I suggest to pick be consistent, pick one or the other, for example "backend" and "backend-db".

Also can you give this the "backend" tag? It will be very useful to have tags for every single service, I will be likely using this to run just sources or sometimes just notifications etc.

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+    image: registry.redhat.io/rhel8/redis-6
+    healthcheck:
+      test: [ "CMD", "redis-cli", "ping" ]
+      interval: 2s
+      timeout: 2s
+      retries: 10
+    expose:
+      - 6379
+    volumes:
+      - ./compose-data/redis:/data
```
lzap, line 30: This needs to login, can you fallback to upstream? Also add appropriate tags, this works for me:

```yaml
  redis:
    profiles:
      - "redis"
      - "sources"
      - "backend"
    image: docker.io/redis:latest
    healthcheck:
      test: [ "CMD", "redis-cli", "ping" ]
      interval: 2s
      timeout: 2s
      retries: 10
    expose:
      - 6379
    volumes:
      - ./data/redis:/data
```

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+      echo -e 'Successfully created the following topics:'
+      bin/kafka-topics.sh --bootstrap-server kafka:29092 --list
+      "
+    depends_on:
+      - kafka
+      - zookeeper
+  
+  backend:
+    env_file:
+      - scripts/compose/api.compose.env
```
lzap, line 97: This can be moved to the new git repo root. Maybe rename this to just `backend.env`.

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+      init-kafka:
+        condition: service_completed_successfully
+    build:
+      context: .
+      dockerfile: build/Dockerfile.dev
+    ports:
+      - ${APP_PORT:-8000}:${APP_PORT:-8000}
+      - ${PROMETHEUS_PORT:-9000}:${PROMETHEUS_PORT:-9000}
+    volumes:
+      - ./:/backend
```
lzap, line 112: This needs to be updated in the new git repo, it is no longer a git checkout. Feel free to use `../provisioning-backend` for now, tho, I would prefer git submodules for this. We can refactor later.

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+    depends_on:
+      migrate:
+        condition: service_completed_successfully
+      provisioning-db:
+        condition: service_healthy
+      init-kafka:
+        condition: service_completed_successfully
+    build:
+      context: .
+      dockerfile: build/Dockerfile.dev
```
lzap, line 107: Path needs to be updated here in the new git repo.

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+      - WATCHPACK_POLLING=true
+    build:
+      context: ../provisioning-frontend
+      dockerfile: Dockerfile.dev
+    volumes:
+      - ../provisioning-frontend/:/prov-frontend
+    ports: 
+      - 1337:1337
+    depends_on:
+    - backend
```
lzap, line 141: If you would use git submodules, frontend can be also here in this repo so you would need update paths here.

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+      context: ../sources-api-go
+    ports: 
+      - 8131:8131
+    env_file:
+      - scripts/compose/sources.compose.env
+    depends_on:
+      redis:
+        condition: service_healthy
+      sources-db:
+        condition: service_healthy
```
lzap, line 156: I would also suggest using git submodules for sources (paths need an update). Can you introduce new profile named "sources-dev" and rename this service tag to that? Then also add this:

```yaml
 sources:
    profiles:
      - "sources"
    container_name: sources
    image: quay.io/cloudservices/sources-api-go:latest
    env_file:
      - sources.env
    expose:
      - 8131
    depends_on:
      redis:
        condition: service_healthy
      sources-db:
        condition: service_healthy
```

After you do `podman login quay.io` this will work so people can simply use the official sources. You can do the same for notifications or other containers as well. Users can decide the profile they want to use any time. It can share the same database that is actually beneficial.

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+  sources-db:
+    restart: on-failure
+    profiles:
+      - "sources"
```
lzap, line 160: Add both "sources" and "sources-dev" profiles here.

docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+      - "migrate"
+    build:
+      context: .
+      dockerfile: build/Dockerfile.dev
+    env_file:
+      - config/api.env
+    depends_on:
+      provisioning-db:
+        condition: service_healthy
+    entrypoint: [go,run,/build/cmd/pbackend migrate]
```
lzap, line 125: Kafka is named "init-kafka", maybe this could be also named "init-backend". Similarily, kafka initialization is in the "kafka" profile, I think this could also be in the "backend" profile too, there is no need to have this separately migration is very quick.

Docker and env file paths needs to be updated, see above.

lzap (approved):
Okay I was testing this with SELinux on and there were some issues. It works with the latest version of podman-compose and permissive mode:

    pip3 install https://github.com/containers/podman-compose/archive/devel.tar.gz

I am dropping some more minor comments, can you move this to the newly created repo: https://github.com/RHEnVision/provisioning-compose that is dedicated just to the compose? It can also be a git submodule for all other projects you want to use (sources, backend, frontend). This way we can even "freeze" versions which are known to work.

Feel free to incorporate those nit pick and directly push into the newly created repo, I think this is good enough so we can start using it and improving. Feel free to add notifications as well.

Thanks this is looking great.

lzap:
Moved to https://github.com/RHEnVision/provisioning-compose/pull/1 ty

PR #541, opened 2023-05-18
fix(HMS-1785): check if public ip is nil during dereference

amirfefer:


internal/clients/http/ec2/ec2_client.go
```diff
@@ -445,8 +445,8 @@ func (c *ec2Client) parseDescribeInstances(respAWS *ec2.DescribeInstancesOutput)
 	for i, instance := range instances {
 		list[i] = &clients.InstanceDescription{
 			ID:         *instance.InstanceId,
-			PublicIPv4: *instance.PublicIpAddress,
-			PublicDNS:  *instance.PublicDnsName,
+			PublicIPv4: ptr.FromOrEmpty(instance.PublicIpAddress),
+			PublicDNS:  ptr.FromOrEmpty(instance.PublicDnsName),
```
amirfefer, line 449: better safe than sorry?

PR #542, opened 2023-05-19
fix(HMS-1800): fix a typo in upload info

lzap:
The provider field was hardcoded to "azure" for some reason. I noticed this today.

```
{"provider":"azure","aws":{"account_id":"399777895069"},"azure":null}
```

### DIFF
--- a/internal/cache/redis.go
+++ b/internal/cache/redis.go
@@ -118,12 +118,13 @@ func Find(ctx context.Context, key string, value Cacheable) error {
 	err = dec.Decode(value)
 	if err != nil {
 		// decode error can be thrown if previous cache entry was JSON-encoded, return not found to overwrite it
-		ctxval.Logger(ctx).Warn().Err(err).Bool("cache", true).Msgf("redis cache decode error: %s", err.Error())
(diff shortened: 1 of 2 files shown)


internal/cache/redis.go
```diff
@@ -118,12 +118,13 @@ func Find(ctx context.Context, key string, value Cacheable) error {
...
 		// decode error can be thrown if previous cache entry was JSON-encoded, return not found to overwrite it
-		ctxval.Logger(ctx).Warn().Err(err).Bool("cache", true).Msgf("redis cache decode error: %s", err.Error())
+		ctxval.Logger(ctx).Warn().Err(err).Bool("cache", true).Msgf("Redis cache decode error: %s", err.Error())
 		metrics.IncCacheHit(prefix, "err")
 		return ErrNotFound
 	}
 
 	metrics.IncCacheHit(prefix, "hit")
+	ctxval.Logger(ctx).Trace().Bool("cache", true).Msgf("Cache hit for key '%s' type %T", key, value)
 	return nil
```
lzap, line 128: I think it is worth tracing these hits just to be able to validate in case there is something wrong with the cache. Caching is hard!

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #546, opened 2023-05-24
fix(HMS-1830): improve concurrency of account creation

lzap:
We made a conscious decision not to use transactions when creating accounts because it would be too slow. There are cases when user that don’t have account created in our database yet comes and performs operation that does many requests, this can lead to this error.

We now saw this once in production, so let’s fix this. Instead of using transactions, which are slow, this patch uses "UPSERT" technique which in Postgres is available through "ON CONFLICT". Here is when not having ORM is a really powerful.

I think we should open up discussion if we can remove account_number from the database. Handling of empty/null values is extra work and it makes the query more complex than it needs to be.

### DIFF
--- a/internal/dao/dao_interfaces.go
+++ b/internal/dao/dao_interfaces.go
@@ -21,11 +21,11 @@ var GetAccountDao func(ctx context.Context) AccountDao
 
 // AccountDao represents an account (tenant)
 type AccountDao interface {
+	// Create is meant for integration tests, use GetOrCreateByIdentity instead.
 	Create(ctx context.Context, pk *models.Account) error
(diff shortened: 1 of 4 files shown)


ezr-ondrej (commented):
LGTM, just wondering if the `GetByOrgID` and `GetByAccountNumber` are used now? I'd guess they should not have right? :thinking:
If those are not used, I'd personaly remove them.

lzap:
The former is used in tests in `WithTenant` so I am removing the latter.

lzap:
Ah sure rebased.

ezr-ondrej (approved):
LGTM, I hope we will be able to simplify even more soon :)

PR #547, opened 2023-05-24
feat(HMS-1828): Add listing templates and creating instances from a template

adiabramovitch:
Adding support for listing launch templates in GCP and launching instances from a specific template name. 
GCP works with launch template names rather than launch template IDs.

### DIFF
--- a/internal/clients/http/gcp/gcp_client.go
+++ b/internal/clients/http/gcp/gcp_client.go
@@ -91,6 +91,54 @@ func (c *gcpClient) newInstancesClient(ctx context.Context) (*compute.InstancesC
 	return client, nil
 }
 
+func (c *gcpClient) NewInstanceTemplatesClient(ctx context.Context) (*compute.InstanceTemplatesClient, error) {
+	client, err := compute.NewInstanceTemplatesRESTClient(ctx, c.options...)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create GCP templates client: %w", err)
+	}
+	return client, nil
+}
+
+func (c *gcpClient) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchTemplate, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListLaunchTemplates")
+	defer span.End()
+
+	logger := logger(ctx)
+	logger.Trace().Msgf("Listing launch templates")
+
+	templatesClient, err := c.NewInstanceTemplatesClient(ctx)
+	if err != nil {
+		logger.Error().Err(err).Msg("Could not get instances client")
+		return nil, fmt.Errorf("unable to get instances client: %w", err)
+	}
+	defer templatesClient.Close()
+
+	req := &computepb.AggregatedListInstanceTemplatesRequest{
+		Project: c.auth.Payload,
+	}
+
+	var templatesList []*clients.LaunchTemplate
+	templates := templatesClient.AggregatedList(ctx, req)
+	for {
+		pair, err := templates.Next()
+		if errors.Is(err, iterator.Done) {
+			break
+		} else if err != nil {
+			logger.Error().Err(err).Msg("An error occurred during listing launch templates")
+			span.SetStatus(codes.Error, err.Error())
+			return nil, fmt.Errorf("cannot list launch templates: %w", err)
+		} else {
+			instancesTemplates := pair.Value.InstanceTemplates
+			for _, template := range instancesTemplates {
+				id := strconv.FormatUint(*template.Id, 10)
+				templatesList = append(templatesList, &clients.LaunchTemplate{ID: id, Name: template.GetName()})
+			}
+		}
+	}
+
+	return templatesList, nil
+}
+
 func (c *gcpClient) InsertInstances(ctx context.Context, params *clients.GCPInstanceParams, amount int64) ([]*string, *string, error) {
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "InsertInstances")
 	defer span.End()
@@ -154,6 +202,11 @@ func (c *gcpClient) InsertInstances(ctx context.Context, params *clients.GCPInst
 				},
 			},
 		},
+	}
+
+	if params.LaunchTemplateName != "" {
+		template := fmt.Sprintf("global/instanceTemplates/%s", params.LaunchTemplateName)
+		req.BulkInsertInstanceResourceResource.SourceInstanceTemplate = &template
 	}
 
 	op, err := client.BulkInsert(ctx, req)
--- a/internal/services/launch_templates_service.go
+++ b/internal/services/launch_templates_service.go
@@ -4,12 +4,41 @@ import (
 	"net/http"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/render"
 )
 
+//nolint:exhaustive
 func ListLaunchTemplates(w http.ResponseWriter, r *http.Request) {
+	sourceId := chi.URLParam(r, "ID")
+
+	sourcesClient, err := clients.GetSourcesClient(r.Context())
+	if err != nil {
+		renderError(w, r, payloads.NewClientError(r.Context(), err))
+		return
+	}
+
+	auth, err := sourcesClient.GetAuthentication(r.Context(), sourceId)
+	if err != nil {
+		renderError(w, r, payloads.NewClientError(r.Context(), err))
+		return
+	}
+
+	switch auth.ProviderType {
+	case models.ProviderTypeAWS:
+		ListLaunchTemplateAWS(w, r)
+	case models.ProviderTypeAzure:
+		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), "azure reservation is not implemented", ProviderTypeNotImplementedError))
+	case models.ProviderTypeGCP:
+		ListLaunchTemplateGCP(w, r)
+	default:
+		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), "provider is not supported", UnknownProviderTypeError))
+	}
+}
+
+func ListLaunchTemplateAWS(w http.ResponseWriter, r *http.Request) {
 	sourceId := chi.URLParam(r, "ID")
 	region := r.URL.Query().Get("region")
 	if region == "" {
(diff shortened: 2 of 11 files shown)


internal/clients/http/gcp/gcp_client.go
```diff
@@ -92,6 +92,55 @@ func (c *gcpClient) newInstancesClient(ctx context.Context) (*compute.InstancesC
 	return client, nil
 }
 
+func (c *gcpClient) NewInstanceTemplates(ctx context.Context) (*compute.InstanceTemplatesClient, error) {
```
lzap, line 95:
```suggestion
func (c *gcpClient) NewInstanceTemplatesClient(ctx context.Context) (*compute.InstanceTemplatesClient, error) {
```

internal/clients/http/gcp/gcp_client.go
```diff
@@ -92,6 +92,55 @@ func (c *gcpClient) newInstancesClient(ctx context.Context) (*compute.InstancesC
...
+	client, err := compute.NewInstanceTemplatesRESTClient(ctx, c.options...)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create GCP regions client: %w", err)
```
lzap, line 98:
```suggestion
		return nil, fmt.Errorf("unable to create GCP templates client: %w", err)
```

internal/clients/http/gcp/gcp_client.go
```diff
@@ -92,6 +92,55 @@ func (c *gcpClient) newInstancesClient(ctx context.Context) (*compute.InstancesC
...
+	if err != nil {
+		logger.Error().Err(err).Msg("Could not get instances client")
+		return nil, fmt.Errorf("unable to get instances client: %w", err)
+	}
+	defer templatesClient.Close()
+
+	req := &computepb.AggregatedListInstanceTemplatesRequest{
+		Project: c.auth.Payload,
+	}
+	templatesList := make([]*clients.LaunchTemplate, 0)
```
lzap, line 120: This is same as

    var templatesList []*clients.LaunchTemplate

I think. Function `append` will automatically allocate new slice when value is `nil`. This is generally preferred in Go codebases I have seen so far.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -92,6 +92,55 @@ func (c *gcpClient) newInstancesClient(ctx context.Context) (*compute.InstancesC
...
+
+	templates := templatesClient.AggregatedList(ctx, req)
+	for {
+		pair, err := templates.Next()
+		if errors.Is(err, iterator.Done) {
+			logger.Trace().Err(err).Msg("Instances iterator has finished")
```
lzap, line 126: Please drop this from the final version, this is quite too verbose.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -92,6 +92,55 @@ func (c *gcpClient) newInstancesClient(ctx context.Context) (*compute.InstancesC
...
+			instancesTemplates := pair.Value.InstanceTemplates
+			for _, template := range instancesTemplates {
+				id := strconv.FormatUint(*template.Id, 10)
+				templatesList = append(templatesList, &clients.LaunchTemplate{ID: id, Name: template.GetName()})
+			}
+		}
+	}
+
+	return templatesList, nil
+}
```
lzap, line 142: This is good for now, but at some point we need to solve pagination.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -131,31 +180,37 @@ func (c *gcpClient) InsertInstances(ctx context.Context, params *clients.GCPInst
...
+				{
+					Name: ptr.To("global/networks/default"),
 				},
 			},
-		},
+			Metadata: &computepb.Metadata{
+				Items: metadata,
+			},
+		}
 	}
```
lzap, line 214: We cannot just throw away users settings, at the moment the UI enforces users to set instance type, amount and pubkey. This code ignores the setting and only applies the template.

We can change this, but this will require:

* updating AWS backend to ignore user input
* updating UI to "grey out" required fields
* discuss this with UX

What I suggest at this point:

* investigate if you can use a template and still provide pubkey/instancetype/amount in the GCP API
* if this can be done easily, let’s do this similarly to AWS
* if not, let’s implement GCP this way and create an EPIC to do all the above for AWS

@ezr-ondrej ?

internal/services/launch_templates_service.go
```diff
@@ -41,7 +59,39 @@ func ListLaunchTemplates(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		renderError(w, r, payloads.NewGCPError(r.Context(), "unable to list GCP launch templates", err))
+		return
+	}
+
+	if err := render.RenderList(w, r, payloads.NewListLaunchTemplateResponse(templates)); err != nil {
+		renderError(w, r, payloads.NewRenderError(r.Context(), "unable to render launch templates list", err))
 		return
 	}
 }
```
lzap, line 97: Nicely done.

scripts/rest_examples/reservation-create-gcp.http
```diff
@@ -10,6 +10,7 @@ X-Rh-Identity: {{identity}}
   "image_id": "https://www.googleapis.com/compute/v1/projects/rhel-cloud/global/images/rhel-9-v20230411",
   "amount": 1,
   "machine_type": "e2-micro",
+  "launch_template_name": "instance-template-1",
```
lzap, line 13: Maybe create this a variable so devs can change template, also the default could be empty - API should ignore an empty template name and treat it as "nothing was provided". Just an idea.

lzap (changes requested):
Looks good, one major concern is about ignoring pubkey/type/amount we need to discuss this maybe online.

internal/clients/http/gcp/gcp_client.go
adiabramovitch: Apologies for any confusion, I realized that I made an error by omitting the metadata section when the template name is present. 

It appears that we do consider the amount(line 181) and public key ID (after the upcoming change), the only missing piece is the **machine type**. 

But, I have noticed that when providing a template name, the created instances use the machine type that the user has chosen and not the machine type that belongs to the template. To address this, my suggestion is to grey out the machine type section if a machine type was provided in the chosen template.

That is also true for the image type, the image type was ovrriden as well and the instances were created from the the image the user has provide and not from the template's image.

lzap:
Errors with:

```
HTTP response body: {"msg":"Invalid request: provider is not supported","trace_id":"d0ff468387501f24ab5be43654641d99","error":"unknown provider type parameter","version":"b25f33d","build_time":"2023-05-28_14:55:49"}
```

This might be relevant.

internal/clients/http/gcp/gcp_client.go
lzap: Problem is that with AWS you can create a launch template that is completely empty - does not define anything other than its name. If we want to go that route, we need to find out if selected template does define instance type, pubkey and other fields which are in question.

That is why we have decided that simply we would enforce the user the always provide all the info and it would always override templates. It is also more simple in the UI for now. At least for AWS, my proposal here is to do the same for GCP.

I suggest to do the same with GCP and we can create and track an epic that would implement what you are suggesting - a UI code that will "disable" some elements. API requests would have empty strings or zeroes for values which were not set.

lzap: The GCP API seems to support overriding, can we do this similarly to AWS at least for now and if you don’t like this feel free to refactor the UI and add support for skipping options when template is set.

https://cloud.google.com/compute/docs/reference/rest/v1/instances/bulkInsert

_Specifies the instance template from which to create instances. You may combine sourceInstanceTemplate with instanceProperties to override specific values from an existing instance template. Bulk API follows the semantics of JSON Merge Patch [described by RFC 7396](https://www.ietf.org/rfc/rfc7396.txt)._

lzap: Created a ticket to track this issue: https://issues.redhat.com/browse/HMS-1861

internal/services/launch_templates_service.go
```diff
@@ -41,7 +59,39 @@ func ListLaunchTemplates(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		renderError(w, r, payloads.NewClientError(r.Context(), err))
+		return
+	}
+
+	gcpClient, err := clients.GetGCPClient(r.Context(), authentication)
+	if err != nil {
+		renderError(w, r, payloads.NewGCPError(r.Context(), "unable to get GCP client", err))
+		return
+	}
```
lzap, line 85: Here you actually do it :)

internal/services/launch_templates_service.go
```diff
@@ -4,12 +4,30 @@ import (
...
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/render"
 )
 
+//nolint:exhaustive
 func ListLaunchTemplates(w http.ResponseWriter, r *http.Request) {
+	provider := r.URL.Query().Get("provider")
```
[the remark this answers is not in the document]
adiabramovitch: I am sorry, I do not understand what you mean. 
I added the parameter so we can have one generic function for listing templates that dispatch to the correct function according to the provider. 

What is your suggestion? 

I can remove this parameter and get the provider from the source itself. is that what you mean? 
`GetAWSAccountIdentity` was not used in listing templates, listing templates were only suitable for listing AWS templates thus, we did not need to know which provider the source belongs to.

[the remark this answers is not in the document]
lzap: > I can remove this parameter and get the provider from the source itself. is that what you mean?

Yes that is what I propose if I am correct.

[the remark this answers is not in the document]
lzap: Here is my reasoning: when you do `/sources/1234/launch_templates?type=gcp` it does not make sense if source 1234 is not gcp. Also, why you would ask the user to provide information twice? Unless I am missing something, I don’t see any reason to do it this way.

lzap:
Also conflicts since we merged second part of the refactoring, sorry about that. Here is a hint for resolving - "accept yours" when git or IDE asks throwing my changes away and then just replace `ctxval` with one of `identity`, `logging` or the ozher package. IDE will help. Sorry!

internal/clients/http/gcp/gcp_client.go
```diff
@@ -91,6 +91,54 @@ func (c *gcpClient) newInstancesClient(ctx context.Context) (*compute.InstancesC
...
+	}
+	return client, nil
+}
+
+func (c *gcpClient) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchTemplate, error) {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListLaunchTemplates")
+	defer span.End()
+
+	logger := logger(ctx)
+	logger.Trace().Msgf("Listing launch templates")
```
lzap, line 107: These "a function was called" are not very useful, in fact, thanks to opentelemetry exactly this message will be written in log (lines 103-104). But its fine.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -156,6 +204,11 @@ func (c *gcpClient) InsertInstances(ctx context.Context, params *clients.GCPInst
 		},
 	}
 
+	if params.LaunchTemplateName != "" {
+		template := fmt.Sprintf("global/instanceTemplates/%s", params.LaunchTemplateName)
```
lzap, line 208: I wonder if there is a possibility to have a different prefix than `global/instanceTemplate` but I think it would be pretty easy to differentiate later using string prefix comparison so this is fine.

lzap (approved):
Thanks.

PR #548, opened 2023-05-24
chore: improve make clean target

lzap:
SSIA might be useful.


lzap:
Let me push one more change, we are struggling with different go versions for linter.

lzap:
Adi stated it did help so merging as I need to rebase my other PR on top of this one.

PR #549, opened 2023-05-25
build: route GCP secrets to ENV

ezr-ondrej:
Adds GCP env variable from credentials in k8s

I think the JSON should be enough in prod environments, as `GCP_PROJECT_ID` is consumed only in the service client, but not the customer client, but could you give me a sanity check here @adiabramovitch pls? :)

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -78,6 +78,12 @@ objects:
                     name: provisioning-azure-acc
                     key: client_secret
                     optional: true
+              - name: GCP_JSON
+                valueFrom:
+                  secretKeyRef:
+                    name: provisioning-gcp-acc
+                    key: json
+                    optional: true
               - name: SENTRY_DSN
                 valueFrom:
                   secretKeyRef:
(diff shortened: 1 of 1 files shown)


lzap:
LGTM

adiabramovitch:
Not sure what you mean
Why we don't need to save the project id on clowdapp.yaml? 
The project id is part of the GCP authentication json, we can take the information from there as well.

ezr-ondrej:
> Why we don't need to save the project id on clowdapp.yaml?

I hoped you could help us figure out if we do need to add PROJECT_ID in clowdapp or not, you've created the config variable and I got a feeling it is redundant, but if used, then for developement only.

https://github.com/RHEnVision/provisioning-backend/commit/741b8207c6ee22d95dc33c84ef001ac413f5e68a#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR82

What I'd love to know is whether I'm right and we do not need to fill out thie PROJECT_ID from clowdapp.yml.

adiabramovitch (approved):
Looks great :)

adiabramovitch (approved):
Looks great :)

adiabramovitch (approved):
Looks great :)

adiabramovitch (approved):
Looks great :)

PR #550, opened 2023-05-25
build: use single binary

lzap:
We build four different binary files: pbapi, pbworker, pbstatuser and pbmigrate. Each about 70MiB in size, we copy them into the container. The most of the size comes with libraries (dependencies), if we only could ship just one binary.

This patch introduces a new binary called `pbackend` which replaces all of the four saving 70% of the old size and trimming compilation time in half (without Go cache). It takes an argument which means which program to start:

```
[lzap@nuc provisioning-backend]$ ./pbackend 
Usage: pbackend [migrate|api|worker|statuser|version]

[lzap@nuc provisioning-backend]$ ./pbackend version
Version 4f2e, Go version go1.19.6, built 2023-05-23T13:09:37Z

[lzap@nuc provisioning-backend]$ ./pbackend migrate
...migrates database...
```

This patch is changing container entrypoints as well as k8s configuration in order to start correct processes. Everything should work as previously without any change. Except one:

We use the `BinaryName` function in logger, it is used for log "stream". So after this is merged, logs will change "stream" from "pbapi" to more human-readable "api". Luckily, we do not use these in prometheus or k8s directly so no monitoring/alert changes should be necessary.

With this change, new feature of Go 1.18 can be used to report build time, go version and sha number. Previously we used build flags, this is no longer necessary.

### DIFF
--- a/cmd/pbackend/api.go
+++ b/cmd/pbackend/api.go
@@ -11,42 +11,24 @@ import (
 
 	"github.com/RHEnVision/provisioning-backend/internal/background"
 	"github.com/RHEnVision/provisioning-backend/internal/cache"
-	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/azure"
-	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/ec2"
-	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/gcp"
-	"github.com/RHEnVision/provisioning-backend/internal/kafka"
-	"github.com/RHEnVision/provisioning-backend/internal/metrics"
-	"github.com/RHEnVision/provisioning-backend/internal/queue/jq"
-	"github.com/RHEnVision/provisioning-backend/internal/random"
-	s "github.com/RHEnVision/provisioning-backend/internal/services"
-
-	// HTTP client implementations
-	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/image_builder"
-	_ "github.com/RHEnVision/provisioning-backend/internal/clients/http/sources"
-	"github.com/RHEnVision/provisioning-backend/internal/telemetry"
-	"github.com/RHEnVision/provisioning-backend/internal/version"
-
-	// DAO implementation, must be initialized before any database packages.
-	_ "github.com/RHEnVision/provisioning-backend/internal/dao/pgx"
-
-	// set GOMAXPROC to ideal value for Kubernetes
-	_ "go.uber.org/automaxprocs"
-
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/db"
+	"github.com/RHEnVision/provisioning-backend/internal/kafka"
 	"github.com/RHEnVision/provisioning-backend/internal/logging"
+	"github.com/RHEnVision/provisioning-backend/internal/metrics"
 	m "github.com/RHEnVision/provisioning-backend/internal/middleware"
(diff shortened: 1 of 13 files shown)


cmd/pbackend/main.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func usage() {
+	fmt.Println("Usage: pbackend [migrate|api|worker|statuser|version]")
+	os.Exit(1)
+}
+
+func ver() {
+	fmt.Printf("Version %s, Go version %s, built %s\n", version.BuildCommit, runtime.Version(), version.BuildTime)
+}
```
lzap, line 54: This is the only main entrypoint, other files in the directory only contain functions named:

* `api()`
* `worker()`
* `statuser()`
* `migrate()`

They are unchanged versions of the respective `main()` counterparts. Except one small change which I will comment.

mk/compile.mk
```diff
@@ -4,37 +4,34 @@ SRC_GO := $(shell find . -name \*.go -print)
...
 .PHONY: run
-run: pbapi ## Build and run backend API
-	./pbapi
+run: pbackend ## Build and run backend API
+	./pbackend $(CMD)
 
 .PHONY: clean
 clean: ## Clean build artifacts
-	-rm pbapi pbmigrate pbworker
+	-rm pbackend pbapi pbmigrate pbworker
```
lzap, line 37: You can now use:

* `make build` - builds the binary (no change except is only builds one)
* `make run` - build and run `version` command
* `make run CMD=api` - run api (or worker or statuser)
* `make run-api` - run api via `go run` (or worker or statuser)

cmd/pbackend/api.go
```diff
@@ -63,6 +45,11 @@ func main() {
 	}
 	defer config.StopFeatureFlags(ctx)
 
+	// set GOMAXPROCs for Kubernetes
+	_, _ = maxprocs.Set(maxprocs.Logger(func(format string, args ...interface{}) {
```
ezr-ondrej, line 49: Wasn't this introduced mainly for the worker? should we do it there as well? :thinking:

ezr-ondrej (commented):
Great change! I personally :heart: it. Great simplification.
I found just one issue, that I don't even know if it's an issue :)

internal/config/helpers.go
```diff
@@ -115,7 +114,4 @@ func DumpConfig(logger zerolog.Logger) {
 		configCopy.Sentry.Dsn = replacement
 	}
 	logger.Info().Msgf("Configuration: %+v", configCopy)
-
-	gmp := runtime.GOMAXPROCS(0)
-	logger.Info().Int("gomaxproc", gmp).Msgf("GOMAXPROC value: %d", gmp)
```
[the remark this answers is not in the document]
ezr-ondrej: should we do it for `worker()` as well?

cmd/pbackend/api.go
lzap: Oh no this tiny library performs some heuristics "guess" of what is the best goroutines max setting for HTTP handling in kubernetes. So it is actually only for API, for workers we don’t have that. At least the documentation states this is for HTTP traffic.

For other pods we use the default value which is I think 3 parallel goroutines (2 cores + 1).

ezr-ondrej: Ah! Cool, makes sense 👍

ezr-ondrej (approved):
I'm good here, I'd wait with merge for post prod deploy on Monday.
Firstly to give more time for review, secondly to make sure everything works as before in stage for a bit longer than over the weekend :)

PR #551, opened 2023-05-28
fix(HMSPROV-451): Use sources filtering mechanism

adiabramovitch:
Using sources filtering instead of fetching all the authentications and sources and for the filtering in our backend. 
Few things: 

1. Some of the tests became meaningless because the data we mock  (from sources) is already filtered so there is no need in checking the cases like: 
- a source of a type different than "Application" 
-  checking the filtering of sources by a provider 

2. Regarding the mapping between our source types (aws, gcp, azure) and sources source types (amazon, google, azure) I am not sure there is a better way to do the mapping. I wanted to rely on their API but I think it gets messy and either way, our service will break if there will be a change in their API (please correct me If I am wrong)
Not sure what is the best solution here, maybe you will have something :)

### DIFF
--- a/internal/clients/http/sources/sources_client.go
+++ b/internal/clients/http/sources/sources_client.go
@@ -5,6 +5,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	stdhttp "net/http"
+	"net/url"
+	"strings"
 
 	"github.com/RHEnVision/provisioning-backend/internal/cache"
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
@@ -90,7 +93,14 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 		return nil, fmt.Errorf("failed to get provisioning app type: %w", err)
 	}
 
-	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, &ListApplicationTypeSourcesParams{}, headers.AddSourcesIdentityHeader, headers.AddEdgeRequestIdHeader)
+	sourcesProviderName := provider.SourcesProviderName()
+	if err != nil {
+		logger.Warn().Err(err).Msg("Failed to get provider name according to sources service")
+		return nil, fmt.Errorf("failed to get provider name according to sources service: %w", err)
+	}
+
+	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, &ListApplicationTypeSourcesParams{}, headers.AddSourcesIdentityHeader,
+		headers.AddEdgeRequestIdHeader, BuildQuery("filter[source_type][name]", sourcesProviderName))
 	if err != nil {
 		logger.Warn().Err(err).Msg("Failed to fetch ApplicationTypes from sources")
 		return nil, fmt.Errorf("failed to get ApplicationTypes: %w", err)
@@ -107,20 +117,13 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 	result := make([]*clients.Source, 0, len(*resp.JSON200.Data))
 
 	for _, src := range *resp.JSON200.Data {
-		sourceTypeName, err := c.GetSourceTypeName(ctx, *src.SourceTypeId)
-		if err != nil {
-			return nil, fmt.Errorf("could not get source type name for source type id %d: %w", src.SourceTypeId, err)
+		newSrc := clients.Source{
+			ID:           ptr.From(src.Id),
+			Name:         ptr.From(src.Name),
+			SourceTypeID: ptr.From(src.SourceTypeId),
+			Uid:          ptr.From(src.Uid),
 		}
-
-		if sourceTypeName == models.ProviderTypeFromString(provider.String()) {
-			newSrc := clients.Source{
-				ID:           ptr.From(src.Id),
-				Name:         ptr.From(src.Name),
-				SourceTypeID: ptr.From(src.SourceTypeId),
-				Uid:          ptr.From(src.Uid),
-			}
-			result = append(result, &newSrc)
-		}
+		result = append(result, &newSrc)
 	}
 
 	return result, nil
@@ -170,9 +173,10 @@ func (c *sourcesClient) GetAuthentication(ctx context.Context, sourceId string) 
 	defer span.End()
 
 	// Get all the authentications linked to a specific source
-	resp, err := c.client.ListSourceAuthenticationsWithResponse(ctx, sourceId, &ListSourceAuthenticationsParams{}, headers.AddSourcesIdentityHeader, headers.AddEdgeRequestIdHeader)
+	resp, err := c.client.ListSourceAuthenticationsWithResponse(ctx, sourceId, &ListSourceAuthenticationsParams{}, headers.AddSourcesIdentityHeader,
+		headers.AddEdgeRequestIdHeader, BuildQuery("filter[resource_type]", "Application", "filter[authtype][starts_with]", "provisioning"))
 	if err != nil {
-		return nil, fmt.Errorf("cannot list source authentication: %w", err)
+		return nil, fmt.Errorf("cannot list provisioning source authentication of type application: %w", err)
 	}
 
 	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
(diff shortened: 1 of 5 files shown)


internal/clients/http/sources/sources_client.go
```diff
@@ -183,20 +190,15 @@ func (c *sourcesClient) GetAuthentication(ctx context.Context, sourceId string)
...
-		for _, auth := range *resp.JSON200.Data {
-			at = append(at, string(*auth.Authtype))
-		}
-		logger.Warn().Msgf("Sources did not return any Provisioning authentication for source(auth types): %s(%v)", sourceId, at)
-		return nil, err
+	if len(*resp.JSON200.Data) != 0 {
+		// TODO: What do we do if we have more than one authentication
+		auth := (*resp.JSON200.Data)[0]
+		authentication := clients.NewAuthenticationFromSourceAuthType(ctx, *auth.Username, string(*auth.Authtype), *auth.ResourceId)
+		return authentication, nil
```
lzap, line 197: Honestly I would not worry about it here. Just check if len is > 1 and take the first element.

internal/clients/http/sources/sources_client.go
```diff
@@ -90,7 +92,16 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
...
+	sourcesProviderName, err := c.getProviderNameBySources(ctx, provider)
+	if err != nil {
+		logger.Warn().Err(err).Msg("Failed to get provider name according to sources service")
+		return nil, fmt.Errorf("failed to get provider name according to sources service: %w", err)
+	}
+
+	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, &ListApplicationTypeSourcesParams{}, headers.AddSourcesIdentityHeader, headers.AddEdgeRequestIdHeader, func(ctx context.Context, req *stdhttp.Request) error {
+		req.URL.RawQuery = url.QueryEscape("filter[source_type][name]") + fmt.Sprintf("=%s", sourcesProviderName)
+		return nil
+	})
```
lzap, line 104: Can you extract this inline function into named function that will return `func(ctx context.Context, req *stdhttp.Request)` with variadic string arguments, so any number of query params would work? I think this will be limitation across all endpoints so we might want to use it. Also pagination will be added soon.

This should look like something:

```
c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, &ListApplicationTypeSourcesParams{}, headers.AddSourcesIdentityHeader, headers.AddEdgeRequestIdHeader, xxx.BuildQuery("filter[source_type][name]", sourcesProviderName))
```

In the future this could be `xxx.BuildQuery("filter[source_type][name]", sourcesProviderName, "limit", limit, "offset", offset)`

internal/clients/http/sources/sources_client.go
```diff
@@ -170,9 +174,12 @@ func (c *sourcesClient) GetAuthentication(ctx context.Context, sourceId string)
 	defer span.End()
 
 	// Get all the authentications linked to a specific source
-	resp, err := c.client.ListSourceAuthenticationsWithResponse(ctx, sourceId, &ListSourceAuthenticationsParams{}, headers.AddSourcesIdentityHeader, headers.AddEdgeRequestIdHeader)
+	resp, err := c.client.ListSourceAuthenticationsWithResponse(ctx, sourceId, &ListSourceAuthenticationsParams{}, headers.AddSourcesIdentityHeader, headers.AddEdgeRequestIdHeader, func(ctx context.Context, req *stdhttp.Request) error {
+		req.URL.RawQuery = url.QueryEscape("filter[resource_type]") + "=Application" + "&" + url.QueryEscape("filter[authtype][starts_with]") + "=provisioning"
+		return nil
+	})
```
lzap, line 180: Here you could use it as well.

If amount of arguments is not even number, it should panic.

internal/clients/http/sources/sources_client.go
```diff
@@ -250,48 +252,15 @@ func (c *sourcesClient) loadAppId(ctx context.Context) (string, error) {
 }
 
 //nolint:exhaustive
```
lzap, line 254: Can this be dropped too?

lzap (changes requested):
Less code, less tests, great!

lzap:
> Regarding the mapping between our source types (aws, gcp, azure) and sources source types (amazon, google, azure) I am not sure there is a better way to do the mapping. I wanted to rely on their API but I think it gets messy and either way, our service will break if there will be a change in their API (please correct me If I am wrong)

I don’t understand, this PR does not do anything with the mapping, does it?. Let’s do it in a separate PR this is fine.

We already have a "cache" for appTypeId, the implementation is little bit weird we use lazy initialization. Since the hyperscalers mapping will add more values, I suggest we refactor this code to be done during initialization - we simply read all the IDs we need to know during application start before we actually start HTTP handlers (so it is thread-safe). Then we can use the mapping from HTTP goroutines (could be some sort of array or map).

internal/clients/http/sources/sources_client_test.go
```diff
@@ -62,30 +62,6 @@ func TestSourcesClient_GetAuthentication(t *testing.T) {
 
 		assert.Equal(t, "1a2b3c4d-5e6f-7a8b-9c8d-7e8f9a8b7c6d", authentication.Payload)
 	})
-
-	t.Run("source with multiple apps", func(t *testing.T) {
```
ezr-ondrej, line 66: Why dont we want to cover this case? it is common for the Superkey method for AWS to end up with state like this, so I wonder why we are removing coverage for it? :thinking:

ezr-ondrej: ah, right we are filtering it all out on the sources side, we won't get this result here, makes sense :+1: IGNORE me :)

ezr-ondrej (commented):
Looks reasonable to me :) I guess it would be nice to have a helper for adding the header, but not a big deal given we have just two usages.

adiabramovitch:
> > Regarding the mapping between our source types (aws, gcp, azure) and sources source types (amazon, google, azure) I am not sure there is a better way to do the mapping. I wanted to rely on their API but I think it gets messy and either way, our service will break if there will be a change in their API (please correct me If I am wrong)
> 
> 
> 
> I don’t understand, this PR does not do anything with the mapping, does it?. Let’s do it in a separate PR this is fine.
> 
> 
> 
> We already have a "cache" for appTypeId, the implementation is little bit weird we use lazy initialization. Since the hyperscalers mapping will add more values, I suggest we refactor this code to be done during initialization - we simply read all the IDs we need to know during application start before we actually start HTTP handlers (so it is thread-safe). Then we can use the mapping from HTTP goroutines (could be some sort of array or map).

I meant in the mapping to the function getProviderNameBySources

internal/clients/http/sources/sources_client.go
```diff
@@ -249,49 +247,19 @@ func (c *sourcesClient) loadAppId(ctx context.Context) (string, error) {
...
+			key := url.QueryEscape(keysAndValues[i])
+			value := url.QueryEscape(keysAndValues[i+1])
+			queryParams = append(queryParams, fmt.Sprintf("%s=%s", key, value))
 		}
+
+		req.URL.RawQuery = strings.Join(queryParams, "&")
+		return nil
 	}
-	return models.ProviderTypeUnknown, fmt.Errorf("cannot find source type name for source type id %s: %w", sourceTypeID, http.SourceTypeNameNotFoundErr)
 }
```
lzap, line 265: There is no need to return error in this case since no other function here does return one. Generally `panic` should not be used, but one of the rare cases when it is useful are static errors (e.g. this one will only happen when someone changes arguments badly).

But it’s okay to leave as is :-)

internal/models/model_constants.go
```diff
@@ -36,6 +36,22 @@ func ProviderTypeFromString(str string) ProviderType {
...
+		return "google"
+	case ProviderTypeAzure:
+		return "azure"
+	case ProviderTypeNoop:
+	case ProviderTypeUnknown:
+	default:
+		return ""
+	}
+	return ""
+}
```
lzap, line 53: Heh the last three cases are not needed but I guess linter was pesky. I hate this, I am all for disabling that for the good. :-D

lzap (approved):
A MASSIVE improvement, great work! Thanks.

PR #552, opened 2023-05-29
build: use PATH for golint

lzap:
Looks like I misread how golangci-lint handles GOROOT, I thought it would use that variable but looks like it always tries to detect the GOROOT via `go env GOROOT`. Problem is, it uses system Go rather than the specific Go version we want.

https://github.com/golangci/golangci-lint/blob/master/pkg/goutil/env.go

This patch adds PATH for the linter command, so the `go` command resolves to the proper version, this fixes the problem.

### DIFF
--- a/mk/code.mk
+++ b/mk/code.mk
@@ -17,7 +17,7 @@ imports: ## Rearrange imports using `goimports`
 
 .PHONY: lint
 lint: ## Run Go language linter `golangci-lint`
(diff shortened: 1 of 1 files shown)


adiabramovitch (approved):
WORKS!!! thanks!

lzap:
Apologies for the confusion, I was testing this with having Go 1.19.6 SDK on my PATH so it worked for me.

PR #553, opened 2023-05-30
refactor: deprecate ctxval.Logger

lzap:
I made a terrible mistake with introducing `ctxval` package. Context function should have been part of package they belong to, this creates unnecessary import hell and cycles around the codebase because pretty much every function uses `ctxval.Logger` function.

Turns out this function is not needed at all, because zerolog provides the way to do the exactly the same:

    logger := zerolog.Ctx(ctx)

is the same as 

    logger := zerolog.Ctx(ctx)

To create a new context (we do this in the middleware) it is also easy:

    ctx := zerolog.WIthContext(ctx)

This patch is a proposal to deprecate the `ctxval.Logger` function pair so we can slowly and gradually get rid of it. I do not plan any "big bang" patch that would replace all the occurances across the codebase, this would create unnecessary conflicts.

### DIFF
--- a/internal/ctxval/logger.go
+++ b/internal/ctxval/logger.go
@@ -4,20 +4,20 @@ import (
 	"context"
 
 	"github.com/rs/zerolog"
-	"github.com/rs/zerolog/log"
 )
 
 // Logger returns the main logger with context fields or the standard global logger
-// when the main logger was not set. Never returns nil.
+// when the main logger was not set. When nil is passed, returns noop logger.
+// Never returns nil.
+//
+// Deprecated: Use zerolog.Ctx or log.Ctx alias instead.
 func Logger(ctx context.Context) *zerolog.Logger {
-	value := ctx.Value(loggerCtxKey)
-	if ctx == nil || value == nil {
-		return &log.Logger
-	}
-	return value.(*zerolog.Logger)
+	return zerolog.Ctx(ctx)
 }
 
 // WithLogger returns context copy with logger.
+//
+// Deprecated: Use log.Logger.WithContext(ctx) instead.
 func WithLogger(ctx context.Context, logger *zerolog.Logger) context.Context {
-	return context.WithValue(ctx, loggerCtxKey, logger)
+	return logger.WithContext(ctx)
 }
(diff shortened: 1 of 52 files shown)


ezr-ondrej:
Should our WithLogger become just a wrapper for zerolog's context fnc? That way, we would get rid of our context value right away

internal/ctxval/logger.go
```diff
@@ -18,6 +20,8 @@ func Logger(ctx context.Context) *zerolog.Logger {
 }
 
 // WithLogger returns context copy with logger.
+//
+// Deprecated: Use log.Logger.WithContext(ctx) instead.
 func WithLogger(ctx context.Context, logger *zerolog.Logger) context.Context {
```
ezr-ondrej, line 25: We could easily drop this function and make sure we use `ctx := zerolog.WIthContext(ctx)` everywhere instead.

internal/ctxval/logger.go
```diff
@@ -9,6 +9,8 @@ import (
 
 // Logger returns the main logger with context fields or the standard global logger
 // when the main logger was not set. Never returns nil.
+//
+// Deprecated: Use zerolog.Ctx or log.Ctx alias instead.
 func Logger(ctx context.Context) *zerolog.Logger {
```
ezr-ondrej, line 14: And make this just a wrapper func for `github.com/rs/zerolog/log.Ctx(ctx)`

ezr-ondrej (commented):
I agree we don't want a big bang, but we can easily drop our ctx value and have just single source of throught, and keep our function just as a deprecated wrapper.

lzap:
Great point, it’s not my day today. Rebased.

lzap:
I pushed replacements in another commit.

lzap:
Let me test for conflicts:

[@Amir Fefer](https://redhat-internal.slack.com/team/U039ENJ904D)
 notification-message - easy one hunk
[@Adi Abramovich](https://redhat-internal.slack.com/team/U0379FA8HHQ)
 fetch-instance-desc - one hunk conflicts but not caused by logger actually
[@Amir Fefer](https://redhat-internal.slack.com/team/U039ENJ904D)
 1804 - no conflits at all
[@Adi Abramovich](https://redhat-internal.slack.com/team/U0379FA8HHQ)
 launch-templates and filtering - no conflicts at all
So I think my concerns were not correct, let’s merge this rebasing will be smooth.

PR #555, opened 2023-05-30
chore: change SLO on the dashboard back

lzap:
We increased it to 5000ms for Summit but now that we have pushed to prod and new buckets include 750, 1000 and 2000ms we can lower it back.

I propose 750ms it has been 100% for last 24 hours. I will raise app-interface change once this is merged.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -57,7 +57,7 @@ var BackgroundJobDuration = prometheus.NewHistogramVec(
 		Name:        "provisioning_background_job_duration",
 		Help:        "task queue job duration (in seconds) by type",
 		ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName, "component": "worker"},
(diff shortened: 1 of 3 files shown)


internal/metrics/metrics.go
```diff
@@ -57,7 +57,7 @@ var BackgroundJobDuration = prometheus.NewHistogramVec(
 		Name:        "provisioning_background_job_duration",
 		Help:        "task queue job duration (in seconds) by type",
 		ConstLabels: prometheus.Labels{"service": version.PrometheusLabelName, "component": "worker"},
-		Buckets:     []float64{1, 4, 5, 10, 30, 60 * 2, 60 * 5, 60 * 10, 60 * 20, 60 * 30, 60 * 60},
+		Buckets:     []float64{0.5, 1, 2, 3, 4, 5, 7, 10, 30, 60 * 2, 60 * 10, 60 * 30},
```
lzap, line 60: Forgot to mention, as part of this PR I updated task job duration, we have 1 to 4 buckets which was not very detailed. Most of our jobs are about 500-1000ms.

PR #557, opened 2023-05-30
fix(HMS-1616): cleanup reservation OpenAPI docs

lzap:
This is the followup and the final cleanup PR of our OpenAPI. After this is merged, our API is 1.0.0 stable.

Review link: https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/RHEnVision/provisioning-backend/ffccbb6ba2c87643a4b62354f89af3ce93b4f1fd/api/openapi.gen.json&nocors


ezr-ondrej:
linter ain't happy, and I found probably missing yaml tag again for `step_titles` vs `steptitles`:
image

ezr-ondrej:
The same for Azure request, I guess the reservation payloads just miss a lot of yaml tags :thinking:

Apart of that, the changes look great and works as expected (just the discrepancy in key names needs to be fixed)

lzap:
Holy crap we actually do not call `Validate` I got a ton of issues but fixed them all!

Changes:

* We now check struct tags in all types - they must be present (otherwise weird defaults will be used)
* We actually perform validation of the whole schema before we finally store it (fixed few problems)
* We update version in the info header to the latest git tag (which will be 1.0.0 soon)
* Added all missing `yaml` tags

lzap:
Oh diff was failing because I used `time.Now()` in the example payloads :-D Fixed.

lzap:
Also I had to fix OpenAPI validation (the git diff) because of the version that was introduced. On CI git clones a shallow copy without tags, so no tag means no version which was causing problems to diff.

Final version: https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/RHEnVision/provisioning-backend/2da51e368085033cf60404b40e40b805716dc53e/api/openapi.gen.json

lzap:
Last minute change - I fixed reservation time it was not initialized correctly (it was zeroes: 0000-00-00).

ezr-ondrej:
I believe there is no need to get QE ack here as we just improve examples, but make no changes to the actual spec.

lzap:
I need this for the demo, merging, thanks.

PR #559, opened 2023-05-31
refactor: split ctxval package

lzap:
This is a followup of https://github.com/RHEnVision/provisioning-backend/pull/553

The package `ctxval` can be now refactored and functions can go into various packages depending on where they belong. I will comment in the PR on where things were moved.

One value does not have a good candidate, it is the Identity/WithIdentity functions, so I moved them into separate package `identity` as it is used across the codebase. Some files do use this new package and the go-middleware identity so to make things clear, a new type alias `identity.Principal` for`identity.XRHID` so there is no need to create package alias when working with identity.


cmd/pbackend/api.go
```diff
@@ -35,7 +35,6 @@ func api() {
 	// initialize cloudwatch using the AWS clients
 	logger, closeFunc := logging.InitializeLogger()
 	defer closeFunc()
-	log.Logger = logger
```
lzap, line 38: This (3 instances) was moved into `InitializeLogger`.

internal/config/ctx.go
```diff
@@ -1,11 +1,17 @@
...
 
+type contextKeyId int
+
+const (
+	unleashContextCtxKey contextKeyId = iota
+)
+
 // UnleashContext returns unleash context or an empty context when not set.
 func UnleashContext(ctx context.Context) ucontext.Context {
 	value := ctx.Value(unleashContextCtxKey)
```
lzap, line 17: Unleash was moved inti `config` package where it is used.

internal/identity/identity.go
```diff
@@ -9,13 +9,20 @@ import (
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
+type Principal = identity.XRHID
```
lzap, line 12: This is the new type alias, everything else should remain the same except there is no need for users of these functions to import "github.com/redhatinsights/platform-go-middlewares/identity".

internal/jobs/ctx.go
```diff
@@ -0,0 +1,20 @@
...
+// copyContext returns a new context with key values copied. Used when the original
+// context has expired but there is still some work to be done.
+func copyContext(ctx context.Context) context.Context {
+	nCtx := context.Background()
+	nCtx = log.Logger.WithContext(nCtx)
+	nCtx = logging.WithTraceId(nCtx, logging.TraceId(ctx))
+	nCtx = logging.WithEdgeRequestId(nCtx, logging.EdgeRequestId(ctx))
+	nCtx = identity.WithAccountId(nCtx, identity.AccountId(ctx))
+	return nCtx
+}
```
lzap, line 20: This is only used in job package so making it not exported.

internal/logging/ctx.go
```diff
@@ -0,0 +1,55 @@
...
+	if value == nil {
+		return ""
+	}
+	return value.(string)
+}
+
+// WithTraceId returns context copy with trace id value.
+func WithTraceId(ctx context.Context, id string) context.Context {
+	return context.WithValue(ctx, requestIdCtxKey, id)
+}
```
lzap, line 55: Most of the context value functions ended up in logging because this is their use.

internal/logging/zerolog.go
```diff
@@ -124,7 +124,10 @@ func InitializeLogger() (zerolog.Logger, func()) {
 			fn()
 		}
 	}
-	return decorate(zerolog.New(io.MultiWriter(writers...))), closeFn
+	logger := decorate(zerolog.New(io.MultiWriter(writers...)))
+	log.Logger = logger
+	zerolog.DefaultContextLogger = &logger
+	return logger, closeFn
 }
```
lzap, line 131: InitializeLogger now initializes both log.Logger and also default loger which `Ctx` function returns when nothing is in the context.

internal/middleware/account_test.go
```diff
@@ -7,9 +7,9 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	identity2 "github.com/RHEnVision/provisioning-backend/internal/identity"
```
lzap, line 10: In tests we do have existing `testing/identity` unfortunately.

lzap:
Few things I would like to improve on after this is merged:

* Initial "Started" and "Completed" log messages do not contain org_id and account_number, because the middleware that extract them is too late in the chain. I think we can split the middleware into two - one that extracts the values early and one that enforces them when needed.
* OpenTelemetry logger does not include account information at all, the context is not available.

internal/middleware/account_test.go
ezr-ondrej: I guess that it would make sense to make it a "stub" and we would have stubbed WithIdentity in the package, that would serve the same purpose, just for setting it up for tests?
I can take a look into that once we merge this :)

lzap:
In general, yeah, perhaps. I have a pending PR to refactor that API quite a bit, waiting for review...

PR #561, opened 2023-06-02
chore: do not allow override of GOROOT

lzap:
When you have GOROOT set in your profile, the makefile would use it instead of wanted GOROOT from the Go 1.19.6 (the current version), this fails the linter because it would use Go version which was set by environmental variable.

We don’t want that, this change will override any previous GOROOT setting from the environment so the proper Go version is ensured.


ezr-ondrej (approved):
ok :shrug: :)

PR #562, opened 2023-06-02
chore: log job cancellation to error level (Glitchtip)

lzap:
I would like to know when a job expired to be able to easily find it by job id. This will create an error log entry and also sentry error so we can track these easily.

### DIFF
--- a/pkg/worker/redis.go
+++ b/pkg/worker/redis.go
@@ -201,7 +201,10 @@ func (w *RedisWorker) processJob(ctx context.Context, job *Job) {
 	ctx = contextLogger(ctx, job)
 	if h, ok := w.handlers[job.Type]; ok {
 		cCtx, cFunc := context.WithTimeout(ctx, config.Worker.Timeout)
-		defer cFunc()
+		defer func() {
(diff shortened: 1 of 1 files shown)


pkg/worker/redis.go
```diff
@@ -201,7 +201,10 @@ func (w *RedisWorker) processJob(ctx context.Context, job *Job) {
 	ctx = contextLogger(ctx, job)
 	if h, ok := w.handlers[job.Type]; ok {
 		cCtx, cFunc := context.WithTimeout(ctx, config.Worker.Timeout)
-		defer cFunc()
+		defer func() {
+			zerolog.Ctx(ctx).Error().Str("job_id_timeout", job.ID.String()).Msg("Job cancelled")
```
ezr-ondrej, line 205: Would this be more meaningful as `job_id: job.ID, timeout: true` ?
Will this include a trace in glitchtip you think, even without `Err()`? 
If not, I'd really love to include also param name of the job, so we also have that info.

lzap: Done.

ezr-ondrej (approved):
Thanks ! :+1:

PR #567, opened 2023-06-07
feat(HMSPROV-1908): Add notifications service to compose config

amirfefer:
In following #539, this PR adds an additional commit for having the notifications service (backend and engine) to the containerized dev environemnt.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ TAGS
 # envdir/envrc (https://direnv.net)
 /.envrc
 /bin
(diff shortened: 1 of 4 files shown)


lzap (approved):
Let’s move this into the newly created repo.

As I mention in the first PR, there is a way to directly consume our quay.io containers if you login, just document that and I think it is good for now. This is much faster as nothing needs to be built.

amirfefer:
Moving this to compose repo

PR #568, opened 2023-06-12
Update deps, types and clients

lzap:
Update it all!


ezr-ondrej (approved):
Wow, the types are much less stable than I thought :/ might be a good idea to figure out a process to not forget updating those :thinking:

Thanks! :+1:

PR #571, opened 2023-06-12
feat(HMS-1932): Add gcp source upload info

adiabramovitch:
For the complete workflow to work from the UI the upload info logic was missing. 
In this PR:
- we are fetching the I am policy from the customer project,
- filtering to get the provisioning service account principal in order to pass into the UI

In the front, we have a validation that the principal we have fetched is equal to the email/principal that was shared with the image.

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -372,6 +372,7 @@
             "account_id": "78462784632"
           },
           "azure": null,
+          "gcp": null,
           "provider": "aws"
         }
       },
@@ -386,6 +387,7 @@
             "subscriptionid": "617807e1-e4e0-4855-983c-1e3ce1e49674",
             "tenantid": "617807e1-e4e0-481c-983c-be3ce1e49253"
           },
+          "gcp": null,
           "provider": "azure"
         }
       }
@@ -974,6 +976,9 @@
               }
             },
             "type": "object"
+          },
+          "gcp": {
+            "nullable": true
           },
           "provider": {
             "type": "string"
--- a/api/openapi.gen.yaml
+++ b/api/openapi.gen.yaml
@@ -348,6 +348,8 @@ components:
                             type: string
                         tenant_id:
                             type: string
+                gcp:
+                    nullable: true
                 provider:
                     type: string
     responses:
@@ -682,6 +684,7 @@ components:
                 aws:
                     account_id: "78462784632"
                 azure: null
+                gcp: null
                 provider: aws
         v1.SourceUploadInfoAzureResponse:
             value:
@@ -692,6 +695,7 @@ components:
                         - MyGroup 42
                     subscriptionid: 617807e1-e4e0-4855-983c-1e3ce1e49674
                     tenantid: 617807e1-e4e0-481c-983c-be3ce1e49253
+                gcp: null
                 provider: azure
 info:
     title: provisioning-api
--- a/internal/clients/source_upload_info.go
+++ b/internal/clients/source_upload_info.go
@@ -23,3 +23,5 @@ type AccountDetailsAzure struct {
 	SubscriptionID string        `json:"subscription_id"`
 	ResourceGroups []string      `json:"resource_groups"`
 }
+
+type AccountDetailsGCP struct{}
--- a/internal/payloads/sources_payload.go
+++ b/internal/payloads/sources_payload.go
@@ -40,6 +40,7 @@ type SourceUploadInfoResponse struct {
 	Provider  string                       `json:"provider" yaml:"provider"`
 	AwsInfo   *clients.AccountDetailsAWS   `json:"aws" nullable:"true" yaml:"aws"`
 	AzureInfo *clients.AccountDetailsAzure `json:"azure" nullable:"true" yaml:"azure"`
+	GcpInfo   *clients.AccountDetailsGCP   `json:"gcp" nullable:"true" yaml:"gcp"`
 }
 
 func (s SourceUploadInfoResponse) Render(_ http.ResponseWriter, _ *http.Request) error {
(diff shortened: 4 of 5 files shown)


internal/clients/http/gcp/gcp_client.go
```diff
@@ -99,6 +109,31 @@ func (c *gcpClient) NewInstanceTemplatesClient(ctx context.Context) (*compute.In
...
+	projectIamPolicy, err := client.Projects.GetIamPolicy(c.auth.Payload, &cm.GetIamPolicyRequest{}).Do()
+	if err != nil {
+		return "", fmt.Errorf("unable to get the project iam policy: %w", err)
+	}
+
+	for _, binding := range projectIamPolicy.Bindings {
+		if strings.Contains(binding.Role, "provisioning_role") {
+			if len(binding.Members) > 0 {
+				// The service account principal
+				return clients.GCPEmail(binding.Members[0]), nil
```
ezr-ondrej, line 130: Do I understand correctly that this will give us the first member of the group that:
1. needs to be called `provisioning_role`
2. we mandaded Customers to create and add our service account to it?

Would this actually result in "when following instructions correctly, returns our service account"?

My point:
1. if it's our service account we want, we probably don't need round trip to GCP
2. if it's our service account we return (and then compare with image.shared_with) how do we guide users that they need to share the image with this account (how would they know)?

adiabramovitch: So first, it is just a proposal, like I explained on Slack it seems like we have a mismatch with image builder that we should fix. 
this proposal aims to return the service account principal which is defined in the customer account and has the "provisioning_role" in the role name (which is not great, we can change that)

> if it's our service account we want, we probably don't need round trip to GCP

I agree, another option of course is to avoid the round trip and just return our service account principal but I did not know if it is a good one because it is a hard-coded/config value 

> if it's our service account we return (and then compare with image.shared_with) how do we guide users that they need to share the image with this account (how would they know)?
This is the issue that we would need to resolve, one of the option is to ass our service account to be shared with the image. 
 I wanted you to see the UI works without problems when the image is shared with the service account for now.

ezr-ondrej: Let's just circut it and return our service account directly, without the GCP roundtrip and we can take it from there :)
I guess it should be easier solution, right?

ezr-ondrej (commented):
BTW we do have exmples for AWS/Azure, do we want to add example here as well, or are we going to do this later (if later, shall we have Jira)?

lzap:
I suggest to replace "email" with "principal". From [documentation](https://cloud.google.com/iam/docs/overview?&_ga=2.199813738.-518601638.1686651754#concepts_related_identity):

_Principal. A principal can be a Google Account (for end users), a service account (for applications and compute workloads), a Google group, or a Google Workspace account or Cloud Identity domain that can access a resource. Each principal has its own identifier, which is typically an email address._

As we discussed on the meeting, let’s get this ready to be merged so we can gradually improve on stage towards the ultimate goal when it fully works there.

adiabramovitch:
> I suggest to replace "email" with "principal". From [documentation](https://cloud.google.com/iam/docs/overview?&_ga=2.199813738.-518601638.1686651754#concepts_related_identity):
> 
> _Principal. A principal can be a Google Account (for end users), a service account (for applications and compute workloads), a Google group, or a Google Workspace account or Cloud Identity domain that can access a resource. Each principal has its own identifier, which is typically an email address._
> 
> As we discussed on the meeting, let’s get this ready to be merged so we can gradually improve on stage towards the ultimate goal when it fully works there.

Image Builder uses email, so I wanted to be consistent but I can change it to principal.

lzap:
I don’t insist, not a big deal.

lzap (approved):
Looks good.

internal/clients/http/gcp/gcp_client.go
lzap: Looks like this was dropped from the patch. But leaving the merge on @ezr-ondrej thanks.

internal/clients/stubs/gcp_stub.go
```diff
@@ -51,6 +51,10 @@ func getServiceGCPClientStub(ctx context.Context) (si clients.ServiceGCP, err er
 	return si, err
 }
 
+func (moc *GCPClientStub) GetServiceAccountPrincipal(ctx context.Context) clients.GcpPrincipal {
+	return clients.GcpPrincipal("serviceAccount:provisioning-stage@red-hat-hcc.iam.gserviceaccount.com")
```
ezr-ondrej, line 55: Not a huge fan of hardcoding it, don't we have the information in the GCP JSON? if it would not be hard to get it from there, I'd very much prefer that, but if we can't do it easily, hardcoding it is fine :)

ezr-ondrej (commented):
Do I get it correctly, that we need this just really temporarily, if we convince IB to give us permissions on their account, we do not need any of this, right?

adiabramovitch:
> Do I get it correctly, that we need this just really temporarily, if we convince IB to give us permissions on their account, we do not need any of this, right?

Right, the image will be automatically shared with our service account and we won't need to do this validation in the frontend

internal/clients/stubs/gcp_stub.go
adiabramovitch: Done

ezr-ondrej: gotcha, might be better not to include real data in stub, but not a big deal :)

adiabramovitch:
> > Do you mean the provisioning project id?
> 
> I mean the project id associated with the source, we do associate project_id right? I don't have a usecase for it now, so might be bit pointless ATM :)


The project ID is saved in a source, so it is a bit redundant to expose the project id (of the customer) in two different ways.

ezr-ondrej:
Yeah, agreed, we are doing it for AWS already, but there we at least have a use case, for GCP we don't, so lets not add that :)

ezr-ondrej (approved):
Nice and easy, @lzap we do not need this endpoint after all so merging this empty response. If you've got concerns with that (I see it might be bit antipatern here), lets discuss what are the other ways to solve this?

PR #572, opened 2023-06-19
refactor: Add username to pubkey body

adiabramovitch:


lzap:
Can you change this to use hardcoded `gce-user` or `google-user` instead? We had a discussion on Slack yesterday and this looks like a more straightforward and easy to understand solution which also matches what we have for AWS and Azure.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -307,3 +309,9 @@ func (c *gcpClient) GetInstanceDescriptionByID(ctx context.Context, id, zone str
 	}
 	return &instanceDesc, nil
 }
+
+func getPubkeyWithUsername(ctx context.Context, pubkey string) string {
+	pubkey = strings.TrimSpace(pubkey)
+	parts := strings.Split(pubkey, " ")
+	return fmt.Sprintf("%s:%s %s", "gcp-user", parts[0], parts[1])
```
lzap, line 316: Interesting :-D Why not:

```suggestion
	return fmt.Sprintf("gcp-user:%s %s", parts[0], parts[1])
```

lzap (changes requested):
No unit test, the function will panic if we ever have incorrectly formatted key in the database.

I think the function could actually be a method of Pubkey - could be useful someday for other clouds.

internal/models/pubkey_model.go
```diff
@@ -56,3 +58,9 @@ func (pk *Pubkey) FindAwsFingerprint(ctx context.Context) string {
 
 	}
 }
+
+func (pk *Pubkey) GetPubkeyWithUsername(ctx context.Context) string {
```
lzap, line 62:
```suggestion
func (pk *Pubkey) BodyWithUsername(ctx context.Context) string {
```

Maybe?

internal/models/pubkey_model.go
```diff
@@ -56,3 +58,9 @@ func (pk *Pubkey) FindAwsFingerprint(ctx context.Context) string {
...
+	pubkey := strings.TrimSpace(pk.Body)
+	parts := strings.Split(pubkey, " ")
+	return fmt.Sprintf("gcp-user:%s %s", parts[0], parts[1])
```
lzap, line 65: This can still panics, can we either check the boundaries or not to add `parts[1]`.

internal/models/pubkey_model.go
lzap: To give you context: getters wit Get are not welcomed in Go.

https://go.dev/doc/effective_go#Getters

internal/models/pubkey_model.go
ezr-ondrej: yeah, I think we are missing a unit test case where the key is missing the optional (comment) part of the key

adiabramovitch: We do not care about the comment (column 2), I adding a gcp-user instead.
Either way I added a condition regarding the size of `parts`.

lzap:
/retest

This should solve the blocker for you @akhil-jha not sure why it failed tho.

ezr-ondrej:
Seemed to have succeeded now, @akhil-jha can we merge, or are we waiting for something? :)

PR #573, opened 2023-06-20
chore: improve job timeout logging

lzap:
It appears that we log "Job cancelled" each time a job finishes. Even for successful jobs. This fixes it.

image


ezr-ondrej (approved):
Right thanks! :+1:

PR #574, opened 2023-06-22
fix(HMS-2002): proper launch statistics

lzap:
Numbers are pretty much unusable for the next 24 hours after deployment. On stage, this is almost every day, on prod this skews the numbers.

Let’s calculate these numbers properly via a background job on regular intervals and export those values as prometheus gauges.

This patch adds database connection and a goroutine to the statuser (which we only have single instance of) which performs two DB queries in every 30 minutes and updates prometheus gauges with the numbers. It will allow us to see:

* Number of success/failed launches per hyperscaler in the last 24 hours
* Number of success/failed launches per hyperscaler in the last 28 days
* Duration of the statistics routine (just in case it gets slower in the future)

### DIFF
--- a/cmd/pbackend/statuser.go
+++ b/cmd/pbackend/statuser.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/db"
 	"github.com/RHEnVision/provisioning-backend/internal/identity"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/notifications"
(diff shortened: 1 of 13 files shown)


cmd/pbackend/statuser.go
```diff
@@ -292,19 +294,19 @@ func statuser() {
...
-				logger.Warn().Err(err).Msg("Not starting metrics service, port already in use")
-			} else if !errors.Is(err, http.ErrServerClosed) {
-				logger.Warn().Err(err).Msg("Metrics service listen error")
+			if errors.As(listenErr, &errInUse) && errInUse == syscall.EADDRINUSE {
+				logger.Warn().Err(listenErr).Msg("Not starting metrics service, port already in use")
+			} else if !errors.Is(listenErr, http.ErrServerClosed) {
+				logger.Warn().Err(listenErr).Msg("Metrics service listen error")
 			}
 		}
 	}()
```
lzap, line 312: Linter wanted me to change because of variable shadowing, probably a new linter enabled after the upgrade.

This is unrelated rename of `err` variables.

internal/metrics/registration.go
```diff
@@ -3,13 +3,29 @@ package metrics
...
 func RegisterStatuserMetrics() {
-	prometheus.MustRegister(TotalSentAvailabilityCheckReqs, AvailabilityCheckReqsDuration, TotalInvalidAvailabilityCheckReqs, CacheHits)
+	prometheus.MustRegister(
+		TotalSentAvailabilityCheckReqs,
+		AvailabilityCheckReqsDuration,
+		TotalInvalidAvailabilityCheckReqs,
+		CacheHits,
+		DbStatsDuration,
+		Reservations24hCount,
+		Reservations28dCount,
```
lzap, line 13: Added these three metrics, everything else is just reformatting.

ezr-ondrej (commented):
If we feel this is a long term solution, I'd feel like the statuser is not a great place to do this :(

ezr-ondrej (commented):
Looks correct tho :)

lzap:
So we thought it would be better to put this (and redis stats) into a separate container. I will refactor it.

lzap:
Rebased, we now have a new pod named "stats" which is dedicated for statistics. It performs:

* Redis job queue statistics
* Database SLO reservations statistics (24h, 28d)

To test this, create the following config:

```
STATS_JOBQUEUE_INTERVAL=2s
STATS_RESERVATIONS_INTERVAL=5s
```

Then configure redis job queue, start "stats" process first and then "api". It will fail to bind prometheus port but that is expected. Now create noop reservation and observe Redis queue metric also DB metrics will show some numbers. Then launch "worker" process and try again, queue should be at zero.

I slightly decreased our memory requirement from 1GB down to few hundred megabytes. Our pods seem to have <50MB of memory in all environments, this will save some resources and also make room for the new stats pod. I introduced also "_SMALL" variants of CPU/MEMORY variables, these are both used for statuser and stats pods to futher save resources as these pods will be doing nothing but waiting for 99% of the time.

ezr-ondrej (approved):
Except the linter this looks ready to me :+1:

lzap:
Rebased.

adiabramovitch:
> Rebased.

Linter failing

lzap:
OMG thanks :-)

lzap:
I see `Name or service not known` error for the GCP test. Not sure if that is relevant tho.

ezr-ondrej:
> I see `Name or service not known` error for the GCP test. Not sure if that is relevant tho.

I guess not, but I'd like to wait for stable pipeline here as it is quite a big change :thinking:

lzap:
Looks green, tho the RHTAP skip makes it less obvious.

ezr-ondrej (approved):
Lets get it in :+1:
Thanks @lzap ! :+1:

lzap:
Thanks dashboard changes tomorrow hopefully.

PR #575, opened 2023-06-22
fix(HMS-1783): allow 2 reservations in one second

lzap:
This caps reservation creation pace per account at 2 operations per second. You can test this via noop reservation, if you create them too quickly this error is returned:

```
{
  "msg": "DAO error: create noop reservation",
  "trace_id": "af0ecea69cbea98d45621f16b86e22c4",
  "error": "failed to create reservation record: failed to create reservation record: ERROR: too many reservations (exceeded 2 per second) (SQLSTATE P0001)",
  "version": "ecc4",
  "build_time": "2023-06-21T15:07:50Z"
}
```

I believe this is the most efficient and consistent way of doing that. I considered implementing this on the API layer but api workers would need to be sharing data via Redis and that is insanely complex compared to just a three line trigger in the database. It has many other advantages as well, it is fast and transaction safe. The only drawback is that some primary keys can be skipped so there can be gaps which is pretty normal in relation databases.


lzap:
Ah sorry did a last minute change and `3:43PM FTL Error running migration error="unable to perform migration: 020_reservations_rate_limit.sql: ERROR: syntax error at or near \"TRIGGER\" (SQLSTATE 42601)" binary=migrate hostname=provisioning-backend-api-5cf7bdcbf9-vzbh6 version=7bfd`

lzap:
Postgres v13 does not support `OR REPLACE` for `TRIGGER` statement. It works locally since latest version does have it. Fixed.

akhil-jha:
Should this is tested by someone from performance team?

lzap:
> Should this be tested by someone from performance team?

Tests were performed and we saw AWS throttling down which ended up with failed reservations. So I created this patch in order to prevent that.

I think the performance team only tests on stage environment anyway.

I added a note to the description tho, this change will not prevent throttling when workers are down for maintenance and then they start up. But I think this is a special case which we don’t need to worry about at this point.

ezr-ondrej:
> Should this be tested by someone from performance team?

Performance team will add higher level testing, I would still love us add low level test on that data layer just creating reservation twice in a row and verifying it fails the second time correctly.
That way we would also test it throws the exception we expect it to throw.

@lzap WDYT?

lzap:
I can create an integration test if that is what you want.

ezr-ondrej:
> I can create an integration test if that is what you want.

That was what I was hoping for, but if you don't think it's useful, I'm ok to merge without it :)

lzap:
I can spam the database in a test but it will be a bit bad test - theoretically DB can lag for a moment and it can sometimes "randomly" pass (thus fail).

There is no need to rush this one in, if the performance team can do another round of testing on two accounts we can merge afterwards. One thing is important - once this is merged it is not easy to disable it or even make it configurable (e.g. not active on stage).

ezr-ondrej:
> theoretically DB can lag for a moment and it can sometimes "randomly" pass (thus fail).

Theoretically, but for a full second? That should be quite rare, shouldn't it?
I'm worried I'm missing something here and thus bit afraid to merge, especially we can't go back :thinking:

I'm happy to add the test and figure out a way out if we see failures in the future.

lzap:
I agree doing this in the database is not perfect but the cost of implementing this on the API level compensates this quite easily I think. Ok how about this.

I refactored the function and split it into two, one function defies a constant so it is easier for us to change the limit in the future. I also updated the trigger in a way that it now also takes provider into account, it does not take source ID into consideration at the moment tho (that would be more complex, doable, but I think preliminary at this point).

I am also updating the initial limit from 2 to 5. Let's say this is a DoS protection rather than being aligned to AWS hard limit. This allows the performance team to still conduct (slower-paced) test if needed.

ezr-ondrej:
I know I'm taking a long time to review this, but ppl (the industry) are moving away DB triggers and functions and I belive it's because it's messy to deal with, so I just have hard time to process this patch :(
I understand this is great alternative compared to other solutions, I just wonder if it's worth it, so I'm trying to think through it properly. Sorry!!

I've tried to write a test for this, but it failed, I thought this shall be limited per account, but it isnt? :thinking: 

The test I wrote for reference:

```go
func TestReservationRate(t *testing.T) {
	reservationDao, ctx := setupReservation(t)

	t.Run("errors out on two fast consecutive reservation on account", func(t *testing.T) {
		defer reset()
		res := newNoopReservation()
		err := reservationDao.CreateNoop(ctx, res)
		require.NoError(t, err)

		for i := 0; i < 4; i++ {
			res = newNoopReservation()
			err = reservationDao.CreateNoop(ctx, res)
			require.NoError(t, err)
		}

		res = newNoopReservation()
		err = reservationDao.CreateNoop(ctx, res)
		require.ErrorContains(t, err, "failed to create reservation record")
	})

	t.Run("does not error out for quick reservations from different account", func(t *testing.T) {
		defer reset()
                res := newNoopReservation()
                err := reservationDao.CreateNoop(ctx, res)
		require.NoError(t, err)

		for i := 0; i < 4; i++ {
			res = newNoopReservation()
			err = reservationDao.CreateNoop(ctx, res)
			require.NoError(t, err)
		}

		res = newNoopReservation()
		res.AccountID = 2
		err = reservationDao.CreateNoop(ctx, res)

                / ------- IT FAILED HERE -------
		require.NoError(t, err)
	})
}
```

lzap:
> the industry is moving away

SQL databases are unbeaten for the moment and while some provide exotic features, I find triggers quite useful for data consistency and I believe they can still prove valuable in today's world. Having said that, this use case is really on the edge but I could not find a better solution.

Being uncomfortable with unknown language is okay, I understand the your concern. Tho I am not introducing any critical business logic or complex code. It is a single query with an exception which nicely propagates back to the client. And moving away from this is as easy as a two lines migration that will remove the function and the trigger.

Looks like something is wrong with your code, I cannot see it but account number one is being passed in:

```
12:01PM ERR Query account_id=1 args=[1,1,1,["Test step"],"Created"] binary=-test.v err="ERROR: too many pending reservations (5) for this provider (maximum 5 per second) (SQLSTATE P0001)" hostname=mone.home.lan module=pgx pid=13387 sql="INSERT INTO reservations (provider, account_id, steps, step_titles, status)\n\t\tVALUES ($1, $2, $3, $4, $5) RETURNING id, created_at"
```

lzap:
Oh AccountID gets overwritten in the DAO function that is why :-)

ezr-ondrej:
> Looks like something is wrong with your code, I cannot see it but account number one is being passed in:

Ah, of course, the code (correctly) ignores accountID passed and takes the one from context :+1:

lzap:
I added the integration test and fixed the problem with account id.

internal/db/seeds/integration.sql
```diff
@@ -17,7 +17,9 @@ $do$;
 
 -- Seed the data
 INSERT INTO accounts(id, account_number, org_id)
-VALUES (1, '1', '1')
+VALUES
+  (1, '1', '1'),
+  (2, '2', '2')
```
ezr-ondrej, line 22: I tried that, but it causes failure in account test, which is bit unfortunate, but I did not want to go to the rabbit hole and fix it, so this is what I come up with, feel free to use, but ofc I'm happy to just drop it ^_^ #598

ezr-ondrej: eh your account tests do not fail :thinking: so I just did something random, I shall stop try to code it appears :facepalm:

internal/dao/tests/account_test.go
```diff
@@ -86,11 +86,11 @@ func TestAccountList(t *testing.T) {
...
-		accounts, err := accDao.List(ctx, 1, 1)
+		accounts, err := accDao.List(ctx, 1, 2)
 		require.NoError(t, err)
 		assert.Equal(t, a2.OrgID, accounts[0].OrgID)
 		assert.Equal(t, a2.AccountNumber, accounts[0].AccountNumber)
-		accounts, err = accDao.List(ctx, 1, 2)
+		accounts, err = accDao.List(ctx, 1, 3)
 		require.NoError(t, err)
 		assert.Equal(t, a3.OrgID, accounts[0].OrgID)
 		assert.Equal(t, a3.AccountNumber, accounts[0].AccountNumber)
```
lzap, line 96: Tests are working because of these changes. These are little bit clunky but I am okay with that.

internal/dao/tests/reservation_test.go
```diff
@@ -348,3 +354,42 @@ func TestReservationFinish(t *testing.T) {
...
+
+	t.Run("does not error out for quick reservations from different account", func(t *testing.T) {
+		defer reset()
+		res := newNoopReservation()
+		err := reservationDao.CreateNoop(ctx, res)
+		require.NoError(t, err)
+
+		for i := 0; i < 4; i++ {
+			res = newNoopReservation()
+			res.Reservation.AccountID = 2
```
ezr-ondrej, line 388: This will be IMO ignored, we need to change the context instead

internal/dao/tests/reservation_test.go
```diff
@@ -348,3 +354,42 @@ func TestReservationFinish(t *testing.T) {
...
+			err = reservationDao.CreateNoop(ctx, res)
+			require.NoError(t, err)
+		}
+
+		require.NoError(t, err)
```
ezr-ondrej, line 393: we do not do the 6th reservation here under different account so we are not really testing much in this test.
There is no clear way to break this test if the tested functionality does not work as expected IMO.

ezr-ondrej (approved):
I'm ok with this, to take remove the trigger with another migration is not that bad.

The test with two accounts has no reason to fail tho so probably not testing what we want it to test, commented inline :)

internal/dao/tests/reservation_test.go
lzap: Oh this is a leftover, it is done above.

ezr-ondrej (approved):
Thanks ! :+1: Lets see it in stage for a week :+1:

ezr-ondrej:
QE ack is not needed here, change only in description of the spec, no api changes.

PR #576, opened 2023-06-23
feat(HMS-1550): improve initialization code for kafka

lzap:
Just put all the code on the same level.

I tried to introduce a helper function but it did not work out, most of other initializations would not benefit from this. Introducing it just for these four lines is not necessary.

### DIFF
--- a/cmd/pbackend/api.go
+++ b/cmd/pbackend/api.go
@@ -71,7 +71,10 @@ func api() {
 		if err != nil {
 			logger.Fatal().Err(err).Msg("Unable to initialize the platform kafka")
 		}
-		notifications.Initialize(ctx)
+
+		if config.Application.Notifications.Enabled {
+			notifications.Initialize(ctx)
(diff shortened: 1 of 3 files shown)


cmd/pbackend/api.go
```diff
@@ -71,7 +71,10 @@ func api() {
 		if err != nil {
 			logger.Fatal().Err(err).Msg("Unable to initialize the platform kafka")
 		}
-		notifications.Initialize(ctx)
```
ezr-ondrej, line 74: Shall we drop the `Initialize` function then?

lzap: Sorry what? I just wrapped it with an if-statement.

ezr-ondrej: Wow, I should learn how to read :face_in_clouds:

ezr-ondrej (approved):

PR #577, opened 2023-06-26
chore(HMS-1847): A job for cloud resources cleanup

mayurilahane:
Ref- https://issues.redhat.com/browse/HMS-1847

Utilize CloudWash and Tekton instance to regularly remove unused resources in our cloud accounts.

### DIFF
--- a/deploy/cloudwash_job.yaml
+++ b/deploy/cloudwash_job.yaml
@@ -1,0 +1,70 @@
+---
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: provisioning-cleanup-resources
+objects:
+- apiVersion: batch/v1
+  kind: Job
+  metadata:
+    name: provisioning-cleanup-${IMAGE_TAG}-${UID}
+    annotations:
+      "ignore-check.kube-linter.io/no-liveness-probe": "probes not required on Job pods"
+      "ignore-check.kube-linter.io/no-readiness-probe": "probes not required on Job pods"
+  spec:
+    backoffLimit: 0
+    template:
+      spec:
+        imagePullSecrets:
+        - name: quay-cloudservices-pull
+        restartPolicy: Never
+        containers:
+        - name: provisioning-cleanup-${IMAGE_TAG}-${UID}
+          image: quay.io/redhatqe/cloudwash
+          imagePullPolicy: Always
+          args:
+          - swach -d azure --all # remove "-d (dry run)" after overall review.
+          env:
+          - name:  CLEANUP_PROVIDERS_AZURE_AUTH_CLIENT_ID
+            valueFrom:
+              secretKeyRef:
+                key: azure_client_id
+                name: iqe-vault
+                optional: true
+          - name: CLEANUP_PROVIDERS_AZURE_AUTH_SECRET_ID
+            valueFrom:
+              secretKeyRef:
+                key: azure_client_secret
+                name: iqe-vault
+                optional: true
+          - name: CLEANUP_PROVIDERS_AZURE_AUTH_TENANT_ID
+            valueFrom:
+              secretKeyRef:
+                key: azure_tenant_id
+                name: iqe-vault
+                optional: true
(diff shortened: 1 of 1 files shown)


mayurilahane:
retest/

deploy/cloudwash_job.yaml
```diff
@@ -0,0 +1,72 @@
...
+            valueFrom:
+              secretKeyRef:
+                key: azure_tenant_id
+                name: iqe-vault
+                optional: true
+          - name: CLEANUP_PROVIDERS_AZURE_AUTH_SUBSCRIPTION_ID
+            valueFrom:
+              secretKeyRef:
+                key: subscription_id
+                name: iqe-vault
```
ezr-ondrej, line 52: Do we have secret in vault called iqe-vault containing all the values in json? :open_mouth:

deploy/cloudwash_job.yaml
```diff
@@ -0,0 +1,72 @@
...
+      "ignore-check.kube-linter.io/no-readiness-probe": "probes not required on Job pods"
+  spec:
+    backoffLimit: 0
+    template:
+      spec:
+        imagePullSecrets:
+        - name: quay-cloudservices-pull
+        restartPolicy: Never
+        containers:
+        - name: provisioning-cleanup-${IMAGE_TAG}-${UID}
```
ezr-ondrej, line 22: Why is the IMAGE_TAG in the name here? :thinking:

ezr-ondrej (commented):
I do not understand tekton pipelines well, so pardon my dumb questions, I do wonder what this should actually be doing? :thinking:

deploy/cloudwash_job.yaml
mayurilahane: Yep here https://vault.devshift.net/ui/vault/secrets/insights/list/secrets/qe/qa/provisioning/

deploy/cloudwash_job.yaml
mayurilahane: So, am not sure either but it should be for the container's unique name.

I think it's like default things which you need to add for every job that will run inside the container /?

I haven't added it at first but then @tpapaioa corrected me there.

ezr-ondrej: Ok :shrug: I'm bit confused why it is an IMAGE_TAG as we do not use it as IMAGE_TAG, but there might be something I do not understand :shrug:

deploy/cloudwash_job.yaml
ezr-ondrej: it's called azure there :thinking: we rename it to iqe-vault in app-interface?
I'd say more descriptive name would be slightly better :stuck_out_tongue:

deploy/cloudwash_job.yaml
mayurilahane: I will let @tpapaioa answer this. :) 
Thanks!

deploy/cloudwash_job.yaml
mayurilahane: Yeah I think you are right 
I can confirm this.

I was mostly studying other files and writing our own so they have the same thing `iqe vault` everywhere. 
then I thought it might be some sort of keyword to connect vault.
 
 for example here 
 https://github.com/RHEnVision/provisioning-backend/blob/main/deploy/testing-integration.yaml#L54

deploy/cloudwash_job.yaml
tpapaioa: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/continuous-testing-in-app-interface.md#job-definition-and-naming

mayurilahane:
The whole discussion around this is [here](https://redhat-internal.slack.com/archives/C04B2FX09FV/p1686855265011929)

mayurilahane:
I think we can merge it now.
cc: @ezr-ondrej @tpapaioa

ezr-ondrej (approved):
Lets try :shrug: this should not affect the deployment, so lets just test it out.

lzap:
Looks like deployment is green for this one.

PR #583, opened 2023-06-30
chore: update dashboard with reservations stats

lzap:
Dashboard change which introduces the newly calculated metrics for 24h/28d which are accurate. They recalculate every 30 minutes. There is also a duration showing how long it took to do the queries just in case this gets slower due to bug in SQL or missing index.


lzap:
I will work on alerts the week after the next week.

adiabramovitch (approved):

PR #584, opened 2023-07-06
feat(HMSPROV-2088): refacor notification's context message

amirfefer:
Currently, we send the generic reservation payload as the notification's context. some data is redundant (i.e. step titles, steps/step) others misleading (i.e. success flag), or might confuse - the provider's id (2) instead of the provider's string (aws).

We should send the minimal information that also will be beneficial, 
This PR replaces the reservation's payload with the `NotificationContext` struct which includes the launch_id and the provider's string, both are used in email templates.

### DIFF
--- a/internal/kafka/notification_msg.go
+++ b/internal/kafka/notification_msg.go
@@ -22,6 +22,11 @@ type NotificationEvent struct {
 	Payload json.RawMessage `json:"payload"`
 }
 
+type NotificationContext struct {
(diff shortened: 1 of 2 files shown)


adiabramovitch:
linter issues

amirfefer:
there are no linter issues, it was a timeout issue

lzap (approved):
Once per month, the linter github action is configured to delete cache which leads to linter exceeding 5 minutes limit on GHA which leads to a failure.

PR #587, opened 2023-07-10
fix(HMS-2002): fix job queue duration registration

lzap:
In the recent changes, I accidentally moved job duration from worker to stats process. The metric no longer exist in Prometheus on stage, this is because this needs to be registered for worker processes where jobs are actually processed not the stats process where no jobs are executed.

This fixes it, after the merge, the metric will re-appear on the dashboard.


ezr-ondrej:
We will need to rebase this for the workaround to get in IMO, or does the bot rebases on main?

akhil-jha:
Needs a rebase, yes.

ezr-ondrej (approved):

PR #592, opened 2023-07-11
refactor: Add GCP job test

adiabramovitch:
This PR  adds missing tests for GCP job: inserting instances and fetching instances description.

### DIFF
--- a/internal/clients/stubs/gcp_stub.go
+++ b/internal/clients/stubs/gcp_stub.go
@@ -2,11 +2,16 @@ package stubs
 
 import (
 	"context"
+	"fmt"
+	"strconv"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 )
 
 type gcpCtxKeyType string
+
+var ipCounter = 10
 
 var (
 	gcpCtxKey        gcpCtxKeyType = "gcp-interface"
@@ -14,13 +19,19 @@ var (
 )
 
 type (
-	GCPClientStub        struct{}
+	GCPClientStub struct {
+		Instances []*string
+	}
 	GCPServiceClientStub struct{}
 )
 
 func init() {
-	clients.GetGCPClient = getCustomerGCPClientStub
+	clients.GetGCPClient = newGCPCustomerClientStub
 	clients.GetServiceGCPClient = getServiceGCPClientStub
+}
+
+func newGCPCustomerClientStub(ctx context.Context, auth *clients.Authentication) (clients.GCP, error) {
+	return getCustomerGCPClientStub(ctx, auth)
 }
 
 // GCPServiceClient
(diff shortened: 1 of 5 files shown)


internal/clients/stubs/gcp_stub.go
```diff
@@ -60,19 +79,23 @@ func (mock *GCPClientStub) Status(ctx context.Context) error {
 }
 
 func (mock *GCPClientStub) GetInstanceDescriptionByID(ctx context.Context, id, zone string) (*clients.InstanceDescription, error) {
```
ezr-ondrej, line 81: I'd expect this method to actually search if the Instance is in the `Instaces` list, so it could actually fail if the instance does not exists.... creating the instance in `Get` method sounds weird :thinking:

internal/clients/stubs/gcp_stub.go
```diff
@@ -60,19 +79,23 @@ func (mock *GCPClientStub) Status(ctx context.Context) error {
...
 }
 
 func (mock *GCPClientStub) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchTemplate, error) {
 	return nil, nil
 }
 
 func (mock *GCPClientStub) InsertInstances(ctx context.Context, params *clients.GCPInstanceParams, amount int64) ([]*string, *string, error) {
-	return nil, nil, NotImplementedErr
+	ids, err := mock.ListInstancesIDsByLabel(ctx, "1")
+	mock.Instances = ids
```
ezr-ondrej, line 93: Shall we rather create the ids here in insert method and in List method just return those?

It seems to me, we are completely ignoring the count here, which does not sound like a real implementation.. If I'd change the amount of instances to create in test to 3, it would fail, right?

internal/jobs/launch_instance_azure_test.go
```diff
@@ -45,7 +45,7 @@ func prepareAzureReservation(t *testing.T, ctx context.Context, pk *models.Pubke
 	}
 	reservation.AccountID = 1
 	reservation.Status = "Created"
-	reservation.Provider = models.ProviderTypeAWS
```
ezr-ondrej, line 48: :orange_heart: good catch!

ezr-ondrej (commented):
Thanks for adding the tests, :orange_heart: it!

The stub seems a bit hacky to me tho :thinking:

internal/clients/stubs/gcp_stub.go
adiabramovitch: Thank you for the feedback I have disregarded the amount, now I understand what didn't make sense to me 😞

internal/clients/stubs/gcp_stub.go
```diff
@@ -60,19 +84,32 @@ func (mock *GCPClientStub) Status(ctx context.Context) error {
...
 }
 
 func (mock *GCPClientStub) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchTemplate, error) {
 	return nil, nil
 }
 
 func (mock *GCPClientStub) InsertInstances(ctx context.Context, params *clients.GCPInstanceParams, amount int64) ([]*string, *string, error) {
-	return nil, nil, NotImplementedErr
+	for i := 0; i < int(amount); i++ {
+		ID := fmt.Sprintf("300394200587658274%s", strconv.Itoa(idCounter))
```
ezr-ondrej, line 103: We wouldn't need the counter, if we use the lenght of the array as lzap pointed out to me in one PR :)

https://github.com/RHEnVision/provisioning-backend/blob/6615db0a2e5cf96cafff28859ac859b285923dff/internal/clients/stubs/azure_stub.go#L72

```suggestion
		ID := fmt.Sprintf("300394200587658274%s", strconv.Itoa(strconv.Itoa(len(mock.Instances)+1)))
```

ezr-ondrej (approved):
Great! Thanks! :orange_heart:

PR #593, opened 2023-07-11
feat(HMS-1955): Add fetch instances description step for GCP

adiabramovitch:
Adding fetching instances description step for GCP.


amirfefer (approved):
Thanks @adiabramovitch 👍 
Tested with https://github.com/RHEnVision/provisioning-frontend/pull/281 and works well

lzap (approved):
TY

PR #595, opened 2023-07-12
chore: add 3h reservation rate and fix duration

lzap:
Duration of db stats was not summed, therefore there were incorrect (multiple) results on the dashboard.

Also, this adds a 3h reservation rate next to 24h/28d. So we can observe all the three.

I still plan on working on the alert which is not working correctly, let's use the calculated 24h for that.


ezr-ondrej (approved):
I still can't read this json :/ Lets just test it out :)

PR #596, opened 2023-07-14
chore: fix typo and replace more identifiers

lzap:
Just a typo and two new Sentry replacers.

### DIFF
--- a/internal/jobs/launch_instance_aws.go
+++ b/internal/jobs/launch_instance_aws.go
@@ -140,7 +140,7 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
 			}
 			ec2Name = pubkey.Name
 		} else {
-			return fmt.Errorf("cannot fetch name of pubkey with fingerpring (%s): %w", fingerprint, err)
+			return fmt.Errorf("cannot fetch name of pubkey with fingerprint (%s): %w", fingerprint, err)
 		}
 	} else {
 		logger.Debug().Msgf("Found pubkey by fingerprint (%s) with name '%s'", fingerprint, ec2Name)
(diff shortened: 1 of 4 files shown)


internal/jobs/launch_instance_aws.go
```diff
@@ -140,7 +140,7 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
 			}
 			ec2Name = pubkey.Name
 		} else {
-			return fmt.Errorf("cannot fetch name of pubkey with fingerpring (%s): %w", fingerprint, err)
+			return fmt.Errorf("cannot fetch name of pubkey with fingerprint (%s): %w", fingerprint, err)
```
ezr-ondrej, line 143: :bulb: what if we log the fingerprint, but do not include it in the error message itself?
We even have it twice there, because we (me) are also adding it to the parent error: https://github.com/lzap/provisioning-backend/blob/80fe90e3dfb0755e6b55ab408c16d9c75c01bcd0/internal/clients/http/ec2/ec2_client.go#L210

If we log it there, wouldn't that be enough?

ezr-ondrej (approved):
I'm not happy adding more and more replacers, but if we can't fix the message, lets have a quick fix :)

Thanks for checking up on the errors :orange_heart:

lzap:
Restarted failed linter, it was a timeout.

I can remove fingerprint from the error message and do a separate debug log statement with the data if you want, but filtering seems like a practical approach to me and you approved it in. What is concerning now for you? Note it always compiles down to just a single regular expression, adding more replacers does not make it anyhow slower.

ezr-ondrej:
> filtering seems like a practical approach to me and you approved it in. What is concerning now for you?

That's why I do not like merge stuff I see as workarounds, those tend to grow :stuck_out_tongue:
I agreed to use it on errors we did not control.
Main concern of mine is that it's pretty complex code, not many ppl in the future will undrstand, the more patterns there is, the harder it is to read IMO.

I see it as ok approach in cases we can't do any better, but I loved your effort to clean up our messages from the dynamic stuff in the first place and it seemed like better approach where we can do something about it :)

Here I see error message that is wrong to the guidelines (it has dynamic information and it even has it twice), so I'd like to improve on it.

But let me merge and propose the patch to fix it myself, I'll try to dig in the code a bit to feel what I'm talking about bit more :)

ezr-ondrej (approved):

PR #599, opened 2023-07-14
chore: remove dynamic fingerprint from error message

ezr-ondrej:
I believe this is cleaner, as the dynamic fingerprint in the error message causes trouble and we do not really care about it.
I'm happy to down level the message even as I think we already log the job error as error.


lzap (approved):
Better indeed. Thanks.

PR #601, opened 2023-07-18
ci: Parametrize replica counts

ezr-ondrej:
Parametrize api, worker and stats serving replicas. Statuser is always needed and there is no need to scale it up or down for now.

We want to allow scaling down of replicas for ephemeral environment.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -198,7 +198,7 @@ objects:
                 cpu: ${CPU_REQUESTS}
                 memory: ${MEMORY_REQUESTS}
         - name: stats
-          replicas: 1
+          replicas: ${{STATS_REPLICAS}}
           metadata:
             annotations:
               ignore-check.kube-linter.io/minimum-three-replicas: "stats pod runs in a single instance"
@@ -490,6 +490,9 @@ parameters:
   - description: Memory limit for each pod (pod restart)
     name: MEMORY_LIMIT
     value: 600Mi
+  - description: Amount of replicas for pod serving stats to optionaly scale to 0
+    name: STATS_REPLICAS
+    value: "1"
   - description: Image tag
     name: IMAGE_TAG
     required: true


deploy/clowdapp.yaml
```diff
@@ -25,7 +25,7 @@ objects:
         iqePlugin: provisioning
       deployments:
         - name: worker
-          replicas: 3
+          replicas: ${WORKER_REPLICAS}
```
Victoremepunto, line 28:
```suggestion
          replicas: ${{WORKER_REPLICAS}}
```

deploy/clowdapp.yaml
```diff
@@ -198,7 +198,7 @@ objects:
                 cpu: ${CPU_REQUESTS}
                 memory: ${MEMORY_REQUESTS}
         - name: stats
-          replicas: 1
+          replicas: ${STATS_REPLICAS}
```
Victoremepunto, line 201:
```suggestion
          replicas: ${{STATS_REPLICAS}}
```

deploy/clowdapp.yaml
```diff
@@ -247,7 +247,7 @@ objects:
                 cpu: ${CPU_REQUESTS}
                 memory: ${MEMORY_REQUESTS}
         - name: api
-          replicas: 3
+          replicas: ${API_REPLICAS}
```
Victoremepunto, line 250:
```suggestion
          replicas: ${{API_REPLICAS}}
```

Victoremepunto (changes requested):
this requires the use of ${{PARAMETER}} syntax to force JSON/YAML format. See https://docs.openshift.com/container-platform/4.12/openshift_images/using-templates.html#templates-writing-parameters_using-templates

lzap (approved):
Sure thanks. Why don’t we do the same for worker/api these could go down from 6 to just 2 saving a lot of resources. I guess this is just a test first and you want to do a followup?

PR #602, opened 2023-07-18
chore: use sum for provisioning db duration rate

lzap:
SSIA

Small typo, fixes the problem after containers are restarted - the pane shows multiple values instead of one.

image


ezr-ondrej (approved):
Thanks! :+1:

PR #603, opened 2023-07-18
fix(HMS-2175): add AWS resource tag

lzap:
We share a team account and sometimes it is not clear who created particular instance. This should make finding owners more easier. It adds AWS instance resource tag in the form of "pr-ID" where "pr" stands for production (or st for stage, ep for ephemeral, de for development) and ID is reservation ID.

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"strconv"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
 	"github.com/RHEnVision/provisioning-backend/internal/clients/http"
@@ -367,7 +368,7 @@ func (c *ec2Client) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchT
 	return res, nil
 }
 
-func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanceParams, amount int32, name *string) ([]*string, *string, error) {
+func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanceParams, amount int32, name *string, reservation *models.AWSReservation) ([]*string, *string, error) {
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "RunInstances")
 	defer span.End()
 
@@ -402,6 +403,10 @@ func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanc
 					{
 						Key:   ptr.To("Name"),
 						Value: name,
+					},
+					{
+						Key:   ptr.To("rhhc:rid"),
+						Value: ptr.To(config.EnvironmentPrefix("r", strconv.FormatInt(reservation.ID, 10))),
 					},
 				},
 			},
--- a/internal/config/helpers.go
+++ b/internal/config/helpers.go
@@ -26,6 +26,24 @@ func Environment() string {
 	return "dev"
 }
 
+// EnvironmentPrefix wraps an identifier (e.g. id, unique id) in the following way.
+// For production environment, it returns "prefix-identifier". For any other environment,
+// it returns "prefix-identifier-env". Examples:
+//
+// * reservation-14572
+// * reservation-9531-stage
+// * reservation-13-ephemeral
+// * reservation-1-dev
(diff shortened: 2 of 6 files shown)


adiabramovitch (commented):
LGTM, just one question

internal/clients/http/ec2/ec2_client.go
```diff
@@ -367,7 +367,7 @@ func (c *ec2Client) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchT
 	return res, nil
 }
 
-func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanceParams, amount int32, name *string) ([]*string, *string, error) {
+func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanceParams, amount int32, name *string, reservation *models.AWSReservation) ([]*string, *string, error) {
```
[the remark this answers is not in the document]
lzap: I am thinking that we might need additional values from the reservation at some point, so passing the whole type means we will not have to update the function again.

internal/config/helpers.go
```diff
@@ -26,6 +26,11 @@ func Environment() string {
 	return "dev"
 }
 
+// EnvironmentPrefix returns the first N characters of Environment.
+func EnvironmentPrefix(n int) string {
+	return Environment()[0:n]
```
avitova, line 31: We should probably test this for the string length. This could panic badly if `len(Environment() < n`).
Can we test this? I am not sure if we mock functions somewhere in the tests, but it should be doable in Go, right?

lzap: Added test, improved the function.

adiabramovitch (approved):

internal/config/helpers.go
adiabramovitch: I think a >= will do

lzap:
Details :-D

Then `>=` it is. Rebased.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -403,6 +403,10 @@ func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanc
 						Key:   ptr.To("Name"),
 						Value: name,
 					},
+					{
+						Key:   ptr.To("rhhc:rid"),
+						Value: ptr.To(fmt.Sprintf("%s-%d", config.EnvironmentPrefix(2), reservation.ID)),
```
ezr-ondrej, line 408: I'd question the use of putting the environment in the tag, if the main target for this feature are our Customers, they do not care much about environment as that should always be prod. Or is there another reason you want to include environment?
I know it's annoying to deal with instances from stage, but won't that be weird to customers to have "pr-....." everywhere?

ezr-ondrej (commented):
I'm not sure if the format is very informative to our Users, I know it's mostly for internal purposes, but I'd like to hear your thoughts on if we can make this valuable to customers as well, or hide it behind some flag to add it only for RH accounts.

internal/config/helpers.go
```diff
@@ -26,6 +26,17 @@ func Environment() string {
 	return "dev"
 }
 
+// EnvironmentPrefix returns the first N characters of Environment.
+func EnvironmentPrefix(n int) string {
+	env := Environment()
+	if len(env) >= n {
```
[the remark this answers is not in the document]
lzap: Yeah I missed that, fixing.

lzap:
Ok refactored it completely from scratch, no shortening, production contains no suffix. Should be non confusing for production users while readable for us.

internal/config/helpers.go
```diff
@@ -26,6 +26,24 @@ func Environment() string {
...
+// * reservation-13-ephemeral
+// * reservation-1-dev
+func EnvironmentPrefix(prefix, identifier string) string {
+	env := Environment()
+
+	if strings.HasPrefix(env, "prod") || env == "" {
+		return prefix + "-" + identifier
+	}
+
+	return prefix + "-" + identifier + "-" + env
```
avitova, line 44: There could be a way where the prefix and the identifier would be empty strings, and therefore, the prefix would be something like "--" + env, but I guess we do not mind?

avitova (approved):
If we generally do not mind having empty prefix/identifier string and returning that, I think we are good here. :) LGTM

internal/config/helpers.go
lzap: Well trash in - trash out :-) Yeah minor issue.

PR #605, opened 2023-07-19
fix(HMS-2002): calculate db stats immediately

lzap:
The Go built-in time Ticker does not trigger immediately and there does not seem to be an elegant way of doing that. We need the metric to immediately appear otherwise there are gaps up to 30 minutes in Prometheus after each restart of the application. This creates ugly gaphs:

image

Part of this patch are changes of the dashboard to prevent "No data" behavior for cases when there are no pending reservations. In these cases, the panels should read "100%".

Finally there was a typo in label name which made the provider Grafana selector not work. This fixes it.

### DIFF
--- a/deploy/Provisioning-1674548289785.json
+++ b/deploy/Provisioning-1674548289785.json
@@ -107,7 +107,7 @@
             "uid": "${datasource}"
           },
           "editorMode": "code",
-          "expr": "1 - (sum(provisioning_reservations_24h_count{result=\"pending\",provider=~\"$type\"}) / sum(provisioning_reservations_24h_count{provider=~\"$type\"}))",
+          "expr": "1 - (sum(provisioning_reservations_24h_count{result=\"pending\",provider=~\"$type\"} or on() vector(0)) / sum(provisioning_reservations_24h_count{provider=~\"$type\"}))",
           "legendFormat": "__auto",
           "range": true,
           "refId": "A"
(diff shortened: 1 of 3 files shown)


deploy/Provisioning-1674548289785.json
```diff
@@ -2149,9 +2149,9 @@
     "list": [
       {
         "current": {
-          "selected": true,
-          "text": "crcp01ue1-prometheus",
-          "value": "crcp01ue1-prometheus"
+          "selected": false,
+          "text": "crcs02ue1-prometheus",
+          "value": "crcs02ue1-prometheus"
         },
```
lzap, line 2155: This happens when you store JSON when the app is switched on prod/stage. Stage is perhaps better.

lzap:
Added 2nd commit which improves AVC throughtput, the graph was not showing relevant numbers. This now clearly shows number of checks per individual hyperscaler:

image

internal/background/db_stats.go
```diff
@@ -18,15 +18,13 @@ func dbStatsLoop(ctx context.Context, sleep time.Duration) {
 	}()
 	ticker := time.NewTicker(sleep)
 
+	// run one tick immediately to prevent prometheus gaps
+	dbStatsObserveTick(ctx)
```
adiabramovitch, line 22: Is That what you meant by there is no other way for the go builds in tick to start immediately?

lzap: Yeah I need to trigger it manually. There is no function in the standard library to do it AFAIK: https://pkg.go.dev/time#NewTicker

adiabramovitch (approved):
LGTM, let's check it out.

PR #606, opened 2023-07-19
fix(HMS-2178): fix context handling in Kafka consumer

lzap:
It looks like some sources availability checks errors with 404 when we call sources. We are supposed to use the same identity which comes from sources and it looks we do.

However, in logging some org_id/account_numbers are not filled or do not match correctly. It looks like handling of identity is not correct and indeed the main consumer loop incorrectly overwrites the global context over and over again. This leads to random behavior when some messages using incorrect identities or logging incorrect data.

The proper way of dealing with contexts is to always create a copy of the global context, this ensures that only copies are passed into processing functions and they do not interfere with other goroutines.

### DIFF
--- a/internal/kafka/kafka.go
+++ b/internal/kafka/kafka.go
@@ -206,25 +206,29 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, since time.Time
 				msg.Key, msg.Topic, msg.Offset, msg.Partition)
 
 			// build new context - identity and trace id
-			ctx, err = identity.WithIdentityFrom64(ctx, header("x-rh-identity", msg.Headers))
+			newLogger := logger.With().Str("msg_id", random.TraceID().String())
+			newCtx, err := identity.WithIdentityFrom64(ctx, header("X-RH-Identity", msg.Headers))
 			if err != nil {
-				logger.Trace().Msgf("Could not extract identity from context to Kafka message: %s", err)
+				errLogger := newLogger.Logger()
+				errLogger.Warn().Err(err).Msgf("Could not extract identity from context to Kafka message")
+				newCtx = errLogger.WithContext(newCtx)
+			} else {
+				id := identity.Identity(newCtx)
+
+				traceId := trace.SpanFromContext(ctx).SpanContext().TraceID()
+				if !traceId.IsValid() {
+					traceId = random.TraceID()
+				}
+				newCtx = logging.WithTraceId(newCtx, traceId.String())
+
+				newLogger = newLogger.
+					Str("trace_id", traceId.String()).
+					Str("account_number", id.Identity.AccountNumber).
+					Str("org_id", id.Identity.OrgID)
+				newCtx = newLogger.Logger().WithContext(newCtx)
 			}
-			identity := identity.Identity(ctx)
 
-			traceId := trace.SpanFromContext(ctx).SpanContext().TraceID()
-			if !traceId.IsValid() {
-				traceId = random.TraceID()
-			}
-			ctx = logging.WithTraceId(ctx, traceId.String())
-
-			newLogger := logger.With().
-				Str("trace_id", traceId.String()).
-				Str("account_number", identity.Identity.AccountNumber).
-				Str("org_id", identity.Identity.OrgID).Logger()
-			ctx = newLogger.WithContext(ctx)
-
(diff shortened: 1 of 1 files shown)


internal/kafka/kafka.go
```diff
@@ -206,25 +206,29 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, since time.Time
 				msg.Key, msg.Topic, msg.Offset, msg.Partition)
 
 			// build new context - identity and trace id
-			ctx, err = identity.WithIdentityFrom64(ctx, header("x-rh-identity", msg.Headers))
-			if err != nil {
-				logger.Trace().Msgf("Could not extract identity from context to Kafka message: %s", err)
+			newLogger := logger.With().Str("msg_id", random.TraceID().String())
```
lzap, line 209: The idea behind this is to have at least some unique identifier just in case there is no identity in the incoming message (thus no span - no trace id). It will be used in the "else" block down below so we can at least correlate logs from the same kafka (anonymous) message.

internal/kafka/kafka.go
```diff
@@ -206,25 +206,29 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, since time.Time
...
+
+				newLogger = newLogger.
+					Str("trace_id", traceId.String()).
+					Str("account_number", id.Identity.AccountNumber).
+					Str("org_id", id.Identity.OrgID)
+				newCtx = newLogger.Logger().WithContext(newCtx)
+			} else {
+				errLogger := newLogger.Logger()
+				errLogger.Warn().Err(err).Msgf("Could not extract identity from context to Kafka message")
+				newCtx = errLogger.WithContext(newCtx)
```
adiabramovitch, line 228: If I understood correctly, you think the issue is because we did not create a new context. 
Why do you change it to err == nil? I would rather be consistent and keep the change like that: 

```suggestion
newLogger := logger.With().Str("msg_id", random.TraceID().String())
newCtx, err := identity.WithIdentityFrom64(ctx, header("X-RH-Identity", msg.Headers))
if err != nil {
errLogger := newLogger.Logger()
				errLogger.Warn().Err(err).Msgf("Could not extract identity from context to Kafka message")
				newCtx = errLogger.WithContext(newCtx)
} else {
		id := identity.Identity(newCtx)

				traceId := trace.SpanFromContext(ctx).SpanContext().TraceID()
				if !traceId.IsValid() {
					traceId = random.TraceID()
				}
				newCtx = logging.WithTraceId(newCtx, traceId.String())

				newLogger = newLogger.
					Str("trace_id", traceId.String()).
					Str("account_number", id.Identity.AccountNumber).
					Str("org_id", id.Identity.OrgID)
				newCtx = newLogger.Logger().WithContext(newCtx)
}
			handler(newCtx, NewMessageFromKafka(&msg))
```

adiabramovitch: It is just a nit, but I find the `err != nil` syntax as a convention and maybe more readable.

lzap: The problem is that the code overwrites `ctx` variable which becomes a shared variable that is passed over into several goroutines. The solution is to make copies.

The if statement change is just a cleanup - previous behavior when there was no identity was also incorrect - it would only log a warning and then continue trying to read identity while now it correctly only read the identity when it is present.

ezr-ondrej: but we could flip the if/else code branches, right?

lzap: I don’t understand what for, please show me full code which would be more readable than this. My goal is:

* When there is no identity available, log a warning into the newly created logger (which has msg_id field)
* Otherwise extract identity, extract trace id and create a new logger with all the fields.
* Call the message handler with either one or the other newly created context.

Honestly, I tried to swap if statements and it looks the same, I cannot drop `else` or at least I can't figure out how that would be more readable.

adiabramovitch: I did not mean to drop the else, I wrote a suggestion with the same changes you have done, just by switching the branches. 
Just said we usually use err != nil.

lzap: Let me rebase I need to push anyways to trigger tests again, swapped.

akhil-jha:
Build failed.

/retest

adiabramovitch (approved):
LGTM, lets see if that fixes the problem

PR #607, opened 2023-07-19
chore: better error when clowder is not present

lzap:
Slightly improves errors, tho I am not sure if these structures are validated (required) or not in clowder.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -214,6 +214,9 @@ func Initialize(configFiles ...string) {
 		cfg := clowder.LoadedConfig
 
 		// database
+		if cfg.Database == nil {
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
I thing the error performance team had hit was about the application we depend on not being deployed in the environment (sources, IB or RBAC is missing and thus it's endpoint not filled)

ezr-ondrej:
looks like it's working for the happy path tho :+1:

lzap:
Yeah I realized that too, but I saw we have similar checks in place so thought it might be useful. Probably not.

ezr-ondrej (approved):
Sure, just in case :shrug: Thanks :+1:

PR #610, opened 2023-07-20
fix(HMS-2002): postpone db stats 10 minutes

lzap:
Reservations metrics are calculated ever 30 minutes and windows are 24hours and 28 days back.

This patch delays that by 10 minutes because from time to time when Azure reservation is in progress, thus in "pending" state, it is collected by Prometheus. Possibly an alert could be triggered when multiple customers would be waiting for reservation to complete. Since reservations takes from few seconds (AWS) to 9 minutes (Azure) on average, delaying the numbers for 10 minutes solves this problem completely.

The patch shorten the delay between database query from 30 to 10 minutes, since queries are about 20ms in total I think we can afford running them more often.

### DIFF
--- a/internal/dao/pgx/stat_pgx.go
+++ b/internal/dao/pgx/stat_pgx.go
@@ -20,10 +20,10 @@ func getStatDao(ctx context.Context) dao.StatDao {
 	return &statDao{}
 }
 
-func (x *statDao) getUsage(ctx context.Context, interval string) ([]*models.UsageStat, error) {
+func (x *statDao) getUsage(ctx context.Context, iStart, iEnd string) ([]*models.UsageStat, error) {
 	query := `select provider, 'success' as result, count(provider) as count
 	from reservations
-	where created_at >= now() - cast($1 as interval)
+	where created_at between now() - cast($1 as interval) and now() - cast($2 as interval)
 	  and success = true
 	group by provider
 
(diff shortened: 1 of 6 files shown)


akhil-jha:
Lint is failing here. pr_check is green.

internal/dao/pgx/stat_pgx.go
```diff
@@ -20,31 +20,31 @@ func getStatDao(ctx context.Context) dao.StatDao {
...
 	select provider, 'pending' as result, count(provider) as count
 	from reservations
-	where created_at >= now() - cast($1 as interval)
+	where created_at between now() - cast($1 as interval) and now() - cast($2 as interval)
 	  and success is null
 	group by provider`
 
 	var result []*models.UsageStat
-	rows, err := db.Pool.Query(ctx, query, interval)
+	rows, err := db.Pool.Query(ctx, query, iStart)
```
avitova, line 47:
```suggestion
	rows, err := db.Pool.Query(ctx, query, iStart, iEnd)
```
I think you want to do this. Without the second argument, it fails.

lzap: Good catch, see, never push untested code :-)

Fixed

avitova (approved):
TY! LGTM

lzap:
Rebased.

adiabramovitch:
LGTM, should we add a test?

lzap:
Added a test, it is primitive there is not much I can do because reservation DAO API does not allow updating of started_at field.

adiabramovitch (approved):

PR #611, opened 2023-07-20
ci(HMS-1947): Update names and add parameter in e2e test template

tpapaioa:
This PR makes the following changes to the post-deploy e2e test template:
- Update resource names to not be specific to the stage environment
- Create an identifier string to pass to the test container's IQE_IBUTSU_SOURCE env value. This is used when reporting test results.

---

Thanks for the contribution, make sure your commit message follows this subject style:

        type: brief summary up to 70 characters or
        type(scope): brief summary up to 70 characters

Type is required, scope is optional. Prefer lower-case and avoid dot at the end.

Find more info about types and scopes at:

https://github.com/RHEnVision/provisioning-backend#contributing

Take a moment to read our contributing and security guidelines:

https://github.com/RedHatInsights/secure-coding-checklist

**Checklist**

- [ ] all commit messages follows the policy above
- [ ] the change follows our security guidelines

### DIFF
--- a/deploy/testing-integration.yaml
+++ b/deploy/testing-integration.yaml
@@ -2,12 +2,12 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: provisioning-stage-e2e
+  name: provisioning-e2e
 objects:
 - apiVersion: batch/v1
   kind: Job
   metadata:
-    name: provisioning-stage-e2e-${IMAGE_TAG}-${UID}
+    name: provisioning-e2e-${IMAGE_TAG}-${UID}
     annotations:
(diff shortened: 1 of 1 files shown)


akhil-jha (approved):
I see it has ENV_FOR_DYNACONF as stage_proxy. I am guessing it's a default value and it will be changed to "prod" depending on the pipeline? @tpapaioa

tpapaioa:
> I see it has ENV_FOR_DYNACONF as stage_proxy. I am guessing it's a default value and it will be changed to "prod" depending on the pipeline? @tpapaioa

Yes, these default values will be overridden as needed for the specific environments in our test deployment saas files.

PR #614, opened 2023-07-24
chore: regenerate example config file

avitova:
Based on [my issue](https://github.com/ilyakaznacheev/cleanenv/issues/119) in the cleanenv package, the mantainer of the project made it possible to sort the variables in the config file. It takes some time before the whole config file is generated, but we do not generate these files as often, so I guess it is okay.

I have regenerated go.mod using `go get -u ./... && go mod tidy`, is this not too many additions though?


lzap:
That is great, thank you.

/retest

lzap:
`AssertionError: IP address field was empty in the GET /reservations/{provider}/{id}`

Seems irrelevant, merging. Thanks.

lzap (approved):

PR #615, opened 2023-07-24
chore: improve logging of EnqueueAvailabilityStatusRequest

lzap:
We see a newly created source on stage not getting through our availability check. It is being skipped or something. There must be something wrong with the internal kafka topic. This should improve logging so we can find the particular place in logs where it should be sent.

I refactored the sending function - it takes asm instead msg so it can log source id properly.

### DIFF
--- a/internal/background/initialize.go
+++ b/internal/background/initialize.go
@@ -12,7 +12,7 @@ import (
 
 // Maximum batch size for each batch send, also an incoming buffered channel size to prevent
 // incoming requests to overload the sender.
-const availabilityStatusBatchSize = 1024
+const availabilityStatusBatchSize = 16
 
 // InitializeApi starts background goroutines for REST API processes.
 // Use context cancellation to stop them.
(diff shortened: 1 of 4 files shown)


avitova (approved):
I think we should merge this so that we know what is wrong, right?

internal/background/initialize.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // Maximum batch size for each batch send, also an incoming buffered channel size to prevent
 // incoming requests to overload the sender.
-const availabilityStatusBatchSize = 1024
+const availabilityStatusBatchSize = 16
```
lzap, line 15: This was way optimistic. Real data from stage are about dozen of status messages captured per tick. It does not need to be that big.

lzap:
Turns out it was a misconfiguration, source checks do work fine for now. If you like the change, feel free to merge it tho. Good logging is key.

adiabramovitch (approved):
LGTM, checked it on a GCP source. 
However, I remember that in my original implementation, there was a reason to receive the message as a generic one, couldn't reproduce any issue for now, so we should just be aware if something goes wrong  :)

lzap:
It still sends generic message, it gets converted in the function, not before calling it.

PR #616, opened 2023-07-24
fix(HMS-2178): followup context bug fix

avitova:
We have missed this, in this codeflow, the `newCtx` would end up as a 'nil' value, which we should not use anymore. This ends up with panic.


internal/kafka/kafka.go
```diff
@@ -211,7 +211,7 @@ func (b *kafkaBroker) Consume(ctx context.Context, topic string, since time.Time
 			if err != nil {
```
avitova, line 211: In case we get an error returned, the newCtx value is nil.

lzap (approved):
Great catch. Thanks.

PR #617, opened 2023-07-25
chore: new sentry replacements

lzap:
We see bunch of Azure and GCP errors in GlitchTip, each error is unique because project id and operation start time. This makes sure all errors are grouped together.


adiabramovitch (approved):
LGTM

PR #621, opened 2023-07-27
fix(HMS-2233): fix reservation id tag for AWS

lzap:
Only worked when Name was provided. Name is not supported from the UI, so this is why it does not work on stage/prod. Also adding user to logging which will be useful for identification of owner of the instance.

### DIFF
--- a/internal/middleware/enforce_identity.go
+++ b/internal/middleware/enforce_identity.go
@@ -55,6 +55,7 @@ func EnforceIdentity(next http.Handler) http.Handler {
 
 		topLevelOrgIDFallback(&jsonData)
 
+		logger.Debug().RawJSON("user", []byte(idRaw)).Msg("Enforcing identity")
 		err = checkHeader(r.Context(), &jsonData, w)
 		if err != nil {
 			return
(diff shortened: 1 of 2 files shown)


internal/middleware/enforce_identity.go
```diff
@@ -55,6 +55,11 @@ func EnforceIdentity(next http.Handler) http.Handler {
 
 		topLevelOrgIDFallback(&jsonData)
 
+		logger.Debug().Msgf("Enforcing identity for %s %s (%s)",
+			jsonData.Identity.User.FirstName,
+			jsonData.Identity.User.LastName,
+			jsonData.Identity.User.Email,
+		)
```
avitova, line 62: I was just checking if this would not error out in case of missing User, but it looks like it is okay. Still, if all the values were empty, would this not be confusing? I am not sure though, it might be okay.

lzap: It is not a pointer so cannot be nil, but you are right that empty values will look strange. Rebasing, let's use the raw JSON. There are no passwords or secret items.

avitova (approved):
LGTM:)

akhil-jha:
@lzap can you rebase? That should fix the pr_check

lzap:
Done.

PR #625, opened 2023-08-02
feat(HMS-2225): Add limit,  offset and metadata

adiabramovitch:
This PR introduces support for pagination with limit and offset parameters to the `List Pubkeys` and` List Sources `endpoints. The pagination functionality has been implemented as a custom middleware, which can be added to specific endpoints as needed. The offset and limit values are added to the context and retrieved in the corresponding handler functions.

To achieve pagination, the PR utilizes Postgres' built-in LIMIT and OFFSET functionality, as well as the offset and limit parameters available in the sources client.

Currently, some tests are missing, which I  intends to add. Before proceeding with adding the tests, I would appreciate your feedback on the proposed approach.

### DIFF
--- a/cmd/spec/example_pubkey.go
+++ b/cmd/spec/example_pubkey.go
@@ -1,6 +1,9 @@
 package main
 
-import "github.com/RHEnVision/provisioning-backend/internal/payloads"
+import (
+	"github.com/RHEnVision/provisioning-backend/internal/page"
+	"github.com/RHEnVision/provisioning-backend/internal/payloads"
+)
 
 var PubkeyRequest = payloads.PubkeyRequest{
 	Name: "My key",
@@ -20,7 +23,7 @@ var PubkeyResponse = payloads.PubkeyResponse{
 var PubkeyListResponse = payloads.PubkeyListResponse{
 	Data: []*payloads.PubkeyResponse{
 		{
-			ID:                1,
+			ID:                3,
 			AccountID:         1,
 			Name:              "My key",
 			Body:              "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEhnn80ZywmjeBFFOGm+cm+5HUwm62qTVnjKlOdYFLHN lzap",
@@ -28,5 +31,12 @@ var PubkeyListResponse = payloads.PubkeyListResponse{
 			Fingerprint:       "gL/y6MvNmJ8jDXtsL/oMmK8jUuIefN39BBuvYw/Rndk=",
 			FingerprintLegacy: "ee:f1:d4:62:99:ab:17:d9:3b:00:66:62:32:b2:55:9e",
 		},
+	},
+	Metadata: page.Metadata{
+		Total: 3,
+	},
+	Links: page.Links{
+		Previous: "/api/provisioning/v1/pubkeys?limit=2&offset=0",
+		Next:     "",
 	},
 }
--- a/cmd/spec/example_reservation.go
+++ b/cmd/spec/example_reservation.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/page"
 	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 )
@@ -62,6 +63,13 @@ var GenericReservationResponsePayloadListExample = payloads.GenericReservationLi
 		&GenericReservationResponsePayloadPendingExample,
 		&GenericReservationResponsePayloadSuccessExample,
 		&GenericReservationResponsePayloadFailureExample,
+	},
+	Metadata: page.Metadata{
+		Total: 3,
+	},
+	Links: page.Links{
+		Previous: "",
+		Next:     "",
 	},
 }
 
--- a/cmd/spec/example_source.go
+++ b/cmd/spec/example_source.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/page"
 	"github.com/RHEnVision/provisioning-backend/internal/payloads"
 )
 
@@ -14,6 +15,13 @@ var SourceListResponse = payloads.SourceListResponse{
 			ID:   "543621",
 			Name: "My other AWS account",
 		},
+	},
+	Metadata: page.Metadata{
+		Total: 4,
+	},
+	Links: page.Links{
+		Previous: "/api/provisioning/v1/sources?limit=2&offset=0",
+		Next:     "",
 	},
 }
 
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -107,6 +107,9 @@ paths:
       operationId: getPubkeyList
       tags:
         - Pubkey
+      parameters:
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Offset'
       description: >
         A pubkey represents an SSH public portion of a key pair with name and body.
         This operation returns list of all pubkeys for particular account.
@@ -132,14 +135,8 @@ paths:
       tags:
         - Source
       parameters:
-      - name: provider
-        in: query
-        schema:
-          type: string
-          enum:
-            - aws
-            - azure
-            - gcp
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Offset'
       responses:
         '200':
           description: Returned on success.
@@ -341,6 +338,9 @@ paths:
       operationId: getReservationsList
       tags:
         - Reservation
+      parameters:
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Offset'
       description: >
         A reservation is a way to activate a job, keeps all data needed for a job to start.
         This operation returns list of all reservations for particular account. To get a
--- a/cmd/spec/path.yaml
+++ b/cmd/spec/path.yaml
@@ -135,6 +135,14 @@ paths:
       tags:
         - Source
       parameters:
+        - name: provider
+          in: query
+          schema:
+            type: string
+            enum:
+                - aws
+                - azure
+                - gcp
         - $ref: '#/components/parameters/Limit'
         - $ref: '#/components/parameters/Offset'
       responses:
--- a/internal/clients/http/sources/sources_client.go
+++ b/internal/clients/http/sources/sources_client.go
@@ -82,7 +82,7 @@ func (c *sourcesClient) Ready(ctx context.Context) error {
 	return nil
 }
 
-func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, provider models.ProviderType) ([]*clients.Source, error) {
+func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, provider models.ProviderType) ([]*clients.Source, int, error) {
 	logger := logger(ctx)
 	params := &ListApplicationTypeSourcesParams{}
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListProvisioningSourcesByProvider")
@@ -91,13 +91,13 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 	appTypeId, err := c.GetProvisioningTypeId(ctx)
 	if err != nil {
 		logger.Warn().Err(err).Msg("Failed to get provisioning type id")
-		return nil, fmt.Errorf("failed to get provisioning app type: %w", err)
+		return nil, 0, fmt.Errorf("failed to get provisioning app type: %w", err)
 	}
 
 	sourcesProviderName := provider.SourcesProviderName()
 	if err != nil {
 		logger.Warn().Err(err).Msg("Failed to get provider name according to sources service")
-		return nil, fmt.Errorf("failed to get provider name according to sources service: %w", err)
+		return nil, 0, fmt.Errorf("failed to get provider name according to sources service: %w", err)
 	}
 
 	offset := page.Offset(ctx).String()
@@ -107,15 +107,15 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 		headers.AddEdgeRequestIdHeader, BuildQuery("filter[source_type][name]", sourcesProviderName, "offset", offset, "limit", limit))
 	if err != nil {
 		logger.Warn().Err(err).Msg("Failed to fetch ApplicationTypes from sources")
-		return nil, fmt.Errorf("failed to get ApplicationTypes: %w", err)
+		return nil, 0, fmt.Errorf("failed to get ApplicationTypes: %w", err)
 	}
 
 	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
 	if err != nil {
 		if errors.Is(err, clients.NotFoundErr) {
-			return nil, fmt.Errorf("list provisioning sources call: %w", http.SourceNotFoundErr)
+			return nil, 0, fmt.Errorf("list provisioning sources call: %w", http.SourceNotFoundErr)
 		}
-		return nil, fmt.Errorf("list provisioning sources call: %w", err)
+		return nil, 0, fmt.Errorf("list provisioning sources call: %w", err)
 	}
 
 	result := make([]*clients.Source, 0, len(*resp.JSON200.Data))
@@ -130,10 +130,15 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 		result = append(result, &newSrc)
 	}
 
-	return result, nil
+	total := 0
+	if resp.JSON200.Meta != nil {
+		total = *resp.JSON200.Meta.Count
+	}
+
+	return result, total, nil
 }
 
-func (c *sourcesClient) ListAllProvisioningSources(ctx context.Context) ([]*clients.Source, error) {
+func (c *sourcesClient) ListAllProvisioningSources(ctx context.Context) ([]*clients.Source, int, error) {
 	logger := logger(ctx)
 	params := &ListApplicationTypeSourcesParams{}
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListAllProvisioningSources")
@@ -142,7 +147,7 @@ func (c *sourcesClient) ListAllProvisioningSources(ctx context.Context) ([]*clie
 	appTypeId, err := c.GetProvisioningTypeId(ctx)
 	if err != nil {
 		logger.Warn().Err(err).Msg("Failed to get provisioning type id")
-		return nil, fmt.Errorf("failed to get provisioning app type: %w", err)
+		return nil, 0, fmt.Errorf("failed to get provisioning app type: %w", err)
 	}
 
 	params.Offset = page.Offset(ctx).IntPtr()
@@ -151,15 +156,15 @@ func (c *sourcesClient) ListAllProvisioningSources(ctx context.Context) ([]*clie
 	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, params, headers.AddSourcesIdentityHeader, headers.AddEdgeRequestIdHeader)
 	if err != nil {
 		logger.Warn().Err(err).Msg("Failed to fetch ApplicationTypes from sources")
-		return nil, fmt.Errorf("failed to get ApplicationTypes: %w", err)
+		return nil, 0, fmt.Errorf("failed to get ApplicationTypes: %w", err)
 	}
 
 	err = http.HandleHTTPResponses(ctx, resp.StatusCode())
 	if err != nil {
 		if errors.Is(err, clients.NotFoundErr) {
-			return nil, fmt.Errorf("list provisioning sources call: %w", http.SourceNotFoundErr)
+			return nil, 0, fmt.Errorf("list provisioning sources call: %w", http.SourceNotFoundErr)
 		}
-		return nil, fmt.Errorf("list provisioning sources call: %w", err)
+		return nil, 0, fmt.Errorf("list provisioning sources call: %w", err)
 	}
 
 	result := make([]*clients.Source, len(*resp.JSON200.Data))
@@ -172,7 +177,13 @@ func (c *sourcesClient) ListAllProvisioningSources(ctx context.Context) ([]*clie
 		}
 		result[i] = &newSrc
 	}
-	return result, nil
+
+	total := 0
+	if resp.JSON200.Meta != nil {
+		total = *resp.JSON200.Meta.Count
+	}
+
+	return result, total, nil
 }
 
 func (c *sourcesClient) GetAuthentication(ctx context.Context, sourceId string) (*clients.Authentication, error) {
--- a/internal/clients/http/sources/sources_client.go
+++ b/internal/clients/http/sources/sources_client.go
@@ -15,6 +15,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/headers"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
+	"github.com/RHEnVision/provisioning-backend/internal/page"
 	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 	"github.com/RHEnVision/provisioning-backend/internal/telemetry"
 	"github.com/rs/zerolog"
@@ -83,6 +84,7 @@ func (c *sourcesClient) Ready(ctx context.Context) error {
 
 func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, provider models.ProviderType) ([]*clients.Source, error) {
 	logger := logger(ctx)
+	params := &ListApplicationTypeSourcesParams{}
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListProvisioningSourcesByProvider")
 	defer span.End()
 
@@ -98,8 +100,11 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 		return nil, fmt.Errorf("failed to get provider name according to sources service: %w", err)
 	}
 
-	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, &ListApplicationTypeSourcesParams{}, headers.AddSourcesIdentityHeader,
-		headers.AddEdgeRequestIdHeader, BuildQuery("filter[source_type][name]", sourcesProviderName))
+	offset := page.Offset(ctx).String()
+	limit := page.Limit(ctx).String()
+
+	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, params, headers.AddSourcesIdentityHeader,
+		headers.AddEdgeRequestIdHeader, BuildQuery("filter[source_type][name]", sourcesProviderName, "offset", offset, "limit", limit))
 	if err != nil {
 		logger.Warn().Err(err).Msg("Failed to fetch ApplicationTypes from sources")
 		return nil, fmt.Errorf("failed to get ApplicationTypes: %w", err)
@@ -130,6 +135,7 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 
 func (c *sourcesClient) ListAllProvisioningSources(ctx context.Context) ([]*clients.Source, error) {
 	logger := logger(ctx)
+	params := &ListApplicationTypeSourcesParams{}
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "ListAllProvisioningSources")
 	defer span.End()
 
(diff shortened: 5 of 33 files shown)


adiabramovitch:
Some tests are failing because the missing offset, limit. 
I have a fix for that but I prefer to stub them in the context, will do that when adding the tests.

internal/middleware/paginate.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/rs/zerolog"
+)
+
+type ctxKeyType int
+
+const (
+	OffsetCtxKey ctxKeyType = iota
+	LimitCtxKey
+)
```
lzap, line 16: The general practice is not to export context keys, make them unexported and provide functions `Offset/WithOffset` and `Limit/WithLimit` that will lookup/store value in the context. This is a nice encapsulation and you also make sure that nobody will mess up with the value in the context. Also the functions will do the conversion to the proper type (int in this case).

Finally, I think it always makes sense to have both values and not one, so in this case create a new type that will have both ints and work with this type instead of individually. This struct could also have a boolean indicating if there is actually pagination or not - this will depend on the endpoint.

```
type Pagination struct {
  Active bool
  Limit int
  Offset int
}
```

The middleware could always insert either pagination that is not active (basically just empty value: `Pagination{}` = not active) or an active one with proper numbers. Then you can always be sure you never get a `nil` out of context and the code will be pretty:

```
pg := middleware.Pagination(ctx)
if pg.Active {
  limit := pg.Limit
  offset := pg.Offset
  // ...
}
```

internal/clients/http/sources/sources_client.go
```diff
@@ -98,8 +101,11 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 		return nil, fmt.Errorf("failed to get provider name according to sources service: %w", err)
 	}
 
-	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, &ListApplicationTypeSourcesParams{}, headers.AddSourcesIdentityHeader,
-		headers.AddEdgeRequestIdHeader, BuildQuery("filter[source_type][name]", sourcesProviderName))
+	limit := strconv.Itoa(ctx.Value(middleware.LimitCtxKey).(int))
+	offset := strconv.Itoa(ctx.Value(middleware.OffsetCtxKey).(int))
```
lzap, line 105: Type assertion like this is dangerous - Value can also return `nil` which will lead to panic. As I suggest below, create a function to do all the hard work of checking for `nil` etc.

internal/middleware/paginate.go
```diff
@@ -0,0 +1,37 @@
...
+			offset = 0
+		}
+		limit, err := strconv.Atoi(r.URL.Query().Get("limit"))
+		if err != nil || limit < 1 {
+			logger.Trace().Err(err).Msg("Limit is missing, setting limit value to 100")
+			limit = 100
+		}
+
+		newCtx := context.WithValue(r.Context(), OffsetCtxKey, offset)
+		newCtx = context.WithValue(newCtx, LimitCtxKey, limit)
```
lzap, line 34: Here you would use the `WithPagination` or `WithOffset/WithLimit` functions (depending if you like the composed type or not).

lzap (changes requested):
Looks great, the only comment I have is not exporting the key (creating a context manipulation functions) and perehaps a composed type would be appropriated in this case.

internal/clients/http/sources/sources_client.go
adiabramovitch: I will create the function as you suggested, however I don't think that it will ever be nil because in the middleware offset and limit are assigned with default values (0, 100)

internal/middleware/paginate.go
```diff
@@ -0,0 +1,37 @@
...
+		logger := zerolog.Ctx(r.Context())
+		offset, err := strconv.Atoi(r.URL.Query().Get("offset"))
+		if err != nil || offset < 0 {
+			logger.Trace().Err(err).Msg("Offset is missing, setting offset value to 0")
+			offset = 0
+		}
+		limit, err := strconv.Atoi(r.URL.Query().Get("limit"))
+		if err != nil || limit < 1 {
+			logger.Trace().Err(err).Msg("Limit is missing, setting limit value to 100")
+			limit = 100
```
avitova, line 30: I suggest we have a constant for that. Not necessary, though.

scripts/rest_examples/pubkeys-list.http
```diff
@@ -1,5 +1,5 @@
 // @no-log
-GET http://{{hostname}}:{{port}}/{{prefix}}/pubkeys/ HTTP/1.1
+GET http://{{hostname}}:{{port}}/{{prefix}}/pubkeys?limit=&offset= HTTP/1.1
```
avitova, line 2: Maybe let's leave here some valid values? This way, it produces error, as the limit and offset are both empty.

scripts/rest_examples/sources-list.http
```diff
@@ -1,4 +1,4 @@
 // @no-log
-GET http://{{hostname}}:{{port}}/{{prefix}}/sources HTTP/1.1
+GET http://{{hostname}}:{{port}}/{{prefix}}/sources?limit=&offset= HTTP/1.1
```
avitova, line 2: ditto

scripts/rest_examples/pubkeys-list.http
adiabramovitch: It does not produce an error, default values are set in the middleware if limit or offset are empty.

avitova: Yeah I am sorry, I meant a warning saying that the values are empty.
Fair enough, I do not think it is that important, so np, just leave it as it is. I would just expect valid values here, but we can leave it as it is:) :+1: Looks good.

internal/clients/http/sources/sources_client.go
adiabramovitch: Can you explain Why should I do the check  in a separate function  and not in the middleware? 
I am always setting a value for offset and limit. 
Do you suggest to transfer  the validation check to the function because of the exported keys?

adiabramovitch:
@lzap Thank for the review, I updated the PR. 
 I removed the exposure of the keys in the context and added functions which are used for adding and getting the limit and offset values. 

the original endpoints (e.g /sources or /pubkeys) are still valid,  limit and offset are assigned with default values

ezr-ondrej (commented):
If I understand this PR properly, it only solves parameters, but does not handle the response, right?

I guess that's fine, as it might help, but in my mind it's still breaking change on the API as now you do not get all the resuts, WDYT? :thinking:

internal/pagination/pagination.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type ctxKeyType int
+
+const (
+	defaultValueLimit             = 100
+	defaultValueOffset            = 0
+	offsetCtxKey       ctxKeyType = iota
+	limitCtxKey
+)
```
lzap, line 17: Can you please create a dedicated `const` block for default values (the first two lines)? I am not sure if it is a good practice to mix iotas with other values.

internal/pagination/pagination.go
```diff
@@ -0,0 +1,53 @@
...
+	logger := zerolog.Ctx(ctx)
+	limit, err := strconv.Atoi(limitStr)
+	if err != nil || limit < 0 {
+		logger.Trace().Msg("Limit is missing, setting offset value to 0")
+		limit = defaultValueLimit
+	}
+	return context.WithValue(ctx, limitCtxKey, limit)
+}
+
+func GetOffset(ctx context.Context) int {
```
lzap, line 41:
```suggestion
func Offset(ctx context.Context) int {
```

Get prefix for getters is considered an anti-pattern in Go.

internal/pagination/pagination.go
```diff
@@ -0,0 +1,53 @@
...
+	if offset, ok := ctx.Value(offsetCtxKey).(int); ok {
+		return offset
+	}
+	return defaultValueOffset
+}
+
+func GetLimit(ctx context.Context) int {
```
lzap, line 48:
```suggestion
func Limit(ctx context.Context) int {
```

Get prefix for getters is considered an anti-pattern in Go.

internal/clients/http/sources/sources_client.go
```diff
@@ -98,8 +101,11 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 		return nil, fmt.Errorf("failed to get provider name according to sources service: %w", err)
 	}
 
-	resp, err := c.client.ListApplicationTypeSourcesWithResponse(ctx, appTypeId, &ListApplicationTypeSourcesParams{}, headers.AddSourcesIdentityHeader,
-		headers.AddEdgeRequestIdHeader, BuildQuery("filter[source_type][name]", sourcesProviderName))
+	offset := strconv.Itoa(pagination.GetOffset(ctx))
+	limit := strconv.Itoa(pagination.GetLimit(ctx))
```
lzap, line 105: Might be pretty useful to create `pagination.OffsetString(ctx)` doing conversion for you. Also pointer (below). Just an idea.

internal/pagination/pagination.go
```diff
@@ -0,0 +1,53 @@
+package pagination
```
lzap, line 1: That is awfully long, I suggest `page` or something shorter: https://go.dev/blog/package-names

internal/pagination/pagination.go
```diff
@@ -0,0 +1,53 @@
...
+	offsetCtxKey       ctxKeyType = iota
+	limitCtxKey
+)
+
+// WithOffset returns context copy with offset value.
+func WithOffset(ctx context.Context, offsetStr string) context.Context {
+	logger := zerolog.Ctx(ctx)
+	offset, err := strconv.Atoi(offsetStr)
+	if err != nil || offset < 0 {
+		logger.Trace().Msg("Offset is missing, setting offset value to 0")
```
lzap, line 24: This log will swallow the reason (the error message) in case user provides an invalid input (e.g. letters instead numbers). I suggest to use `Err` function and rephrase the message to "Offset missing or invalid, setting...".

The same problem below.

internal/services/pubkey_service.go
```diff
@@ -49,7 +50,10 @@ func CreatePubkey(w http.ResponseWriter, r *http.Request) {
 func ListPubkeys(w http.ResponseWriter, r *http.Request) {
 	pubkeyDao := dao.GetPubkeyDao(r.Context())
 
-	pubkeys, err := pubkeyDao.List(r.Context(), 100, 0)
+	offset := pagination.GetOffset(r.Context())
+	limit := pagination.GetLimit(r.Context())
```
lzap, line 54: Maybe Offset64/Limit64 too? :-D

lzap (changes requested):
Couple of smaller nitpicks, but it is great overall.

@ezr-ondrej this PR does only prepare for pagination and it is all about handling input, does not return valid output. It is not an API breaking change (yet) so I think we can merge this. I would suggest to do OpenAPI change in separate PR so we can continue development but the feature will be still undocumented (therefore "invisible" to the users).

I suggest to do a preparation PR for the output part of this task - to introduce "result" and empty "metadata" into all relevant endpoints so we can sync this API breaking change with IQE. We might even need to bump the version to V2 tho I propose to just do it while in preview.

internal/pagination/pagination.go
adiabramovitch: Do you mean like that? 
`logger.Err(err).Msg("Offset is missing or invalid, setting offset value to 0")
`

internal/services/pubkey_service.go
adiabramovitch: Do you mean to do the conversion here like that:
 ```
int64(pagination.GetOffset(r.Context())) 
```
or to return int64 from the Offset function?

Comment (lzap) [reply internal/services/pubkey_service.go:54]: Just an idea (you don’t have to implement it if you don’t like it):

    func GetOffset64(context.Context) int64

File: internal/pagination/pagination.go (hunk 12 above)
Comment (lzap) [reply internal/pagination/pagination.go:24]: Exactly.

**Review by ezr-ondrej — COMMENTED** [review]
Given the limit is always set and can't be ommited (if I read correctly), I believe we should get the metadata field in either first, or at the same time.

When we limit the amount of returned resources, I believe it's important to also inform "how many more is there".

And by first I mean we can just get it in at the same time, but we should have it ready.
But IMO metadata with just `total`, where now it would equal to the returned resources would be an easy start :)

**Comment by lzap** [comment]
I would suggest to avoid total so users will not build their clients in a way that is always there. You can tell whether there is another page or not from metadata `next` link and we should properly document this behavior.

Single page result:

```
{
 "data": [...],
 "metadata": {
   "next": "",
   "previous": "",
 }
}
```

Multi-page result:

```
{
 "data": [...],
 "metadata": {
   "next": "/xxx?offset=100&limit=100",
   "previous": "/xxx?offset=0&limit=100",
 }
}
```

AWS multi-page result:

```
{
 "data": [...],
 "metadata": {
   "next": "/xxx?token=XXX&limit=100",
   "previous": "",
 }
}
```

File: cmd/spec/path.yaml [hunk 14]
```diff
@@ -107,6 +107,21 @@ paths:
...
+            default: 100
+          required: false
+          description: The number of items to return
+        - in: query
+          name: offset
+          schema:
+            type: integer
+            default: 0
+          required: false
+          description: The number of items to skip before starting to collect the result set.
```
lzap, line 124: Can you research if there is a way to create a `component` that can be referenced in parameters? That would save us a lot of copy-pasting.

internal/page/page.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"strconv"
+
+	"github.com/RHEnVision/provisioning-backend/internal/ptr"
+	"github.com/rs/zerolog"
+)
+
+type (
+	ctxKeyType  int
+	LimitOffset int
```
lzap, line 13: Can you unexport this? Context types and keys should not be exported to prevent data manipulation from outside of the package.

internal/page/page.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func Offset(ctx context.Context) LimitOffset {
+	if off, ok := ctx.Value(offsetCtxKey).(int); ok {
+		return LimitOffset(off)
+	}
+	return LimitOffset(defaultValueOffset)
+}
+
+func (o LimitOffset) IntP() *int {
```
lzap, line 62:
```suggestion
func (o LimitOffset) IntPtr() *int {
```

is perhaps a little bit more clean.

internal/page/page.go
```diff
@@ -0,0 +1,72 @@
...
+	return ptr.To(int(o))
+}
+
+func (o LimitOffset) Int64() int64 {
+	return int64(o)
+}
+
+func (o LimitOffset) String() string {
+	return strconv.Itoa(int(o))
```
lzap, line 71: The type cast is not needed I think.

lzap (changes requested):
Looks great, small nitpicks and we are good!

internal/page/page.go
lzap: Oh right.

cmd/spec/example_pubkey.go
```diff
@@ -29,4 +32,8 @@ var PubkeyListResponse = payloads.PubkeyListResponse{
 			FingerprintLegacy: "ee:f1:d4:62:99:ab:17:d9:3b:00:66:62:32:b2:55:9e",
 		},
 	},
+	Metadata: &page.Metadata{
+		Previous: "/api/provisioning/v1/pubkeys?limit=100&offset=0",
+		Next:     "/api/provisioning/v1/pubkeys?limit=100&offset=2",
```
ezr-ondrej, line 37: This is not correct, right? the offset should be `100`, or limit `2` right? :thinking: 
Having examples correct goes long way of explaining the funcionality :)

internal/page/page.go
```diff
@@ -74,3 +79,33 @@ func (o limitOffset) Int64() int64 {
 func (o limitOffset) String() string {
 	return strconv.Itoa(int(o))
 }
+
+type Metadata struct {
+	Previous string `json:"previous,omitempty"`
+	Next     string `json:"next,omitempty"`
+}
+
+func MetadataResponse(ctx context.Context, r *http.Request, total int) *Metadata {
```
ezr-ondrej, line 88: Am I missing where we are ommiting the next link when it's the last page?
Or how do we approach the last and first page? I thought we would have the next and prev missing, right?

internal/page/page.go
```diff
@@ -74,3 +79,33 @@ func (o limitOffset) Int64() int64 {
...
+	nextOffset := math.Min(total, offset+limit)
+	prevOffset := math.Max(0, offset-limit)
+
+	q := r.URL.Query()
+
+	// set the "prev" link with offset - limit or 0
+	q.Set("limit", strconv.Itoa(limit))
+	q.Set("offset", strconv.Itoa(prevOffset))
+	params, _ := url.PathUnescape(q.Encode())
+	prev := fmt.Sprintf("%v?%v", r.URL.Path, params)
```
ezr-ondrej, line 100: I'm not sure if there is no URL native way to format the path (copy the URL object and manipulate it for prev and next) insted of doing the string transpolation.
I guess it's quite safe, but I'd rather see native way, it's easy to make a mistake here :shrug:

ezr-ondrej (commented):
Just few nits, looking great overall :+1:

cmd/spec/example_pubkey.go
adiabramovitch: It is correct if you have only two pubkeys :) 
What would you expect to see if you have seen the last page?

adiabramovitch: Maybe it is more correct to put the limit (100)

internal/page/page.go
adiabramovitch: So I thought to show the next as the minimux between (limit +offset) and the total records. 
but I might be wrong here. 
btw, sources are not showing an empty string when there is no next page.

internal/page/page.go
adiabramovitch: I tried the naive way, but I think this is a good approavh and safer one. 
I saw that sources used that way.

cmd/spec/example_pubkey.go
ezr-ondrej: Ah, in that case that goes to the discussion if we should have next empty, as it basically does not make sense to have next, that will give you empty result IMO :thinking:

internal/page/page.go
ezr-ondrej: I'm not sure we want to strictly follow exactly what sources are doing :innocent:
I'm not sure what would be best approach, but empty link sounds as a good approach to me, it's easy to recognize for client that there are no more results IMO :shrug:
As we do not have total nor page count, it is impossible to recognize for client, they'd need to do the last request for empty page, which sounds like a waste of an API call

internal/page/page.go
ezr-ondrej: I'm not sure we  want to strictly follow what Sources are doing :innocent:

adiabramovitch: I agree :) but I find it better I guess but I can change it if you prefer.

ezr-ondrej: nah, it's fine :) If you like this better, lets keep it, not a blocker for sure! :+1:

internal/page/page.go
adiabramovitch: I wanted to follow common practice. 
IB and sources do not show an empty link, for example: 
```
  "links": {
    "first": "/api/image-builder/v1.0/composes?offset=0&limit=100",
    "last": "/api/image-builder/v1.0/composes?offset=4&limit=100"
  },
  "meta": {
    "count": 5
  }

```
And If I am talking about common practices I would change it to first and last like both of them expose.

lzap: Seeing something in a different project does not make anything a common practice :-D

I feel like returning empty links makes life easier for the HTTP client (our customer), so little bit of extra effort might be worthwhile.

internal/clients/interface.go
```diff
@@ -13,10 +13,10 @@ var GetSourcesClient func(ctx context.Context) (Sources, error)
 // Sources interface provides access to the Sources backend service API
 type Sources interface {
 	// ListProvisioningSourcesByProvider returns sources filtered by provider that have provisioning credentials assigned
-	ListProvisioningSourcesByProvider(ctx context.Context, provider models.ProviderType) ([]*Source, error)
+	ListProvisioningSourcesByProvider(ctx context.Context, provider models.ProviderType) ([]*Source, *int, error)
 
 	// ListAllProvisioningSources returns all sources that have provisioning credentials assigned
-	ListAllProvisioningSources(ctx context.Context) ([]*Source, error)
+	ListAllProvisioningSources(ctx context.Context) ([]*Source, *int, error)
```
lzap, line 19: Use `int` not pointer here, that is so much more efficient. You need to check for nils anyway in functions, if Sources for any reason returns no metadata just return 0.

internal/models/model_constants.go
```diff
@@ -31,6 +34,8 @@ func ProviderTypeFromString(str string) ProviderType {
 		return ProviderTypeAzure
 	case "gcp":
 		return ProviderTypeGCP
+	case "":
+		return ProviderTypeAll
 	default:
 		return ProviderTypeUnknown
 	}
```
lzap, line 41: I don’t like introducing a constant that is very similar if not the exactly the same as `ProviderTypeUnknown`. I created it exactly for this purpose, if you don’t like the name then rename it. Maybe `ProviverTypeUnspecified` is more appropriate for your use case.

internal/page/page.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// set the "next" link with limit+offset or total set for the next page
+	q.Set("offset", strconv.Itoa(nextOffset))
+	params, _ = url.PathUnescape(q.Encode())
+	next := fmt.Sprintf("%v?%v", r.URL.Path, params)
```
lzap, line 105: Uh, this is very ugly. Let’s take a look how you can cleanly create URL parameters in Go: 

```
	q := url.Values{}
	q.Add("clean", "abc")
	q.Add("dirty", "100% $999")
	q.Add("base64", "123/ABC+abc=")

	// prints: base64=123%2FABC%2Babc%3D&clean=abc&dirty=100%25+%24999
	println(q.Encode())
```

Here is a working example to play with: https://go.dev/play/p/eQU_2AZoI5W

Now, the main issue with base64 encoding is that it contains three characters which have special meaning in URLs: `/+=` and these needs to be escaped. There is no need to ever use PathUnescape, I am assuming that the AWS client library does perform proper URL decoding and return the token in the correct (normal) format. Something like `123/ABC+abc=` and not `123%2FABC%2Babc%3D` therefore you can just simply build a new params and call `Encode` to get the proper result.

My suggestion: just use `url.Values` and `Add`/`Encode` methods and it will work fine.

internal/payloads/pubkey_payload.go
```diff
@@ -25,7 +26,8 @@ type PubkeyResponse struct {
 	FingerprintLegacy string `json:"fingerprint_legacy,omitempty" yaml:"fingerprint_legacy,omitempty"`
 }
 type PubkeyListResponse struct {
-	Data []*PubkeyResponse `json:"data" yaml:"data"`
+	Data     []*PubkeyResponse `json:"data" yaml:"data"`
+	Metadata *page.Metadata    `json:"metadata" yaml:"metadata"`
```
lzap, line 30: I believe I did set a wrong standard of using pointers in our database model types, long term I would like to avoid pointers going forward. Little bit of copying is actually more performant than a pointer since the data can live on the stack and GC has less work. I’d say let’s keep it this way for now, this is an easy refactoring in the future.

lzap (dismissed):
Looking good, I have some concerns about parameters and one constant.

internal/models/model_constants.go
adiabramovitch: But my problem is not the naming, there are two cases here which I want to differentiate: 
- If we get an empty provider we should call  `ListAllProvisioningSources`
-  If the provider is ibm for example I want to render an error of not supported provider. 

So I want to add the `ProviderTypeAll` case and we can change the naming to  `ProviderTypeUnsupported`

internal/page/page.go
adiabramovitch: ok I changed it to empty links if we are listing the first page and the last one.

internal/models/model_constants.go
ezr-ondrej: Why do we want to return all sources? I would agree, that for such a case `ProviderTypeAll` makes sense, but do we have a usecase for all sources?

adiabramovitch: Yes, when the provider is not mentioned we are listing all the sources. 
as you can see in `sources_service.go`

ezr-ondrej: but there it's already solved, do we need to fix it in this PR? sounds like the change is controversial and not really needed in given PR, in such cases, I'm up for extracting it, so it can be discussed properly separately :)

adiabramovitch: I added the switch case for few reasons: 

1. Introducing a new switch case for `ProviderTypeAll` aligns with the existing pattern of switch cases we have in the reservations service. 

2. The `ListSources` function is more  readable and cleaner.

3. separation of logic. The `ListSources`  function was previously handling two different logics within a single, lengthy function. By isolating the `ProviderTypeAll` case, we can clearly differentiate the distinct behaviors. 

4. The need to access` r.Path` information for rendering accurate metadata inside the `ListProvisioningX` function in the `sources_client.go` files. I did not want to add another parameter to the `ListProvisioningX` functions related to the endpoint path. I thought we should wrap the client function with a wrapper function and get the r and w parameters.

adiabramovitch: And moreover, the "" provider was not provided is not the same case as "ibm".

cmd/spec/example_reservation.go
```diff
@@ -63,6 +64,13 @@ var GenericReservationResponsePayloadListExample = payloads.GenericReservationLi
 		&GenericReservationResponsePayloadSuccessExample,
 		&GenericReservationResponsePayloadFailureExample,
 	},
+	Metadata: page.Metadata{
+		Total: 2,
+	},
+	Links: page.Links{
+		Previous: "/api/provisioning/v1/reservations?limit=100&offset=0",
```
ezr-ondrej, line 71: Isn't this (the result in example the first page, so the Previous should be alse empty, right?

Also I see total 2, but there are three results (Pending, Success, Failure) :)

internal/page/page.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// First page: offset is 0
+	if offset == 0 {
+		prev = ""
+	} else {
+		prevOffset := math.Max(0, offset-limit)
+		q := url.Values{}
+		q.Add("limit", strconv.Itoa(limit))
+		q.Add("offset", strconv.Itoa(prevOffset))
+		prev = fmt.Sprintf("%v?%v", r.URL.Path, q.Encode())
```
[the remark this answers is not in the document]
adiabramovitch: This is the suggestion @lzap suggested :)

cmd/spec/example_reservation.go
adiabramovitch: Yes, copy-past mistake :) fixing

internal/page/page.go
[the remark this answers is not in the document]
ezr-ondrej: Ah, oki! :)

cmd/spec/example_pubkey.go
```diff
@@ -29,4 +32,11 @@ var PubkeyListResponse = payloads.PubkeyListResponse{
 			FingerprintLegacy: "ee:f1:d4:62:99:ab:17:d9:3b:00:66:62:32:b2:55:9e",
 		},
 	},
+	Metadata: page.Metadata{
+		Total: 1,
+	},
+	Links: page.Links{
+		Previous: "/api/provisioning/v1/pubkeys?limit=100&offset=0",
```
ezr-ondrej, line 39: Here it should be empty as well, right?

cmd/spec/example_source.go
```diff
@@ -15,6 +16,13 @@ var SourceListResponse = payloads.SourceListResponse{
 			Name: "My other AWS account",
 		},
 	},
+	Metadata: page.Metadata{
+		Total: 2,
+	},
+	Links: page.Links{
+		Previous: "/api/provisioning/v1/sources?limit=2&offset=0",
```
ezr-ondrej, line 23: Empty?

cmd/spec/example_pubkey.go
ezr-ondrej: for the pubkey, you'd also need another key, as currently we have only one key in the result... 
But here, you can do a trick or rather demonstrate a feature and do `Total = 3`, that way it will make sense :)

adiabramovitch: yes, so just s sources the total will be four and pubkey will be total of 3

ezr-ondrej: Yup :)

ezr-ondrej (approved):
LGTM, thanks @adiabramovitch ! :pray:

internal/page/page.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (o limitOffset) Int64() int64 {
+	return int64(o)
+}
+
+func (o limitOffset) String() string {
+	return strconv.Itoa(int(o))
+}
+
+func APIInfoResponse(ctx context.Context, r *http.Request, total int) *Info {
```
lzap, line 96: This could have been just `NewInfo`, the API prefix is not very informative, pretty much all functions in this app could be prefixed with API :-)

internal/page/page.go
```diff
@@ -0,0 +1,132 @@
...
+	Next     string `json:"next"`
+}
+type Metadata struct {
+	Total int `json:"total"`
+}
+
+type Info struct {
+	Metadata Metadata `json:"metadata"`
+	Links    Links    `json:"links"`
+}
```
lzap, line 42: The `Info` type name is pretty generic, I would suggest to find a better name for this.

adiabramovitch: That is what me and @ezr-ondrej decided on slack after seeing examples in other services and online examples.

lzap: I saw that, my point is that if this was not done or if you included links within the metadata structure, then the root type could have been named just "metadata" with its all associated functions/methods.

PR #626, opened 2023-08-02
feat(HMS-2248): Add token support for list templates

adiabramovitch:
Depends on #625 
This PR introduces pagination support for AWS and GCP by returning the received token when fetching the corresponding page. **To validate the retrieval of AWS pages, the token needs to be escaped since it is in base64 format.**

However, tests are currently missing, andI plan on adding those once will get some feedback. Additionally, tests are failing due to the initial commit missing limit and offset. I plan to stub them and proceed with adding the tests once this implementation is agreed upon.

Please let me know if you have any feedback or suggestions regarding this approach before proceeding with the test implementation.  
**For easy review just take a look at the changes made by the second commit, once the dependent PR will be merged i'll rebase.**


lzap (changes requested):
The same concern.

Now that I see we have two combinations:

* limit + offset
* limit + token

I think composed type does not make too much sense, checking for `nil` or better for `0` and/or empty string is also fine.

Still, I recommend to create the functions that will check for `nil` (in case context does not contain any value) and return `0` or `""` in that case so checks are easy.

akhil-jha:
Launch template test failing. Not able to list the templates.

adiabramovitch:
I will update this PR once #626 will get merged, there are a lot of conflicts but feel free to test it out.

internal/payloads/launch_template_payload.go
```diff
@@ -9,8 +9,8 @@ import (
 
 // See clients.LaunchTemplate
 type LaunchTemplateResponse struct {
-	ID   string `json:"id" yaml:"id"`
-	Name string `json:"name" yaml:"name"`
+	Templates []*clients.LaunchTemplate
+	NextToken string
```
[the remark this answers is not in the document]
adiabramovitch: I agree, but I think we first need the basic to work and add the metadata support in a different PR.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -100,44 +101,45 @@ func (c *gcpClient) NewInstanceTemplatesClient(ctx context.Context) (*compute.In
...
 	if err != nil {
 		logger.Error().Err(err).Msg("Could not get instances client")
-		return nil, fmt.Errorf("unable to get instances client: %w", err)
+		return nil, nil, fmt.Errorf("unable to get instances client: %w", err)
 	}
 	defer templatesClient.Close()
 
-	req := &computepb.AggregatedListInstanceTemplatesRequest{
-		Project: c.auth.Payload,
+	limit := ctx.Value(middleware.LimitCtxKey).(int)
```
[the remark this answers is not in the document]
adiabramovitch: Fixed

ezr-ondrej (commented):
I wonder (and this is really just a thought) if we actually need to do this? Launch templates will usually not be in hundreds and if they do, it does not make sense to paginate in the UI, but search, so I guess we would be fine here not implementing pagination, but only search and some indication of "there is more", but no way of fetching more.
I guess that would quite simplify the whole pagination effort and IMO it would be good enough solution.

lzap:
> I wonder (and this is really just a thought) if we actually need to do this? Launch templates will usually not be in hundreds and if they do, it does not make sense to paginate in the UI

I share the sentiment, but I would propose to do at least one pagination for every single type (SQL, sources, AWS, GCP) so we know for sure the design is good. We might not even use it in the UI but we will have backend API tests, documented API and confidence that it works.

api/openapi.gen.json
```diff
@@ -331,7 +331,14 @@
...
               "name": "XXL large backend API"
             }
-          ]
+          ],
+          "links": {
+            "next": "",
+            "previous": ""
+          },
+          "metadata": {
+            "total": 0
```
adiabramovitch, line 340: Not sure why it created the total here if it is empty and I am using the omitempty tag. 
Wanted to change the metadata struct to pointer but it is the same discussion as on Slack

cmd/spec/path.yaml
```diff
@@ -280,6 +282,14 @@ paths:
...
           required: true
           description: Hyperscaler region
+        - in: query
+          name: token
+          schema:
+            type: string
+            default: ""
+          required: false
+          description: The token to request the next page of results, empty token for first page.
+        - $ref: '#/components/parameters/Limit'
```
lzap, line 292: I suggest to also create a component for token, we only have it once for now but this can prevent copy pasting.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -338,34 +339,37 @@ func (c *ec2Client) DescribeInstanceDetails(ctx context.Context, InstanceIds []s
 	return instanceDetailList, nil
 }
 
-func (c *ec2Client) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchTemplate, error) {
+func (c *ec2Client) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchTemplate, *string, error) {
```
lzap, line 342: Why pointer? My first idea is empty string = no token.

internal/clients/http/ec2/ec2_client.go
```diff
@@ -338,34 +339,37 @@ func (c *ec2Client) DescribeInstanceDetails(ctx context.Context, InstanceIds []s
...
+			Name: ptr.From(awsTemplate.LaunchTemplateName),
 		}
+		res = append(res, &t)
 	}
-
-	return res, nil
+	nextToken := resp.NextToken
+	if nextToken == nil {
+		nextToken = ptr.To("")
+	}
```
lzap, line 371: AWS library overuses pointers and I believe the reason is it is auto generated.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -100,44 +101,42 @@ func (c *gcpClient) NewInstanceTemplatesClient(ctx context.Context) (*compute.In
...
+	}
+
+	templatesList := make([]*clients.LaunchTemplate, 0, len(lst))
+	for _, template := range lst {
+		id := strconv.FormatUint(*template.Id, 10)
+		templatesList = append(templatesList, &clients.LaunchTemplate{ID: id, Name: template.GetName()})
 	}
 
-	return templatesList, nil
+	return templatesList, &nextToken, nil
```
lzap, line 139: Ditto - the same concern.

internal/page/page.go
```diff
@@ -33,12 +34,12 @@ type Links struct {
 	Next     string `json:"next"`
 }
 type Metadata struct {
-	Total int `json:"total"`
+	Total int `json:"total,omitempty"`
 }
 
 type Info struct {
```
lzap, line 40: My suggestion would be renaming this to `Metaroot` instead of `Info`, that would include `NewMetaroot`.

Alternatively, I suggest to rename Metadata to just Data (keeping variable and JSON field as "metadata") and Info to Metadata.

Finally, my third suggestion is simply move links into metadata as this sets a bad practice - it feel like it is okay to pollute the root JSON namespace with fields while a single "metadata" field could be a nice solution with no other fields added.

internal/page/page.go
```diff
@@ -89,11 +106,15 @@ func (o limitOffset) Int64() int64 {
...
+func (o limitOffset) Int32() int32 {
+	return int32(o)
+}
+
 func (o limitOffset) String() string {
 	return strconv.Itoa(int(o))
 }
 
-func APIInfoResponse(ctx context.Context, r *http.Request, total int) *Info {
+func NewInfo(ctx context.Context, r *http.Request, total int) *Info {
```
lzap, line 117: Maybe `NewOffsetInfo` and the other one `NewTokenInfo`? (Alternatively `NewOffsetMetadata` if you choose to do the refactoring I recommend.)

internal/routes/all_routes.go
```diff
@@ -72,10 +72,9 @@ func MountAPI(r *chi.Mux) {
 				// TODO DEPRECATED: move this to outside of /sources (see below)
 				r.Get("/instance_types", s.ListInstanceTypes)
 
-				// TODO DEPRECATED: replaced with upload_info
```
lzap, line 75: Don’t remove deprecation warning, I believe this was for account_identity endpoint.

lzap (changes requested):
I only have naming conventions suggestions. I still do propose the refactoring (single root JSON field) but if you like to keep it this way I just suggest to get rid if the too generic Info (I suggest Metaroot/Metadata).

cmd/spec/parameters.go
```diff
@@ -26,3 +26,12 @@ var OffsetQueryParam = Parameter{
 	Required:    false,
 	In:          "query",
 }
+
+var TokenQueryParam = Parameter{
+	Name:        "token",
+	Description: "The token used for requesting the next page of results; empty for the first page",
```
lzap, line 32:
```suggestion
	Description: "The token used for requesting the next page of results; empty for the last page",
```

internal/clients/interface.go
```diff
@@ -83,8 +83,8 @@ type EC2 interface {
 	// ListInstanceTypesWithPaginator lists all instance types.
 	ListInstanceTypes(ctx context.Context) ([]*InstanceType, error)
 
-	// ListLaunchTemplates lists all launch templates.
-	ListLaunchTemplates(ctx context.Context) ([]*LaunchTemplate, error)
+	// ListLaunchTemplates lists all launch templates return the next page token.
```
lzap, line 86:
```suggestion
	// ListLaunchTemplates lists all launch templates and returns the next page token.
```

internal/clients/interface.go
```diff
@@ -156,5 +156,6 @@ type GCP interface {
 
 	GetInstanceDescriptionByID(ctx context.Context, id, zone string) (*InstanceDescription, error)
 
-	ListLaunchTemplates(ctx context.Context) ([]*LaunchTemplate, error)
+	// ListLaunchTemplates lists all launch templates and return the next page token.
```
lzap, line 159:
```suggestion
	// ListLaunchTemplates lists all launch templates and returns the next page token.
```

lzap (changes requested):
This is so much cleaner, only documentation typos.

@ezr-ondrej are you happy with this?

lzap:
Tests are failing with `iqe_provisioning_api.exceptions.ApiAttributeError: V1ListLaunchTemplateResponse has no attribute 'metadata' at ['received_data']['metadata']` and I guess there is a problem that "metadata" root field is missing after the refactoring.

There are two options, you can either create "Metaroot" type for the root field, or make use of [Go anonymous struct](https://dev.to/wagslane/use-anonymous-structs-for-json-marshalling-in-go-282g) directly in `NewMetadata` functions.

adiabramovitch:
> 

It is a problem in the iqe tests in my opinion. @akhil-jha ?

cmd/spec/parameters.go
adiabramovitch: That was not what I meant here, I meant that an empty token leads to the first page.

lzap: Oh it is a bit unclear but it is fine, thanks.

internal/payloads/launch_template_payload.go
```diff
@@ -14,28 +15,21 @@ type LaunchTemplateResponse struct {
...
-}
-
-func (s *LaunchTemplateResponse) Bind(_ *http.Request) error {
-	return nil
-}
-
-func (s *LaunchTemplateResponse) Render(_ http.ResponseWriter, _ *http.Request) error {
-	return nil
+	Data     []*LaunchTemplateResponse `json:"data" yaml:"data"`
+	Metadata page.Metadata             `json:"metadata" yaml:"metadata"`
```
lzap, line 19: Ah here it is, so the root is correct actually. Tests need to be updated I guess.

lzap (approved):

lzap:
I think IQE REST client needs to be regenerated, let’s wait what @akhil-jha or @mshriver have to say.

lzap:
@adiabramovitch you closed the PR?

PR #627, opened 2023-08-03
refactor: Change launch template name to id in GCP

adiabramovitch:
Upon thorough examination, it appears that the template "id" format is supported, which means it can be formatted as "global/instanceTemplates/ID" instead of using the "NAME" format.

@akhil-jha, would you be kind enough to test this and verify if there are any noticeable changes now? I personally inspected the request from the UI, and it seems that the `launch_template_id` field is properly populated and sent to the backend as expected.


akhil-jha (approved):

akhil-jha:
Please rebase! :)

PR #632, opened 2023-08-08
fix(HMS-2295): add numeric rid to GCP

lzap:
We currently do not store reservation id, the number, which is useful for figuring out owner of the instance.

While on it, I also noticed that we use fairly long "rhhhc" prefix, I propose to shorten it just to "rh".

Additionally, I noticed that AWS keys allows colons while GCP does not. Therefore I propose to use "rh-" (with dash characters) for both AWS and GCP.

The "rid" key is already used for unique identifier we generate during GCP launch, so I renamed it to "rh-uuid". It serves for identifying instances, "rh-rid" is now used for reservation numeric id. It could be used to the the same as well, but I do not want to change code too much.

Finally, keypair id was renamed to "rh-kid" to avoid too generic "id".

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -166,7 +166,7 @@ func (c *ec2Client) ImportPubkey(ctx context.Context, key *models.Pubkey, tag st
 			ResourceType: types.ResourceTypeKeyPair,
 			Tags: []types.Tag{
 				{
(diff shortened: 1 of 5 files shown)


lzap:
> we can remove uuid label and use rid to filter.

Now that I think about it, UUID is safer. An attacker could actually craft an instance with `r-XXX` and wait until it is "gathered" as someone else's instance. Then user could sign int and we have MITM attack. Let’s keep UUID.

PR #633, opened 2023-08-08
fix(HMS-1453): platform RBAC support enabled

lzap:
This PR enables RBAC, it will fail tests.

The following roles are required in order to perform tasks:

https://github.com/RedHatInsights/rbac-config/blob/master/configs/stage/roles/provisioning.json

These are the possible permissions:

https://github.com/RedHatInsights/rbac-config/blob/master/configs/stage/permissions/provisioning.json

Note these roles and permissions are not in production yet, I will link prod PR in a moment.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -524,7 +524,7 @@ parameters:
     value: "debug"
   - description: RBAC checking enabled
     name: APP_RBAC_ENABLED
-    value: "false"
+    value: "true"
   - description: RBAC checking enabled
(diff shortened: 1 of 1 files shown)


lzap:
Prod promote: https://github.com/RedHatInsights/rbac-config/pull/406

ezr-ondrej:
@lzap can we pls flip it to false in prod (app-interface) explicitly, just so we don't forget? :)

lzap:
Added to app-interface.

PR #636, opened 2023-08-09
chore: update deps, clients, tools and all types

lzap:
Review commit by commit.

Make sure to do `make install-tools` after merge so we are all on the same tool versions.


ezr-ondrej (approved):
Not really xD my browser got stuck, so I trust our automation is set up correctly :shrug:

PR #639, opened 2023-08-09
fix(HMS-1453): disable RBAC for sources

lzap:
We have to disable RBAC for sources because image builder UI performs two calls:

- List all sources (for the sources dropdown)
- Get image_info of a source (to get account number or subscription UUID + resource groups)

Since Image Builder doesn't have RBAC implementation yet, there is no role we can add permissions to. Therefore the easiest way is to disable sources RBAC for now and enable it later.


ezr-ondrej (approved):
I'm all for this as intermediate solution it's good enough IMO

PR #644, opened 2023-08-14
fix(HMS-2327): fix empty resource perm check

lzap:
The app perform `reservation:read` check via middleware and then additional `reservation.provider:read` check in the service function. However, it is possible to read a generic reservation where the provider is set to empty string. In this case, check is performed on an incorrect resource `reservation.` (note the trailing dot) which does not match anything and the request ends up with access denied:

image

This refactoring fixes that, the function now takes the verb first and then arbitrary amount of resources, removes empty strings and joins them with `.` characters so this is avoided for the future.

### DIFF
--- a/internal/services/permissions.go
+++ b/internal/services/permissions.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


lzap:
Added a unit test too.

PR #645, opened 2023-08-14
fix(HMS-2033): cache for RBAC

lzap:
Introduces cache for ACLs. I am decreasing the cache application expiration to 10 minutes which is more reasonable for ACLs. Other than that, it works the same as with account (numbers).


ezr-ondrej (commented):
Linter is unhappy.

I like this as I presume it will be a major speed up, but do you think we could let it "boil" for a bit in prod without the cache so we know how much of a speed up it is?
Or would the statistics tell us anyway for the cache misses?

lzap:
> Or would the statistics tell us anyway for the cache misses?

The application cache has built-in metrics for each and every supported Go type. It shows hit/miss ratio, for example for accounts it is over 90% all the time on stage.

lzap:
Amended the liner changes.

internal/clients/http/rbac/rbac_client.go
```diff
@@ -76,57 +79,74 @@ func (c *rbac) GetPrincipalAccess(ctx context.Context) (clients.RbacAcl, error)
...
-		for _, a := range resp.JSON200.Data {
-			result = append(result, clients.NewAccess(a.Permission))
-		}
-		offset += *FetchLimit
 
-		maxQueries -= 1
-		if maxQueries <= 0 {
-			zerolog.Ctx(ctx).Warn().Msg("Maximum amount of RBAC requests reached, giving up")
-			break
+		err = cache.Set(ctx, orgID+accountNumber, &result)
```
ezr-ondrej, line 142: How big can the result be? I se we are storing it all, so we are serializing all permissions the user has.
I see those are filtered, so it should be just permissions to our app, so I guess not a big issue, just verifying :)

lzap: This is a slice of

```
type Access struct {
	Resource string `json:"resource"`
	Verb     string `json:"verb"`
}
```

and on our platform we have several dozens of permissions. In the worst case scenario, perhaps hundreds of elements in the slice. This gets serialized by Gob encoder and stored as blob, it cannot get faster than that.

What is your concern? The bigger the result is, the better caching actually is because transferring a binary blob to Redis will be likely faster than HTTP/JSON (maybe with pagination in the future). Once this is merged, opentelemetry will clearly answer how much faster this is (or not).

ezr-ondrej: My concern is huge documents in Redis, I've heard stories, that redis is not particularly great at handling huge documents, but given I do not have enough info to what "huge" means, it's just an unsupported worry :)
I think the main issue is that Redis is not concurrent, but not sure at all :)

Aren't we filtering the results to only include provisioning permissions tho?

> The bigger the result is, the better caching actually is because transferring

I see biggest value in saving round trip and the computation needed to return the result, I'd not be so sure that the download itself will be significantly faster for big documents (once the connection is established) :thinking:

lzap: > My concern is huge documents in Redis

We are not storing documents, but binary blobs. Maximum value size is 512 MB of each entry, according to google.

> Aren't we filtering the results to only include provisioning permissions tho?

Actually, we only query permissions for "provisioning" so I guess this is the answer to your concern:

```
		params := GetPrincipalAccessParams{
			Application: "provisioning",
			Limit:       FetchLimit,
			Offset:      &offset,
		}
```

Average size of an entry will be less than a memory page.

Okay, enough theory, let’s merge this so I still have a day to provide a fix. Stage logs should answer us what is faster - RBAC API or Redis. It can be anything actually.

ezr-ondrej (approved):
Thanks! :+1:

lzap:
For the record, from the stage logs it seems that a Redis fetch takes less than 2ms whereas RBAC API call takes over 100 ms.

Conclusion: Cache is 100x times faster :-)

PR #646, opened 2023-08-15
chore: log EC2 client STS as warn

lzap:
Otherwise it creates too many GlitchTip events:

https://glitchtip.devshift.net/insights/issues/958355?project=28


ezr-ondrej (approved):
Thanks! :pray:

PR #647, opened 2023-08-15
chore: lower reservation limit to 2

lzap:
This lowers down the limit to 2 which was tested as the only one that works.

Also changes the trigger function not to include the amount of pending reservations in the error message to avoid confusing GlitchTip.

Do not merge until performance team is done with testing.

### DIFF
--- a/internal/dao/tests/reservation_test.go
+++ b/internal/dao/tests/reservation_test.go
@@ -359,7 +359,7 @@ func TestReservationRate(t *testing.T) {
 	rdao, ctx := setupReservation(t)
 	t.Run("allows slow reservations", func(t *testing.T) {
 		defer reset()
-		for i := 1; i <= 5; i++ {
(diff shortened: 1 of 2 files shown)


ezr-ondrej:
What if I have two AWS sources and I (in theory) want to do 2 per second on one and 2 from the other, AWS would be fine with that, but we would limit it :thinking:
I guess it does not matter much, but scoping the limit to source might be something to think about :shrug:

lzap:
Source id is unfortunately stored in the detail JSON, should be possible to do tho if ever needed. I would rather keep this simple and in case customers hit these limitations I would rather suggest to redesign the job queue in a way that it can do throttling itself.

ezr-ondrej:
We've had our performance team to run tests with three accounts and they are hitting the ceiling of 5/account.

I'd rather have it stay bit higher as is, to account for requests for multiple sources, I don't see a reason to be this strict :)

PR #651, opened 2023-08-22
fix(HMS-2258): Check errors after defer

avitova:
This is the first part of the error handling fix, and also the next thing I have noticed. Sometimes, when we use defer, we do not check for errors returned from such functions. I do not think this is correct. If we have a reason we don't do that, let me know. Otherwise, I will look around the codebase to check this in more places.

### DIFF
--- a/internal/clients/http/image_builder/image_client.go
+++ b/internal/clients/http/image_builder/image_client.go
@@ -50,7 +50,11 @@ func (c *ibClient) Ready(ctx context.Context) error {
 		logger.Error().Err(err).Msg("Readiness request failed for image builder")
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		if tempErr := resp.Body.Close(); tempErr != nil {
+			logger.Error().Err(tempErr).Msg("Readiness request for image builder: response body close error")
+		}
+	}()
 
 	err = http.HandleHTTPResponses(ctx, resp.StatusCode)
 	if err != nil {
--- a/internal/clients/http/rbac/rbac_client.go
+++ b/internal/clients/http/rbac/rbac_client.go
@@ -57,7 +57,11 @@ func (c *rbac) Ready(ctx context.Context) error {
 		logger.Error().Err(err).Msgf("Readiness request failed for RBAC: %s", err.Error())
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		if tempErr := resp.Body.Close(); tempErr != nil {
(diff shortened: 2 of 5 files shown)


internal/clients/http/image_builder/image_client.go
```diff
@@ -50,7 +50,11 @@ func (c *ibClient) Ready(ctx context.Context) error {
 		logger.Error().Err(err).Msg("Readiness request failed for image builder")
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		if tempErr := resp.Body.Close(); tempErr != nil {
+			err = tempErr
```
avitova, line 55: Fyi, we have to do this, because the defer is called after the end of the function. Therefore, if we did not use temporary error, and Close() would not return any error, the error variables would get overwritten with the 'nil' value.

internal/clients/http/sources/sources_client.go
```diff
@@ -232,7 +236,11 @@ func (c *sourcesClient) loadAppId(ctx context.Context) (string, error) {
 		logger.Warn().Err(err).Msg("Failed to fetch ApplicationTypes from sources")
 		return "", fmt.Errorf("failed to fetch ApplicationTypes: %w", err)
```
ragevou, line 237:
```suggestion
		return "", fmt.Errorf("Failed to fetch ApplicationTypes: %w", err)
```

internal/clients/http/rbac/rbac_client.go
```diff
@@ -57,7 +57,11 @@ func (c *rbac) Ready(ctx context.Context) error {
 		logger.Error().Err(err).Msgf("Readiness request failed for RBAC: %s", err.Error())
```
ragevou, line 57:
```suggestion
		logger.Error().Err(err).Msgf("Readiness request failed for the RBAC: %s", err.Error())
```

internal/clients/http/sources/sources_client.go
adiabramovitch: Errors should start with a small letter :)

avitova:
Thank you for the comments. I have found more places where we could log this. :) Let me know what y'all think.

ezr-ondrej (approved):
Nice! :)

ezr-ondrej:
Thank you! great find!

lzap:
Try to avoid `temp` or `temporary` variable names in general, this could easily be `closeErr` or `clsErr` or whatever. But thanks, nice one :-D

PR #653, opened 2023-08-22
chore(HMS-1847): Updating cronjob spec

mayurilahane:
The `cron job` param was defined in `app-interface/cloudwash_saas_file ` before and it gives errors
Which says the correct location for it will `provisioning-backend` repo
Thanks, @tpapaioa for finding it


mshriver (approved):

ezr-ondrej:
Thanks! :)

PR #655, opened 2023-08-24
fix(HMS-2258): Refactor error handling

avitova:
I do not think that our HandleHTTPResponses is bad in any way, but I think it should be more clear, that it handles only non2xx responses. Renaming might be enough, but if you have any other ideas, let me know.
Next, when using generated functions for requests, i.e.
`resp, err := ...`, the response should not be nil. But I was a bit afraid to count on that. Let me know if you think we need to check for the response.
If you think of better ways to handle nil pointers, let me know. But I think we'd have to check for the fields anyways.


internal/clients/http/image_builder/image_client.go
```diff
@@ -50,9 +50,12 @@ func (c *ibClient) Ready(ctx context.Context) error {
 		logger.Error().Err(err).Msg("Readiness request failed for image builder")
 		return err
 	}
+	if resp == nil {
+		return fmt.Errorf("ready call: response not available")
+	}
```
adiabramovitch, line 55: I don't have a problem with that, but I am wondering is that possible when err is nil?

adiabramovitch:
@akhil-jha Can you take a look at the error? I think it is related to RBAC

akhil-jha:
The PR needs a rebase.

internal/clients/http/image_builder/image_client.go
lzap: Here is docs of the client Do method:

https://pkg.go.dev/net/http#Client.Do

Here is the generated code:

https://github.com/RHEnVision/provisioning-backend/blob/main/internal/clients/http/image_builder/client.gen.go#L1088

My quick conclusion is that response cannot be nil, unless there was an error.

avitova: Yeah, sure. That makes sense. I was just thinking...of course, in the Client.Do documentation we know this is how the function works. And we should probably be able to depend on that. I was thinking if this was not redundant and we could have a conversation on that for sure!:)
In this case, however, we are using a different function from a generated code,. I did not find any docs that would guarantee us that those functions, if the template possible changes, would have the same behaviour that we expect from the Client.Do.
Any thoughts on this or links to such docs?

lzap: Good thinking, I am not against checking it.

avitova:
Rebased and deleted the HandleHTTPStatusCode function.

lzap (commented):
Looks good.

avitova:
@lzap Thank you for the comments. I incorporated everything, ready for re-review.

internal/clients/http/image_builder/image_client.go
```diff
@@ -45,21 +44,20 @@ func (c *ibClient) Ready(ctx context.Context) error {
...
-	err = http.HandleHTTPResponses(ctx, resp.StatusCode)
-	if err != nil {
-		return fmt.Errorf("ready call: %w", err)
+	if resp == nil {
+		return fmt.Errorf("ready call: empty response: %w", clients.UnexpectedBackendResponse)
 	}
+
+	if resp.StatusCode() < 200 || resp.StatusCode() > 299 {
+		return fmt.Errorf("ready call: %w: %d", clients.UnexpectedBackendResponse, resp.StatusCode())
+	}
```
lzap, line 59: I think you can now use `resp.JSON200 == nil` (in all the three functions) to keep the consistency and prevent copy and paste error.

lzap (commented):
Looks good.

internal/clients/http/image_builder/image_client.go
avitova: Yeah, that is possible, but we do not access resp.JSON200 structure, so we technically do not need this to be only 200, right? Or do we want to error out on things like 201 etc., in case of readiness call?

internal/clients/http/image_builder/image_client.go
```diff
@@ -45,21 +44,20 @@ func (c *ibClient) Ready(ctx context.Context) error {
...
+	resp, err := c.client.GetReadinessWithResponse(ctx, headers.AddImageBuilderIdentityHeader, headers.AddEdgeRequestIdHeader)
 	if err != nil {
 		logger.Error().Err(err).Msg("Readiness request failed for image builder")
 		return err
 	}
-	defer func() {
-		if tempErr := resp.Body.Close(); tempErr != nil {
-			logger.Error().Err(tempErr).Msg("Readiness request for image builder: response body close error")
-		}
-	}()
```
adiabramovitch, line 57: Why did you remove the defer?

avitova: I have replaced the calls with `XXX**WithResponse**` function, as [Lukas recommended](https://github.com/RHEnVision/provisioning-backend/pull/655#discussion_r1315926042). The generated code takes care of closing the body and only returns fields we need that we do not have to close. :)

adiabramovitch (approved):

lzap (approved):
Thank you.

PR #657, opened 2023-08-29
chore: improve user data logging

lzap:
The whole user data (cloud init) is sent to the message field breaking the whole Kibana UI:

image

This fixes it.


ezr-ondrej (approved):
I like it, but linter is unhappy, let me fix that, seems like just imports ordering.

PR #658, opened 2023-08-30
chore: add compile daemon targets

lzap:
So I looked into the compile daemon problem:

I figured out that ability to run multiple "actions" within our pbackend application is more significant refactoring that I thought. Adding just go command to main functions will not work because every single "action" (api, worker, statuser) initializes logging, db, redis and other resources differently. I don’t think such a refactoring is worth the effort.

Then I looked into creating a Makefile target that would run individual "actions", this did not work because different compile daemon processes were overwriting pbackend binary which was causing issues.

The solution is this PR where every binary is different so compile daemon works even if you try to run multiple instances of it. This creates more binaries but the Go cache is shared so compilation is at least much faster.

In the compose, it should be possible to run these from a shell wrapper:

```bash
#!/bin/bash
make cd-stats &
make cd-statuser &
make cd-worker &
make cd-api
```

Something like that: single container, multiple processes, automatic recompilation. I am not going to work on compose part tho, take it from here if you are interested or close this PR.

### DIFF
--- a/mk/compile.mk
+++ b/mk/compile.mk
@@ -20,12 +20,32 @@ run-api: check-go ## Run backend API using `go run`
 	$(GO) run ./cmd/pbackend api
 
 .PHONY: run-worker
-run-worker: check-go ## Run backend API using `go run`
+run-worker: check-go ## Run worker using `go run`
 	$(GO) run ./cmd/pbackend worker
 
 .PHONY: run-statuser
-run-statuser: check-go ## Run backend API using `go run`
+run-statuser: check-go ## Run statuser using `go run`
 	$(GO) run ./cmd/pbackend statuser
+
+.PHONY: run-stats
+run-stats: check-go ## Run stats using `go run`
+	$(GO) run ./cmd/pbackend stats
+
+.PHONY: cd-api
+cd-api: check-go ## Run backend API using compile daemon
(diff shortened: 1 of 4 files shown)


akhil-jha:
pr_check didn't run? 🤔

ezr-ondrej:
The bot was lazy :shrug: xD

mk/compile.mk
```diff
@@ -20,13 +20,33 @@ run-api: check-go ## Run backend API using `go run`
...
+run-worker: check-go ## Run worker using `go run`
 	$(GO) run ./cmd/pbackend worker
 
 .PHONY: run-statuser
-run-statuser: check-go ## Run backend API using `go run`
+run-statuser: check-go ## Run statuser using `go run`
 	$(GO) run ./cmd/pbackend statuser
 
+.PHONY: run-stats
+run-statuser: check-go ## Run stats using `go run`
```
ezr-ondrej, line 31: Typo?
```suggestion
run-stats: check-go ## Run stats using `go run`
```

lzap: Yes this was a typo.

ezr-ondrej (approved):
@amirfefer do you want to implement this in compose?
It would make sense IMO, but up to you, I'm all for merging this even if we don't but I'll let you have a look first :)

adiabramovitch:
We have decided for now to stick with three containers, if needed we will get back to it in the future :)

PR #662, opened 2023-09-04
fix: Redis nil pointers checking

avitova:
This is just a small change. I might wait if I find some other missing nil pointer checking. But these are our own functions with pointer parameters, so I thought I should create a PR. We can also avoid pointers in general and make the functions accept `Job` instead of `*Job`

### DIFF
--- a/internal/jobs/launch_instance_aws.go
+++ b/internal/jobs/launch_instance_aws.go
@@ -10,6 +10,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/dao"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/RHEnVision/provisioning-backend/internal/notifications"
+	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 	"github.com/RHEnVision/provisioning-backend/internal/userdata"
 	"github.com/RHEnVision/provisioning-backend/pkg/worker"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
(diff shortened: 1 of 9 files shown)


internal/jobs/launch_instance_aws.go
```diff
@@ -42,8 +43,13 @@ type LaunchInstanceAWSTaskArgs struct {
 	ARN *clients.Authentication
 }
 
-// Unmarshall arguments and handle error
+// HandleLaunchInstanceAWS unmarshalls arguments and handles error
```
avitova, line 46: Just so that Golang's comment formatting is happy. Comment above a function should start with the function name. Not a big issue, but since I was already refactoring...

internal/ptr/errors.go
```diff
@@ -0,0 +1,7 @@
+package ptr
+
+import (
+	"errors"
+)
+
+var JobNotFound = errors.New("job not found")
```
avitova, line 7: I wanted to put this into the "jobs" package, but since we are using a pointer for a job in both "worker" and "jobs" packages, it is not possible (circular dependency - jobs package imports worker, so we can not import the error from the jobs package in the worker package). If you think of a better placement, feel free to recommend it.

avitova (commented):
Apparently, we use pointers for a job more than I thought so there should probably be an error for JobNotFound or something like that. Or again, we could dereference it right away and not have pointers as a parameter.

internal/ptr/errors.go
adiabramovitch: I would create `errors.go `under jobs, I don't see a problem when I checked locally. 
and if you need at to the worker as well and you have a cyclic import you can create an errors.go under worker as well.

lzap: There is no need to share the error. Make a copy in each package.

If you ask me, creating a error value just to pass it into the logger is probably too much, I know the linter can complain so either disable it for that particular line or just make two copies.

avitova:
> If you ask me, creating a error value just to pass it into the logger is probably too much

Yeah, this was actually a really good point, thank you, Lukáš! We do not really need to create/use the error in the jobs package.

lzap (approved):

internal/jobs/launch_instance_aws.go
```diff
@@ -42,8 +42,13 @@ type LaunchInstanceAWSTaskArgs struct {
...
 func HandleLaunchInstanceAWS(ctx context.Context, job *worker.Job) {
+	if job == nil {
+		zerolog.Ctx(ctx).Error().Msg("No job for HandleLaunchInstanceAWS")
```
adiabramovitch, line 48: nit: If you prefer it, maybe we should add a logger? 
```suggestion
	logger := zerolog.Ctx(ctx)
	logger.Debug().Msg("Started instance launch AWS job")
	if job == nil {
			logger.Debug().Msg("No job for HandleLaunchInstanceAWS")
		}
```

adiabramovitch: ditto on other providers

internal/queue/stub/stub.go
```diff
@@ -58,6 +58,9 @@ func (h hollowEnqueuer) Enqueue(_ context.Context, _ *worker.Job) error {
 }
 
 func (s *stubEnqueuer) Enqueue(ctx context.Context, job *worker.Job) error {
+	if job == nil {
+		return nil
```
adiabramovitch, line 62: Shouldn't we return an error here if the job is nil?

adiabramovitch (commented):
Just small small wondering/nits, feel free to disregard if you don't want to add them

avitova:
@adiabramovitch All the comments are valid. As this is a kind of code enhancement and refactoring, this all should be incorporated. Thank you!

adiabramovitch:
@avitova Thanks!

PR #664, opened 2023-09-05
fix(HMS-1847): Cronjob template update

mayurilahane:
Updated cronjob template

### DIFF
--- a/deploy/cloudwash_job.yaml
+++ b/deploy/cloudwash_job.yaml
@@ -13,54 +13,61 @@ objects:
       "ignore-check.kube-linter.io/no-readiness-probe": "probes not required on Job pods"
   spec:
     schedule: ${CRON_SCHEDULE}
-    backoffLimit: 0
-    template:
+    jobTemplate:
       spec:
-        imagePullSecrets:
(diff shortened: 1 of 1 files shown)


lzap:
I think `chore` does not support HMS references, use `fix(HMS-XXX)`.

ezr-ondrej (approved):
I'm good here, but @mshriver seems to want to take a look, holding off merge :)

deploy/cloudwash_job.yaml
```diff
@@ -13,54 +13,57 @@ objects:
...
+        template:
+          spec:
+            imagePullSecrets:
+              - name: quay-cloudservices-pull
+            restartPolicy: Never
+            containers:
+              - name: provisioning-cleanup-${IMAGE_TAG}-${UID}
+                image: quay.io/redhatqe/cloudwash
+                imagePullPolicy: Always
+                args:
```
mshriver, line 28: I believe you need to specify `command` here, not args, because the cloudwash container doesn't have an ENTRYPOINT defined.

mshriver (approved):

PR #669, opened 2023-09-07
Delete simple pointers

avitova:
Sometimes, we use pointers with no big reason. We'd dereference again and again. Sometimes it is not that needed. I split such cases into commits so that you can see which changes are related to one usage of such a variable.
I will no longer devote time to this, but these were easy changes, and there was no need to use pointers here, so I just wanted to delete these.
Before merging, I think I should merge the commits into one.

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -368,7 +368,7 @@ func (c *ec2Client) ListLaunchTemplates(ctx context.Context) ([]*clients.LaunchT
 	return res, nil
 }
 
-func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanceParams, amount int32, name *string, reservation *models.AWSReservation) ([]*string, *string, error) {
+func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanceParams, amount int32, name string, reservation *models.AWSReservation) ([]*string, *string, error) {
 	ctx, span := otel.Tracer(TraceName).Start(ctx, "RunInstances")
 	defer span.End()
 
@@ -408,10 +408,10 @@ func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanc
 		},
 	}
 
-	if name != nil {
+	if name != "" {
 		t := types.Tag{
 			Key:   ptr.To("Name"),
-			Value: name,
+			Value: &name,
 		}
 		input.TagSpecifications[0].Tags = append(input.TagSpecifications[0].Tags, t)
 	}
--- a/docs/dev-environment.md
+++ b/docs/dev-environment.md
@@ -116,7 +116,7 @@ Statuser process (`pbstatuser`) is a custom executable that runs in a single ins
 
 [Sources](https://github.com/RedHatInsights/sources-api-go) is an authentication inventory. Since it only requires Go, Redis and Postgres, we created a shell script that automatically checks out sources from git, compiles it, installs and creates postgres database, seeds data and starts the Sources application.
 
-Follow [instructions](../scripts/README.sources) to perform the setup. Note that configuration via `sources.local.conf` is **required** before the setup procedure. This has been written and tested for Fedora Linux, in other operating systems perform all the commands manually.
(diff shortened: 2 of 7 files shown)


avitova:
Solved, AWS ReservationID was quite a lot to handle. :) Let's leave it as it is.

internal/clients/instance_type_info.go
```diff
@@ -13,15 +13,15 @@ type InstanceTypeInfo struct {
...
+func (iii *InstanceTypeInfo) InstanceTypesForZone(region, zone string, supported bool) ([]*InstanceType, error) {
 	names, err := iii.RegionalAvailability.NamesForZone(region, zone)
 	if err != nil {
 		return nil, err
 	}
 	result := make([]*InstanceType, 0, len(names))
 	for _, name := range names {
 		rt := iii.RegisteredTypes.Get(name)
-		if supported != nil && *supported != rt.Supported {
+		if supported != rt.Supported {
```
ezr-ondrej, line 24: I wonder ... could have this been used as "not passing a value => include all"?
Now there is no way to get both `Supported = true` and `Supported = false`.

I'm not sure if we were using it that way, nor I'm saying it was a good use of pointer. Tho now the function changed behaviour, so just making sure it is intended :wink:

avitova: Yeah, you are correct. Interesting. I think it makes sense as it is, even though I struggled to understand the logic. Thank you for the comment. I will leave out this part for now.

adiabramovitch:
Can you squash the commits?

internal/clients/instance_type_info.go
adiabramovitch: Are you changing it back?

adiabramovitch (approved):
LGTM, just two questions.

avitova:
@adiabramovitch Thank you. Since reviews are more or less here, I have done a rebase. :+1:

internal/clients/http/ec2/ec2_client.go
```diff
@@ -408,13 +408,11 @@ func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanc
...
 
-	if name != nil {
-		t := types.Tag{
-			Key:   ptr.To("Name"),
-			Value: name,
-		}
-		input.TagSpecifications[0].Tags = append(input.TagSpecifications[0].Tags, t)
+	t := types.Tag{
+		Key:   ptr.To("Name"),
+		Value: &name,
```
ezr-ondrej, line 413: Actually do we know what this does?
I'm guessing we don't want to have blank tag here, right?
So if it would end up tagging the instance with `name: ''`, I'd keep the condition, just checked for empty string instead of nil

internal/testing/identity/dummy_identity.go
```diff
@@ -20,7 +19,7 @@ const (
 	DefaultAccountNumber = "1"
 )
 
-var xRhId = newIdentity(DefaultOrgId, ptr.To(DefaultAccountNumber))
```
ezr-ondrej, line 23: I guess this is fine as it won't have too much consequences, but I wonder:

We still use pointer to account_number in `models.Account`, but changing it here in tests.
I'd rather see it kept the same, tho migrating the account_number to string in prod might be bit tricky...
Currently we do not rely on the account_number anywhere IMO, so it's not a big deal, but what are your thoughts on having this difference?

ezr-ondrej (commented):
Sorry for the delay, I was just scanning through it the first time, did not do the full review.
I see few potential weeknesses of the migration in few places.
I guess we don't have much test coverage for the code we're changing, so it doesn't help xD

I'm not sure of consequences of any of it, so take my review just as a raised eyebrow, feel free to dismiss if you think it's safe :)

avitova (commented):
I do not want to break anything, so let's leave the tests as they are. :) Adding a commit to update our docs.

ezr-ondrej (approved):
:orange_heart: Thank you! Great excercise, I'd love to do more of these, but I guess we should improve the test coverage to have easier time to do so.
All the points I've raised should have failed a unit test IMHO :thinking:

PR #671, opened 2023-09-07
fix(HMS-2519): error handling of jobs

lzap:
Looks like our error handling of the last step was not right. In some cases, reservations are kept in "pending" state when the instance info could not be retrieved. This fixes it.

Also moves handling of notifications into a helper function, we only have jobs for reservations so it is fine for now to have it here.

Finally, adds open telemetry to AWS and GCP similarly as it is already implemented for Azure.


ezr-ondrej (approved):
Nice! :)
Except the Import ordering, do you have different imports ordering set up in goland then the linter wants maybe? :thinking:

.golangci.yml
```diff
@@ -20,7 +20,6 @@ linters:
     - gci # too strict
     - contextcheck # too restrictive
     - depguard # too verbose
-    - tagalign # cleanenv is unreadable
```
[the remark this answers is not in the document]
avitova: Happened to me. Check your golangci-lint version.
Tagalign is available from v1.53.0.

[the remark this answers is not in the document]
lzap: Ah I had an old version of the linter locally :-D

internal/jobs/launch_instance_azure.go
```diff
@@ -95,6 +91,7 @@ func DoEnsureAzureResourceGroup(ctx context.Context, args *LaunchInstanceAzureTa
 
 	azureClient, err := clients.GetAzureClient(ctx, args.Subscription)
 	if err != nil {
+		logger.Error().Err(err).Msg("cannot create new Azure client")
```
avitova, line 94:
```suggestion
		logger.Error().Err(err).Msg("Cannot create new Azure client")
```

internal/jobs/launch_instance_gcp.go
```diff
@@ -150,6 +159,9 @@ func DoLaunchInstanceGCP(ctx context.Context, args *LaunchInstanceGCPTaskArgs) e
 }
 
 func FetchInstancesDescriptionGCP(ctx context.Context, args *LaunchInstanceGCPTaskArgs) error {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "FetchInstancesDescriptionAWS")
```
avitova, line 162:
```suggestion
	ctx, span := otel.Tracer(TraceName).Start(ctx, "FetchInstancesDescriptionGCP")
```

internal/jobs/launch_instance_gcp.go
```diff
@@ -56,26 +58,27 @@ func HandleLaunchInstanceGCP(ctx context.Context, job *worker.Job) {
...
+		finishWithError(ctx, args.ReservationID, jobErr)
+		return
 	}
+
 	finishJob(ctx, args.ReservationID, jobErr)
 }
 
 // DoLaunchInstanceGCP is a job logic, when error is returned the job status is updated accordingly
 func DoLaunchInstanceGCP(ctx context.Context, args *LaunchInstanceGCPTaskArgs) error {
+	ctx, span := otel.Tracer(TraceName).Start(ctx, "FetchInstancesDescriptionAWS")
```
avitova, line 79:
```suggestion
	ctx, span := otel.Tracer(TraceName).Start(ctx, "FetchInstancesDescriptionGCP")
```

internal/jobs/launch_instance_aws.go
```diff
@@ -273,13 +297,15 @@ func FetchInstancesDescriptionAWS(ctx context.Context, args *LaunchInstanceAWSTa
 		for _, instance := range instancesDescriptionList {
 			errRetry := rDao.UpdateReservationInstance(ctx, args.ReservationID, instance)
 			if errRetry != nil {
+				span.SetStatus(codes.Error, "cannot update instance description")
 				return fmt.Errorf("cannot update instance description: %w", err)
```
avitova, line 301:
```suggestion
				return fmt.Errorf("cannot update instance description: %w", errRetry)
```

internal/jobs/launch_instance_aws.go
```diff
@@ -60,33 +62,32 @@ func HandleLaunchInstanceAWS(ctx context.Context, job *worker.Job) {
...
 		return
 	}
+
 	jobErr = FetchInstancesDescriptionAWS(ctx, &args)
 	if jobErr != nil {
-		nc.FailedLaunch(ctx, args.ReservationID, jobErr)
-	} else {
-		nc.SuccessfulLaunch(ctx, args.ReservationID)
+		finishWithError(ctx, args.ReservationID, jobErr)
 	}
```
lzap, line 81: This is the actual bug - instead of calling `finishWithError` we just sent a notification and skipped on the error.

avitova (commented):
Some more things. Sorry for not noticing earlier. They are more or less only nits.

internal/jobs/launch_instance_azure.go
lzap: Lesson learned for my next project - all logging start with lower case :-)

internal/jobs/launch_instance_aws.go
ezr-ondrej: hmmm when thinking about it, shall we fail here? as this step is not required... and we will report error on it, Users might get confused :thinking:

lzap:
Great finds.

I also found one suspicions code that I want to remove, we only finish a reservation in case all steps were processed. But the `finishJob` function is always called at the very end of each job. So I think this was a mistake and I remove that. I added a error log in case a job that did not run all jobs is finished.

Finally, I also added panic handing via `recover`, in case there was a panic (nil pointer) the reservation is immediately finished with failure flag and the panic value (string) is logged and stored in status.

internal/jobs/common.go
```diff
@@ -31,25 +38,26 @@ func finishWithSuccess(ctx context.Context, reservationId int64) {
...
 	metrics.IncReservationCount(reservation.Provider.String(), "success")
 
-	// if this was the last step, set the success flag
-	if reservation.Step >= reservation.Steps {
-		logger.Info().Msgf("All jobs executed, marking job as success")
-
-		err = rDao.FinishWithSuccess(ctx, reservationId)
-		if err != nil {
-			logger.Warn().Err(err).Msg("unable to update job status: finish")
-		}
```
lzap, line 52: This smells, I do not remember why I wrote this, finish function should be only called as the very last step therefore it does not matter at which step we are. I removed this and now it simply finishes properly, tho, there is some extra logging above just to inform us in case we finish a reservation that did not execute all steps.

internal/jobs/common.go
```diff
@@ -15,11 +17,16 @@ import (
 const TraceName = telemetry.TracePrefix + "internal/jobs"
 
 var ErrTypeAssertion = errors.New("type assert error")
+var ErrPanicInJob = errors.New("panic during job")
```
ezr-ondrej, line 20: No idea what linter is complaining about :thinking: 
My guess would be that it wants:
```suggestion
var (
  ErrTypeAssertion = errors.New("type assert error")
  ErrPanicInJob = errors.New("panic during job")
)
```

ezr-ondrej: Yup, now it's happy! :tada:

internal/jobs/launch_instance_aws.go
```diff
@@ -100,11 +108,13 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
 	resDao := dao.GetReservationDao(ctx)
 	awsReservation, err := resDao.GetAWSById(ctx, args.ReservationID)
 	if err != nil {
+		span.SetStatus(codes.Error, "cannot get aws reservation by id")
 		return fmt.Errorf("cannot get aws reservation by id: %w", err)
 	}
 
 	pubkey, err := pkDao.GetById(ctx, args.PubkeyID)
 	if err != nil {
+		span.SetStatus(codes.Error, "cannot upload aws key")
```
adiabramovitch, line 117:
```suggestion
		span.SetStatus(codes.Error, "cannot get key")
```

lzap: Okay rebased

internal/jobs/common.go
```diff
@@ -31,25 +38,26 @@ func finishWithSuccess(ctx context.Context, reservationId int64) {
...
 	if err != nil {
 		logger.Warn().Err(err).Msg("unable to update job status: get by id")
 		return
 	}
-	logger.Debug().Msgf("Job step: %d/%d", reservation.Step, reservation.Steps)
+	if reservation.Step == reservation.Steps {
+		logger.Info().Msgf("Finishing reservation with success at step %d/%d", reservation.Step, reservation.Steps)
+	} else {
+		logger.Error().Msgf("Finishing reservation with success at step %d/%d", reservation.Step, reservation.Steps)
+	}
```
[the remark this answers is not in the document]
lzap: Feels weird but I want this to be logged into Glitchtip because a reservation was finished while expected amount of steps was not fulfilled. Could be a warning but we do not send warnings to Glitchtip.

lzap:
Fixed linter.

avitova (approved):
LGTM :+1:

PR #672, opened 2023-09-08
chore: remove logging of bad input values

lzap:
Our security team runs regular penetration tests and glitchtip is cluttered with errors, I don’t think we need any logging for this it is pretty normal to expect number to convert to number and fall back to default.

I think it is worth putting limit and offset into metadata tho, because when you don’t provide it it might be useful to see the default value. But not in this PR.

image


ezr-ondrej (approved):
Thank you, agreed! 👍

PR #674, opened 2023-09-08
chore: shorten message field for unhandled panics

lzap:
Looks like we are getting way too many glitchtip events in case of many of panics:

image

I am moving the stacktrace into a different field so we keep only the error message. This should hopefully solve this.


ezr-ondrej (approved):
Nice find :+1:
I wondered why the panic looks so ugly :))

PR #675, opened 2023-09-08
refactor: remove deprecated /account_identity

lzap:
SSIA

Can we do that now @ezr-ondrej ? All code gone from IB at this point?

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -1996,46 +1996,6 @@
         ]
       }
     },
-    "/sources/{ID}/account_identity": {
-      "get": {
-        "deprecated": true,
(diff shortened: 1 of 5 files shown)


ezr-ondrej (approved):
Yep, @lucasgarfield (pinging for awareness in case it causes troubles) already refactored it to use the `upload_info` instead.

https://github.com/RedHatInsights/image-builder-frontend/blob/0822a696197ebf63659650794e519bf004c965a0/src/Components/CreateImageWizard/formComponents/AwsAccountId.tsx#L10

https://github.com/osbuild/image-builder/blob/c27adabf487e2ced3adcd29ad2e62f73c1662bfd/internal/v1/handler.go#L659C6-L659C6

PR #676, opened 2023-09-08
chore: add status and steps to floorist

lzap:
This will allow us to understand possible issues with pending reservations. Hopefully all is resolved now, I still see 4 pending at the moment on stage so I would like to investigate.


ezr-ondrej (approved):
LGTM, FYI @avitova :)

PR #679, opened 2023-09-11
fix(HMS-2545): add rh labels to all providers

lzap:
We already have "rh-rid" for EC2 and GCP, this patch adds the same for Azure.

On top of that, this patch also adds "rh-org" with organization number. This was requested by Mike and I hope it will help to solve the problem.

### DIFF
--- a/internal/clients/http/azure/create_vm.go
+++ b/internal/clients/http/azure/create_vm.go
@@ -45,7 +45,7 @@ func (c *client) BeginCreateVM(ctx context.Context, networkInterface *armnetwork
 		return "", err
 	}
 
-	vmAzureParams := c.prepareVirtualMachineParameters(vmParams.Location, armcompute.VirtualMachineSizeTypes(vmParams.InstanceType), networkInterface, vmParams.ImageID, vmParams.Pubkey.Body, vmParams.UserData, vmName)
+	vmAzureParams := c.prepareVirtualMachineParameters(vmParams, networkInterface, vmName)
 
 	poller, err := vmClient.BeginCreateOrUpdate(ctx, vmParams.ResourceGroupName, vmName, *vmAzureParams, nil)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/clients/http/ec2/ec2_client.go
```diff
@@ -408,6 +409,10 @@ func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanc
 					Key:   ptr.To("rh-rid"),
 					Value: ptr.To(config.EnvironmentPrefix("r", strconv.FormatInt(reservation.ID, 10))),
 				},
+				{
+					Key:   ptr.To("rh-org"),
+					Value: ptr.To(identity.Identity(ctx).Identity.OrgID),
```
ezr-ondrej, line 414: @mshriver is this what you were after? :)

ezr-ondrej (commented):
Linter does not like the imports :)

internal/clients/http/azure/create_vm.go
```diff
@@ -33,7 +37,7 @@ const (
 	vmPollFrequency       = 10 * time.Second
 )
 
-func (c *client) BeginCreateVM(ctx context.Context, networkInterface *armnetwork.Interface, vmParams clients.AzureInstanceParams, vmName string) (string, error) {
+func (c *client) BeginCreateVM(ctx context.Context, reservation *models.AzureReservation, networkInterface *armnetwork.Interface, vmParams clients.AzureInstanceParams, vmName string) (string, error) {
```
ezr-ondrej, line 40: I'd rather not pass the reservation in it's entirety just for this (ideally not at all). It's ofc just a feeling, if you think we should, I'm ok :)

Tho could we add the tags in vmParams? We could add a map there and pass the prepared map of tags :thinking:

WDYT?

lzap: Good idea, I have missed that.

lzap:
Fixed params, fixed linter.

lzap:
All green.

internal/clients/http/azure/create_vm.go
```diff
@@ -453,7 +454,7 @@ func (c *client) prepareVirtualMachineParameters(location string, instanceType a
 						PublicKeys: []*armcompute.SSHPublicKey{
 							{
 								Path:    to.Ptr(fmt.Sprintf("/home/%s/.ssh/authorized_keys", adminUsername)),
-								KeyData: to.Ptr(sshKeyBody),
+								KeyData: to.Ptr(vmParams.Pubkey.Body),
```
avitova, line 457: Can Pubkey be 'nil' here in some cases? :thinking: 
Since this is a pointer, we do not have to makeup values, so we could in theory do:
```
keydata := nil
if vmParams.Pubkey != nil {
    keydata = to.Ptr(vmParams.Pubkey.Body)
}
...
KeyData: keydata,
```
This violates the happy code path we want to keep, but this is just an idea.

internal/jobs/launch_instance_azure.go
```diff
@@ -154,6 +158,10 @@ func DoLaunchInstanceAzure(ctx context.Context, args *LaunchInstanceAzureTaskArg
 		Pubkey:            pubkey,
 		InstanceType:      clients.InstanceTypeName(reservation.Detail.InstanceSize),
 		UserData:          userData,
+		Tags: map[string]*string{
+			"rh-rid": ptr.To(config.EnvironmentPrefix("r", strconv.FormatInt(reservation.ID, 10))),
```
avitova, line 162: Just a reminder that this is a pointer but we do not return nil anywhere, so it is okay for now.

avitova (commented):
LGTM, thanks!:) I have few reminders, but nothing important if we do not intend to return nil.

ezr-ondrej (approved):
I'm good, I don't see the pointers as a big deal as those should never be nil (and if they  are we should fail sooner).
But I'll wait for @lzap's response here to see what he thinks :)

internal/jobs/launch_instance_azure.go
lzap: Yeah it is because Azure requires exactly this type, otherwise I would need to copy the values manually.

lzap:
> I'm good, I don't see the pointers as a big deal as those should never be nil (and if they are we should fail sooner). But I'll wait for @lzap's response here to se what he thinks :)

Well, I love how Anna is going after pointers, this is great. Pointers are often overused, as you can see I am introducing unnecessary pointer in the code just because Azure library needs it so I am avoiding copying. Let me take a look.

internal/clients/http/azure/create_vm.go
lzap: You are right, pubkey does not need to be pointer in the vmParams type, I can remove that. We fetch it from the database and in case it does not exist it is an error, so it is guaranteed to be present.

lzap:
Rebased.

ezr-ondrej:
@avitova are you good with this going in? :)

avitova:
Yeah, sure:) @ezr-ondrej

ezr-ondrej:
Thanks @lzap and @avitova ! :tada:

PR #681, opened 2023-09-12
fix(HMS-2555): create reservation right before job

lzap:
This fixes the problem when source does not have proper app associated, then the reservation is created but never finishes because API errors out before a new job is submitted.


adiabramovitch (approved):
Thanks

PR #682, opened 2023-09-12
docs: Split the IAM policy to external file

ezr-ondrej:


docs/aws-iam-role-policy.json
avitova, line 1: A nitpick, but maybe let's drop the 'iam' part since these have different prefixes?:D

avitova (commented):
Should this actually be in the 'docs' directory? We'd need to use it even in the code, so something tells me maybe this should be in the config? :thinking: 
We should maybe think of this, once we change the place for the doc, we'd have to think of changing URLs on several places.
Also, please, could you refer this change in the [CONTRIBUTING](https://github.com/RHEnVision/provisioning-backend/blob/main/CONTRIBUTING.md#basic-guidelines-for-code-contributions) doc so that the part about changing AWS permissions is relevant? Thank youu

ezr-ondrej:
Agreed, it's not well placed and you're totaly right we should place it well, moving it will be hard.
Not sure if config is good, as it's not really config for the app...
Any preference between:

- `config`
- `templates`
- `distro`
- `build`
- `docs`

I'm fine with either, I'm  not sure what's the correct place :)

lzap:
Permissions are in fact part of IAM, I am okay with anything. I will not start review on this one, you decide @avitova thanks.

avitova (approved):
You know what? I think that the ide should eventually be able to notify us if we find out later that the placement is not the best. And lzap is also right about the 'iam' name. :+1: I think we are good. ;)

lzap:
Please merge as you are the only reviewer on this PR, I removed myself from reviewers list :-)

avitova:
Cool. Thank you, Ondrej!:)

PR #684, opened 2023-09-15
chore: print pubkey on error

lzap:
A little bit more detailed logging. Public keys can be safely logged (private keys are not stored at all in our DB).

### DIFF
--- a/internal/jobs/launch_instance_aws.go
+++ b/internal/jobs/launch_instance_aws.go
@@ -143,29 +143,40 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
 	// check presence on AWS first
 	fingerprint := pubkey.FindAwsFingerprint(ctx)
 	ec2Name, err := ec2Client.GetPubkeyName(ctx, fingerprint)
-	if err != nil {
-		span.SetStatus(codes.Error, "key error")
+
+	if errors.Is(err, http.ErrPubkeyNotFound) {
+		logger.Trace().Msgf("Pubkey with fingerprint '%s' not found, importing type %s key '%s'", fingerprint, pubkey.Type, pubkey.Body)
 
 		// if not found on AWS, import
-		if errors.Is(err, http.ErrPubkeyNotFound) {
-			pkr.Tag = ""
-			pkr.RandomizeTag()
-			pkr.Handle, err = ec2Client.ImportPubkey(ctx, pubkey, pkr.FormattedTag())
+		pkr.Tag = ""
+		pkr.RandomizeTag()
+		var retryCount int
+		origName := pubkey.Name
+	retry:
+		pkr.Handle, err = ec2Client.ImportPubkey(ctx, pubkey, pkr.FormattedTag())
 
-			if errors.Is(err, http.ErrDuplicatePubkey) {
-				// key not found by fingerprint but importing failed for duplicate err so fingerprints do not match
-				logger.Warn().Msgf("Pubkey with fingerprint %s not found on AWS, but importing the key failed: %s", pubkey.Fingerprint, err.Error())
-				return fmt.Errorf("cannot upload aws pubkey because of duplicate: %w", err)
-			} else if err != nil {
-				return fmt.Errorf("cannot upload aws pubkey: %w", err)
+		if errors.Is(err, http.ErrDuplicatePubkey) {
+			retryCount += 1
+			pubkey.Name = fmt.Sprintf("%s%d", origName, retryCount+1)
+			if retryCount < 4 {
+				goto retry
 			}
-			ec2Name = pubkey.Name
-		} else {
-			logger.Error().Err(err).Str("pubkey_fingerprint", fingerprint).Msg("Cannot fetch name of pubkey by its fingerprint")
-			return fmt.Errorf("cannot fetch name of pubkey by its fingerprint: %w", err)
+			// key not found by fingerprint but importing failed for duplicate err so fingerprints do not match
+			span.SetStatus(codes.Error, "import duplicate key error")
+			return fmt.Errorf("key with fingerprint %s not found on AWS, but importing the key failed: %w", pubkey.Fingerprint, err)
 		}
+
+		if err != nil {
+			return fmt.Errorf("cannot upload aws pubkey: %w", err)
+		}
+
+		ec2Name = pubkey.Name
+	} else if err != nil {
+		span.SetStatus(codes.Error, "import key error")
+		logger.Error().Err(err).Str("pubkey_fingerprint", fingerprint).Msg("Cannot fetch name of pubkey by its fingerprint")
+		return fmt.Errorf("cannot fetch name of pubkey by its fingerprint: %w", err)
 	} else {
-		logger.Debug().Msgf("Found pubkey by fingerprint (%s) with name '%s'", fingerprint, ec2Name)
+		logger.Debug().Msgf("Found pubkey by fingerprint '%s' with name '%s'", fingerprint, ec2Name)
 	}
 
 	// update the AWS key name in reservation details


internal/jobs/launch_instance_aws.go
```diff
@@ -144,7 +144,7 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
 	fingerprint := pubkey.FindAwsFingerprint(ctx)
 	ec2Name, err := ec2Client.GetPubkeyName(ctx, fingerprint)
 	if err != nil {
-		span.SetStatus(codes.Error, "key error")
+		logger.Debug().Msgf("Pubkey with fingerprint '%s' not found, importing type %s key '%s'", fingerprint, pubkey.Type, pubkey.Body)
```
ezr-ondrej, line 147:
```suggestion
		logger.Trace().Msgf("Pubkey with fingerprint '%s' not found, importing type %s key '%s'", fingerprint, pubkey.Type, pubkey.Body)
```

Maybe? :)

lzap:
So the problem turns out in the NAME not Fingerprint. I think I can refactor this patch and implement "try again with a different name".

ezr-ondrej:
That would be awesome! 🧡

lzap:
How about this? :-)

internal/jobs/launch_instance_aws.go
```diff
@@ -144,27 +144,37 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
...
 		if errors.Is(err, http.PubkeyNotFoundErr) {
 			pkr.Tag = ""
 			pkr.RandomizeTag()
+			retryCount := 0
+			origName := pubkey.Name
+		retry:
 			pkr.Handle, err = ec2Client.ImportPubkey(ctx, pubkey, pkr.FormattedTag())
 
 			if errors.Is(err, http.DuplicatePubkeyErr) {
+				retryCount += retryCount + 1
```
avitova, line 159: This would be a sequence of: 0, 1, 3, 7, etc. Is that intentional?
```suggestion
				retryCount += 1
```

lzap: Haha made my day, no obviously this was a typo.

lzap:
Rebased, please test I did not test this.

internal/jobs/launch_instance_aws.go
```diff
@@ -144,27 +144,37 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
...
 			pkr.RandomizeTag()
+			retryCount := 1
+			origName := pubkey.Name
+		retry:
 			pkr.Handle, err = ec2Client.ImportPubkey(ctx, pubkey, pkr.FormattedTag())
 
 			if errors.Is(err, http.DuplicatePubkeyErr) {
+				retryCount += 1
+				pubkey.Name = fmt.Sprintf("%s%d", origName, retryCount)
+				if retryCount < 4 {
```
avitova, line 161: I feel like this might get confusing. This says that the retryCount is < 4, but this retry block is actually performed twice. This is just by the fact that we are starting from 1 and also checking for "<".
Just by the looks of it, I would expect the ImportPubkey to be invoked at least three times. Can we change this a bit so it is more clear how many times we actually run this?

avitova (commented):
Sorry for late review. I got a bit lost in the readability of the function, so I have few ideas how to make this better. Feel free to not deal with it. But I would probably do something with the retryCount readability. :)

internal/jobs/launch_instance_aws.go
```diff
@@ -144,27 +144,37 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
...
-		span.SetStatus(codes.Error, "key error")
+		logger.Trace().Msgf("Pubkey with fingerprint '%s' not found, importing type %s key '%s'", fingerprint, pubkey.Type, pubkey.Body)
 
 		// if not found on AWS, import
 		if errors.Is(err, http.PubkeyNotFoundErr) {
 			pkr.Tag = ""
 			pkr.RandomizeTag()
+			retryCount := 1
+			origName := pubkey.Name
+		retry:
```
ezr-ondrej, line 155: Why do we need a sorcery here? xD
Wouldn't simple for loop with break suffice here?
I'd personally prefer it as more ppl might be more familiar with it and thus having easier time to read it :)

lzap: Loop introduces a new indentation level, labels are feature of the language and I did some research and one of the main use across the community is for retries.

If you insist tho, I can remove and introduce a loop.

internal/jobs/launch_instance_aws.go
lzap: This was not intentional, you are right. Rebased a change with starting from zero.

internal/jobs/launch_instance_aws.go
```diff
@@ -144,27 +144,37 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
...
 			} else if err != nil {
 				return fmt.Errorf("cannot upload aws pubkey: %w", err)
 			}
 			ec2Name = pubkey.Name
 		} else {
+			span.SetStatus(codes.Error, "import key error")
 			logger.Error().Err(err).Str("pubkey_fingerprint", fingerprint).Msg("Cannot fetch name of pubkey by its fingerprint")
 			return fmt.Errorf("cannot fetch name of pubkey by its fingerprint: %w", err)
 		}
 	} else {
```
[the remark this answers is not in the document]
lzap: Yep, this code is clunky. Let me refactor it completely.

lzap:
How about this?

Vastly simplified, removed multiple nested levels, still uses goto which everyone have strong opinions on (we can talk about it).

Please test before merge, this can break things up quite a bit.

avitova (approved):
Nice! So, I have tested all of the flows, and everything seems to be working correctly.

internal/jobs/launch_instance_aws.go
```diff
@@ -143,28 +143,40 @@ func DoEnsurePubkeyOnAWS(ctx context.Context, args *LaunchInstanceAWSTaskArgs) e
...
+		pkr.Tag = ""
+		pkr.RandomizeTag()
+		var retryCount int
+		origName := pubkey.Name
+	retry:
+		pkr.Handle, err = ec2Client.ImportPubkey(ctx, pubkey, pkr.FormattedTag())
+
+		if errors.Is(err, http.DuplicatePubkeyErr) {
+			retryCount += 1
+			pubkey.Name = fmt.Sprintf("%s%d", origName, retryCount+1)
```
avitova, line 160: Just an idea: add a separator between the key name and new suffix. If we had a key named "key1", it suddenly becomes a "key12" if we upload new. But that is nothing major. :thinking:

lzap: I don’t thing we ever want to retry more than 3-5 times. If you insist I can rebase, this one looks more "human" tho.

ezr-ondrej:
needs rebase after the error rename

internal/jobs/launch_instance_aws.go
avitova: No, I think that the amount of retries is sufficient. I was overthinking the suffix appending. I think this is okay. :+1:

lzap:
Done.

ezr-ondrej (approved):
Looks sane to me :+1:
@avitova acked the last version if I understand correctly, so lets get it in.

:warning: I'd like to hold off until we figure out the Kafka issue in prod.

ezr-ondrej:
Kafka now works, we will want to play with it more, but got time.
Thanks @lzap and @avitova! Merging :)

ezr-ondrej:
This actually fixed https://issues.redhat.com/browse/HMS-711 :) I'll link it.

PR #686, opened 2023-09-21
fix(HMS-2640): Support missing values when using launch template

adiabramovitch:
There's an issue concerning GCP launch templates that involves different behaviors between the UI and API.
In the API, we  allow empty values for machine type and image ID so that users can launch instances with the properties they've defined in their templates, a feature not available in the UI.

This small fix aims to support both behaviors. On one hand, if a user provides a machine type or image ID, the instance will launch with those specified values. On the other hand, if a user selects a launch template and leaves the machine type and image ID blank, the instance will launch using the information provided in the template.

This adjustment might serves as a temporary solution for utilizing templates in GCP. I also believe it's essential to include a test for this behavior. It think it should ideally be part of the launch templates tests in IQE, but please correct me if I am wrong.

### DIFF
--- a/internal/clients/http/gcp/gcp_client.go
+++ b/internal/clients/http/gcp/gcp_client.go
@@ -191,17 +191,6 @@ func (c *gcpClient) InsertInstances(ctx context.Context, params *clients.GCPInst
 					"rh-uuid": params.UUID,
 					"rh-org":  identity.Identity(ctx).Identity.OrgID,
 				},
-				Disks: []*computepb.AttachedDisk{
-					{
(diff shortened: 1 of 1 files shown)


avitova (approved):
LGTM. :) I think we might want to have a test for these kinds of values. Not necessarily now, but do you think it is possible to test the gcp client in general?

lzap:
Agreed, this is vague, I don’t even know if we have the same behavior across the board. API tests would be great when template is present to endure the same behavior for both AWS and GCP. @akhil-jha 

@avitova merge at will, thanks for the review.

PR #689, opened 2023-09-26
fix(HMS-2640): Support missing values when using aws launch template

adiabramovitch:
Continue #686 
My AWS setup is broken so I have asked @avitova to check it out :)

### DIFF
--- a/internal/clients/http/ec2/ec2_client.go
+++ b/internal/clients/http/ec2/ec2_client.go
@@ -396,10 +396,16 @@ func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanc
 		LaunchTemplate: templateSpec,
 		MaxCount:       ptr.To(amount),
 		MinCount:       ptr.To(amount),
-		InstanceType:   params.InstanceType,
-		ImageId:        ptr.To(params.AMI),
 		KeyName:        &params.KeyName,
 		UserData:       &encodedUserData,
+	}
+
+	if params.InstanceType != "" {
+		input.InstanceType = params.InstanceType
+	}
+
+	if params.AMI != "" {
+		input.ImageId = ptr.To(params.AMI)
 	}
 
 	input.TagSpecifications = []types.TagSpecification{


internal/clients/http/ec2/ec2_client.go
```diff
@@ -396,12 +396,18 @@ func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanc
...
 		UserData:       &encodedUserData,
 	}
 
+	if params.InstanceType != "" {
+		input.InstanceType = params.InstanceType
+	}
+
+	if params.AMI != "" {
+		input.ImageId = ptr.To(params.AMI)
+	}
```
avitova, line 409: Okay, this behaviour should be slightly different.
If we do not provide instance type and launch template, this errors out anyway. So, I am not sure if we want to have this case.

> WRN Invalid request: Both instance type and launch template are missing error="instance type or launch template not set"

If we do not provide the ImageId at all (either nothing or a pointer to an empty string), this does not return 400, but ends up with 

> ERR Finishing reservation with error at step 2/3 error="cannot run instances: cannot run instances: operation error EC2: RunInstances, https response error StatusCode: 400, api error MissingParameter: The request must contain the parameter ImageId". But we would create the reservation anyway and return the reservation payload.


Since we can not make a call with an empty ImageId, we should cover this a bit earlier. What do y'all think?

avitova (commented):
These cases are both solved differently. So I think that we do not have to check for the InstanceType being empty string, because it would not go through anyway.
For the Image Id, the behaviour is a bit different, but kind of a same story. :thinking:

internal/clients/http/ec2/ec2_client.go
adiabramovitch: First, have you created a launch templates with an image id and with an instance type? We first need to check if the instance is being created from the properties in the template.

adiabramovitch: And that is what I have said- this is an extra mile step, we should check for two cases:
1. Launch template provides instance type/image id and the fields for instance type/image id are empty 

2. Launch templates have the instance type/image id fields propagated and also the fields in the requests are propagated as well. The instance is supposed to be created from the values of the fields not of the template

avitova:
I am sorry for the late response. I have finally been able to look into this a bit more.
So yeah, if you create this with a launch template that provides us with the instance type and image id, it is possible to launch through the API (even though on the frontend, we still require the instance type value. So the 'next' button is disabled when not providing it. But yeah, if we are about to do this, we can work on this later.
The response, though, returns these values empty. Is that desired behaviour?

adiabramovitch:
> I am sorry for the late response. I have finally been able to look into this a bit more. So yeah, if you create this with a launch template that provides us with the instance type and image id, it is possible to launch through the API (even though on the frontend, we still require the instance type value. So the 'next' button is disabled when not providing it. But yeah, if we are about to do this, we can work on this later. The response, though, returns these values empty. Is that desired behaviour?

I can add omit-empty but I am not sure what is the preferred behavior.
 @ezr-ondrej Do you want to express you thoughts about the empty values?

internal/clients/http/ec2/ec2_client.go
```diff
@@ -396,12 +396,18 @@ func (c *ec2Client) RunInstances(ctx context.Context, params *clients.AWSInstanc
...
 		MinCount:       ptr.To(amount),
-		InstanceType:   params.InstanceType,
-		ImageId:        ptr.To(params.AMI),
 		KeyName:        &params.KeyName,
 		UserData:       &encodedUserData,
 	}
 
+	if params.InstanceType != "" {
+		input.InstanceType = params.InstanceType
+	}
```
ezr-ondrej, line 405: I believe this one is a no-op.

lzap:
We already have a check for AWS in the service code:

```
	// Either Launch Template or Instance Type must be set. Both can be set too, in that case, instance type overrides the launch template.
	if payload.InstanceType == "" && payload.LaunchTemplateID == "" {
		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), "Both instance type and launch template are missing", ErrBothTypeAndTemplateMissing))
		return
	}
```

The issue is about GCP. I am going to close this one and take a look on GCP now, will raise a new PR thanks Adi!

PR #690, opened 2023-09-27
fix(HMS-2674): Error when name pattern is invalid

adiabramovitch:
Google API does not allow upper case name pattern, this PR fixes the issue and verifies the name pattern is lower cased, if not we error out.

### DIFF
--- a/internal/services/gcp_reservation_service.go
+++ b/internal/services/gcp_reservation_service.go
@@ -3,6 +3,8 @@ package services
 import (
 	"fmt"
 	"net/http"
+	"regexp"
+	"strings"
 
 	"github.com/RHEnVision/provisioning-backend/internal/preload"
 	"github.com/google/uuid"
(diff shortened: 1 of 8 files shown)


internal/services/gcp_reservation_service.go
```diff
@@ -152,3 +166,12 @@ func CreateGCPReservation(w http.ResponseWriter, r *http.Request) {
...
 }
+
+func isValidNamePattern(namePattern string) bool {
+	if strings.ToLower(namePattern) != namePattern {
+		return false
+	}
+	// Define a regular expression pattern to match valid RFC-1035-compatible names
+	pattern := regexp.MustCompile(`^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$`)
+	return pattern.MatchString(namePattern)
+}
```
lzap, line 177: You could have create a package-level unexported variable and compile the expression only once on package import (via `init`), because typically regexp compilation is slower operation than matching by far. But it does not a huge difference in this use case.

lzap (approved):
We could also check if there are no separate patterns like `xxx-####-####` that is also invalid according to Google, but this is good enough.

Nicely done, thanks.

lzap:
@akhil-jha Adi requested your review, feel free to merge. This is a cosmetic thing, looks good. Untested tho :-)

akhil-jha (approved):

PR #691, opened 2023-09-27
fix(HMS-1847): Adding configmap to job

mayurilahane:
For adding config maps into a job:
1. have already created config map yaml in app-interface
2. then referenced it to the namespace in app-interface itself.
**3. Now have added its mount point in this MR's job file**

### DIFF
--- a/deploy/cloudwash_job.yaml
+++ b/deploy/cloudwash_job.yaml
@@ -31,6 +31,14 @@ objects:
                   - "-d"
                   - "azure"
                   - "--all"
+                volumeMounts:
+                  - name: config-volume
+                    mountPath: /opt/app-root/src/cloudwash/settings.yaml
+                    subPath: settings.yaml
+            volumes:
+              - name: config-volume
+                configMap:
+                  name: cloudwash-config
                 env:
                   - name: CLEANUP_PROVIDERS_AZURE_AUTH_CLIENT_ID
                     valueFrom:


deploy/cloudwash_job.yaml
```diff
@@ -31,6 +31,17 @@ objects:
...
                   - "--all"
+                volumeMounts:
+                  - name: config-volume
+                    mountPath: /opt/app-root/src/cloudwash/settings.yaml
+                    subPath: settings.yaml
+                volumes:
+                  - name: config-volume
+                    configMap:
+                      name: cloudwash-config
+                envFrom:
```
mshriver, line 42: This block shouldn't be necessary, the expected configmap is getting subPath mounted

mayurilahane: I removed the config map block only 
did you mean the whole volume block?

mshriver: just the `envFrom` map - but the `volumes` map is out of place too, its not part of Container spec, only `volumeMounts`.   `volumes` is part of the Pod spec, that map starts at L20.

deploy/cloudwash_job.yaml
```diff
@@ -31,6 +31,14 @@ objects:
                   - "-d"
                   - "azure"
                   - "--all"
+                volumeMounts:
+                  - name: config-volume
+                    mountPath: /opt/app-root/src/cloudwash/settings.yaml
+                    subPath: settings.yaml
```
mshriver, line 37: suspect of this config with mountPath having the file, but I'm going to accept at this point to run in SRE cluster.

mshriver (approved):

PR #692, opened 2023-10-03
fix(HMS-2703): allow to configure TLS without CA cert

lzap:
Our app expects CA cert to be present to configure TLS, in case it is blank it will not configure TLS and will attempt to connect via plain text.

The recent stage migration changed the configuration, CA cert is no longer provided. All our pods currently fail with errors "TLS is required".

This patch changes the app behavior, it is not possible to enable TLS separately via flag (enabled by default) and also skip TLS server cert validation (false at the moment, can be useful for dev setup). When CA is blank, it will pass `nil` to the Kafka client library which has documented behavior to "use host OS cert store" which is what we want.

When we merge this, it should fix the problem on stage as long as Amazon Managed Kafka CA is in the host OS cert store. If not, the workaround will be to disable cert validation and ask the devprod team to pass Amazon CA via Clowder or we could do the same in `clowdapp.yaml`.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -103,6 +103,8 @@ objects:
                 value: ${APP_CACHE_TYPE}
               - name: WORKER_QUEUE
                 value: ${WORKER_QUEUE}
+              - name: KAFKA_TLS_ENABLED
+                value: ${KAFKA_TLS_ENABLED}
             resources:
               limits:
                 cpu: ${{CPU_LIMIT}}
@@ -190,6 +192,8 @@ objects:
                 value: ${APP_INSTANCE_PREFIX}
               - name: APP_CACHE_TYPE
                 value: ${APP_CACHE_TYPE}
+              - name: KAFKA_TLS_ENABLED
+                value: ${KAFKA_TLS_ENABLED}
             resources:
               limits:
                 cpu: ${{CPU_LIMIT}}
@@ -239,6 +243,8 @@ objects:
                 value: ${APP_INSTANCE_PREFIX}
               - name: APP_CACHE_TYPE
                 value: ${APP_CACHE_TYPE}
+              - name: KAFKA_TLS_ENABLED
+                value: ${KAFKA_TLS_ENABLED}
             resources:
               limits:
                 cpu: ${{CPU_LIMIT}}
@@ -365,6 +371,8 @@ objects:
                 value: ${APP_CACHE_TYPE}
               - name: WORKER_QUEUE
                 value: ${WORKER_QUEUE}
+              - name: KAFKA_TLS_ENABLED
+                value: ${KAFKA_TLS_ENABLED}
             resources:
               limits:
                 cpu: ${{CPU_LIMIT}}
@@ -572,3 +580,6 @@ parameters:
   - description: Notification service enabled
     name: APP_NOTIFICATIONS_ENABLED
     value: "true"
+  - description: Kafka TLS connection
+    name: KAFKA_TLS_ENABLED
+    value: "false"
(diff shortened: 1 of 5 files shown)


internal/kafka/kafka.go
```diff
@@ -39,7 +39,7 @@ var (
 
 func createSASLMechanism(saslMechanismName string, username string, password string) (sasl.Mechanism, error) {
 	switch strings.ToLower(saslMechanismName) {
-	case "plain":
+	case "plain", "none":
```
lzap, line 42: Unrelated, noticed we use "none" in some code so added it. Not too much useful, blank value will do as well.

ezr-ondrej (approved):
Sounds reasonable, can you refresh the example conf pls? :pray:

lzap:
Done.

ezr-ondrej:
/retest

But failures look  related: https://ci.int.devshift.net/job/RHEnVision-provisioning-backend-pr-check/1861/console

akhil-jha:
Same failures

lzap:
Seems like it does not work, I will try to ignore server cert with a flag to see if that helps.

lzap:
Amended a change where I set the default value of `TlsSkipValidate` to `true` temporarily to see if this actually helps.

lzap:
It looks TLS is not available in ephemeral that is why tests fail, added condition for ephemeral.

internal/kafka/kafka.go
```diff
@@ -76,30 +76,34 @@ func InitializeKafkaBroker(ctx context.Context) error {
...
 
 		pemCerts := config.Kafka.CACert
-		pool := x509.NewCertPool()
+		pool = x509.NewCertPool()
 		if ok := pool.AppendCertsFromPEM([]byte(pemCerts)); !ok {
 			logger.Warn().Msg("Could not add an CA cert to the pool")
 		}
+	}
+
+	if config.Kafka.TlsEnabled && !config.InEphemeralClowder() {
```
ezr-ondrej, line 96: Isn't there a way to figure out wether we have TLS from clowder config?

I guess we can do some follow-ups, if this passes, lets merge as-is, but it looks bit suspicius to me :)

lzap: If this works, we can remove this line and properly disable TLS in app-interface for ephemeral. But I just want to know if this actually works or not.

lzap:
And it works, I suggest to try the temporary "ephmeral" hack and if it fixes the stage I will prepare app-interface MR and clowder PR.

lzap:
You know what, let me introduce clowder variables in this PR, the TLS setting will default to "false" anyways so it should pass tests. App interface MR incoming.

deploy/clowdapp.yaml
```diff
@@ -572,3 +580,6 @@ parameters:
   - description: Notification service enabled
     name: APP_NOTIFICATIONS_ENABLED
     value: "true"
+  - description: Kafka TLS connection
+    name: KAFKA_TLS_ENABLED
+    value: "false"
```
lzap, line 585: Review carefully... :-)

lzap: I tend to write app options to be developer friendly. Many of other flags are sort of "run the app in standalone mode".

lzap: Also note this is ephemeral friendly, for prod/stage I filed an MR to enable these explicitly.

ezr-ondrej: oki, it's just bit more prone for mistakes in prod, but it's fine :)
Worst case scenario, we would not be allowed to connect to kafka without TLS :shrug: :+1:

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #693, opened 2023-10-03
chore: Show image_ids in floorist

avitova:
We should be able to track image_ids in the floorist now.
Though: I could not find a gcp floorist data, is that an intention?


ezr-ondrej (approved):
LGTM! :orange_heart:

PR #694, opened 2023-10-03
feat: Allow Azure resource group customization

ezr-ondrej:
This makes the Azure resource group work as expected. The static image is expected in the same group.
Composer image is still looked up.

Fixes HMS-1772

You  won't believe it, but I've even  tested it properly and it works! xD


lzap (approved):
Great work, thanks.

PR #696, opened 2023-10-04
chore: improve kafka startup logging

lzap:
It looks like we use TLS on stage but not on prod and ephemeral.

This is small logging improvement that should confirm it as we do not dump configuration on prod.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -155,17 +155,16 @@ var config struct {
 		Dsn string `env:"DSN" env-default:"" env-description:"data source name (empty value disables Sentry)"`
 	} `env-prefix:"SENTRY_"`
 	Kafka struct {
-		Enabled       bool     `env:"ENABLED" env-default:"false" env-description:"kafka service enabled"`
-		TlsEnabled    bool     `env:"TLS_ENABLED" env-default:"false" env-description:"enable TLS or use plaintext when false"`
-		TlsSkipVerify bool     `env:"TLS_SKIP_VERIFY" env-default:"false" env-description:"do not verify TLS server certificate"`
-		Brokers       []string `env:"BROKERS" env-default:"localhost:9092" env-description:"kafka hostname:port list of brokers"`
-		AuthType      string   `env:"AUTH_TYPE" env-default:"" env-description:"kafka authentication type (mtls, sasl or empty)"`
-		CACert        string   `env:"CA_CERT" env-default:"" env-description:"kafka TLS CA certificate path (use the OS cert store when blank)"`
-		SASL          struct {
-			Username         string `env:"USERNAME" env-default:"" env-description:"kafka SASL username"`
-			Password         string `env:"PASSWORD" env-default:"" env-description:"kafka SASL password"`
-			SaslMechanism    string `env:"MECHANISM" env-default:"" env-description:"kafka SASL mechanism (scram-sha-512, scram-sha-256 or plain)"`
-			SecurityProtocol string `env:"PROTOCOL" env-default:"" env-description:"kafka SASL security protocol"`
+		Enabled          bool     `env:"ENABLED" env-default:"false" env-description:"kafka service enabled"`
+		AuthType         string   `env:"AUTH_TYPE" env-default:"" env-description:"kafka authentication type (MTLS, SASL or empty)"`
+		SecurityProtocol string   `env:"PROTOCOL" env-default:"" env-description:"kafka SASL security protocol (PLAINTEXT, SSL, SASL_PLAINTEXT, or SASL_SSL, empty means PLAINTEXT)"`
+		TlsSkipVerify    bool     `env:"TLS_SKIP_VERIFY" env-default:"false" env-description:"do not verify TLS server certificate"`
+		Brokers          []string `env:"BROKERS" env-default:"localhost:9092" env-description:"kafka hostname:port list of brokers"`
+		CACert           string   `env:"CA_CERT" env-default:"" env-description:"kafka TLS CA certificate path (use the OS cert store when blank)"`
+		SASL             struct {
+			Username      string `env:"USERNAME" env-default:"" env-description:"kafka SASL username"`
+			Password      string `env:"PASSWORD" env-default:"" env-description:"kafka SASL password"`
+			SaslMechanism string `env:"MECHANISM" env-default:"" env-description:"kafka SASL mechanism (scram-sha-512, scram-sha-256 or plain)"`
 		} `env-prefix:"SASL_"`
 	} `env-prefix:"KAFKA_"`
 }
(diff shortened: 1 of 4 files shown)


internal/kafka/kafka.go
```diff
@@ -98,7 +106,6 @@ func NewKafkaBroker(ctx context.Context) (Broker, error) {
 
 		//nolint:gosec
 		tlsConfig = &tls.Config{
-			MinVersion:         tls.VersionTLS12,
```
lzap, line 101: Unrelated, TLS 1.2 is the default value when this field is `nil` so useless.

internal/kafka/kafka.go
```diff
@@ -81,7 +81,15 @@ func NewKafkaBroker(ctx context.Context) (Broker, error) {
 	var saslMechanism sasl.Mechanism
 
 	logger := zerolog.Ctx(ctx)
-	logger.Debug().Msgf("Setting up Kafka transport: %v", config.Kafka.Brokers)
+	logger.Info().Msgf("Setting up Kafka transport: %v TLS:%b Skip:%b CA:%b Auth:%s SASLMech:%s SASLProto:%s",
```
ezr-ondrej, line 84: I guess %b should have been %t?
```suggestion
	logger.Info().Msgf("Setting up Kafka transport: %v TLS:%t Skip:%t CA:%t Auth:%s SASLMech:%s SASLProto:%s",
```

internal/kafka/kafka.go
```diff
@@ -81,7 +81,15 @@ func NewKafkaBroker(ctx context.Context) (Broker, error) {
...
 	logger := zerolog.Ctx(ctx)
-	logger.Debug().Msgf("Setting up Kafka transport: %v", config.Kafka.Brokers)
+	logger.Info().Msgf("Setting up Kafka transport: %v TLS:%t Skip:%t CA:%t Auth:%s SASLMech:%s SASLProto:%s",
+		config.Kafka.Brokers,
+		config.Kafka.TlsEnabled,
+		config.Kafka.TlsSkipVerify,
+		config.Kafka.CACert != "",
+		config.Kafka.AuthType,
+		config.Kafka.SASL.SaslMechanism,
+		config.Kafka.SASL.SecurityProtocol,
```
ezr-ondrej, line 91: aren't we more interested in what's in the clowder config directly?

ezr-ondrej (approved):
sure, let's get it in :)

lzap:
Do not merge, I have some new info about clowder.

lzap:
After chat with Derek Horton, he suggest to provide an empty TLS config value as the library is supposed to automatically detect if TLS is present on a cluster or not and use the safest transport. He mentioned that he uses this on their project and it works even after Kafka migration from this weekend.

Therefore this patch drops the newly introduced setting completely, I don’t have a TLS Kafka setup in my home lab so we need to test this. Ephemeral is plaintext and stage is TLS so if both works prod should be fine. Then I will remove these flags.

lzap:
**DO NOT MERGE.**

Looks like Derek was not right, might not work as he described.

lzap:
I am reverting the commit back to the initial one - only improves logging and drops a line that has no effect. Looks like our solution is the best for now. I asked via email if we can use one of clowder's fields to detect TLS, let’s wait until I get some reply.

lzap:
_Regarding how apps should know whether SSL is on or off moving forward... I think a good example we have is the one Dan Kuc just pushed for malware-detection[1]. You are correct that the only values Clowder sets on this field is either "SSL"[2] or "SASL_SSL"[3,4]. In short, I think you can use this logic to determine if your Kafka client should use SSL or not:

* if kafka.brokers[0].securityProtocol is defined and its value is "SSL" or "SASL_SSL", then SSL should be ON

Your client should be able to use SSL with or without the 'cacert' field present. If the field is present, then use that as your client's trust store. If not, then ideally your client falls back to using the "system default trust store" -- which should already trust well-known root CA's like Digicert, Verisign, LetsEncrypt, and so on._

So I will go ahead and drop our own field and rely on this field instead.

lzap:
So the latest version drops the KAFKA_TLS_ENABLED flag and uses SecurityProtocol flag to check if it contains "SSL" string. If it does, it configures TLS.

The same way it also checks for SASL in the AuthType which should contain this string if SASL should be used.

I am keeping the verification skip flag tho, can be useful for local TLS testing. Once this is confirmed to work on both stage and ephemeral we can remove flags from app-interface. I believe this is the final version.

internal/config/config.go
```diff
@@ -155,17 +155,16 @@ var config struct {
...
-		TlsSkipVerify bool     `env:"TLS_SKIP_VERIFY" env-default:"false" env-description:"do not verify TLS server certificate"`
-		Brokers       []string `env:"BROKERS" env-default:"localhost:9092" env-description:"kafka hostname:port list of brokers"`
-		AuthType      string   `env:"AUTH_TYPE" env-default:"" env-description:"kafka authentication type (mtls, sasl or empty)"`
-		CACert        string   `env:"CA_CERT" env-default:"" env-description:"kafka TLS CA certificate path (use the OS cert store when blank)"`
-		SASL          struct {
-			Username         string `env:"USERNAME" env-default:"" env-description:"kafka SASL username"`
-			Password         string `env:"PASSWORD" env-default:"" env-description:"kafka SASL password"`
-			SaslMechanism    string `env:"MECHANISM" env-default:"" env-description:"kafka SASL mechanism (scram-sha-512, scram-sha-256 or plain)"`
-			SecurityProtocol string `env:"PROTOCOL" env-default:"" env-description:"kafka SASL security protocol"`
+		Enabled          bool     `env:"ENABLED" env-default:"false" env-description:"kafka service enabled"`
```
ezr-ondrej, line 158: should we delete it right away, given we are  now quite sure it's probably not gonna be used? :)

internal/kafka/kafka.go
```diff
@@ -93,18 +100,17 @@ func NewKafkaBroker(ctx context.Context) (Broker, error) {
 		}
 	}
 
-	if config.Kafka.TlsEnabled && !config.InEphemeralClowder() {
+	if strings.Contains(strings.ToUpper(config.Kafka.SecurityProtocol), "SSL") {
```
ezr-ondrej, line 103: I'd love to have the values in constants, but I guess we can contribute that to https://github.com/RedHatInsights/app-common-go and then use it.
So good for now :+1:

ezr-ondrej (commented):
Just two nits and only one of them addressable xD
Apart of it great effort here to find out how to do it properly! :+1:

internal/config/config.go
lzap: Delete what, enabled flag? How are you going to test Kafka in dev environment (without clowder)?

internal/kafka/kafka.go
lzap: Ehm "middleware" ehm.

lzap:
So we got the following info for the field: _PLAINTEXT, SSL, SASL_PLAINTEXT, or SASL_SSL_ I can incorporate this in the help text if necessary.

internal/kafka/kafka.go
ezr-ondrej: right xD :)

internal/config/config.go
lzap: Still not sure what you mean here, rebasing anyeway.

lzap:
Rebased the docs-only change.

lzap:
Fails on `unexpected backend data` that is the error we talked about on scrum. Something is deleting source prior this request. @akhil-jha any ideas?

ezr-ondrej:
Thanks! :orange_heart:

PR #697, opened 2023-10-06
feat(HMS-2438): AWS private IPv4/IPv6

lzap:
Adds private IPv4 and IPv6 into the instance detail for AWS.

There is slight inconsistency how we name things, in the API type we name the field "PublicIPv4" while the JSON/YAML is in fact "ipv4". The same applies for DNS. Therefore I am renaming the Go variable to "IPv4" and introducing new "PublicIPv4". Internally, we store the data in JSONB as "PrivateIPv4" and "PublicIPv4", I do not want to change that at this point and introduce new version of the document. 

Luckily, in the API we use the model type to send the results, so everything looks great:

* private_ipv4
* private_ipv6
* public_ipv4
* public_dns

I also added an example with real values and tested this in and out.


lzap:
I believe the error is relevant, the patch added two new fields so IQE stubs need to be regenerated I think. @akhil-jha ?

_iqe_provisioning.ApplicationProvisioningNotFoundException: Cannot find reservation within context to resolve getattr for instances_

api/openapi.gen.yaml
```diff
@@ -930,7 +948,7 @@ info:
     description: Provisioning service API
     license:
         name: GPL-3.0
-    version: 1.7.0
+    version: 1.8.0
 paths:
     /availability_status/sources:
         post:
```
lzap, line 954: As you can see, no changes only new fields (the change hunk is just a doc example).

internal/jobs/launch_instance_aws.go
```diff
@@ -308,20 +308,27 @@ func FetchInstancesDescriptionAWS(ctx context.Context, args *LaunchInstanceAWSTa
...
+			)
 			errRetry := rDao.UpdateReservationInstance(ctx, args.ReservationID, instance)
 			if errRetry != nil {
 				span.SetStatus(codes.Error, "cannot update instance description")
 				return fmt.Errorf("cannot update instance description: %w", errRetry)
 			}
 		}
 		return nil
-	}, 1000, 500, 500, 1000, 2000)
+	}, 500, 500, 1000, 1000, 2000)
```
lzap, line 331: I tested numbers as short as 10ms and AWS is blazing fast, it looks like it has the IP addresses prepared way before we even attempt to get. So shortening from 1 second initial lookup to just 1/2 second. Also added some logging, it was useful for debugging.

internal/models/reservation_model.go
```diff
@@ -171,8 +171,10 @@ type AzureReservation struct {
 }
 
 type ReservationInstanceDetail struct {
-	PublicDNS  string `json:"public_dns"`
-	PublicIPv4 string `json:"public_ipv4"`
+	PublicDNS   string `json:"public_dns,omitempty"`
+	PublicIPv4  string `json:"public_ipv4,omitempty"`
+	PrivateIPv4 string `json:"private_ipv4,omitempty"`
+	PrivateIPv6 string `json:"private_ipv6,omitempty"`
 }
```
lzap, line 178: This struct is used to store data in the JSONB column, added `omitempty` there is no reason to waste SQL resources with empty fields.

scripts/rest_examples/http-client.env.json
```diff
@@ -6,11 +6,13 @@
...
+    "source_id_azure": "2",
+    "source_id_gcp": "3",
     "pubkey_id": "1",
     "region": "us-east-1",
     "launch_template_id": "",
     "reservation-get-id": "1",
-    "azure-source-id": "2",
     "azure-resource-group": "redhat-deployed"
   }
 }
```
lzap, line 18: Slight update for consistency so people can easily change their stage source IDs for testing.

api/openapi.gen.yaml
ezr-ondrej: I guess it still needs update of the API client in IQE :)

api/openapi.gen.json
```diff
@@ -28,8 +28,10 @@
           "instances": [
             {
               "detail": {
-                "publicdns": "",
-                "publicipv4": "10.0.0.88"
+                "privateipv4": "172.31.36.10",
+                "privateipv6": "",
```
ezr-ondrej, line 32: can we add this to example as well, or does this default to blank?

ezr-ondrej (approved):
LGTM, but I guess we need to update the IQE client for the tests to pass.

internal/models/reservation_model.go
ezr-ondrej: It might be the reason why the IQE fails actually .... that now we  do not return instances at all? :thinking:

ezr-ondrej: Testing out without it: https://github.com/RHEnVision/provisioning-backend/pull/698

lzap: Right, rebased.

ezr-ondrej:
Yeah, it  was it... now it still fails, but specifically because of the privateIPv4 field

lzap:
Yeah, solved nothing. Shall I add omits back? I feel like that is slightly better.

ezr-ondrej:
> Yeah, solved nothing. Shall I add omits back? I feel like that is slightly better.

It did solve it, now there is a different error, which I think I know how to fix :)

ezr-ondrej:
Maybe it did not...  xD let me test locally with this PR if I figure out how that is

akhil-jha (approved):
All green!

api/openapi.gen.json
lzap: You mean IPv6, I can add one, sure. It is unconfigured on new accounts of course :-D

lzap:
Added `omitempty` back, added an IPv6 example from the documentation range.

internal/jobs/launch_instance_aws.go
```diff
@@ -308,20 +308,27 @@ func FetchInstancesDescriptionAWS(ctx context.Context, args *LaunchInstanceAWSTa
...
+		logger.Trace().Msgf("AWS returned %d instance details", len(instancesDescriptionList))
 
 		if len(instancesDescriptionList) == 0 {
 			return ErrTryAgain
 		}
 
-		for _, instance := range instancesDescriptionList {
+		for i, instance := range instancesDescriptionList {
+			logger.Trace().Msgf("Instance id %s private IPv4:%s public IPv4:%s DNS:%s",
+				instancesIDList[i],
```
avitova, line 319: !. Are we sure that the instancesIDList would have the same order as instancesDescriptionList? :thinking: 
instancesDescriptionList comes from a call of ec2Client.DescribeInstanceDetails. It parses a response from ec2 client, right?
For example, from the description of the DescribeInstances function:
> If you specify an instance that you do not own, it is not included in the output.
If we have had instancesIDList with such ID, this would probably break the order.

2. We are iterating instancesIDList within a loop to iterate through instancesDescriptionList. In theory, based on the upper quote, it should be true that
`len(instancesIDList) >  len(instancesDescriptionList)`
But I guess we are playing with fire here.

avitova (commented):
Very cool and quick work! Looks good, but I do have one nitpick.

internal/jobs/launch_instance_aws.go
avitova: Suggestion: Let's use instance.ID here, if that is possible.

avitova (approved):
Thank you. LGTM:)

lzap:
All green, merging thanks all.

PR #700, opened 2023-10-09
chore: requery account for transaction isolation

lzap:
So I came up with this nice trick on how to quickly create accounts without explicit transactions. Works great, except it does not :-) When there are actually two requests with the same account at the same time, the query is executed in an implicit transactions which cannot be turned off and I totally forgot about it.

What happens is that one request will fail with `pgx error: scanning one: no rows in result set` (https://glitchtip.devshift.net/insights/issues/973561) because although the conflict will trigger, the default isolation level in postgres will not allow to see the inserted record.

The solution is to execute the same query one more time, this only happens rarely, I saw 20 events in production in one week. But it is worth fixing this.

This is hard to test or reproduce, so review carefully. Maybe with some arbitrary sleep via `pg_sleep()` function but I find this difficult to figure out. I think the code is good (TM).

### DIFF
--- a/internal/dao/pgx/account_pgx.go
+++ b/internal/dao/pgx/account_pgx.go
@@ -9,6 +9,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/db"
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 	"github.com/georgysavva/scany/v2/pgxscan"
+	"github.com/rs/zerolog"
 )
 
 func init() {
@@ -74,6 +75,15 @@ func (x *accountDao) GetOrCreateByIdentity(ctx context.Context, orgId string, ac
 	err := pgxscan.Get(ctx, db.Pool, result, query, orgId, account)
 	if err != nil {
 		return nil, fmt.Errorf("pgx error: %w", err)
+	}
+
+	// Step 4: requery in case transaction isolation (simultaneous transactions)
+	if result.ID == 0 && result.OrgID == "" {
+		zerolog.Ctx(ctx).Warn().Bool("requery", true).Msgf("Organization id %s requery", orgId)
+		err = pgxscan.Get(ctx, db.Pool, result, query, orgId, account)
+		if err != nil {
+			return nil, fmt.Errorf("pgx requery error: %w", err)
+		}
 	}
 
 	return result, nil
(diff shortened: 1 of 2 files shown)


internal/dao/pgx/account_pgx.go
```diff
@@ -76,6 +77,15 @@ func (x *accountDao) GetOrCreateByIdentity(ctx context.Context, orgId string, ac
 		return nil, fmt.Errorf("pgx error: %w", err)
 	}
 
+	// Step 4: requery in case transaction isolation (simultaneous transactions)
+	if result.ID == 0 && result.OrgID == "" {
```
ezr-ondrej, line 81: Query again makes sense (it will increase the chances of success :) ), but will we ever trigger this, given we return on error few lines before?

internal/dao/pgx/account_pgx.go
```diff
@@ -72,7 +74,17 @@ func (x *accountDao) GetOrCreateByIdentity(ctx context.Context, orgId string, ac
...
+	if errors.Is(err, dao.ErrNoRows) {
+		// Step 4: requery in case transaction isolation (simultaneous transactions)
+		if result.ID == 0 && result.OrgID == "" {
+			zerolog.Ctx(ctx).Warn().Bool("requery", true).Msgf("Organization id %s requery", orgId)
+			err := pgxscan.Get(ctx, db.Pool, result, query, orgId, account)
+			if err != nil {
+				return nil, fmt.Errorf("pgx requery error: %w", err)
+			}
+		}
+	} else if err != nil {
```
ezr-ondrej, line 87: Linter doesn't like shadowing the var.
We can use the same var and use the same if for catching all the "leftover" errors in the main, right?

Smthing like:
```suggestion
			err = pgxscan.Get(ctx, db.Pool, result, query, orgId, account)
		}
	}
	if err != nil {
```


or just use `requeryErr` var would also work :)

ezr-ondrej (commented):
I'm good, but the linter ain't happy now.

internal/dao/pgx/account_pgx.go
lzap: Rebased

lzap:
All good.

ezr-ondrej (approved):
Thanks @lzap ! :+1:

PR #701, opened 2023-10-09
build(HMS-2743): Enable rbac caching

akhil-jha:
Relates to https://github.com/RHEnVision/provisioning-backend/pull/699


lzap (approved):
Interesting that is does not work, let's see.

Please share logs with me thanks.

PR #702, opened 2023-10-09
chore: add extra sources body logging

lzap:
I am trying to understand why sources randomly returns 404, there should be something in the body it looks like:

https://github.com/RedHatInsights/sources-api-go/blob/main/middleware/error_handling.go#L22C25-L22C25

This will print a warning log statement with full body as a field. @akhil-jha then let’s reproduce once again.

### DIFF
--- a/internal/clients/http/sources/sources_client.go
+++ b/internal/clients/http/sources/sources_client.go
@@ -77,6 +77,9 @@ func (c *sourcesClient) Ready(ctx context.Context) error {
 	if resp == nil {
 		return fmt.Errorf("ready call: empty response: %w", clients.ErrUnexpectedBackendResponse)
 	}
+	if resp.JSON400 != nil {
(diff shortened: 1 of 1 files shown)


akhil-jha (approved):

internal/clients/http/sources/sources_client.go
```diff
@@ -77,6 +77,9 @@ func (c *sourcesClient) Ready(ctx context.Context) error {
 	if resp == nil {
 		return fmt.Errorf("ready call: empty response: %w", clients.ErrUnexpectedBackendResponse)
 	}
+	if resp.JSON400 != nil {
+		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("Get authentication from sources returned 4xx")
```
ezr-ondrej, line 81: I've failed to actually send this comment for some reason:

Why are we logging it in separate field and not the message? :thinking:

PR #703, opened 2023-10-09
feat(HMS-2439): Add private IP address in Azure details

ezr-ondrej:
Fetch Azure private IP address as well as public one.

### DIFF
--- a/internal/clients/http/azure/create_vms.go
+++ b/internal/clients/http/azure/create_vms.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 	"github.com/google/uuid"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/codes"
@@ -37,7 +38,8 @@ func (c *client) CreateVMs(ctx context.Context, vmParams clients.AzureInstancePa
 			return nil, err
 		}
 
-		vmDescriptions[i].IPv4 = *publicIP.Properties.IPAddress
+		vmDescriptions[i].IPv4 = ptr.FromOrEmpty(publicIP.Properties.IPAddress)
+		vmDescriptions[i].PrivateIPv4 = ptr.FromOrEmpty(networkInterface.Properties.IPConfigurations[0].Properties.PrivateIPAddress)
 
 		resumeTokens[i], err = c.BeginCreateVM(ctx, networkInterface, vmParams, vmName)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


lzap (changes requested):
Something is not right, this should not pass since you haven’t modified OpenAPI. What is wrong? This should fail.

lzap:
Aaah I understand now, the PrivateIPv4 is a shared model, no need to change OpenAPI.

The only concern I have is the nil exceptions which can possibly occur. It is very unlikely that an internal IPv4 is not set, not in 2023, it is still another two decades until full IPv6 adoption.

lzap:
Can you add example similarly as I did in AWS please?

ezr-ondrej:
Added example and check for nil pointer :)

ezr-ondrej:
> Aaah I understand now, the PrivateIPv4 is a shared model, no need to change OpenAPI.

Yup, that's why I was so keen to implement this, the hard part was already done by you! xD

internal/clients/http/azure/create_vms.go
```diff
@@ -38,6 +38,11 @@ func (c *client) CreateVMs(ctx context.Context, vmParams clients.AzureInstancePa
 		}
 
 		vmDescriptions[i].IPv4 = *publicIP.Properties.IPAddress
+		// can be nil if IPv6 is default in the future
+		privateIPv4 := networkInterface.Properties.IPConfigurations[0].Properties.PrivateIPAddress
```
avitova, line 42: > 		// can be nil if IPv6 is default in the future

If this happends, would it end up with only PrivateIPAddress being nil? Everything else here are pointers/accessing IPConfigartions array that might be empty. Is this all safe?:D

lzap: We have a `ptr.FromOrEmpty` generic helper, can you use that instead?

ezr-ondrej: it is not safe for sure xD it's YOLO I'll try to improve at least the ones I'm touching.
But the structure should be pretty stable, Azure uses pointers everywhere, but it either fails, or the data are there, so it's "a bit safer then complete YOLO" xD

ezr-ondrej: Will use the `ptr.FromOrEmpty` :+1:

ezr-ondrej: Done, I'll try to fix-up other occurences when I touch other Azure code :)

lzap (approved):
Thanks, merge at will @avitova

internal/clients/http/azure/create_vms.go
avitova: Yeah, I do believe that for Azure this would not be inconsistent.

avitova:
Thanks! LGTM:)

PR #704, opened 2023-10-09
feat(HMS-2440): private IPv4 address for GCP

ezr-ondrej:


scripts/rest_examples/reservation-create-gcp.http
```diff
@@ -10,7 +10,6 @@ X-Rh-Identity: {{identity}}
   "image_id": "https://www.googleapis.com/compute/v1/projects/rhel-cloud/global/images/rhel-9-v20230411",
   "amount": 1,
   "machine_type": "e2-micro",
-  "launch_template_id": "{{launch_template_id}}",
```
lzap, line 13: Intentional? Just checking, might be useful.

internal/clients/http/gcp/gcp_client.go
```diff
@@ -316,6 +316,7 @@ func (c *gcpClient) GetInstanceDescriptionByID(ctx context.Context, id, zone str
 	instanceId := strconv.FormatUint(instance.GetId(), 10)
 	instanceDesc := clients.InstanceDescription{ID: instanceId}
 	for _, n := range instance.NetworkInterfaces {
+		instanceDesc.PrivateIPv4 = *n.NetworkIP
```
lzap, line 319: Similarly, I would perhaps check for nil unless there is a documentation remark somewhere that this is never unallocated pointer.

lzap (approved):
Other than that, LGTM.

lzap:
Can you add example similarly as I did in AWS please?

scripts/rest_examples/reservation-create-gcp.http
ezr-ondrej: Just removed for testing, commited by mistake, thanks for catching that! :orange_heart:

internal/clients/http/gcp/gcp_client.go
```diff
@@ -316,6 +316,9 @@ func (c *gcpClient) GetInstanceDescriptionByID(ctx context.Context, id, zone str
 	instanceId := strconv.FormatUint(instance.GetId(), 10)
 	instanceDesc := clients.InstanceDescription{ID: instanceId}
 	for _, n := range instance.NetworkInterfaces {
+		if n.NetworkIP != nil {
```
avitova, line 319: I suggest to even check for n != nil, but I do not know what documentation says.
```suggestion
		if n != nil && n.NetworkIP != nil {
```

lzap: It is rare to put `nil` into slice/map I would not worry too much about thi.

We have a `ptr.FromOrEmpty` generic helper, can you use that instead?

ezr-ondrej: I've used the generic helper, but did not fix the range, I guess it'd be very weird to have nil in an array.

lzap (approved):
I agree, @avitova?

internal/clients/http/gcp/gcp_client.go
avitova: Yeah, this was just a suggestion, we do not have to check necessarily:)

avitova (approved):
LGTM;)

PR #706, opened 2023-10-11
fix(HMS-2757): improve unexpected row error

lzap:
An error striked me this morning: "failed to save AWS pubkey name to DB: expected 1 row: unexpected affected rows" which means detail reservation could not be updated after AWS key was updated. It looks like we fetch AWS reservation detail from the database but we do not load `reservation_id` field which is okay unless we need to update such record. This field is used in the WHERE query to locate the correct record.

I am adding few more log statements to see little bit more and adding the missing field.

During this, I noticed that our reservation - detail association is set up in a way that one reservation can have multiple details for different providers. But we have never done that, our current implementation is 1:1, we might drop this in the future. But it is okay for now.

### DIFF
--- a/internal/dao/pgx/pubkey_pgx.go
+++ b/internal/dao/pgx/pubkey_pgx.go
@@ -88,7 +88,7 @@ func (x *pubkeyDao) Update(ctx context.Context, pubkey *models.Pubkey) error {
 		return fmt.Errorf("pgx error: %w", err)
 	}
 	if tag.RowsAffected() != 1 {
-		return fmt.Errorf("expected 1 row: %w", dao.ErrAffectedMismatch)
+		return fmt.Errorf("expected 1 row, got %d: %w", tag.RowsAffected(), dao.ErrAffectedMismatch)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


ezr-ondrej:
are you sure, as I went looking for usage and we do not load reservation_id anywhere to the model, we use `aws_reservation_id` which I'm not sure exists, but won't load `reservation_id` field data I'm guessing:
https://github.com/lzap/provisioning-backend/blob/3b30277346e6dfe862a428a62c3afd62cefed026/internal/models/reservation_model.go#L85


and on usage, we fetch the `ID` of `Reservation` to use in the where condition: https://github.com/lzap/provisioning-backend/blob/3b30277346e6dfe862a428a62c3afd62cefed026/internal/jobs/launch_instance_aws.go#L142

lzap:
You are right that was an invalid conclusion, I renamed this only to improve logging which might come handy.

I believe the root cause is this: https://github.com/RHEnVision/provisioning-backend/pull/707

But we need to confirm, I am not sure.

ezr-ondrej (approved):
Thanks @lzap ! :)

PR #707, opened 2023-10-11
fix(HMS-2757): use transactions for reservations

lzap:
During investigation on the linked bug, I noticed this. I am not sure if this is root cause of the problems we currently see on stage, but chances are this is actually caused by the performance testing and throttle limiting.

This needs to be fixed no matter of what.

### DIFF
--- a/internal/dao/pgx/reservation_pgx.go
+++ b/internal/dao/pgx/reservation_pgx.go
@@ -27,9 +27,17 @@ func getReservationDao(ctx context.Context) dao.ReservationDao {
 }
 
 func (x *reservationDao) CreateNoop(ctx context.Context, reservation *models.NoopReservation) error {
-	reservation.Provider = models.ProviderTypeNoop
-	if err := x.createGenericReservation(ctx, &reservation.Reservation); err != nil {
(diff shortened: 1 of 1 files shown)


internal/dao/pgx/reservation_pgx.go
```diff
@@ -27,23 +27,31 @@ func getReservationDao(ctx context.Context) dao.ReservationDao {
...
+			return fmt.Errorf("failed to create reservation record: %w", err)
+		}
+
+		return nil
+	})
+
+	if txErr != nil {
+		return fmt.Errorf("pgx tx error: %w", txErr)
 	}
 	return nil
```
lzap, line 42: To make things simple and because pgx.Tx and pgx.Pool are incompatible types, I am putting this in tx block although it does not need to be. I think I could have used pgx.Querier or some interface, not sure.

lzap: Nah Querier is not from the pgx package, I would rather not use it. These are two different types.

ezr-ondrej (approved):
Thanks! :orange_heart:

lzap:
And it works...


image

PR #708, opened 2023-10-11
ci: Image stub script from gitlab

ezr-ondrej:


ezr-ondrej:
All right, we are a GO

PR #710, opened 2023-10-11
chore: correlation for worker logger

lzap:
During investigation of a problem, I noticed that worker is missing jobID from all pgx log messages therefore you will not see some records which are useful. This fixes it and unifies how context (and logger) is created across jobs. It also adds trace/edge IDs so logs can be correlated across api/worker to see the whole workflow of a reservation.

I cannot figure out why tests are failing, it is too late my brain does not work anymore. Will take a look tomorrow.

### DIFF
--- a/internal/db/connection.go
+++ b/internal/db/connection.go
@@ -65,7 +65,7 @@ func Initialize(ctx context.Context, schema string) error {
 	poolConfig.MaxConnIdleTime = config.Database.MaxIdleTime
 
 	if config.Telemetry.Enabled {
-		poolConfig.ConnConfig.Tracer = otelpgx.NewTracer()
+		poolConfig.ConnConfig.Tracer = otelpgx.NewTracer(otelpgx.WithIncludeQueryParameters())
 	} else {
 		logLevel, configErr := tracelog.LogLevelFromString(config.Database.LogLevel)
 		if configErr != nil {
@@ -74,16 +74,31 @@ func Initialize(ctx context.Context, schema string) error {
 
 		if logLevel > 0 {
 			zeroLogger := pgxlog.NewLogger(log.Logger,
-				pgxlog.WithContextFunc(func(ctx context.Context, logWith zerolog.Context) zerolog.Context {
+				pgxlog.WithContextFunc(func(ctx context.Context, zx zerolog.Context) zerolog.Context {
+					jobId := logging.JobId(ctx)
+					if jobId != "" {
+						zx = zx.Str("job_id", jobId)
+					}
+					reservationId := logging.ReservationId(ctx)
+					if reservationId != 0 {
+						zx = zx.Int64("reservation_id", reservationId)
+					}
 					traceId := logging.TraceId(ctx)
 					if traceId != "" {
-						logWith = logWith.Str("trace_id", traceId)
+						zx = zx.Str("trace_id", traceId)
+					}
+					requestId := logging.EdgeRequestId(ctx)
+					if requestId != "" {
+						zx = zx.Str("request_id", requestId)
 					}
 					accountId := identity.AccountIdOrNil(ctx)
 					if accountId != 0 {
-						logWith = logWith.Int64("account_id", accountId)
+						zx = zx.Int64("account_id", accountId)
 					}
-					return logWith
+					principal := identity.Identity(ctx)
+					zx = zx.Str("org_id", principal.Identity.OrgID)
+					zx = zx.Str("account_number", principal.Identity.AccountNumber)
+					return zx
 				}))
 			poolConfig.ConnConfig.Tracer = &tracelog.TraceLog{
 				Logger:   zeroLogger,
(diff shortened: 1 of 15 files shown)


lzap:
This is weird, unit tests are failing locally for me. Not on CI tho :-)

ezr-ondrej:
/retest
weird error ... public_ip missing, I'm guessing omitempty is not delt with properly on OpenAPI level :thinking:

ezr-ondrej:
Huh, I've started reviewing and then decided to actually learn about tracing bit more.
I've ended up with following: #711 unfortunatelly the rename makes it hard to read :/

Let's discuss tomorrow? I guess I'd prefer a meeting, I'm too deep and in need of second brain xD

lzap:
Found a typo, rebased.

internal/db/connection.go
```diff
@@ -74,16 +74,31 @@ func Initialize(ctx context.Context, schema string) error {
...
 					if accountId != 0 {
-						logWith = logWith.Int64("account_id", accountId)
+						zx = zx.Int64("account_id", accountId)
 					}
-					return logWith
+					principal := identity.Identity(ctx)
+					zx = zx.Str("org_id", principal.Identity.OrgID)
+					zx = zx.Str("account_number", principal.Identity.AccountNumber)
+					return zx
 				}))
```
lzap, line 102: This hunk ensures that all pgx logging has a proper context and will appear in Kibana correctly.

internal/identity/identity.go
```diff
@@ -13,7 +13,11 @@ type Principal = identity.XRHID
 
 // Identity returns identity header struct or nil when not set.
 func Identity(ctx context.Context) Principal {
-	return identity.Get(ctx)
+	val := ctx.Value(identity.Key)
+	if val == nil {
+		return Principal{}
+	}
+	return val.(Principal)
```
lzap, line 20: The `Get` from middleware project does not handle `nil` correctly, so this is a copy.

lzap:
Ah test do fail locally on the `main` branch as well. Likely my local environment, need to take a look.

internal/db/connection.go
lzap: Just saw your PR and realized some of this code should live in worker.contextLogger instead. Moving.

internal/jobs/ctx.go
```diff
@@ -18,3 +21,25 @@ func copyContext(ctx context.Context) context.Context {
...
+		Str("org_id", principal.Identity.OrgID).
+		Str("account_number", principal.Identity.AccountNumber).
+		Str("trace_id", traceID).
+		Str("request_id", edgeID).
+		Int64("reservation_id", reservationID).
+		Str("job_id", jobID.String()).
+		Logger())
+	ctx = logger.WithContext(ctx)
+	return ctx, logger
+}
```
lzap, line 45: Just saw your PR and realized some of this code should live in worker.contextLogger instead. Moving.

lzap:
I just moved most of the logging/context handling into worker/ package where it belongs. Only reservation ID is in the jobs/ package now.

lzap:
One additional change: `otelpgx.WithIncludeQueryParameters()` will cause the otel version of the pgx tracer to include parameters in logs. We do not store any sensitive data, this will vastly help me to understand the issue I am solving.

pkg/worker/redis.go
```diff
@@ -96,7 +83,11 @@ func (w *RedisWorker) Enqueue(ctx context.Context, job *Job) error {
 		}
 	}
 
-	logger := loggerWithJob(ctx, job)
```
ezr-ondrej, line 99: Why did you removed the function? :thinking:

pkg/worker/redis.go
```diff
@@ -198,17 +193,16 @@ func (w *RedisWorker) fetchJob(ctx context.Context) {
...
 	}
-
 	defer atomic.AddInt64(&w.inFlight, -1)
-	logger := loggerWithJob(ctx, job)
 
-	logger.Info().Msg("Dequeued job from Redis")
-	ctx = contextLogger(ctx, job)
+	ctx, logger := contextLogger(origCtx, job)
+	defer recoverAndLog(ctx)
+	logger.Info().Msgf("Dequeued job %s %s from Redis", job.Type.String(), job.ID)
```
ezr-ondrej, line 204: We are now not logging the attributes on Dequeue, is that intentional?

ezr-ondrej (approved):
I believe this should work! :)
Left few nits, but we can address in follow-ups :+1:

ezr-ondrej:
> One additional change: `otelpgx.WithIncludeQueryParameters()` will cause the otel version of the pgx tracer to include parameters in logs. We do not store any sensitive data, this will vastly help me to understand the issue I am solving.

We do not yet, but we should keep it in mind, there might be some sensitive data in the future.

pkg/worker/redis.go
lzap: Yeah there is replacement now, only for two cases there is not yet enough information so I replaced it with plain code.

lzap:
Hmm unfortunately, I don’t see query parameters being logged at all. Other than that it works and with your patch the trace will be actually correct :-)

PR #711, opened 2023-10-11
fix(HMS-2785): pass otel spans properly 

ezr-ondrej:
Otel needs a root spans, if that is not passed from external service, we should start one ourselves.
Also jobs shall retrieve these spans and not start new one.

For Jobs:
Screenshot from 2023-10-13 12-50-48

And for Kafka messages:
Screenshot from 2023-10-13 17-24-42


internal/logging/ctx.go
```diff
@@ -42,14 +44,9 @@ func WithEdgeRequestId(ctx context.Context, id string) context.Context {
...
 		return ""
 	}
-	return value.(string)
-}
-
-// WithTraceId returns context copy with trace id value.
-func WithTraceId(ctx context.Context, id string) context.Context {
-	return context.WithValue(ctx, requestIdCtxKey, id)
+	return value.String()
 }
```
lzap, line 52: This makes sense, also delete the key constant.

internal/middleware/telemetry.go
```diff
@@ -0,0 +1,36 @@
...
+
+		ctx, span = otel.Tracer(TraceName).Start(ctx, r.URL.Path)
+
+		// Store TraceID in response headers for easier debugging
+		w.Header().Set("X-Trace-Id", span.SpanContext().TraceID().String())
+
+		next.ServeHTTP(w, r.WithContext(ctx))
+	}
+	return http.HandlerFunc(fn)
+}
```
lzap, line 36: Yes, that was a hack this is more appropriate. I did not have a chance to look at proper telemetry setup.

pkg/worker/job.go
```diff
@@ -77,9 +80,14 @@ func contextLogger(ctx context.Context, job *Job) context.Context {
...
+	var span trace.Span
+	ctx, span = otel.Tracer(telemetry.TracePrefix+"pkg/worker").Start(ctx, string(job.Type))
+
 	accountId := job.AccountID
 	id := job.Identity
 	logger := zerolog.Ctx(ctx).With().
+		Str("trace_id", span.SpanContext().TraceID().String()).
 		Str("job_id", job.ID.String()).
 		Int64("account_id", accountId).
 		Str("account_number", id.Identity.AccountNumber).
```
lzap, line 93: Yeah this makes a lot of sense. In my patch, I put trace_id at reservation level but I think this is much better place. I will move it here but I will not add any otel related stuff and leave this to you so you can further develop this.

Great idea, in the end, we will have one trace (aka transaction in GlitchTip) and spans for api pod then one big span for the worker pod and additional spans for individual functions.

lzap (commented):
Great idea, let’s merge my PR first where I want to modify the `contextLogger` function in worker package to pass trace_id into logs so you can continue on this.

internal/logging/ctx.go
ezr-ondrej: Done :)

internal/kafka/avail_status_msg.go
```diff
@@ -25,27 +23,3 @@ func NewAvailabilityStatusMessage(msg *GenericMessage) (*AvailabilityStatusMessa
...
-		// Keep headers written in lowercase to match sources comparison.
-		Headers: GenericHeaders(
-			"content-type", "application/json",
-			"x-rh-identity", identity.IdentityHeader(ctx),
-			"x-rh-sources-org-id", id.Identity.OrgID,
-			"x-rh-sources-account-number", id.Identity.AccountNumber,
-			"event_type", "availability_status",
-		),
-	}, nil
-}
```
lzap, line 51: I am assuming no changes were made here.

lzap (approved):
This is so much better than what I came up with. Great work.

internal/kafka/avail_status_msg.go
ezr-ondrej: There is the one additional header.

ezr-ondrej:
> Image problem again.

My fault this time, working on a fix

lzap:
Merged thanks.

PR #714, opened 2023-10-12
chore: log HTTP protocol in request start

lzap:
SSIA

Our 3scale API uses HTTP/2 for client communication but internally I believe it relies on HTTP 1.1. This should confirm this.


avitova (approved):
Thank you:)

PR #715, opened 2023-10-12
chore: add otel pgx parameters to log

lzap:
We do not include slices in our telemetry2zerolog implementation, the otelpgx library does use string slice for params. This was the reason why it was missing. This fixes it.


avitova (approved):
LGTM:)

lzap:
Thanks, I do have a small followup :-D

PR #716, opened 2023-10-12
chore: add more otel params

lzap:
Add few more important correlation params into otel logger, also store job id and type in context for later use. Finally, rename one context function to better reflect what it returns.

### DIFF
--- a/internal/identity/account.go
+++ b/internal/identity/account.go
@@ -22,8 +22,8 @@ func AccountId(ctx context.Context) int64 {
 	return value.(int64)
 }
 
-// AccountIdOrNil returns current account model or 0 when not set.
-func AccountIdOrNil(ctx context.Context) int64 {
+// AccountIdOrZero returns current account model or 0 when not set.
+func AccountIdOrZero(ctx context.Context) int64 {
 	value := ctx.Value(accountIdCtxKey)
 	if value == nil {
 		return 0
--- a/internal/telemetry/zerolog_exporter.go
+++ b/internal/telemetry/zerolog_exporter.go
@@ -3,6 +3,8 @@ package telemetry
 import (
 	"context"
 
+	"github.com/RHEnVision/provisioning-backend/internal/identity"
+	"github.com/RHEnVision/provisioning-backend/internal/logging"
 	"github.com/rs/zerolog"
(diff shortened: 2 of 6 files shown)


internal/telemetry/zerolog_exporter.go
```diff
@@ -26,11 +28,22 @@ func (e *loggerExporter) ExportSpans(ctx context.Context, spans []trace.ReadOnly
...
+			Str("account_number", identity.Identity(ctx).Identity.AccountNumber).
+			Str("job_id", logging.JobId(ctx)).
+			Str("job_type", logging.JobType(ctx)).
 			Dur("duration", duration)
 
+		accountId := identity.AccountIdOrZero(ctx)
+		if accountId != 0 {
+			t = t.Int64("account_id", accountId)
+		}
+
```
lzap, line 46: This is the gist of the patch - I need to have in our zerolog exporter (which is used on stage/prod) request id which I can use to correlate all the records. It was not present at all, only trace_id which is unfortunately different because we cannot use it between api and worker pods (it gets closed).

internal/identity/account.go
```diff
@@ -22,8 +22,8 @@ func AccountId(ctx context.Context) int64 {
 	return value.(int64)
 }
 
-// AccountIdOrNil returns current account model or 0 when not set.
-func AccountIdOrNil(ctx context.Context) int64 {
+// AccountIdOrZero returns current account model or 0 when not set.
+func AccountIdOrZero(ctx context.Context) int64 {
```
lzap, line 26: This had wrong name, it does not return `nil` but `0` in fact.

internal/telemetry/zerolog_exporter.go
ezr-ondrej: The job_id and job_type should be optional (set only when available) IMO, not to set them from all processes, having them blank may confuse us.

lzap: Yeah, rebased. All fields are now compared and only added when needed.

ezr-ondrej (approved):
Thanks! :+1:

ezr-ondrej:
Ah sorry! I meant to merge it, thanks Lukáš!

PR #717, opened 2023-10-13
chore: Add GCP provider to floorist

avitova:
I noticed we do not make statistics of gcp provider. :eyes: I understand that the floorplan is old, but I hope we can do this now.
Please, do not merge yet, I am thinking if I would add something else. But could you please check if this would not break the process?

note: We only track whether or not the users use any template, but we do not track what is in the template. We could add the template to the floorplan in the future, but we'd most likely have to export the template for each reservation detail. I remember @lzap wanted to track this, so I am just saying it is possible. But it seems it would complicate the data, so I would postpone it a bit.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -461,8 +461,7 @@ objects:
             from reservations r
             join aws_reservation_details d on r.id = d.reservation_id
             join accounts a on r.account_id = a.id
-            where provider = provider_type_aws()
-            order by r.created_at)
+            where provider = provider_type_aws())
 
             union all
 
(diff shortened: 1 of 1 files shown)


lzap:
Also I noticed we do union of three sorted results, not sure how useful is sorting this way. We can probably remove sorting alltogether.

avitova:
Yeah, true. There is probably no need to sort these. :+1: I also noticed just now, as I found it weird there is no data yet for this provider. But that is okay, we will see even the old ones, once this is merged and we get the data. :+1: You are right, let's merge now, and add more later. :)

avitova:
/retest

We could also sort it on top of all that if we find the need to do that. But I personally do not need that for the statistics, so I think we are good.

ezr-ondrej:
The tests are broken, working on a fix, but those do not test floorist anyway, so lets just merge :)

vkrizan:
Hey,
I think that these changes could be the reason why your floorist pipeline is failing:
```
psycopg2.errors.SyntaxError: each UNION query must have the same number of columns
LINE 5: (select 'gcp' as provider, r.id, r.creat...
^
```

ezr-ondrej:
Thanks a bunch @vkrizan ! :orange_heart:
We've already fixed it in https://github.com/RHEnVision/provisioning-backend/pull/728 but we did not want to release end of the week, so it only got to prod today, hopefully it should be fixed today's run :)

avitova:
Thank you, @vkrizan, you are right.
[This one](https://github.com/RHEnVision/provisioning-backend/pull/731) was the actual fix, gcp had less columns than the rest of the providers. :+1:

PR #718, opened 2023-10-13
chore: Sanitize job logging 

ezr-ondrej:
:tipping_hand_person:  depends on #711 and thus contains those changes as well

Cleans up job logs.
Removes job_args from most of the job logs except Enqueue and Dequeue.
Removes double logging of job id and type.

I was touching this code quite heavily in #711 so I wanted to follow-up with what I feel might improve the logging a bit.
:warning: These are my opinions and maybe some of the logs were there on purpose, happy to discuss it! :)

### DIFF
--- a/pkg/worker/job.go
+++ b/pkg/worker/job.go
@@ -105,8 +105,7 @@ func initJobContext(origCtx context.Context, job *Job) (context.Context, *zerolo
 		Str("account_number", job.Identity.Identity.AccountNumber).
 		Str("request_id", job.EdgeID).
 		Str("job_id", job.ID.String()).
-		Str("job_type", job.Type.String()).
-		Interface("job_args", job.Args)
+		Str("job_type", job.Type.String())
 
 	if config.Telemetry.Enabled {
 		ctx = otel.GetTextMapPropagator().Extract(ctx, job.TraceContext)
--- a/pkg/worker/redis.go
+++ b/pkg/worker/redis.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	"github.com/RHEnVision/provisioning-backend/internal/metrics"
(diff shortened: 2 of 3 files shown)


pkg/worker/job.go
```diff
@@ -105,8 +105,7 @@ func initJobContext(origCtx context.Context, job *Job) (context.Context, *zerolo
 		Str("account_number", job.Identity.Identity.AccountNumber).
 		Str("request_id", job.EdgeID).
 		Str("job_id", job.ID.String()).
-		Str("job_type", job.Type.String()).
-		Interface("job_args", job.Args)
+		Str("job_type", job.Type.String())
```
lzap, line 108: I thought it was useful to have it in all records, but I am fine.

pkg/worker/redis.go
```diff
@@ -188,29 +189,31 @@ func (w *RedisWorker) fetchJob(ctx context.Context) {
...
+	defer atomic.AddInt64(&w.inFlight, -1)
+
 	w.processJob(ctx, &job)
 }
 
 func (w *RedisWorker) processJob(origCtx context.Context, job *Job) {
 	if job == nil {
 		return
 	}
-	defer atomic.AddInt64(&w.inFlight, -1)
```
lzap, line 200: Why are you moving this? I am fine, just curious.

lzap:
These are good changes, yeah.

pkg/worker/redis.go
ezr-ondrej: I've moved it to `fetchJob` so it's together with the +1 ... if felt bit disjoined.
And if I understand it correctly, it should work the same, right?

pkg/worker/job.go
ezr-ondrej: All the arguments? It felt bit too extensive, that can get quite large, can't it?

lzap (approved):
Thanks.

PR #719, opened 2023-10-13
chore: start spans only when telemetry is enabled

ezr-ondrej:
Extract span start into a separate function.
This also obsoletes the Tracer naming per package, which is actually correct
According to Tracer docs https://pkg.go.dev/go.opentelemetry.io/otel/trace@v1.16.0#TracerProvider
the passed name should be a name of a tracing library, not the package being intrumented.

Follows #711 with cleanup of `TraceName` we got it wrong according to [docs](https://pkg.go.dev/go.opentelemetry.io/otel/trace@v1.16.0#TracerProvider).
More info is also in these nice docs: https://uptrace.dev/opentelemetry/go-tracing.html#tracer

### DIFF
--- a/docs/opentelemetry.md
+++ b/docs/opentelemetry.md
@@ -34,7 +34,7 @@ Keep in mind that logging level must be set to "trace" (-1) as all the tracing d
 
 ## Features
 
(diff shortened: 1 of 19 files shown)


lzap (approved):
SUperb.

PR #720, opened 2023-10-16
chore: log stats structure alongside totals

lzap:
Something does not work right, the numbers in the logs and prometheus/graphana do not match. This should show more insights.

### DIFF
--- a/internal/background/db_stats.go
+++ b/internal/background/db_stats.go
@@ -51,7 +51,7 @@ func dbStatsTick(ctx context.Context) error {
 		return fmt.Errorf("stats error: %w", err)
 	}
 
-	success, failure := 0, 0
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
sure :+1: Thanks

internal/background/db_stats.go
```diff
@@ -51,17 +51,20 @@ func dbStatsTick(ctx context.Context) error {
...
+	success, failure, pending := 0, 0, 0
 	for _, s := range stats.Usage24h {
 		if s.Result == "success" {
 			success += 1
 		}
 		if s.Result == "failure" {
 			failure += 1
 		}
+		if s.Result == "pending" {
+			pending += 1
```
ezr-ondrej, line 63: Should this be `pending += s.Count` ?

PR #724, opened 2023-10-17
chore: Include reservation template ID in Floorplan

avitova:
I will probably add the whole template later, but for now let's just see if the "is not null" does not misbehave, since we are using an empty string and not a null value.


ezr-ondrej (approved):
Thanks @avitova ! :)

We should be bit careful, as these are already customer data, but template id won't hurt for sure :+1: 

I read add full template later, so I'll just note: I'd love to see what customizations were used in the template, but I think we'll need to fetch that during the reservation, as we do not have that in our DB currently :)

PR #725, opened 2023-10-17
fix(HMS-2757): Do not cascade on pubkey delete

avitova:
So I think this should do. I forgot to set pubkey_id nullable and could not find rollback, so I have two migrations.
Tested scenario: create key, create a reservation with the key, delete the key, GetReservationById.
In this sequence of steps, this all works as expected; we do not have any pubkey id, and GetReservationById returns a response with no pubkey_id. All occurrences of PubkeyId for both responses and pubkey models should be covered by checking for nil pointers, and we should keep in mind, that we do allow empty pointers here.

### DIFF
--- a/cmd/spec/example_reservation.go
+++ b/cmd/spec/example_reservation.go
@@ -86,7 +86,7 @@ var AwsReservationRequestPayloadExample = payloads.AWSReservationRequest{
 }
 
 var AwsReservationResponsePayloadPendingExample = payloads.AWSReservationResponse{
-	PubkeyID:         42,
+	PubkeyID:         ptr.ToInt64(42),
 	SourceID:         "654321",
 	Region:           "us-east-1",
 	InstanceType:     "t3.small",
@@ -99,7 +99,7 @@ var AwsReservationResponsePayloadPendingExample = payloads.AWSReservationRespons
 
 var AwsReservationResponsePayloadDoneExample = payloads.AWSReservationResponse{
 	ID:               1305,
-	PubkeyID:         42,
+	PubkeyID:         ptr.ToInt64(42),
 	SourceID:         "654321",
 	Region:           "us-east-1",
 	InstanceType:     "t3.small",
@@ -133,7 +133,7 @@ var AzureReservationRequestPayloadExample = payloads.AzureReservationRequest{
 
 var AzureReservationResponsePayloadPendingExample = payloads.AzureReservationResponse{
 	ID:            1310,
-	PubkeyID:      42,
+	PubkeyID:      ptr.ToInt64(42),
 	SourceID:      "654321",
 	ResourceGroup: "myCustom Azure RG",
 	Location:      "useast",
@@ -147,7 +147,7 @@ var AzureReservationResponsePayloadPendingExample = payloads.AzureReservationRes
 
 var AzureReservationResponsePayloadDoneExample = payloads.AzureReservationResponse{
 	ID:            1310,
-	PubkeyID:      42,
+	PubkeyID:      ptr.ToInt64(42),
 	SourceID:      "654321",
 	ResourceGroup: "myCustom Azure RG",
 	Location:      "useast",
@@ -179,7 +179,7 @@ var GCPReservationRequestPayloadExample = payloads.GCPReservationRequest{
 
 var GCPReservationResponsePayloadPendingExample = payloads.GCPReservationResponse{
 	ID:               1305,
-	PubkeyID:         42,
+	PubkeyID:         ptr.ToInt64(42),
 	SourceID:         "654321",
 	Zone:             "us-east-4",
 	MachineType:      "e2-micro",
@@ -193,7 +193,7 @@ var GCPReservationResponsePayloadPendingExample = payloads.GCPReservationRespons
 
 var GCPReservationResponsePayloadDoneExample = payloads.GCPReservationResponse{
 	ID:               1305,
-	PubkeyID:         42,
+	PubkeyID:         ptr.ToInt64(42),
 	SourceID:         "654321",
 	Zone:             "us-east-4",
 	MachineType:      "e2-micro",
--- a/internal/dao/dao_errors.go
+++ b/internal/dao/dao_errors.go
@@ -38,4 +38,7 @@ var (
 
 	// ErrReservationRateExceeded is returned when SQL constraint does not allow to insert more reservations
 	ErrReservationRateExceeded = usrerr.New(429, "rate limit exceeded", "too many reservations, wait and retry")
+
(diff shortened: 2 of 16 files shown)


internal/migrations/sql/021_remove_delete_cascade.sql
```diff
@@ -0,0 +1,17 @@
+ALTER TABLE aws_reservation_details
+DROP CONSTRAINT aws_reservation_details_pubkey_id_fkey,
+ADD CONSTRAINT aws_reservation_details_pubkey_id_fkey
+FOREIGN key (pubkey_id) REFERENCES pubkeys(id) ON DELETE SET NULL;
+ALTER TABLE aws_reservation_details ALTER COLUMN pubkey_id DROP NOT NULL;
```
lzap, line 5: Is the double drop of `NOT NULL` here and then again in `022_pubkey_nullable.sql` intentional? I am trying to figure out why would you do that, but I feel like it is just an oversight.

internal/dao/dao_errors.go
```diff
@@ -38,4 +38,7 @@ var (
 
 	// ErrReservationRateExceeded is returned when SQL constraint does not allow to insert more reservations
 	ErrReservationRateExceeded = usrerr.New(429, "rate limit exceeded", "too many reservations, wait and retry")
+
+	// ErrPubkeyNotFound is returned when a nil pointer to a pubkey is used for reservation detail
+	ErrPubkeyNotFound = usrerr.New(404, "pubkey not found", "no pubkey found")
```
lzap, line 43: Maybe to stress out the fact that this was caused very likely by deletion (the usecase we see quite often):

```suggestion
	ErrPubkeyNotFound = usrerr.New(404, "pubkey not found", "no pubkey found (already deleted?)")
```

But I have to admit it looks weird, 404 is a 404 period. So feel free to let this wild idea go :-)

cmd/spec/example_reservation.go
```diff
@@ -86,7 +86,7 @@ var AwsReservationRequestPayloadExample = payloads.AWSReservationRequest{
 }
 
 var AwsReservationResponsePayloadPendingExample = payloads.AWSReservationResponse{
-	PubkeyID:         42,
+	PubkeyID:         func() *int64 { i := int64(42); return &i }(),
```
lzap, line 89: This is interesting, please explain why you haven’t used our utility function `ptr.ToInt64(42)` or alternatively `ptr.To(int64(v))`.

lzap (changes requested):
Perfect work, there are few nitpicks but overall it is perfect.

cmd/spec/example_reservation.go
avitova: Oh yeah! I definitely overlooked the ptr.ToInt64 function. This makes perfect sense to use since we have it. Thanks for the comment!

internal/migrations/sql/021_remove_delete_cascade.sql
avitova: Yeah, you are right, this was not intentional. Dropping these lines.

internal/dao/dao_errors.go
avitova: I have modified the message a bit to put the information out there.

avitova:
Thank you for the comments. All of them are addressed, as they were valid. :tada:

internal/dao/dao_errors.go
avitova: One more idea, we might want to incorporate a message telling users what can they do to fix this, right?

lzap: I thought it cannot be fixed, but sure that is the intention yes.

internal/dao/tests/reservation_test.go
```diff
@@ -58,7 +58,7 @@ func newGCPReservation() *models.GCPReservation {
 			AccountID: 1,
 			Status:    "Created",
 		},
-		PubkeyID: 1,
+		PubkeyID: func(i int64) *int64 { return &i }(1),
```
lzap, line 61: Perhaps here too?

internal/services/reservations_service.go
```diff
@@ -23,6 +23,7 @@ var (
 	ErrBothTypeAndTemplateMissing = errors.New("instance type or launch template not set")
 	ErrUnsupportedRegion          = errors.New("unknown region/location/zone")
 	ErrInvalidNamePattern         = errors.New("name pattern is not RFC-1035 compatible")
+	ErrPubkeyNotFound             = errors.New("no pubkey found, it may have been already deleted")
```
lzap, line 26: For plain Go errors I would keep it simple:

```suggestion
	ErrPubkeyNotFound             = errors.New("no pubkey found")
```

Only in "user message" (above) it makes sense I think.

lzap (changes requested):
There are two more anonymous functions but other than that all good.

lzap (approved):
Please check the linter and this can be merged, thank you.

lzap:
That should be resolved now.

/retest

ezr-ondrej:
/retest

I suspect the gcp is hitting public IP quota and we are then not having public IP present: https://ci.int.devshift.net/job/RHEnVision-provisioning-backend-pr-check/1975/

lzap:
Maybe if we expose private IPs to the UI, we can not rely on public IP at all. Would that help @akhil-jha ?

ezr-ondrej:
Ah, I guess I've caused conflict here :( Sorry! :(

internal/payloads/reservation_payload.go
```diff
@@ -55,7 +55,7 @@ type AWSReservationResponse struct {
 	ID int64 `json:"reservation_id" yaml:"reservation_id"`
 
 	// Pubkey ID.
-	PubkeyID int64 `json:"pubkey_id" yaml:"pubkey_id"`
+	PubkeyID *int64 `json:"pubkey_id,omitempty" yaml:"pubkey_id"`
```
ezr-ondrej, line 58:
```suggestion
	PubkeyID *int64 `json:"pubkey_id,omitempty" yaml:"pubkey_id,omitempty"`
 ```
I guess we should keep yaml and json consistent?

File: internal/payloads/reservation_payload.go [hunk 7]
```diff
@@ -121,7 +121,7 @@ type GCPReservationResponse struct {
...
```
ezr-ondrej, line 124: ditto

internal/payloads/reservation_payload.go
avitova: Sure, this slipped me through.

avitova:
Updated the yaml omitempty fields and rebased.

ezr-ondrej:
Thank you @avitova ! :orange_heart:

PR #722, opened 2023-10-17
chore: update deps

lzap:
Hopefully resolves https://github.com/advisories/GHSA-rcjv-mgp8-qvmr

Had to pin two dependencies because of 1.19 incompatibility already.

PR #726, opened 2023-10-18
chore: use OTLP instead of Jaeger exporter

lzap:
A small refactoring of tracing - we had to move from deprecated Jaeger exporter to OTLP exporter. With that, a small improvement - all errors of the tracing library now go into our Zerolog (thus platform logs). This works great, tested. However, I do not have Jaeger installed anymore, can you test it actually works @ezr-ondrej please?

FYI I renamed config group so now it should be:

```
TELEMETRY_OLTP_ENABLED=true
TELEMETRY_OLTP_HOSTNAME=localhost
```

This needs to be merged before we can upgrade deps: https://github.com/RHEnVision/provisioning-backend/pull/722


ezr-ondrej (approved):
Make sense, thanks @lzap ! :+1:

PR #728, opened 2023-10-18
Fix: SQL Union on different datatypes

avitova:
I do not wish to make a nightmare commit history. But I guess we are not able to test this until we find out that we broke the process. And it seems to be broken right now, we are not getting the latest data, unfortunately. I think I have found the reason, after incorporating the gcp, we still could have fetched the data, so I hope that it is only this.
I think that since I have recreated the field into a string field, it was wrong to leave the boolean for Azure.


lzap:
Tests for this are hard, my bad. I usually simply try the SQL in the console to validate it, however, as you can see I am lazy reviewer. Thanks.

lzap (approved):

PR #730, opened 2023-10-18
feat(HMS-2255): launch to location of the resource group by default

ezr-ondrej:
We should allow users to follow Azure default and use the resource group location by default.

This PR also makes the default Resource Group the image resource group.


akhil-jha:
> Failed: `response` not updated with instances
> 
> Maybe we need to update a test - images are now being launched in the location of the resource group rather than some hardcoded default, @akhil-jha

Probably true.
I'll look at it tomorrow.

internal/preload/instance_type.go
```diff
@@ -79,7 +79,7 @@ func (p *instanceType) ValidateRegion(region string) bool {
 		panic(fmt.Errorf("unable to read availability dir: %w", err))
 	}
 
-	fullName := region + ".yaml"
+	fullName := region + "_1.yaml"
```
ezr-ondrej, line 82: :warning: change for the future, we now validate region, not location.

In the past, we've validated for location, but then actually ignored the param down the line and forced `eastus`, but we actually need regions, not locations in all the calls (even tho azure calls them Location :facepalm:)

internal/services/azure_reservation_service.go
```diff
@@ -147,9 +149,10 @@ func CreateAzureReservation(w http.ResponseWriter, r *http.Request) {
 		EdgeID:    logging.EdgeRequestId(r.Context()),
 		AccountID: identity.AccountId(r.Context()),
 		Args: jobs.LaunchInstanceAzureTaskArgs{
+			// Ignore location for now
+			//Location:          reservation.Detail.Location,
```
ezr-ondrej, line 153: We have ignored location previously and always substituted it for `eastus`, now we're just explicit about it.

internal/services/azure_reservation_service.go
```diff
@@ -34,13 +34,16 @@ func CreateAzureReservation(w http.ResponseWriter, r *http.Request) {
...
-	}
-	if !preload.AzureInstanceType.ValidateRegion(payload.Location) {
-		renderError(w, r, payloads.NewInvalidRequestError(r.Context(), "Unsupported location", ErrUnsupportedRegion))
-		return
+	// validate region
+	// it needs to be in region format, but also
+	if payload.Location != "" && !preload.AzureInstanceType.ValidateRegion(payload.Location+"_1") {
+		// Location format is accepted now for backwards compatibility, but we should deprecate it
+		if preload.AzureInstanceType.ValidateRegion(payload.Location) {
+			logger.Warn().Msgf("Azure region passed with location suffix (%s), this is deprecated behaviour format", payload.Location)
```
ezr-ondrej, line 42: ⚠️ change for the future, we now validate region, not location.

In the past, we've validated for location, but then actually ignored the param down the line and forced eastus, but we actually need regions, not locations in all the calls (even tho azure calls them Location 🤦)

lzap:
iqe_provisioning_api.exceptions.ApiAttributeError: V1AWSReservationResponseDetail has no attribute 'public_ipv4' at ['['received_data', 'instances', 0, 'detail']']['public_ipv4']

Hmmm does not look related.

internal/clients/http/image_builder/image_client.go
```diff
@@ -87,41 +87,41 @@ func (c *ibClient) GetAWSAmi(ctx context.Context, composeID string) (string, err
...
 	}
 
 	azureUploadRequest, err := composeStatus.Request.ImageRequests[0].UploadRequest.Options.AsAzureUploadRequestOptions()
 	if err != nil {
-		return "", fmt.Errorf("failed to decode Azure upload request from IB: %w", err)
+		return "", "", fmt.Errorf("failed to decode Azure upload request from IB: %w", err)
 	}
-	return fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/images/%s", azureUploadRequest.ResourceGroup, uploadOptions.ImageName), nil
+	return azureUploadRequest.ResourceGroup, uploadOptions.ImageName, nil
 }
```
lzap, line 125: Multiple return values (except error) typically mean we should think about introducing a type. But this is still okay I believe.

internal/clients/interface.go
```diff
@@ -6,6 +6,12 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/models"
 )
 
+type AzureResourceGroup struct {
+	ID       string
+	Name     string
+	Location string
```
lzap, line 12: You removed bunch of locations to region, shall we remove this one as well?

internal/jobs/launch_instance_azure.go
```diff
@@ -72,6 +73,13 @@ func HandleLaunchInstanceAzure(ctx context.Context, job *worker.Job) {
 		logger.Debug().Msg("Resource group has not been set, defaulting to 'redhat-deployed'")
 		args.ResourceGroupName = DefaultAzureResourceGroupName
 	}
+	if args.Location != "" {
+		// Match for location - availability zone - suffix and remove it
+		res, e := regexp.MatchString(`_[1-6]\z`, args.Location)
+		if e == nil && res {
+			args.Location = args.Location[0 : len(args.Location)-2]
+		}
+	}
```
lzap, line 82: Can you explain this?

internal/services/azure_reservation_service.go
lzap: Does this need to be part of this PR? Also I am not sure if I am following here. Can you explain?

lzap (changes requested):
I am not sure about the terminology. We have the "location" term and then there are "region" and "zone". To me, it looks like "location" is simply "region" and "zone" concatenated into one string, maybe we could adopt this in the codebase.

We need to figure out a good approach, the current proposal simply ditches the location completely, however, it is possible to technically have a VM in a different zone than resource group. Sure, it is not recommended, but it is possible. I would suggest that we keep this possibility at least in the API.

There are some changes I do not understand.

internal/clients/http/image_builder/image_client.go
ezr-ondrej: Yeah, I was just lazy here, let me know if you insist, I agree struct would work better here :thinking:

internal/clients/interface.go
ezr-ondrej: I went back on the decision, lets keep Location, but explain it's region, not region with availability zone, as it souns like it the first look.

internal/jobs/launch_instance_azure.go
ezr-ondrej: I want to strip the availability zone suffix here, to keep supporting passing this as we forced that on users in the past unfortunatelly, so some clients migh have that already (`eastus_1`, etc)... I don't want to break those clients.

Basically this: https://go.dev/play/p/V6nK3LY4ajN

internal/services/azure_reservation_service.go
ezr-ondrej: Well, we required region+zone in the past, but then we dropped the value and  used hard coded `eastus`, so it worked, but our client expects only region (without the zone suffix).

So this patch accepts both, but when user passes the version with suffix, we now warn about it, so we can keep track of how many usages of that there is.

api/openapi.gen.json
```diff
@@ -66,7 +66,7 @@
           "amount": 1,
           "image_id": "composer-api-081fc867-838f-44a5-af03-8b8def808431",
           "instance_size": "Basic_A0",
-          "location": "useast_1",
+          "location": "useast",
```
ezr-ondrej, line 69: I forgot to update this, I guess this speaks the main message what most of this patch does.

internal/services/azure_reservation_service.go
lzap: Ok this is fine, if you want to go an extra mile I think you can refactor `ValidateRegion` to accept variadic string and concatenate by `_` because this is exactly how it works for all the clouds.

ezr-ondrej:
Rebased and all :green_apple:

internal/services/azure_reservation_service.go
ezr-ondrej: It's actually dash (`-`) for EC2 and GCP and underscore `_` for Azure :face_exhaling: xD

lzap:
Okay, thanks a bunch for working on this.

PR #731, opened 2023-10-19
chore: Add zone for GCP provider

avitova:


ezr-ondrej (approved):
lets try :stuck_out_tongue:

lzap:
Yes this works, although I am not sure how useful is to see template ID instead of just boolean. I guess you can group by templates now.

avitova:
@lzap Yeah, and I was also worried, that we are seeing True, while the string for template id is there but it is empty.

PR #735, opened 2023-10-23
chore: Fix typo in a chart label

ezr-ondrej:
The last panel says 24h, but it should be 28d IMO :)

image


lzap (approved):
Yeah thanks I noticed as well.

PR #736, opened 2023-10-25
chore: fix account requery

lzap:
The scanning function returns no rows error so the previous patch was not working. Indeed, there are no "requery" log statements on stage.

It is really hard to write a test for this, I would need to create many goroutines, we have tests for successful cases.


ezr-ondrej (approved):
Thanks @lzap ! LGTM, lets test it out :)

PR #738, opened 2023-10-27
docs: Add descriptions to all tags

ezr-ondrej:
Add descriptions to tags, as some libraries read the tags from tag list only.

Also moves the availability endpoint to Source tag, as it felt kinda lonely in the tag by itself and Source felt like a good home for it :)

For example RH api-catalog:

image

### DIFF
--- a/api/openapi.gen.json
+++ b/api/openapi.gen.json
@@ -1397,7 +1397,7 @@
           }
         },
         "tags": [
-          "AvailabilityStatus"
+          "Source"
         ]
       }
     },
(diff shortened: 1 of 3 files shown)


lzap (approved):
Thanks.

lzap:
iqe_provisioning_api.exceptions.ApiAttributeError: V1AWSReservationResponseDetail has no attribute 'public_ipv4' at ['['received_data', 'instances', 0, 'detail']']['public_ipv4']

/retest

ezr-ondrej:
:green_apple: Looks like some glitch - missing public_ipv4 quota, I guess we should keep the empty value for it :thinking:

PR #740, opened 2023-10-30
Removing optional param from secretKeyRef

mayurilahane:
removing `Optional: True` from "secretKeyRef" block


mshriver:
This will aggressively fail the job/pod if the secrets aren't available.

PR #744, opened 2023-11-07
chore: add pre-commit action

lzap:
This adds pre-commit action and two hooks: one for security token leaks and one for whitespace/end of lines. Not sure how this will work out, we might consider only adding the security related.


ezr-ondrej (commented):
I'm all for it, but why did you removed the icons? :(
Can we just merge commit message and commit check into one "Commit checks" and keep the icon?
Icons are fun! xD

lzap:
> I'm all for it, but why did you removed the icons? :(

Oh I forgot these were not working well with my Vim, I will put them back then.

lzap:
Fixed generators, there are multiple commits fixing various whitespace/eol issues in our code actually.

.github/workflows/gotest.yml
```diff
@@ -30,8 +30,16 @@ jobs:
           make check-commits
         working-directory: code
 
+  precommit:
```
ezr-ondrej, line 33: this seems to do the "same" thing as commit message (checking validity of a commit) and it even had the same icon! :stuck_out_tongue:  would it make sense to combine it?

ezr-ondrej:
> Rebased to get IQE passing but I think it will never be all green for this PR.

I'm happy to force merge it to get around the required checks (and then update them in settings for future PRs) but I believe IQE should go green, because there is no change that should break it, right?

.github/workflows/gotest.yml
ezr-ondrej: I'm happy to propose it in a follow up :)

lzap: I actually had it combined in my initial version but I could not figure out how to run action in different path (`./code`) instead of root, so I did split it. Feel free to refactor.

ezr-ondrej:
Thanks a bunch Lukas! :orange_heart: this is a great first step to have it enforced by GHA :+1:

ezr-ondrej (approved):
:orange_heart: Thanks! :+1:

.github/workflows/gotest.yml
ezr-ondrej: I've proposed it in https://github.com/RHEnVision/provisioning-backend/pull/745 but I'm actually not sure if its much better, but have a look and we can decide there :)

PR #745, opened 2023-11-07
ci: Combine commit checks into one check

ezr-ondrej:
I'm opening up just for a discussion, happy to close this one.

This is just a feeling based combination to safe bit of resources on setting up containers.

OTOH the separation makes the failure bit more readable without opening the logs :)


lzap (approved):
It is a bit ugly since we now do checkout into "code" directory and then again into the pwd but whatever.

PR #746, opened 2023-11-07
chore: update to Go 1.20, oapi-gen/v2 and deps

lzap:
Big update since Go 1.20 is live on ubi9, this allows us to unpin oapi-gen and finally upgrade to v2 which drops ton of dependencies because they have broken down the project into subprojects. I had to regenerate all OpenAPI clients too. Finally, updated all tools and deps to test it out.


go.mod
```diff
@@ -154,11 +141,7 @@ require (
 	github.com/shopspring/decimal v1.3.1 // indirect
 	github.com/spf13/cast v1.5.1 // indirect
 	github.com/stretchr/objx v0.5.1 // indirect
-	github.com/twitchyliquid64/golang-asm v0.15.1 // indirect
 	github.com/twmb/murmur3 v1.1.8 // indirect
-	github.com/ugorji/go/codec v1.2.11 // indirect
-	github.com/valyala/bytebufferpool v1.0.0 // indirect
-	github.com/valyala/fasttemplate v1.2.2 // indirect
```
lzap, line 161: Many of dependencies dropped because of openapi-gen changes (echo and its subdependencies).

mk/code.mk
```diff
@@ -17,7 +17,7 @@ imports: ## Rearrange imports using `goimports`
 
 .PHONY: lint
 lint: ## Run Go language linter `golangci-lint`
-	GOROOT=$(GOROOT) PATH=$(GOROOT)/bin:${PATH} GOCACHE=$(GOCACHE) $(GOLINT) run
+	GOROOT=$(GOROOT) PATH=$(GOROOT)/bin:${PATH} GOCACHE=$(GOCACHE) $(GOLINT) run --path-prefix $(shell pwd)
```
lzap, line 20: This prefix files with absolute patch which makes the output "clickable" in GoLand directly.

lzap:
So the newer linter is again more strict, bunch of refactoring:

* It now insist on `require.NoError` instead of `assert.NoError`
* It insist on `assert/require.Len` when testing for length
* It insist on using "getters" for protocol buffers for some reason
* Had to disable one linter which was doing a lot of false positives

ezr-ondrej (approved):
:orange_heart: it ! I love even the code improvements, but being on Go 1.20 and splitting off the oapi-codegen deps is amazing! :+1:

PR #747, opened 2023-11-07
fix(HMS-1847): Removing provider from env var of cloudwash pod spec

mayurilahane:
As per settings.yaml of cloudwash 
there is no Provider block now so need to remove it from the env var's format


lzap:
But I think Akhil is not working maybe @mshriver can give this a look. There are conflicts for some reason, you need to rebase this.

mshriver (approved):

PR #749, opened 2023-11-08
chore: fix postgres logging

lzap:
A small overlook, I decreased this value a bit few hours ago (from info to warning).

### DIFF
--- a/config/api.env.example
+++ b/config/api.env.example
@@ -76,7 +76,7 @@
 #     	cloudwatch logging stream (default "")
 #   DATABASE_HOST string
 #     	main database hostname (default "localhost")
-#   DATABASE_LOG_LEVEL string
+#   DATABASE_LOGGING_LEVEL string
 #     	logging level of database logs (default "info")
 #   DATABASE_MAX_CONN int32
 #     	connection pool maximum size (default "50")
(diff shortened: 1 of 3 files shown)


ezr-ondrej:
My proposal for improving the test is still in the IQE plugin repo, shall we discuss it? :)

deploy/clowdapp.yaml
```diff
@@ -43,7 +43,7 @@ objects:
                 value: ${LOGGING_LEVEL}
               - name: REST_ENDPOINTS_TRACE_DATA
                 value: ${REST_ENDPOINTS_TRACE_DATA}
-              - name: DATABASE_LOGGING_LEVEL
+              - name: DATABASE_LOG_LEVEL
```
ezr-ondrej, line 46: You've probably missed my comment:

Now the clowder parameter is named differently then the env value, do you think it would make sense to name them the same?
Renaming the paramter to LOGGING might be bit easier (just one line PR and two line MR to app-interface), right?

lzap: Right, okay, I am taking 180 deg turn, how about this?

ezr-ondrej (approved):
sure, that makes me happy! :orange_heart:

PR #750, opened 2023-11-10
feat(HMS-1674): Verify Azure source availability

ezr-ondrej:
Implement Azure availability check.
We are checking for subscription metadata.
If we are able to fetch that, we consider the source available. It's not bullet proof, but it covers basics ;)


internal/clients/http/azure/azure_client.go
```diff
@@ -3,17 +3,23 @@ package azure
...
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
+	"github.com/RHEnVision/provisioning-backend/internal/clients/http"
 	"github.com/RHEnVision/provisioning-backend/internal/config"
+	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 	"github.com/RHEnVision/provisioning-backend/internal/telemetry"
 )
 
+const AZURE_SERVICE_ACCOUNT_OBJECT_ID = "bf9c3e0e-4bbd-425b-adb1-f005d3dfe9f9"
```
ezr-ondrej, line 21: Ugly, unfortunatelly this is the attribute we do not have in config, we do have application ID.

We could:
a) resolve applicationID to objectID through azure sdk,
b) store the applicationID in config as new value used just for this check.

lzap: Storing it in the config makes sense, assuming it does not change too often.

lzap (approved):
Okay thanks.

internal/clients/http/azure/azure_client.go
ezr-ondrej: Oki, moved to config

lzap:
Relevant test error I believe.

lzap (approved):
Code looks good. Failure on CI tho.

ezr-ondrej:
/retest

Right, I've updated the account in ephemeral, lets :pray:

lzap:
Bad luck.

lzap:
Relevant failure error.

ezr-ondrej:
I've completely changed the approach here. I've tried to verify the correct role assignment, but it's bit too ambitious as Azure lighthouse works bit differently.
I've decided to simplify this and only fetch the subscription metadata and get TenantID from there.
It does not verify the permissions completely, but it does improve the situation quite a bit.

ezr-ondrej:
:green_apple:

lzap:
Thanks.

PR #752, opened 2023-11-14
fix(HMS-2933): Improve some HTTP error codes

avitova:
Hopefully this should be okay.
I am not entirely sure, if ParseUint would prevent out of range error. Maybe sources use ParseInt, and I just added a bigger range for them.
Also, I think, that they validate with different regex: https://github.com/RedHatInsights/sources-api-go/blob/4a3475e59b8f515dbda70c423826b2a3e4d07cb6/middleware/id_validation.go#L14, and I do not think that this is correct tbh, so I did this my way.
This is still WIP, but I wanted you to be able to comment on this in advance.

### DIFF
--- a/internal/clients/http/sources/sources_client.go
+++ b/internal/clients/http/sources/sources_client.go
@@ -114,18 +114,27 @@ func (c *sourcesClient) ListProvisioningSourcesByProvider(ctx context.Context, p
 	}
 
 	if resp == nil {
-		return nil, 0, fmt.Errorf("failed to get ApplicationTypes: empty response: %w", clients.ErrUnexpectedBackendResponse)
+		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("list application type sources returned no response")
+		return nil, 0, fmt.Errorf("failed to get ApplicationTypes from sources: empty response: %w", clients.ErrUnexpectedBackendResponse)
 	}
-	if resp.JSON404 != nil || resp.JSON400 != nil {
-		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("Get authentication from sources returned 4xx")
+
+	if resp.JSON404 != nil {
+		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("list application type sources returned 4xx")
+		return nil, 0, fmt.Errorf("failed to get ApplicationTypes from sources: %w", clients.ErrNotFound)
+	}
+	if resp.JSON400 != nil {
+		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("list application type sources returned 4xx")
+		return nil, 0, fmt.Errorf("failed to get ApplicationTypes from sources: %w", clients.ErrBadRequest)
 	}
 
 	if resp.JSON200 == nil {
-		return nil, 0, fmt.Errorf("failed to get ApplicationTypes: %w", clients.ErrUnexpectedBackendResponse)
+		logger.Warn().RawJSON("response", resp.Body).Msg("list application type sources returned non-200 response")
+		return nil, 0, fmt.Errorf("failed to get ApplicationTypes from sources: %w", clients.ErrUnexpectedBackendResponse)
 	}
 
 	if resp.JSON200.Data == nil {
-		return nil, 0, fmt.Errorf("list provisioning sources call: %w", clients.ErrNoResponseData)
+		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("list application type sources returned no data")
+		return nil, 0, fmt.Errorf("failed to get ApplicationTypes from sources: %w", clients.ErrNoResponseData)
 	}
 
 	result := make([]*clients.Source, 0, len(*resp.JSON200.Data))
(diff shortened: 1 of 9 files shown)


ezr-ondrej:
Nice, this is what we probably should do with all the id params. :orange_heart: 
Do we also want to correct the return codes on the sources non-200 response, or would that be another PR? :)

avitova:
How to test this:
These requests returned 500:
GET: /api/provisioning/v1/sources/{ID}/launch_templates
GET: /api/provisioning/v1/sources/zj{{=9243*9806}}zj/launch_templates 
/api/provisioning/v1/pubkeys (for example, add a new field in the request body, that should not be there) - on create/update

ezr-ondrej:
> I do not mind doing a followup, but I can include it in here as well.

No real rush ;) We are on target with SLO now, so we can do it properly and do follow-up :)

internal/middleware/id_validation.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	// Checking for out of range error
+	_, err := strconv.ParseUint(id, 10, 64)
+	if err != nil {
+		return fmt.Errorf("%w: id too high", ErrInvalidId)
+	}
+
+	return nil
+}
```
lzap, line 26: Do not assume that primary keys in databases are higher than 0, in fact, when a database reaches the maximum value it starts from -MAX back to zero. This is at least behavior of postgres. So I would drop this check only keeping `ParseUint` and checking for error. In addition, I would expect this function to actually error out when called with negative number anyway.

I would not bother too much about the error message, we have properly documented what is expected, just returning "invalid id" is enough.

internal/services/aws_iam_service.go
```diff
@@ -15,6 +17,10 @@ import (
 func ValidatePermissions(w http.ResponseWriter, r *http.Request) {
 	logger := zerolog.Ctx(r.Context())
 	sourceId := chi.URLParam(r, "ID")
+	if err := middleware.DigitsOnly(sourceId); err != nil {
+		renderError(w, r, payloads.NewURLParsingError(r.Context(), "id parameter invalid", err))
+	}
+
 	region := r.URL.Query().Get("region")
 
 	if region == "" {
```
lzap, line 26: I highly suggest to avoid creating cyclic dependencies, what if middleware wants to import this package? Then we have a problem. I suggest to find a new home for the validation function and if there is no such place simply create a brand new `validate` or similar package. It might sound weird, but this is going to pay off in the future.

You can take advantage of the fact that there is no link between parent and child package in go, so you can go ahead and create `payload/validate` package and it will be a completely separate entity but still you will not "pollute" the root folder.

lzap (changes requested):
Looks good, tho is could be simplified a bit.

internal/middleware/id_validation.go
lzap: Now that I think about it, you should not use `ParseUint` because our primary keys are `int64`, just parse a regular 64 signed number and report errors.

lzap:
Any updates @avitova ? Thanks.

internal/middleware/id_validation.go
avitova: Yeah, sorry. There was even an invalid string, this regex does not check against 0, but numbers overall. The second check is against numbers that might overflow.

The second check would not fail for negative numbers. So I think we can keep both.

internal/services/aws_iam_service.go
avitova: Let me know if you feel better about the structure now:)

internal/clients/http/sources/sources_client.go
```diff
@@ -222,9 +222,15 @@ func (c *sourcesClient) GetAuthentication(ctx context.Context, sourceId string)
...
-	if resp.JSON404 != nil || resp.JSON400 != nil {
+	if resp.JSON404 != nil {
 		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("Get authentication from sources returned 4xx")
+		return nil, fmt.Errorf("get source authentication call: %w", clients.ErrNotFound)
 	}
+	if resp.JSON400 != nil {
+		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("Get authentication from sources returned 4xx")
+		return nil, fmt.Errorf("get source authentication call: %w", clients.ErrBadRequest)
+	}
+
```
avitova, line 233: These two would be the most common cases - sources return 404, 400 or 2xx payloads. I think any other case could be a larger failure that we might want to know about anyway.
Therefore, after calling GetAuthentication, I feel like we want to check for ErrBadRequest and ErrNotFound, and the rest, like ErrNoResponseData, ErrUnexpectedBackendResponse should be left as is. :)

internal/clients/http/sources/sources_client.go
```diff
@@ -222,9 +222,15 @@ func (c *sourcesClient) GetAuthentication(ctx context.Context, sourceId string)
...
 		return nil, fmt.Errorf("get source authentication call: empty response: %w", clients.ErrUnexpectedBackendResponse)
 	}
-	if resp.JSON404 != nil || resp.JSON400 != nil {
+	if resp.JSON404 != nil {
 		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("Get authentication from sources returned 4xx")
+		return nil, fmt.Errorf("get source authentication call: %w", clients.ErrNotFound)
 	}
+	if resp.JSON400 != nil {
+		logger.Warn().Bytes("body", resp.Body).Int("status", resp.StatusCode()).Msg("Get authentication from sources returned 4xx")
+		return nil, fmt.Errorf("get source authentication call: %w", clients.ErrBadRequest)
```
avitova, line 231: The missing return at this resp.JSON404 != nil || resp.JSON400 != nil branch was the cause for the invalid returns.
The next **if resp.JSON200 == nil** returned ErrUnexpectedBackendResponse.
We are now validating id, so this would not happen for invalid ids at all. For 404 and 400 we are now able to return the correct status.

ezr-ondrej (approved):
I did not tested this, but code looks exactly what I've imagined! :orange_heart:

internal/payloads/validation/id.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	// Checking for out of range error
+	_, err := strconv.ParseInt(id, 10, 64)
+	if err != nil {
+		return ErrInvalidId
+	}
+
+	return nil
+}
```
lzap, line 26: I still do not understand why there is a regular expression needed, both `strconv.Atoi` and `strconv.ParseInt` will return an error in case the string does not contain a valid number: https://go.dev/play/p/XFoPrKfctrC

On top of that, the proposed regular expression will not work for negative numbers.

internal/payloads/validation/id.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+)
+
+var (
+	idValidation = regexp.MustCompile("^[0-9]+$")
+	ErrInvalidId = fmt.Errorf("invalid id")
```
lzap, line 11:
```suggestion
	ErrInvalidId = errors.New("invalid id")
```

internal/payloads/validation/id.go
```diff
@@ -0,0 +1,26 @@
...
+
+func DigitsOnly(id string) error {
+	if !idValidation.MatchString(id) {
+		return ErrInvalidId
+	}
+
+	// Checking for out of range error
+	_, err := strconv.ParseInt(id, 10, 64)
+	if err != nil {
+		return ErrInvalidId
```
lzap, line 22:
```suggestion
		return fmt.Errorf("%w: %s", ErrInvalidId, err.Error())
```

With Go 1.21 you could actually wrap both errors but we are on 1.20 at the moment so string will do.

internal/payloads/validation/id_test.go
```diff
@@ -0,0 +1,59 @@
...
+	err = DigitsOnly(id3)
+	require.NoError(t, err)
+}
+
+func TestDigitsOnlyWithSymbolsFails(t *testing.T) {
+	id1 := "-1111"
+	err := DigitsOnly(id1)
+	if err == nil {
+		t.Errorf("no error for invalid input: %s", id1)
+	}
```
lzap, line 28: As I said, negative primary key is a valid key. It is rather theoretical because we will never run out of positive numbers this century but still.

lzap (changes requested):
Can you explain why the regular expression is needed?

internal/payloads/validation/id_test.go
avitova: Oh, I see. Okay, I am sorry. I misunderstood your previous comment on this issue. :+1:

internal/payloads/validation/id.go
avitova: I was afraid about how Golang would handle numbers from other alphabets since we never really know what things from libraries do under the hood.
I thought that the regex would not hurt, and wanted to make sure, that some exotic things like Persian numbers would not pass the validation. But I have added some more values into the test before removing the regex, so it should be hopefully okay.

ezr-ondrej (commented):
Can we add the log of the failure in `JSON200  == nil` that we have in `Get authentication` also in `ListProvisioningSourcesByProvider` and `ListAllProvisioningSources` for easier debugging in the future?

avitova:
Yeah, and you have correctly reminded me that this has the exact handling, and we should return the error even for 404 and 400 at those two places as well. @ezr-ondrej

lzap (approved):
Very extensive testing indeed :-) Nice.

I would put few empty lines between those if statements and the error message is about "id" while function is called `DigitsOnly` which is a slight inconsistency but 
that is okay. Merging, thanks.

PR #753, opened 2023-11-21
fix(HMS-3000): update log statement for kafka

lzap:
Code-wise everything is okay and we already use multiple kafka brokers. It is just we haven’t deployed into stage for quite a bit and I do not see any of the messages, this small fix will trigger deployment so I can investigate the string on stage.


lzap:
Fixed pre-commit hook complaining on the generated CHANGELOG file.

ezr-ondrej (approved):
Lets see it :+1:

PR #754, opened 2023-11-22
fix(HMS-1829): translate source_type to provider

ezr-ondrej:
We want to translate sources type into our provider type. We do not want to expose sources internals, but rather translate to user readable data.

I need to make extra call to sources to be able to translate the ID into string.
I feel slightly dirty, any advice on how to improve on one extra call to sources appreciated :)
It's static within an environment so we might be able to cache it pretty intensly, advice welcome :)


internal/clients/http/sources/sources_client.go
```diff
@@ -274,6 +286,40 @@ func (c *sourcesClient) GetProvisioningTypeId(ctx context.Context) (string, erro
 	return appTypeId, nil
 }
 
+func (c *sourcesClient) GetSourceTypes(ctx context.Context) (map[ID]string, error) {
+	sourceTypes, err := c.loadSourceTypes(ctx)
```
ezr-ondrej, line 290: I guess that if we don't find a better solution, we can cache this whole map[ID]string :shrug: 

@lzap opinion whether its a good idea? :thinking:

lzap: Hmmm how about refactoring [global type id](https://github.com/RHEnVision/provisioning-backend/blob/main/internal/cache/redis.go#L36) we hold in cache package as a global variable into struct type `AppConstants` or something like that having ID as one field and this map as another.

All you need is to implement the Cacheable interface and register the type as a GOB serializeable type, all [in the same file](https://github.com/RHEnVision/provisioning-backend/blob/main/internal/cache/redis.go#L36).

lzap:
Other than that looks okay.

ezr-ondrej:
> Other than that looks okay.

Alright, sounds good! Done :sunglasses:
I've added it as second commit, for easier review and cleaner history :)

ezr-ondrej:
When going over it, I've noticed I missed adding the value in examples, so I've squashed that change in the first commit, but apart of that, first commit stays the same.

lzap:
Needs IQE refresh.

lzap (approved):
Ah sorry I mentioned "register a type for GOB serialization", that was not necessary. I meant to keep the map just as a regular global variable and not storing it in Redis at all.

Tho granted doing it this way in `cache` named package is weird, it should have been somewhere else from day one. It sort of ended up there because I created it there and then you had a good point NOT to store it in Redis actually :-)

Whatever, this will work too, if you want you can refactor it back and put it somewhere into appropriate place, maybe `cache/constant`

ezr-ondrej:
> Whatever, this will work too, if you want you can refactor it back and put it somewhere into appropriate place, maybe `cache/constant`

What about this? I've put it in the client package itself, as there is no need to expose it outside of `clients/sources/cache`.
What do you think? The naming is bit wobbly, happy to change it! :)

akhil-jha:
Looking into the failure.

ezr-ondrej:
:green_apple: :tada:

PR #755, opened 2023-11-24
fix(HMS-2933): Fix validator DAO error

avitova:
This should fix the 500 status


ezr-ondrej:
Ugh, great find! :+1:
Nasty issue, would be nice to have a test for this :thinking:
Happy to do it as follow-up tho :)

ezr-ondrej (approved):
Yeah, on the second thought, lets fix this ASAP ;)
We can work on test coverage later on.

Thanks ! :orange_heart:

PR #756, opened 2023-11-24
fix(2933): Unify GetAuthentication handling

avitova:
I will look into more NewClientErrors generated incorrectly

### DIFF
--- a/internal/payloads/error_payload.go
+++ b/internal/payloads/error_payload.go
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/RHEnVision/provisioning-backend/internal/clients"
+
 	"github.com/RHEnVision/provisioning-backend/internal/usrerr"
 
 	"github.com/RHEnVision/provisioning-backend/internal/logging"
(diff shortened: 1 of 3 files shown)


internal/payloads/error_payload.go
```diff
@@ -199,3 +201,13 @@ func NewGCPError(ctx context.Context, message string, err error) *ResponseError
...
+
+func HandleGetSourcesAuthenticationError(ctx context.Context, err error) *ResponseError {
+	if errors.Is(err, clients.ErrNotFound) {
+		return NewNotFoundError(ctx, "unable to get authentication for sources", err)
+	}
+	if errors.Is(err, clients.ErrBadRequest) {
+		return NewResponseError(ctx, http.StatusBadRequest, "unable to get authentication from sources", err)
+	}
+	return NewClientError(ctx, err)
+}
```
lzap, line 213: This is okay, but can you do it in `NewClientError` function instead? We already do this for user messages, we can do the similar check too. Then there is no need to change all the services:

```
func NewClientError(ctx context.Context, err error) *ResponseError {
	if response := findUserResponse(ctx, "Client error", err); response != nil {
		return response
	}
	log.Ctx(ctx).Error().Msgf("Unknown client error: %s", err)
	return NewResponseError(ctx, 500, "backend client error", err)
}
```

lzap (approved):
Great stuff, cheers!

ezr-ondrej (commented):
Sorry for being late to the party, but the test doesn't test the code we are updating, does it?
It rather tests the ID validation that we've introduced in the previous PR.
It's still perfect to have such test, don't get me wrong, it's just unrelated I feel like :)
But it's fine :)

lzap:
What you mean, it does appear to test for `StatusBadRequest`.

ezr-ondrej:
> What you mean, it does appear to test for `StatusBadRequest`.

I mean all the tests are testing https://github.com/RHEnVision/provisioning-backend/blob/c6442c2e0c462142e3b19cfbdc4fa024aae83b80/internal/services/sources_service.go#L87 which is indead BadRequest, but not caused by the cases that we are covering here :)

The test are simulating invalid ID, not these errors :)

avitova:
Yeah, that is correct, I should have probably created a different commit/PR for that.

PR #757, opened 2023-11-24
test: pubkey validation returns correct status

ezr-ondrej:
This is not perfect, because it duplicates the validation funcionality in the stub, so it still have untested paths (when we'd change validation in DAO, changed the DAO test, but forgotten to keep the stub validation up to date).

But IMO it's better then what we have no - no tests for validation on the handler level :)


lzap (approved):

ezr-ondrej:
:green_apple:

lzap (approved):
Nice. TY.

PR #758, opened 2023-11-24
refactor: extract common funcionality of listing sources

ezr-ondrej:
Extracts funcionality that was duplicated between ListSources and ListSourcesByProvider. Keeps the unique bits and thus both functions.
This way it will be easy to merge the functions once sources support the filtering through OpenAPI client.


lzap (approved):
This one's messy, I get the idea and tests are passing, thanks for noticing!

PR #759, opened 2023-11-28
fix(HMS-3185): fail with 403 on failed assume role

ezr-ondrej:
AssumeRole is supposed to assume the role that client is giving us. This fixes the error code when this operation fails.

It could have also been 401 as the role might be considered an input. Given this is a permissions operation, I went with 403.

Let me know if you think 401 (or any other code) would be more fitting.

I've went bit overboard to mimic the error that AWS returns in our stub.
But now it should test the real thing pretty closely.

### DIFF
--- a/internal/payloads/error_payload.go
+++ b/internal/payloads/error_payload.go
@@ -8,8 +8,8 @@ import (
 	"strings"
 
 	"github.com/RHEnVision/provisioning-backend/internal/clients"
-
 	"github.com/RHEnVision/provisioning-backend/internal/usrerr"
+	"github.com/aws/smithy-go"
 
 	"github.com/RHEnVision/provisioning-backend/internal/logging"
 	"github.com/RHEnVision/provisioning-backend/internal/version"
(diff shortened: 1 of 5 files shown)


internal/payloads/error_payload.go
```diff
@@ -194,6 +194,11 @@ func NewStatusError(ctx context.Context, message string, err error) *ResponseErr
 }
 
 func NewAWSError(ctx context.Context, message string, err error) *ResponseError {
+	var awsOpErr *smithy.OperationError
+	if errors.As(err, &awsOpErr) && awsOpErr.OperationName == "AssumeRole" {
+		message = fmt.Sprintf("AWS assume role failed: %s", message)
+		return NewResponseError(ctx, http.StatusForbidden, message, err)
+	}
```
lzap, line 201: This does not look correct, operation name does not say anything about resulting HTTP code. Isn’t better to search for `GenericAPIError` code "AccessDenied"?

lzap (changes requested):
Tests failed, irrelevant tho I think.

internal/payloads/error_payload.go
ezr-ondrej: I was actually talking about the middle layer error (ResponseError), you're right, GenericAPIError makes more sense here, fixed! :+1:

lzap (approved):

ezr-ondrej:
We are :green_apple: lets get it in

PR #760, opened 2023-11-28
fix(HMS-1847): Removing location and resource group

mayurilahane:
Have added a new vault path for only secrets.
https://vault.devshift.net/ui/vault/secrets/insights/show/secrets/qe/global/hms/azure

and removed the location and resource group field of auth from the previous path.

now pod spec will fetch the location and resource group value from the configmap file instead of overriding it from the secrets file.

### DIFF
--- a/deploy/cloudwash_job.yaml
+++ b/deploy/cloudwash_job.yaml
@@ -56,16 +56,6 @@ objects:
                       secretKeyRef:
                         key: subscription_id
                         name: provisioning-azure-auth-secret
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
The commit message is missing space after `ci:` also, can we have the ticket reference in the commit message as well (not just the PR title)?

mayurilahane:
@mshriver needs ack here

mshriver (approved):

akhil-jha (approved):

PR #761, opened 2023-12-01
fix(HMS-1847):Aws_cloudwash_pipeline

mayurilahane:
parameterizing pod template to use Azure as well AWS for the cleanup.

### DIFF
--- a/deploy/cloudwash_job.yaml
+++ b/deploy/cloudwash_job.yaml
@@ -2,7 +2,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: provisioning-cleanup-resources
+  name: provisioning-cleanup-cloud-resources
 objects:
 - apiVersion: batch/v1
   kind: CronJob
@@ -22,7 +22,7 @@ objects:
               - name: quay-cloudservices-pull
             restartPolicy: Never
             containers:
-              - name: provisioning-cleanup
+              - name: provisioning-azure-cleanup
                 image: quay.io/redhatqe/cloudwash
                 imagePullPolicy: Always
                 command:
@@ -37,7 +37,7 @@ objects:
                     mountPath: /opt/app-root/src/cloudwash/settings.yaml
                     subPath: settings.yaml
                 env:
-                  - name: CLEANUP_AZURE__AUTH__CLIENT_ID
+                  - name: CLEANUP_AZURE__AUTH__CLIENT_ID # Azure creds
                     valueFrom:
                       secretKeyRef:
                         key: azure_client_id
@@ -57,6 +57,31 @@ objects:
                       secretKeyRef:
                         key: subscription_id
                         name: provisioning-azure-auth-secret
+
+              - name: provisioning-aws-cleanup
+                image: quay.io/redhatqe/cloudwash
+                imagePullPolicy: Always
+                command:
+                  - swach
+                args:
+                  - "-d"
+                  - "aws"
+                  - "--vm"
+                volumeMounts:
+                  - name: config-volume
+                    mountPath: /opt/app-root/src/cloudwash/settings.yaml
+                    subPath: settings.yaml
+                env:
+                  - name: CLEANUP_AWS__AUTH__ACCESS_KEY # AWS creds
+                    valueFrom:
+                      secretKeyRef:
+                        key: access_key_id
+                        name: provisioning-aws-auth-secret
+                  - name: CLEANUP_AWS__AUTH__SECRET_KEY
+                    valueFrom:
+                      secretKeyRef:
+                        key: secret_access_key
+                        name: provisioning-aws-auth-secret
             volumes:
               - name: config-volume
                 configMap:


deploy/cloudwash_job_aws.yaml
```diff
@@ -0,0 +1,64 @@
...
+            restartPolicy: Never
+            containers:
+              - name: provisioning-cleanup
+                image: quay.io/redhatqe/cloudwash
+                imagePullPolicy: Always
+                command:
+                  - swach
+                args:
+                  - "-d"
+                  - "aws"
```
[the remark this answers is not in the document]
mayurilahane: and what about env var?
it varies for every hypervisor
aws needs 2 var
azure needs 6 var

deploy/cloudwash_job.yaml
```diff
@@ -66,6 +64,16 @@ objects:
                       secretKeyRef:
                         key: location
                         name: provisioning-azure-auth-secret
+                  - name: CLEANUP_AWS__AUTH__ACCESS_KEY # AWS creds
+                    valueFrom:
+                      secretKeyRef:
+                        key: access_key_id
+                        name: cloudwash-secret
```
mshriver, line 71: Do you have an app-interface MR for the namespace to include this secret?

mayurilahane: yep here
https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/89921/diffs

mayurilahane:
@mshriver or @akhil-jha wanna give it a go ?

deploy/cloudwash_job.yaml
```diff
@@ -73,3 +80,6 @@ parameters:
   description: "Schedule for cron job"
   value: ''
   required: true
+- name: CLOUDWASH_ARGS
+  required: true
+  value: '["aws", "azure", "--nics", "--vms", "--discs", "--images"]'
```
akhil-jha, line 85: We don't want to delete images.

akhil-jha: `'["aws", "azure", "--nics", "--vms", "--discs", "--images"]' ` is not a valid command. 
Can you confirm it?

akhil-jha (commented):
I would say having a working azure cleanup takes higher priority than add all the clouds for now.

deploy/cloudwash_job.yaml
mayurilahane: Thanks for the review!
I would like to get some feedback from @mshriver too.
and yes those are parameterized lists for both the cloud resources.
that's why it's like combined params but yeah let's discuss it when mike is back.

mayurilahane: looks like cloudwash will only support one cloud at a time.
and as per akhil mentioned i am creating a new container for other resource 
Thanks !

mayurilahane: ref : https://redhat-internal.slack.com/archives/C04B2FX09FV/p1705412248915979?thread_ts=1705355987.988969&cid=C04B2FX09FV

akhil-jha: Please don't post the internal links on github

deploy/cloudwash_job.yaml
```diff
@@ -66,6 +67,30 @@ objects:
                       secretKeyRef:
                         key: location
                         name: provisioning-azure-auth-secret
+            containers:
```
mshriver, line 70: This line is breaking the spec - the containers map is already defined on L24, this should be an additional item in that map.

deploy/cloudwash_job.yaml
```diff
@@ -56,16 +57,30 @@ objects:
...
-                  - name: CLEANUP_AZURE__AUTH__RESOURCE_GROUPS
+              - name: provisioning-aws-cleanup
+                image: quay.io/redhatqe/cloudwash
+                imagePullPolicy: Always
+                command:
+                  - swach
+                args:
+                  - "-d" # Need confirmation from jakub or akhil then only we can remove this flag.
+                  - "aws"
+                  - "--all"
```
akhil-jha, line 68: Only vms for aws and gcp. 
The structure of azure vs aws/gcp is different on how we are billed.

akhil-jha: and Yes, we can remove the `-d` parameter.

akhil-jha: We don't care about the vm prefix for these clouds.

mayurilahane: added only vms flag and removed -d flag 
thanks!

deploy/cloudwash_job.yaml
```diff
@@ -57,6 +57,30 @@ objects:
                       secretKeyRef:
                         key: subscription_id
                         name: provisioning-azure-auth-secret
+              - name: provisioning-aws-cleanup
+                image: quay.io/redhatqe/cloudwash
+                imagePullPolicy: Always
+                command:
+                  - swach
+                args:
+                  - "-d"
```
mshriver, line 66: thanks for starting with dryrun here

mshriver (approved):

mshriver (approved):

PR #766, opened 2024-01-10
fix(HMS-3369): Implement replacing as beforeSend hook

ezr-ondrej:
With implementing the replace functionality as Sentry hook,
we can implement it bit more efficiently for only messages that are sent to sentry.

It also allows us to use the sentry writer as LevelWriter and leverage WriteLevel function.
WriteLevel is bit more streamlined as it allows to skip filtering when not needed.

### DIFF
--- a/internal/logging/sentry.go
+++ b/internal/logging/sentry.go
@@ -2,28 +2,25 @@ package logging
 
 import (
 	"fmt"
-	"io"
 
 	"github.com/RHEnVision/provisioning-backend/internal/config"
 	sentrywriter "github.com/archdx/zerolog-sentry"
 	"github.com/getsentry/sentry-go"
+	"github.com/rs/zerolog"
 )
 
 // sentryWriter creates a zerolog writer for sentry.
 // Uses github.com/archdx/zerolog-sentry which is very simple wrapper.
-func sentryWriter(dsn string) (io.Writer, func(), error) {
-	sWriter, err := sentrywriter.New(dsn)
+func sentryWriter(dsn string) (zerolog.LevelWriter, func(), error) {
+	replacer := NewSentryReplacer()
+
+	sWriter, err := sentrywriter.New(dsn, sentrywriter.WithBeforeSend(replacer.Replace))
 	if err != nil {
 		return nil, func() {}, fmt.Errorf("cannot initialize sentry: %w", err)
 	}
 	sentry.ConfigureScope(func(scope *sentry.Scope) {
 		scope.SetTag("stream", config.BinaryName())
 	})
-	fWriter := NewSentryReplacer(sWriter)
 
-	closeFunc := func() {
-		_ = fWriter.Close()
-		_ = sWriter.Close()
-	}
-	return fWriter, closeFunc, nil
+	return sWriter, func() { _ = sWriter.Close }, nil
 }
--- a/internal/logging/sentry_replacer.go
+++ b/internal/logging/sentry_replacer.go
@@ -1,12 +1,10 @@
 package logging
 
 import (
-	"bytes"
-	"fmt"
-	"io"
 	"regexp"
 	"strings"
-	"sync"
+
+	"github.com/getsentry/sentry-go"
 )
 
 var filters = []string{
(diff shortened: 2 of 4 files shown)


internal/logging/sentry_replacer.go
```diff
@@ -30,13 +32,18 @@ var replacement = []byte{'?'}
...
 	mu     sync.Mutex
 	closed bool
 }
 
-func NewSentryReplacer(w io.Writer) *SentryReplacer {
+func NewSentryReplacer(w zerolog.LevelWriter, levels ...zerolog.Level) *SentryReplacer {
+	lvls := make(map[zerolog.Level]struct{}, len(levels))
+	for _, lvl := range levels {
+		lvls[lvl] = struct{}{}
+	}
```
lzap, line 46: Where `lvls` is actually used?

internal/logging/sentry_replacer.go
```diff
@@ -45,13 +52,56 @@ func NewSentryReplacer(w io.Writer) *SentryReplacer {
...
+		line, sr.buf = sr.buf[:idx+1], sr.buf[idx+1:]
+		line = sr.re.ReplaceAll(line, replacement)
+
+		_, err := sr.w.Write(line)
+		if err != nil {
+			// This is not strictly the error related to the incoming `p`, but the best we can do.
+			return 0, fmt.Errorf("cannot filter: %w", err)
+		}
+	}
+}
```
lzap, line 88: So it looks like level gets replaced by the library:https://github.com/archdx/zerolog-sentry/blob/a5e7c6e3cfc1dabe917a8ce22c4a7642a0c9bafd/writer.go#L94

So we cannot delegate here, these must be two different functions.

Can you find a way to reuse this code? It looks like a complex one. How about making a helper function and then passing in an (anonymous) function that will:

* call `Write(line)`
* call `WriteLevel(level, line)` where level would be captured via closure

A callback approach is not the only one, that is the first one which comes to my mind.

internal/logging/sentry_replacer_test.go
```diff
@@ -57,50 +58,50 @@ func TestUUIDSplit(t *testing.T) {
...
 	require.Equal(t, "The resource ? was not found\n", buf.String())
 }
 
 func TestAzureTime(t *testing.T) {
 	buf := bytes.NewBufferString("")
-	repl := NewSentryReplacer(buf)
+	repl := NewSentryReplacer(zerolog.LevelWriterAdapter{Writer: buf})
 	_, _ = repl.Write([]byte("'start time': '2023-06-24T19:34:34.2581206+00:00'\n"))
 	require.Equal(t, "'start time': '?'\n", buf.String())
 }
```
lzap, line 107: There are zero tests for the newly introduced `WriteLevel` function, just add an extra `require` to each test.

lzap (changes requested):
Looks good, nice catch, just reuse the code.

internal/logging/sentry_replacer.go
```diff
@@ -45,13 +52,56 @@ func NewSentryReplacer(w io.Writer) *SentryReplacer {
...
 
+	// skip the filter unless we will actually write to sentry
+	if _, ok := sr.levels[level]; !ok {
+		//nolint:wrapcheck
+		return sr.w.WriteLevel(level, p)
+	}
+
+	sr.mu.Lock()
+	defer sr.mu.Unlock()
+
```
avitova, line 104: Is there a reason to do this before locking a mutex? I am not quite sure I am getting the goal of this PR.
But it seems like we are working with code achievable from goroutines. We have a pointer here, and executing several sr.w.WriteLevel on the same pointer in different threads might eventually make a mess.
Or is it not the case?

avitova (commented):
LGTM, I have only one comment. Let me know what the intentions are with that part of code. :)

internal/logging/sentry_replacer.go
ezr-ondrej: The reason is that if the requested level is not in enabled levels, we just pass the call down to the underlying logger.

The main reason for the mutex here is the internal buffer where we keep the logs in between calls. Tho to be fair I'm not sure why its needed and I'm not  sure how the logger works, but IMO when we do not touch that buffer, we do not need to worry about locking.

The underlying logger should IMO be thread safe by its own. It's only a theory tho, so maybe @lzap can chim in here :)

internal/logging/sentry_replacer_test.go
ezr-ondrej: Will do :+1:

internal/logging/sentry_replacer.go
lzap: Yeah the mutex is only needed for the buffer which is shared.

From the zerolog docs, it seems that the library does not affect thread safety, it guarantees that every single log event = single write which is thread safe as long as your underlaying writer is thread safe (e.g. appending to a file in Linux).

internal/logging/sentry_replacer.go
ezr-ondrej: Yeah, I did think about few approaches but all felt bit dirty, I'll try something in a follow-up commit.

With callbacks I see bit of an issue of creating closure (I'm not sure how it works in Go TBH), because closures can take quite a bit of a space on stack, which might go against the minimal memory footprint of zerolog.

internal/logging/sentry_replacer.go
ezr-ondrej: just to filter out the levels we do not want to filter on in the `WriteLevel`. I've copied this from https://github.com/archdx/zerolog-sentry/blob/a5e7c6e3cfc1dabe917a8ce22c4a7642a0c9bafd/writer.go#L78 but it might not be the best approach. I'd love to use the library internals instead, but it's not really possible.

Maybe we can just copy the library and merge the funcionality? :shrug:

ezr-ondrej:
I've found a way to simlify the filtering significantly by using `BeforeSend` hook https://docs.sentry.io/platforms/go/configuration/filtering/

It's less generic code, but it works for what we want IMO and is much easier to implement and read.

internal/logging/sentry.go
```diff
@@ -2,28 +2,25 @@ package logging
...
 		scope.SetTag("stream", config.BinaryName())
 	})
-	fWriter := NewSentryReplacer(sWriter)
 
-	closeFunc := func() {
-		_ = fWriter.Close()
-		_ = sWriter.Close()
-	}
-	return fWriter, closeFunc, nil
+	return sWriter, func() { _ = sWriter.Close }, nil
```
lzap, line 25: I suggest rather to turn off this pesky warning in your IDE than this, Go compiler will not insist on this.

internal/logging/sentry_replacer.go
```diff
@@ -26,70 +24,20 @@ var filters = []string{
...
-	if err != nil {
-		return fmt.Errorf("cannot close filter: %w", err)
-	}
-
-	sr.closed = true
-	return nil
+func (sr *SentryReplacer) Replace(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
+	event.Message = sr.re.ReplaceAllString(event.Message, replacement)
+	return event
 }
```
lzap, line 43: Holy crap that is it? My God I was so focused on the `io.Writer` interface that I did not even dig this up. Nice.

lzap (approved):
Perfect, ton of lines replaced with one, love it.

PR #771, opened 2024-01-18
fix: Double the availability request channel size

ezr-ondrej:
We are seeing major slow down in queing up the availability status checks after certain amount of checks are queued up.

This doubles the channel size in hope the requests can be queued up in channel in the meantime when the previous batch is being send to kafka. This is just a guess for the cause, thus this also adds telemetry, so we have bit more info.

Refs HMS-3415


lzap (approved):
Perfect, thanks.

PR #772, opened 2024-01-19
feat: async avl check kafka send

lzap:
I believe these delays are caused by slow Kafka batch send. This patch makes the sending async, this requires full copy of the buffer to prevent random misbehaviour and a wait group to ensure all messages are flushed on restart. Additionally, I have added new duration metric to actually see how slow this is. Also I slightly increase the buffer size and incoming queue, although I think it will be only the background processing what will help to solve the situation as the problem is really in pace of incoming request being much faster than sending.

### DIFF
--- a/internal/background/availability_queue.go
+++ b/internal/background/availability_queue.go
@@ -3,17 +3,19 @@ package background
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/RHEnVision/provisioning-backend/internal/kafka"
+	"github.com/RHEnVision/provisioning-backend/internal/metrics"
 	"github.com/RHEnVision/provisioning-backend/internal/telemetry"
 	"github.com/rs/zerolog"
 	"go.opentelemetry.io/otel/codes"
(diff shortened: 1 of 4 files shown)


ezr-ondrej:
Love it, I think this solves it permanently, just one ask: could you please keep the batch size now to 16, so we keep the pressure on the system and see the metrics.
Also we already have a ticket for it https://issues.redhat.com/browse/HMS-3415, feel free to use :)

lzap:
> Love it, I think this solves it permanently, just one ask: could you please keep the batch size now to 16, so we keep the pressure on the system and see the metrics.

Well, with background processing the symptoms will be fully eliminated and HTTP requests will never be blocked, but let me decrease it to smaller number like 32. Anything we will choose without testing is just a guess, I do not think this is worth investing time. Number of goroutines is currently exported by goruntime metrics so we can keep an eye on it anytime we want.

ezr-ondrej (approved):
Pefrect, thanks, lets give it a go and see :)

PR #773, opened 2024-01-22
chore: further increase AVL batch size

lzap:
It looks like on stage the largest cumulated checks is around 175, this gets sent in 7 batches. This should be similar on prod as well as there are also two goroutines sending the requests from sources. Increasing to 64 will half number of goroutines spawned.


ezr-ondrej (approved):
Jup, thanks :)

PR #778, opened 2024-02-01
Adding env var for RG and location

mayurilahane:
Ref - https://issues.redhat.com/browse/HMS-3447
To be more consistent including resource group and location value in vault.


deploy/cloudwash_job.yaml
```diff
@@ -57,6 +58,16 @@ objects:
                       secretKeyRef:
                         key: subscription_id
                         name: provisioning-azure-auth-secret
+                  - name: CLEANUP_AZURE__AUTH__RESOURCE_GROUPS
+                    valueFrom:
+                      secretKeyRef:
+                        key: resource_group
+                        name: cloudwash-secret
```
mshriver, line 65: copypasta from your local env - the secre is called `provisioning-azure-auth-secret`

mshriver (approved):

PR #781, opened 2024-03-05
ci: Parallel test execution disabled

adarshdubey-star:
Disabling parallel test execution considering current failure for hms-integration testing workflow.


akhil-jha (approved):
ACK, Let's wait for pr_check

ezr-ondrej (approved):
I see we wait for @mshriver ack, but looks ok to me :)

PR #787, opened 2024-03-26
ci(IQE-2748): Update sel testing image backend

akhil-jha:


ezr-ondrej (commented):
Off topic question: so the pipeline tests frontend as well, it's just defined twice?
Can't we use the same pipeline definition whether it's triggered by backend or frontend merge? :thinking:

ezr-ondrej (approved):
The change itself is fine ofc :)

akhil-jha:
> Off topic question: so the pipeline tests frontend as well, it's just defined twice? Can't we use the same pipeline definition whether it's triggered by backend or frontend merge? 🤔

This yaml is used to hms-integration testing. 
We can use the same file for backend and frontend, for that we'd have to change the deploy.yml in app-interface, and make it point to a common file. But then in case we have to change the testing for backend and frontend, things will get tough.

akhil-jha:
I will merge this PR when frontend(merged) PR testing and deploy is done

ezr-ondrej:
> But then in case we have to change the testing for backend and frontend, things will get tough.

I can't imagine a scenario when it would be useful, but we shouldn't base decisions on my lack of imagination, so if you feel it's useful, lets keep it that way! :)

PR #788, opened 2024-03-26
ci(IQE-2748): Update sel testing image backend

akhil-jha:
Continuing from https://github.com/RHEnVision/provisioning-backend/pull/787 (I messed up my remote :) )


ezr-ondrej (approved):
It's a different image then we have in frontend (or we have the remote defined here).
Is this one correct, or the one in FE? xD

PR #790, opened 2024-04-11
chore: prevent possible leaks

lzap:
This should prevent possible credentials leak into build container. Although it is not being uploaded anywhere, it is better to be on the safe side.


lzap:
You know what, this is useless :-D

ezr-ondrej:
xD yeah, the build container is thrown away, so it should not leak anything :)

PR #791, opened 2024-04-11
chore: lower severity on missing identity

lzap:
We have over 30k issues on stage, this is not a severe issue that is worth reporting.


ezr-ondrej (approved):
Agreed, it's not failure in code.

PR #792, opened 2024-05-03
feat(HMS-3416): Switch to service accounts for dev auth

ezr-ondrej:
For developement we have been using basic auth to authenticate against stage services. This switches the auth to Service accounts.


internal/headers/access_token.go
```diff
@@ -0,0 +1,62 @@
...
+		return "", err
+	}
+
+	data := url.Values{}
+	data.Add("grant_type", grant_type)
+	data.Add("scope", scopes)
+	data.Add("client_id", clientId)
+	data.Add("client_secret", clientSecret)
+
+	req, err := http.NewRequest(http.MethodPost, provider.Endpoint().TokenURL, strings.NewReader(data.Encode()))
```
ezr-ondrej, line 40: This is me being bit fancy just to resolve a static URL, for which we need extra module :thinking: 

provider.Endpoint().TokenURL is quite static value.

internal/headers/access_token.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	token := tokenResponse{}
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse a token response: %w", err)
+	}
+	res.Body.Close()
+	json.Unmarshal(body, &token)
+
```
lzap, line 58: Check return value, also use `defer` to ensure body is closed.

internal/headers/headers.go
```diff
@@ -11,15 +10,14 @@ import (
...
-	if username != "" && password != "" {
-		zerolog.Ctx(ctx).Warn().Msgf("Username/password authentication: %s", username)
-		req.Header.Add("Authorization", "Basic "+basicAuth(username, password))
+func addIdentityHeader(ctx context.Context, req *http.Request, issuerUrl, clientID, clientSecret string) error {
+	if clientID != "" && clientSecret != "" {
+		token, err := getToken(ctx, issuerUrl, clientID, clientSecret)
+		if err != nil {
+			return err
+		}
+		zerolog.Ctx(ctx).Warn().Msgf("Service account authentication: %s", clientID)
```
lzap, line 19: This should not be warning, maybe info or debug.

lzap (changes requested):
Normally, I would insist on caching the token but it has some drawbacks too and since this is only for dev setups, doing it on every request is fine. Thanks.

internal/headers/headers.go
ezr-ondrej: This is warning, because we don't want that to happen in Stage/Prod, so this makes it easier to spot in case by some accident we allow it there.

ezr-ondrej: Or that was the thinking why it was a warning for basic auth, I've kept the level.

ezr-ondrej:
> I would insist on caching the token but it has some drawbacks too and since this is only for dev setups

Yeah exactly, I was thinking about it, but then decided against it for 1) I was lazy and 2) probably not as necessary for dev only.

lzap (approved):
Tested with sources, thanks. Took me a while to prepare my new dev setup, apologies.

PR #793, opened 2024-05-09
feat: config as second arg

lzap:
In my new setup, I am trying JetBrains Fleet. It always seem to start the app in a temporary directory, I need to be able to pass config explicitly. How about this?


ezr-ondrej (approved):
Sure, linter is unhappy, but makes sense to me :+1:

lzap:
Actually, I figured a way to set a working dir in Fleet.

PR #794, opened 2024-05-09
chore: upgrade to 1.21

lzap:
Bump to Go Toolbox 1.21.6 from RHEL 9.4.


lzap:
Omg I knew this is gonna hit us some day. And it hits hard. [Go 1.21 updated how](https://tip.golang.org/doc/go1.21) `init()` functions are called. Unfortunately, we heavily rely on init functions for stub clients, job queue and other stuff instead calling explicit initializers. It is more convenient for ability to run tests individually. However, ordering is a mess:

```
$ GODEBUG=inittrace=1 make test 2>&1| grep init | grep provisioning
init github.com/RHEnVision/provisioning-backend/internal/version @4.1 ms, 0 ms clock, 144 bytes, 1 allocs
init github.com/RHEnVision/provisioning-backend/internal/models @8.6 ms, 0.028 ms clock, 13400 bytes, 184 allocs
init github.com/RHEnVision/provisioning-backend/internal/userdata @8.6 ms, 0.042 ms clock, 14536 bytes, 147 allocs
init github.com/RHEnVision/provisioning-backend/internal/metrics @12 ms, 0.064 ms clock, 16192 bytes, 304 allocs
init github.com/RHEnVision/provisioning-backend/internal/config @13 ms, 0.001 ms clock, 16 bytes, 1 allocs
init github.com/RHEnVision/provisioning-backend/internal/dao @13 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/logging/testing @13 ms, 0.005 ms clock, 1792 bytes, 7 allocs
init github.com/RHEnVision/provisioning-backend/internal/testing/identity @13 ms, 0.004 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/dao/stubs @13 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/clients/http/image_builder @17 ms, 0.006 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/clients/http/rbac @17 ms, 0.013 ms clock, 1392 bytes, 5 allocs
init github.com/RHEnVision/provisioning-backend/internal/clients/http/sources @17 ms, 0.004 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/clients/stubs @17 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/kafka @17 ms, 0.012 ms clock, 6728 bytes, 26 allocs
init github.com/RHEnVision/provisioning-backend/pkg/worker @17 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/queue/stub @17 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/testing/initialization @17 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/middleware @17 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/preload @17 ms, 64 ms clock, 31668344 bytes, 570515 allocs
init github.com/RHEnVision/provisioning-backend/internal/clients/http/gcp @82 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/queue/jq @82 ms, 0 ms clock, 0 bytes, 0 allocs
init github.com/RHEnVision/provisioning-backend/internal/background @82 ms, 0 ms clock, 2784 bytes, 2 allocs
```

I need to figure out how to solve this.

ezr-ondrej:
> I need to figure out how to solve this.

Good topic for a tech discussion I'd say :)

lzap:
I have half of the patch done, will finish it on Monday. I got rid of all `init` functions, but it is quite big one.

lzap:
So I was working for two days on complete refactoring of `init` trying to make it explicit, but I ran into way too many import cycle problems that manifested only after side effect importing was done for some reason. I am falling back now to figure out a different way out.

cmd/spec/example_reservation.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/RHEnVision/provisioning-backend/internal/ptr"
 )
 
-var ReservationTime time.Time = MustParseTime("2013-05-13T19:20:25Z")
+var ReservationTime = MustParseTime("2013-05-13T19:20:25Z")
```
lzap, line 12: There are couple of changes like this, I did not pay attention and IDE fixed these for me. These are not relevant, will not hurt I guess.

lzap:
So the solution was rather simple, but I spent quite some time trying to do a big-bang refactoring, I ran into many import cycle problems and it was a dead end.

The solution I am proposing is to move all `GetXXX` function `init()` initializers into files named `0init.go` and include or exclude them depending on if unit tests are executed. Because it is not possible to guarantee the order of when package is being imported it sometimes loaded stubs and then overwritten the `GetXXX` function with real implementation later on. I tried to only set once, but that did not do the trick so the final solution is to actually avoid this `init()` code alltogether. When running real code (no build tags) or integration tests then only the relevant `init()` function will be called.

Lessons learned:

* Never ever use `init()` this way anymore, it is fine for local initialization but this is a pain to debug or work with.
* Think more about inter-package dependencies to avoid import cycles. Specifically the design when we have a common package with many common code and types (e.g. `dao`) and then packages which use that (e.g. `dao/pxy` and `dao/stub`) are prone to import cycles.

cmd/spec/example_reservation.go
ezr-ondrej: I like it, that's the way C++ headed last time I lookes as well, no useless explicity :)

ezr-ondrej (approved):
Cool! lets do it! :tada:

PR #805, opened 2024-07-18
fixes(HMS-4439): resource limits (ADR-046)

lzap:
See ADR-046 for more details.

This app is very small, minimal traffic, these are the smallest values I have seen in the examples so let’s go with this.


ezr-ondrej (approved):
Nice :+1:

PR #806, opened 2024-07-31
chore: update deps

lzap:
SSIA

There are some shenanigans with recent OpenAPI changes in IB, not updating clients. After the update we got this:

```
lzap@dev:~/provisioning-backend$ make test
/home/lzap/go/bin/go1.21.9 test -tags=test ./...
# github.com/RHEnVision/provisioning-backend/internal/clients/http/image_builder
internal/clients/http/image_builder/client.gen.go:3631:6: CreateBlueprintResponse redeclared in this block
```

It looks like recent changes around blueprint (experimental subpath) broke it. I do not understand why the type is redeclared.


ezr-ondrej:
We'll need to regenerate anyway, the generator now generates nullable params.

lzap:
> We'll need to regenerate anyway, the generator now generates nullable params.

Yeah. The problem is not with prov API, it is with IB API.

lzap:
This is the YAML that is causing the issue: https://gist.github.com/lzap/55d2144c19d85b1072b6d32cfbe9a6c9

lzap:
Filed an issue: https://github.com/oapi-codegen/oapi-codegen/issues/1713

ezr-ondrej (approved):
Thanks @lzap ! :orange_heart:

PR #825, opened 2024-11-04
build: Add rpm-signature-scan task for Konflux pipeline

ezr-ondrej:


ezr-ondrej:
This is no-negotiable and copy-paste, the build pipelines are green, so lets merge :)

PR #887, opened 2025-05-26
chore: Use lzap/gobump tool to bump deps

ezr-ondrej:
The current implementation opens a potential vector of attack.


ezr-ondrej:
Some tests passed, so this is not broken :)

PR #890, opened 2025-06-16
ci: Konflux build pipeline service account migration

ezr-ondrej:
Update Konflux pipeline definitions


lzap (approved):
All changes shown in your diff are valid YAML.
You have used proper indentation, colon spacing, and quoting as required. If you want to check the entire files or ensure they work with your CI/CD tools, you can use tools like yamllint, kubeval, or Tekton’s own validation, but from a pure YAML syntax perspective, these changes are correct.

If you want me to check anything else or run a specific linter, let me know!

PR #897, opened 2025-10-22
chore(decom): Block post/put methods

avitova:
Related to issues:
https://issues.redhat.com/browse/HMS-5864
https://issues.redhat.com/browse/HMS-5865


avitova:
I do not even think this makes sense to be done? @lzap

lzap (approved):
Nicely done.

I was wondering if you could slap it as a global middleware since it selectively only blocks some methods, but this is fine if you want to be explicit.

